PR #26, opened 2024-02-04
Add ServiceFactory to support  quick setup for tools model&agent. 

DavdGao:
## Description
 
**Target**: Use `ServiceFactory` to help generate functions and JSON SCHEMA format dictionary that models can use. 

**Usage**: Taking bing search as an example, the first argument to `ServiceFactory.get` is the function, following by keyword arguments that should be specified by developers. Then `ServiceFactory.get` returns 
- a function for llms to use and 
- a dict to introduce the function in JSON Schema format, which is compatible to OpenAI API

For open-source models, developers need to convert `doc_dict` to their required format. 

```python
_, doc_dict = ServiceFactory.get(
          bing_search,
          num_results=3,
          bing_api_key="xxx",
      )

print(doc_dict)
```
```json
{
    "type": "function",
    "function": {
    "name": "bing_search",
    "description": "Search question in Bing Search API and return the searching results",
      "parameters": {
          "type": "object",
          "properties": {
              "question": {
                  "type": "string",
                  "description": "The search query string.",
              }
          },
          "required": ["question"],
      }
   }
}
```
## Note
- The descriptions are extracted from docstring, which means for developer-defined functions, docstring is required. 
- The types and default values of the arguments are extracted from function definition rather than docstring, which may be more accurate. 
- Documents will be updated after the modifications are confirmed. 

Besides the major modification, these changes are also included in this PR:
1. Add unit tests for service factory. 
2. Rename `service/code` to `service/code_` to avoid duplicated name during debug in pycharm. 
3. Modify some service functions for llm to use. 

## Checklist


src/agentscope/service/code_/exec_python.py
pan-x-c, line 1: It feels like `execute_code` is better than `code_`.
In addition, sphinx doc needs to be updated simultaneously.

pan-x-c (approved):
LGTM

PR #46, opened 2024-03-04
add tongyi model wrapper

qbc2016:
As the title says.


tests/model_test.py
```diff
@@ -52,7 +52,7 @@ def test_load_model_configs(self) -> None:
             {
                 "model_type": "openai",
                 "config_name": "gpt-4",
-                "model": "gpt-4",
+                "model_name": "gpt-4",
```
pan-x-c, line 55: Add an assert in openaiwrapper  to ensure that the `model_name` field is provided

PR #59, opened 2024-03-11
Add Gemini Chat and Embedding models in AgentScope

DavdGao:
## Description

- Add support for gemini chat and embedding python API in AgentScope.
- The type of `raw` field of `ModelResponse` is modified into `Any`. Please check whether the distribution is affected @pan-x-c 

TODO:
README.md and tutorial will be updated together with ollama and dashscope. 

## Checklist


src/agentscope/models/model.py
```diff
@@ -103,16 +105,21 @@ def image_urls(self) -> Sequence[str]:
         return self._image_urls
 
     @property
-    def raw(self) -> dict:
-        """Raw dictionary field."""
+    def raw(self) -> Any:
```
pan-x-c, line 108: The `Any` type is too free. It is recommended to give a warning if using this field and add another field named `dictionary` or `json` in `ModelResponse` to inherit the original `raw` field.

DavdGao: The response from Gemini is in neither json nor string format, and we cannot ensure that responses from all model APIs can be turned into json or string in the future.
(In my view, since the response is from remote model services, we can assume that raw response is serializable.)

src/agentscope/models/gemini_model.py
```diff
@@ -0,0 +1,293 @@
...
+try:
+    import google.generativeai as genai
+except ImportError:
+    genai = None
+
+
+class GeminiWrapperBase(ModelWrapperBase):
+    """The base class for Google Gemini model wrapper."""
+
+    _generation_method = None
```
ZiTao-Li, line 22: Seems only for Gemini?  Leave a comment about how it will be used?

tests/z_gemini_test.py
ZiTao-Li, line 1: Why the name starts with "z_"?

ZiTao-Li (commented):
See two minor inline comments. Other parts LGTM.

tests/z_gemini_test.py
DavdGao: Currently, all of our unit tests run within a single process, which leads to them sharing the same singleton classes such as _Runtime, FileManager, and Monitor. This setup implies that if one unit test alters these singleton classes, it can influence subsequent tests. Therefore, to prevent such interference, we have rearranged the unit tests so that the tests for Gemini are executed last.

DavdGao: This is only a temporary solution and we need to handle singleton classes in unit tests better in the future.

src/agentscope/models/gemini_model.py
DavdGao: This argument is only for gemini api, which aims at filtering different models in `list_models` function.

DavdGao: More explanation is added.

pan-x-c (approved):
LGTM

PR #74, opened 2024-03-15
Fix error in agent when `use_memory` is `False`

DavdGao:
## Description

- Fix the error when `self.memory` is `None` in `reply` function. 
- Disable the default value of `model_config_name` of built-in agents, so that a model must be provided when using agent.

## Checklist

### DIFF
--- a/src/agentscope/agents/dialog_agent.py
+++ b/src/agentscope/agents/dialog_agent.py
@@ -15,8 +15,8 @@ class DialogAgent(AgentBase):
     def __init__(
         self,
         name: str,
-        sys_prompt: Optional[str] = None,
-        model_config_name: str = None,
+        sys_prompt: str,
+        model_config_name: str,
         use_memory: bool = True,
         memory_config: Optional[dict] = None,
         prompt_type: Optional[PromptType] = PromptType.LIST,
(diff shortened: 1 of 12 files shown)


src/agentscope/agents/dialog_agent.py
```diff
@@ -68,13 +68,13 @@ def reply(self, x: dict = None) -> dict:
             response to the user's input.
         """
         # record the input if needed
-        if x is not None:
+        if not self.memory:
```
qbc2016, line 71: Should it be `if self.memory`? BTW, if use_memory is False, does `self.memory` defined?

DavdGao: Thanks! The error is fixed

ZiTao-Li (approved):
lgtm

src/agentscope/agents/dialog_agent.py
```diff
@@ -68,13 +68,13 @@ def reply(self, x: dict = None) -> dict:
             response to the user's input.
         """
         # record the input if needed
-        if x is not None:
+        if self.memory:
```
qbc2016, line 71: If set `use_memory=False` when initializing an agent, `self.memory` is not defined, which will cause `AttributeError: 'DialogAgent' object has no attribute 'memory'`

DavdGao: The bug is fixed accordingly.

PR #94, opened 2024-03-21
Support agent_id in agent server

pan-x-c:
## Description

In the previous version of AgentScope, each agent server contained only one agent instance.
When the main process is finished, the already running Agent server needs to be stopped to clear the local state of each agent, even if the code of these agents has not been modified.

To avoid the above scenario, this PR introduces `agent_id` into the agent server. (#128 )
After the agent server is started, it can instantiate multiple agent instances with different local states to serve multiple main processes. 
All agents on the same server share the same computing resources and Class, but can have different initialization parameters and local states. 
These instances are distinguished using `agent_id`.
The current RPC message should contain an additional `agent_id` field to determine which agent instance to call.

All the above modifications are completely transparent to users, and the AgentScope distributed user interface remains unchanged compared to the previous version.

## Checklist


DavdGao (commented):
A few questions need to be discussed: 
1. What's the difference between `runtime_id` and `session_id`? We should avoid introducing similar concepts repeatedly.
2. What's shared between two agent instances in a single server? 
3. How about the data, storage and memory isolations in a server? Especially for singleton classes. 
4. About "...After the agent server is started, it can instantiate multiple identical agent instances with different local states to serve multiple main processes...", why do we support identical agents in a server rather than multiple different agents? 
5. How is the concept "state" defined here?

DavdGao (approved):
lgtm

PR #112, opened 2024-03-28
deal with audio2text when result["output"] is None

qslia:
I'm trying to use the code below to do audio2text. But there is something wrong with streaming audio output when recording audio. So this is my solution.

def f_audio_mic(audio):
    text = audio2text(audio)
    print(text)
    return text

with gr.Blocks() as test:
    audio_mic = gr.Audio(type='filepath', streaming=True, format='wav')
    text1 = gr.Textbox()
    audio_mic.stream(
        f_audio_mic,
        inputs=audio_mic,
        outputs=[text1]
    )
test.launch()

### DIFF
--- a/src/agentscope/web/studio/utils.py
+++ b/src/agentscope/web/studio/utils.py
@@ -183,6 +183,8 @@ def audio2text(audio_path: str) -> str:
     )
 
     result = rec.call(audio_path)
+    if not result["output"]:
+        return ""
     return " ".join([s["text"] for s in result["output"]["sentence"]])
 
 


qslia:
This is my new solution. and it worked, so I closed the pull request.

def f_audio1(audio1):  
    text = audio2text(audio1)
    return text

with gr.Blocks() as demo:
    audio1 = gr.Audio(label='audio', sources="microphone", type='filepath')
    text = gr.Textbox()
    button1 = gr.Button()
    button1.click(f_audio1, [audio1], [text])

demo.launch()

PR #135, opened 2024-04-02
Agent servers support running any type of agent

pan-x-c:
## Description

Based on #94 

1. rename `RpcServerSideWrapper` into `AgentPlatform`
2. Simplify the setup of agent platform. Only `host` and `port` are necessary. (`custom_agents` is required, if using agent class written by the user)
3. AgentPlatform is completely decoupled from Agent. Multiple different types of agents can run on the same agent platform now.
4. RpcAgentServer now starts in an async way
5. Deprecated the `launch_server` parameter in `to_dist` method. When the `host` and `port` are not passed in, the server will be started automatically
6. Support using `to_dist=True` to initialize any `AgentBase` subclass to convert the agent into its distributed version
7. Support agent nesting
8. Add a large scale simulation example
9. Rewrite the distributed deployment tutorial

## Checklist


src/agentscope/agents/agent.py
```diff
@@ -15,16 +16,102 @@
...
+                cls.registry[name] = cls
+        super().__init__(name, bases, attrs)
 
     def __call__(cls, *args: tuple, **kwargs: dict) -> Any:
+        if len(args) > 0:
+            logger.warning(
+                "Please only use keywords arguments to initialize your agents."
+                "\ne.g.\nagent = YourAgentClass(name='agent_name', ...)",
+            )
+        if "to_dist" in kwargs:
```
ZiTao-Li, line 44: Do we need to check to_dist == True here?

examples/distributed_simulation/README.md
```diff
@@ -0,0 +1,99 @@
+# Distributed Large Scale Simulation
+
+> **WARNING:**
+> **This example will consume a huge amount of tokens.**
+> **Please do not use model API that requires payment.**
```
ZiTao-Li, line 5: A little bit misleading. "Using model API with this example can introduce high cost." should be enough as a warning.

ZiTao-Li (commented):
* The rpc_argent.py may be too large. Wondering if it can be split into several smaller files in a module?
* See inline comments

pan-x-c:
> * The rpc_argent.py may be too large. Wondering if it can be split into several smaller files in a module?
> * See inline comments

Currently, classes in the `rpc_agent.py` file are highly coupled. If they are separated into multiple files, circular references may occur. It might be better to keep the status quo for now

examples/distributed_simulation/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+> **WARNING:**
+> **This example will consume a huge amount of tokens.**
+> **Using paid model API with this example can introduce a high cost.**
+> **Users with powerful GPUs (A800 or better) can use local inference services (such as vLLM) to run this example,**
+> **while CPU inference services such as ollama is not recommended.**
+
+This example is a large scale simulation to demonstrate the scalability of AgentScope's distributed mode. From this example, you can learn:
+
+- How to run a large number of agent servers in a GPU cluster.
```
DavdGao, line 11: BTW, give an example number here, e.g. 
1. In our test, we setup xxx agents in x machines equipped with xx CPU cores. 
2. Totally xxx seconds is used with xxx agents.


src/agentscope/agents/agent.py
```diff
@@ -15,16 +16,102 @@
...
+                logger.warning(
+                    f"Agent class with name [{name}] already exists.",
+                )
+            else:
+                cls.registry[name] = cls
+        super().__init__(name, bases, attrs)
 
     def __call__(cls, *args: tuple, **kwargs: dict) -> Any:
+        if len(args) > 0:
+            logger.warning(
```
DavdGao, line 40: When is this function called, and does this warning means we must initialize all agents (including centralized and distributed agents) with keyword arguments?

tests/rpc_agent_test.py
```diff
@@ -292,29 +319,19 @@ def test_multi_rpc_agent(self) -> None:
...
             name="b",
         )
         # rpc agent c
-        agent_c = DemoRpcAgentAdd(
+        agent_c = DemoRpcAgentAdd(  # pylint: disable=E1123
             name="c",
-        ).to_dist(
-            host=host,
-            port=port2,
+            to_dist=True,
```
DavdGao, line 334: How should we initialize a remote agent in the new setting? Pass a to_dist argument or calling to_dist function?

DavdGao (commented):
Plz see inline comments.

tests/rpc_agent_test.py
pan-x-c: Both methods are okay, see tutorial "Advanced Usage of `to_dist`"

docs/sphinx_doc/en/source/tutorial/208-distribute.md
```diff
@@ -12,78 +12,161 @@ This tutorial will introduce the implementation and usage of AgentScope distribu
...
 # Create an agent service process
-server_a = RpcAgentServerLauncher(
-    agent_class=AgentA,
-    agent_kwargs={
-        "name": "A"
-        ...
-    },
-    host="a.b.c.d",
-    port=12001,
+server = RpcAgentServerLauncher(
```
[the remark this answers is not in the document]
pan-x-c: `RpcAgentServerLauncher` is only used to launch the RPC agent server.
And `AgentPlatform` (named `RpcServerSideWrappe` before) is the RPC servicer that provides RPC service for RPC agent server.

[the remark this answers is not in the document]
DavdGao: Cannot find the implementation (e.g. AgentPlatform class) in this PR.

docs/sphinx_doc/en/source/tutorial/208-distribute.md
```diff
@@ -12,78 +12,161 @@ This tutorial will introduce the implementation and usage of AgentScope distribu
 
 ## Usage
 
-In AgentScope, the process that runs the application flow is called the "main process", and all agents will run in separate processes.
-According to the different relationships between the main process and the agent process, AgentScope supports two distributed modes: Master-Slave and Peer-to-Peer mode.
-In the Master-Slave mode, developers can start all agent processes from the main process, while in the Peer-to-Peer mode, the agent process is independent of the main process and developers need to start the agent service on the corresponding machine.
+In AgentScope, the process that runs the application flow is called the "main process", and each agent can run in a separate process named "agent server process".
+According to the different relationships between the main process and the agent server process, AgentScope supports two modes for each agent: **Subprocess** and **Independent** mode.
+In the Subprocess mode, agent server processes will be automatically started from the main process, while in the Independent mode, the agent server process is independent of the main process and developers need to start the agent server process on the corresponding machine.
```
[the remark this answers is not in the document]
pan-x-c: 2. It's not a daemon process, using this name is misleading.

docs/sphinx_doc/en/source/tutorial/208-distribute.md
```diff
@@ -12,78 +12,161 @@ This tutorial will introduce the implementation and usage of AgentScope distribu
...
+    name="B"
+    # ...
+).to_dist()
 ```
 
-#### Peer-to-Peer Mode
+#### Independent Mode
 
-In the Peer-to-Peer mode, we need to start the service of the corresponding agent on the target machine first. For example, deploy an instance of `AgentA` on the machine with IP `a.b.c.d`, and its corresponding port is 12001. Run the following code on this target machine:
+In the Independent mode, we need to start the agent server process on the target machine first.
```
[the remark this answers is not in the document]
pan-x-c: description is in line 15

src/agentscope/agents/agent.py
pan-x-c: Only agents that need to use `to_dist=True` have this requirement.

DavdGao (commented):
I have two issues in this modification:
1. If it's appropriate to add a new entry for distributed agent. I know directly instantiate a distributed agent object will save time, but the in this way the local mode and distributed mode are mixed. 
2. If the new entry is necessary. Please move the initialization arguments to the constructor of `AgentBase` class, and add `kwargs` to the agent classses under the library `agentscope.agents`.

examples/distributed_simulation/main.py
```diff
@@ -0,0 +1,203 @@
+# -*- coding: utf-8 -*-
+""" Average """
```
DavdGao, line 2: Plz provide a more specific description

src/agentscope/agents/agent.py
```diff
@@ -40,6 +102,7 @@ def __init__(
         model_config_name: str = None,
         use_memory: bool = True,
         memory_config: Optional[dict] = None,
+        to_dist: Optional[Union[dict, bool]] = False,
```
DavdGao, line 105: Just `host: str=None` and `port: int=None` is fine (users don't need to search what parameters should be contained here and their format)

DavdGao (commented):
plz see inline comments, others lgtm

src/agentscope/agents/agent.py
```diff
@@ -40,6 +139,7 @@ def __init__(
         model_config_name: str = None,
         use_memory: bool = True,
         memory_config: Optional[dict] = None,
+        to_dist: Optional[Union[DistConf, bool]] = False,
```
DavdGao, line 142: Why bool here?

DavdGao (approved):
lgtm

PR #136, opened 2024-04-02
Update ReActAgent for qwen and other models

DavdGao:
### About Prompt
1. update format function for `DashScopeChatWrapper`, `OllamaGenerationWrapper` and `GeminiChatWrapper`
2. move model response related classes into a single file
3. update docs

### About `ReActAgent`
1. update prompt for `ReActAgent` to adapt to qwen-max
2. add a re-correct processing in `ReActAgent`
3. add a `ResponseParser` to handle the conversation from text to dict and list
4. add a `ResponseParsingError` to enable self-correct operation

## Checklist

### DIFF
--- a/src/agentscope/agents/react_agent.py
+++ b/src/agentscope/agents/react_agent.py
@@ -23,28 +23,27 @@ DEFAULT_TOOL_PROMPT = """The following tool functions are available in the forma
 ## Tool Functions:
 {function_prompt}
 
-## Notice:
-1. Fully understand the tool function and its arguments before using it.
-2. Only use the tool function when it's necessary.
-3. Check if the arguments you provided to the tool function is correct in type and value.
-4. You can't take some problems for granted. For example, where you are, what's the time now, etc. But you can try to use the tool function to solve the problem.
+## What You Should Do:
+1. First, analyze the current situation, and determine your goal.
+2. Then, check if your goal is already achieved. If so, try to generate a response. Otherwise, think about how to achieve it with the help of provided tool functions.
+3. Respond in the required format.
+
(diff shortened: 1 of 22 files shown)


src/agentscope/models/dashscope_model.py
```diff
@@ -330,31 +262,53 @@ def advanced_format(
...
+                )
 
-        return [{"role": "system", "content": prompt_str}]
+        dialogue_history = "\n".join(dialogue)
+
+        user_content_template = "## Dialogue History\n{dialogue_history}"
+
+        messages.append(
+            {
+                "role": "system",
```
qbc2016, line 304: If this is the only message for model, the role should be change to "user". Otherwise, it will raise  an error.

src/agentscope/agents/react_agent.py
```diff
@@ -138,17 +147,39 @@ def reply(self, x: dict = None) -> dict:
...
-                prompt,
-                parse_func=parse_func,
-                max_retries=3,
-            ).raw
+            try:
+                hint_msg = Msg("system", TOOL_HINT_PROMPT, role="system")
+                self.memory.add(hint_msg)
+
+                # Generate LLM response
+                prompt = self.model.advanced_format(self.memory.get_memory())
```
qbc2016, line 155: In the definition of advanced_format in dashscope_model.py, the role in the return value should be "user"?

src/agentscope/models/dashscope_model.py
DavdGao: Done

pan-x-c (approved):
LGTM

PR #147, opened 2024-04-08
Fix "File exists error" when multiple agents are called at the same time

pan-x-c:
## Description

When multiple agents in the same agent server are called simultaneously, they will try to create the same directory at the same time and cause errors.

## Checklist

### DIFF
--- a/src/agentscope/file_manager.py
+++ b/src/agentscope/file_manager.py
@@ -46,8 +46,7 @@ class _FileManager:
         """Get the path of the subdir and create the subdir if it does not
         exist."""
         path = os.path.join(self.dir, _runtime.runtime_id, subdir)
-        if not os.path.exists(path):
-            os.makedirs(path)
+        os.makedirs(path, exist_ok=True)
         return path
 
     def _get_file_path(self, file_name: str) -> str:
@@ -89,8 +88,7 @@ class _FileManager:
         """Set the directory for saving files."""
         self.dir = save_dir
         runtime_dir = os.path.join(save_dir, _runtime.runtime_id)
-        if not os.path.exists(runtime_dir):
-            os.makedirs(runtime_dir)
+        os.makedirs(runtime_dir, exist_ok=True)
 
         self.save_api_invoke = save_api_invoke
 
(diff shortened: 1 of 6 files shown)


src/agentscope/utils/monitor.py
```diff
@@ -244,16 +244,17 @@ def __init__(
 
 
 @contextmanager
-def sqlite_transaction(db_path: str) -> Generator:
+def sqlite_transaction(db_path: str, timeout: float = 30.0) -> Generator:
```
DavdGao, line 247: Why we need a timeout argument here?

src/agentscope/file_manager.py
```diff
@@ -46,8 +46,7 @@ def _get_and_create_subdir(self, subdir: str) -> str:
         """Get the path of the subdir and create the subdir if it does not
         exist."""
         path = os.path.join(self.dir, _runtime.runtime_id, subdir)
-        if not os.path.exists(path):
```
DavdGao, line 49: since we have a "if" statement here to determine the existence of the path, why do we have "File exists error" here?

DavdGao (commented):
please see inline comments

src/agentscope/utils/monitor.py
pan-x-c: the default timeout is only 5 seconds, and when the number of agents is large, it may cause timeout errors, e.g.`database is locked.`

src/agentscope/file_manager.py
pan-x-c: multiple threads reach this line and find the file does not exist, so they try to create the file at the same time.

src/agentscope/utils/monitor.py
DavdGao: So this timeout is actually how long to wait for the db lock?

src/agentscope/file_manager.py
pan-x-c: Yes

src/agentscope/utils/monitor.py
pan-x-c: Yes

DavdGao (approved):
lgtm

PR #197, opened 2024-05-08
[Hotfix] Fix memory loading error

ZiTao-Li:
---
name: Fix memory loading error
---
## Description

Fix the error reported in #194 

The reason of this error is the temporary memory did not ensure the loaded content to be in message types in AS. Thus, this PR make the following changes to fix the issue:

1. When exporting the memory, use the `serialize` function provided in `message.py` instead of `json.dump`, and use `deserialize` instead of `json.load` when export/load memory from files.
2. When adding new pieces of message or dict into memory, manually check if they are Msg/Tht

## Checklist

### DIFF
--- a/src/agentscope/memory/memory.py
+++ b/src/agentscope/memory/memory.py
@@ -35,6 +35,8 @@ class MemoryBase(ABC):
     def update_config(self, config: dict) -> None:
         """
         Configure memory as specified in config
+        Args:
+            config (`dict`): Configuration of resetting this memory
         """
         self.config = config
 
@@ -45,8 +47,16 @@ class MemoryBase(ABC):
         filter_func: Optional[Callable[[int, dict], bool]] = None,
     ) -> list:
         """
-        Return a certain range (`recent_n` or all) of memory, filtered by
-        `filter_func`
+        Return a certain range (`recent_n` or all) of memory,
+        filtered by `filter_func`
+        Args:
+            recent_n (int, optional):
+                indicate the most recent N memory pieces to be returned.
+            filter_func (Optional[Callable[[int, dict], bool]]):
+                filter function to decide which pieces of memory should
+                be returned, taking the index and a piece of memory as
+                input and return True (return this memory) or False
+                (does not return)
         """
 
     @abstractmethod
@@ -56,6 +66,10 @@ class MemoryBase(ABC):
     ) -> None:
         """
         Adding new memory fragment, depending on how the memory are stored
+        Args:
+            memories (Union[Sequence[dict], dict, None]):
+                Memories to be added. If the memory is not in MessageBase,
+                it will first be converted into a message type.
         """
 
     @abstractmethod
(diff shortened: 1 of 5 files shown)


src/agentscope/memory/temporary_memory.py
```diff
@@ -46,14 +47,28 @@ def add(
...
         memories_idx = set(_.id for _ in self._content if hasattr(_, "id"))
         for memory_unit in record_memories:
+            if not issubclass(type(memory_unit), MessageBase):
+                try:
+                    if (
+                        "name" in memory_unit
+                        and memory_unit["name"] == "thought"
+                    ):
+                        memory_unit = Tht(**memory_unit)
+                    else:
```
pan-x-c, line 65: Add an `elif` entry for `PlaceholderMessage`

```
elif isinstance(memory_unit, PlaceholderMessage)
    memory_unit.update_value()
    memory_unit = Msg(**memory_unit)
```

src/agentscope/memory/memory.py
```diff
@@ -63,7 +68,7 @@ def delete(self, index: Union[Iterable, int]) -> None:
     @abstractmethod
     def load(
         self,
-        memories: Union[str, dict, list],
+        memories: Union[str, list, MessageBase],
         overwrite: bool = False,
```
DavdGao, line 72: Add more docstring to describe the arguments. For example, if `overwrite` is `False` and the content of memory is not empty, how will we load the new memory (to the beginning or end)?

DavdGao: Similar problem with the export interface, e.g. what's the use of the argument `to_mem`?

A suggestion: place `file_path` before `to_mem`, which is a more common used argument.

src/agentscope/memory/temporary_memory.py
```diff
@@ -120,16 +135,16 @@ def export(
 
     def load(
         self,
-        memories: Union[str, dict, list],
+        memories: Union[str, list, MessageBase],
         overwrite: bool = False,
     ) -> None:
         if isinstance(memories, str):
```
DavdGao, line 141: Please check if this is a bug when loading Tht objects
image

tests/memory_test.py
```diff
@@ -84,29 +66,23 @@ def test_delete(self, mock_logging: MagicMock) -> None:
...
-        self.assertEqual(
-            self.memory.get_memory(),
-            [{"invalid_key": "invalid_value"}],
+        with self.assertRaises(Exception) as context:
+            self.memory.add(self.invalid)
+        self.assertTrue(
+            f"Cannot add {self.invalid} to memory" in str(context.exception),
         )
 
     def test_load_export(self) -> None:
```
DavdGao, line 76: Add test for loading and exporting Tht objects. Or just remove Tht class, since we never use it.

src/agentscope/memory/memory.py
```diff
@@ -63,7 +68,7 @@ def delete(self, index: Union[Iterable, int]) -> None:
     @abstractmethod
     def load(
         self,
-        memories: Union[str, dict, list],
```
DavdGao, line 66: Suggest to be more specific, e,g. `Union[str, List[MessageBase], MessageBase]`

Similar problem with the typing of the return values in export interface.

DavdGao (commented):
Please see inline comments

src/agentscope/memory/temporary_memory.py
ZiTao-Li: This `elif` won't be trigger because of the outter if-check.
Added another `if` for `PlaceholderMessage`.

DavdGao (approved):
LGTM

PR #255, opened 2024-05-24
Add DashScope API based multimodal service functions in AgentScope library

PengHongyiNTU:
This PR introduces three new services to the service toolkit: dashscope_text_to_image, dashscope_image_to_text, and dashscope_text_to_audio.

also fix a small error in examples/conversation_with_customized_services


garyzhang99:
The title of this PR isn't quite accurate.

garyzhang99:
Although it might be a bit of a hassle, please consider adding corresponding tests for the new added service functions. 
You may refer to the examples in `tests/execute_python_code_test.py` and `tests/dashscope_test.py`, utilizing mocks to simulate the return values of the relevant `dashscope` libraries. This way, if any bugs arise in the future due to changes in the library, we can quickly pinpoint them.

src/agentscope/service/multi_modality/dash_scope.py
```diff
@@ -0,0 +1,229 @@
...
+        print(image_to_text(image_url, query_prompt)) gives:
+        {'status': 'SUCCESS', 'content': 'A beautiful sunset in the mountains'}
+    """
+    dashscope.api_key = api_key
+    if not isinstance(image_urls, tuple):
+        image_urls = (image_urls,)
+    contents = []
+    for image_url in image_urls:
+        # check image url is local content or remote content
+        if not image_url.startswith(("http://", "https://")):
```
DavdGao, line 126: Suggested to reuse the DashScopeMultimodalModelWrapper in AgentScope, which already handles the local/web url and prompt construct

src/agentscope/service/multi_modality/dash_scope.py
```diff
@@ -0,0 +1,229 @@
...
+        prompt = "A beautiful sunset in the mountains"
+        print( dashscope_text_to_image(prompt)) gives:
+        {'status': 'SUCCESS',
+        'content': {'fig_paths':
+            ['FIG_RELATEIVE_PATH1',
+            'FIG_RELATEIVE_PATH2']}}
+    """
+
+    dashscope.api_key = api_key
+    res = dashscope.ImageSynthesis.call(
```
DavdGao, line 57: Suggested to reuse the model wrapper in AgentScope

src/agentscope/service/multi_modality/dash_scope.py
```diff
@@ -0,0 +1,229 @@
...
+        err_msg += res.message
+        return ServiceResponse(
+            ServiceExecStatus.ERROR,
+            {"error": err_msg},
+        )
+
+
+def dashscope_image_to_text(
+    image_urls: Union[str, Tuple[str, ...]],
+    prompt: str,
```
DavdGao, line 98: The docstring misses the parameter `prompt`, and should we consider to provided a default prompt here? e.g. describe the figure.

DavdGao (commented):
Please see inline comments.

src/agentscope/service/multi_modality/dash_scope.py
```diff
@@ -0,0 +1,229 @@
...
+    api_key: str,
+    model: str = "qwen-vl-plus",
+) -> ServiceResponse:
+    """Generate text based on an image.
+
+    Args:
+        image_urls (`str`): the url of single or multiple images.
```
xieyxclack, line 105: I am not sure, whether `image_urls` can contain both local paths and remote urls?

xieyxclack (commented):
LGTM. 
BTM, it may be not convenient for users to have to pass the same api_key every time they use dashscope_xx_to_xx. Is it possible for users to set up an api_key with export api_key=xxx instead? @DavdGao

src/agentscope/service/multi_modality/dash_scope.py
DavdGao: The local url requires special prefix in the url

src/agentscope/service/multi_modality/dash_scope.py
PengHongyiNTU: The DashScopeMultimodalModelWrapper handles it through a protected member function called _convert_url. Therefore, I choose to stay with the original version.

PengHongyiNTU: Can also use Wrapper.format but it require the input to be a MessageBase. Seems a bit heavy for me.

PengHongyiNTU (commented):
Need modify test accordingly

DavdGao (approved):
LGTM

PR #304, opened 2024-06-19
Add OpenAI as services 

PengHongyiNTU:
Feature Addition: OpenAI's Multimodal Services includes Text2Img, Img2Text, ImgEditing, ImgVariation, Text2Audio, Audio2Text


src/agentscope/service/multi_modality/openai_services.py
```diff
@@ -0,0 +1,536 @@
...
+        )
+    except Exception as e:
+        return ServiceResponse(
+            ServiceExecStatus.ERROR,
+            str(e),
+        )
+
+
+def openai_create_image_variation(
+    image_url: str,
```
DavdGao, line 248: So here we support both local and web URLs?

DavdGao (commented):
lgtm, please update the service function section in tutorial accordingly, and add unit tests for openai services

src/agentscope/service/multi_modality/openai_services.py
PengHongyiNTU: Yes.  As `agentscope.models.openai_model` does not provide wrappers for editing and creating image variation, I use `_parse_url` function that takes either local or web URLs as input, and return objects `BytesIO` .

DavdGao (commented):
LGTM, please update the tutorial file in both Chinese and English

DavdGao (approved):
lgtm

PR #325, opened 2024-07-04
Fix display bug in workstation for group nodes

PatrickG1014:
## Description

Fix display bug in AgentScope workstation for group nodes. Specifically, when there are multiple levels of nested nodes, hide the nodes below first, then hide the node above two levels, and then open the upper node. The hidden nodes below will still be displayed, as shown in the figure.

截屏2024-07-04 14 30 42

For unhiding, it should be determined whether the child node's `contentBox` is marked as hidden to determine whether to unhide it.

### DIFF
--- a/src/agentscope/studio/static/js/workstation.js
+++ b/src/agentscope/studio/static/js/workstation.js
@@ -962,11 +962,17 @@ function hideShowGroupNodes(groupId, show) {
         groupInfo.data.elements.forEach(elementNodeId => {
             const elementNode = document.getElementById(`node-${elementNodeId}`);
             const childNodeInfo = editor.getNodeFromId(elementNodeId);
+            const contentBox = elementNode.querySelector('.box') ||
+                elementNode.querySelector('.box-highlight');
             if (elementNode) {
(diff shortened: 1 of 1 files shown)


qbc2016 (commented):
LGTM.

src/agentscope/studio/static/js/workstation.js
```diff
@@ -962,11 +962,17 @@ function hideShowGroupNodes(groupId, show) {
...
+                elementNode.querySelector('.box-highlight');
             if (elementNode) {
                 elementNode.style.display = show ? '' : 'none';
             }
             if (childNodeInfo.class === 'GROUP') {
-                hideShowGroupNodes(elementNodeId, show);
+                if (!show) {
+                    hideShowGroupNodes(elementNodeId, show);
+                } else if (contentBox && !contentBox.classList.contains('hidden')) {
+                    hideShowGroupNodes(elementNodeId, show);
```
qbc2016, line 974: BTW, since the behaviors of  "if" and "else if" conditions are the same, you can merge them.

pan-x-c (approved):
LGTM

PR #341, opened 2024-07-16
Add new parsers for uncertain tag names and quantities

DavdGao:
## Description

- Disable the `tool` argument in ReActAgent; 
- Add a new RegexTaggedContentParser for uncertain tag names and quantities. For example, when calling functions, each function differs each other in terms of arguments, and we cannot know what tags to extract in advanced. 
```
<function>bing_search</function>
<query>Some news in July</query>
```
- Update docs accordingly

## Checklist


src/agentscope/parsers/regex_tagged_content_parser.py
```diff
@@ -0,0 +1,169 @@
...
+            "<content>" in tagged_content_pattern
+        ), "The tagged content pattern should contain a named group 'content'."
+
+        self.tagged_content_pattern = tagged_content_pattern
+        self._format_instruction = format_instruction
+        self.try_parse_json = try_parse_json
+        self.required_keys = required_keys or []
+
+    @property
+    def format_instruction(self) -> str:
```
garyzhang99, line 113: Not an important issue, but should we have a built in format instruction template? Or at least let the user know what their own format instruction should be link?

src/agentscope/agents/react_agent.py
```diff
@@ -129,13 +94,22 @@ def __init__(
...
 
         # Initialize a parser object to formulate the response from the model
-        self.parser = MarkdownJsonDictParser(
-            content_hint={
-                "thought": "what you thought",
-                "speak": "what you speak",
-                "function": service_toolkit.tools_calling_format,
-            },
-            required_keys=["thought", "speak", "function"],
+        self.parser = RegexTaggedContentParser(
```
garyzhang99, line 97: Instead of just remove the former `MarkdownJsonDictParser`, should we give an argument to user, and let the user choose which parser they want? e.g. (parser: Literal["json", "tag"] = "tag")

src/agentscope/agents/react_agent.py
```diff
@@ -129,13 +94,22 @@ def __init__(
...
+            format_instruction="""Respond with specific tags as outlined below:
+
+- When not calling any tool functions:
+<thought>what you thought</thought>
+<speak>what you speak</speak>
+
+- When calling tool functions, note the "arg_name" should be replaced with the actual argument name:
+<thought>what you thought</thought>
+<speak>what you speak</speak>
```
garyzhang99, line 106: I am not sure whether the <speak> here is a must for ReAct agent. Other frameworks found that <thought> + <action> performs the best. And in practice, both <thought> and <action> should be spoken by the agent to let the user know the internal process of agent anyway.

src/agentscope/parsers/regex_tagged_content_parser.py
DavdGao: The key problem in regex tagged content parser is that we cannot know the required keys in advanced. So here we can provide a general format instruction here.

src/agentscope/agents/react_agent.py
DavdGao: Improved by providing a response field in the last.

PR #362, opened 2024-07-24
Add customized agent in workstation.

PatrickG1014:
## Description

Support user-defined agent classes in the workstation, read the parameters of the __init__ method of the customized agent class and display them in the node (str, int, or float).

截屏2024-07-24 17 52 48

截屏2024-07-24 17 52 34


src/agentscope/studio/static/html-drag-components/agent-customizedagent.html
```diff
@@ -0,0 +1,34 @@
...
+        <div class="readme">
+            {{agent._cls_name}}
+            <div>Node ID: <span class="node-id">ID_PLACEHOLDER</span></div>
+        </div>
+
+        {% for key, value in agent.items() %}
+        {% if key != '_cls_name' %}
+        {% if value.type_hint == 'str' %}
+        <label>{{key}}</label>
+        <input type="text" df-args-{{key}} placeholder="{{value.default}}" data-required="true">
```
qbc2016, line 24: When the key is customized or just "sys_prompt", it should be allowed to use Enter. Using <textarea> may be a better way.

DavdGao:
Given that our online workstation will support customized agents in the future, we've decided to leave the implementation of these agents entirely to the workstation itself, rather than importing them at launch. Therefore, we will close this PR for now.

DavdGao:
Given that our online workstation will support customized agents in the future, we've decided to leave the implementation of these agents entirely to the workstation itself, rather than importing them at launch. Therefore, we will close this PR for now.

PR #378, opened 2024-08-02
Add environment modules

pan-x-c:
### Add Environment Modules (`Env` and its sub-classes)
`Env` is a key concept of AgentScope, representing global data shared among agents.

Each env has name and value, and multiple envs can be organized into a tree structure, where each env can have multiple children envs and one parent env.

Different implementations of envs may have different event functions, which are marked by `@event_func`.
Users can bind `EventListener` to specific event functions, and the listener will be activated when the event function is called.

Similar to `AgentBase` sub-classes, `Env` sub-classes also support the `to_dist` call, and the parameter is the same as `AgentBase`. 

#### Move `to_dist` into `RpcMeta`

`RpcMeta` is a meta-class that integrates all `to_dist` related methods/parameters.
Now, `to_dist` function/parameter can be used directly if the class is a subclass of `RpcMeta`.
The original `to_dist` and `_AgentMeta` is removed.

#### Extend `RpcAgent` into `RpcObject`

`RpcObject` is an enhanced version of `RpcAgent`, and can be used to represent any object (unlike `RpcAgent` which can only represent `Agent`)
Users can call any public methods or get any attributes on the `RpcObject`, and the `RpcObject` will forward the request to the real object running in the RPC server.
The original `RpcAgent` is removed. 

#### Add `@async_func` and `AsyncResult` to replace the `reply` and `Placeholder`

Because the `RpcObject` supports calling any public methods and returns any results, the old `Placeholder` cannot cover this complex scenario. Therefore, `@async_func` and `AsyncResult` are added.

Functions decorated with `@async_func` will automatically return `AsyncResult` objects in distributed mode.
The usage of `AsyncResult` is the same as `Placeholder`, which is instantiated only when its internal objects are accessed.
`AsyncResult` can be used to represent any type of data, not just `Msg`.
The original `PlaceholderMessage` is removed.

Through the above modification, the `agent` and `message` modules are decoupled from the distributed mode.
Note that **the user interface of the distributed mode is unchanged**, all modifications above are completely hidden from the user, and all previous distributed examples can be run without any modification.

### Add Guess Two Third of the Average Game example

The code of "Very Large-Scale Multi-Agent Simulation in AgentScope" is released under `examples/paper_large_scale_simulation`

## Checklist


DavdGao (commented):
1. Move the environment modules into examples like what rag and react did before, and we will merge it into the library when it's ready.
2. Will it enter an infinite loop? For example, agentA and agentB subscribe to the locations of each other and trigger each other, leading to an endless actions cycle.

agentA move ==> agentB listen ==> agent B move ==> agentA listen ==> agentB move ....

DavdGao (commented):
Please move the environemnt module into example first.

pan-x-c:
Because the gRPC server relies on the `RpcEnv`,`rpc_env.py` and the modules it depends on can't be moved to `examples`, all other modules have been moved to `examples/environments`

DavdGao:
> Because the gRPC server relies on the `RpcEnv`,`rpc_env.py` and the modules it depends on can't be moved to `examples`, all other modules have been moved to `examples/environments`

The current implementation is too complex, and we didn't decide on the final implementation of this environment module. 

However, in this PR, the env module is actually embedded in the main library. If we can decouple it from the main library, we should consider to create a branch to maintain the simulation related code.

examples/paper_large_scale_simulation/scripts/stop_vllm.sh
garyzhang99, line 1: This file is empty.

examples/paper_large_scale_simulation/scripts/start_vllm.sh
```diff
@@ -0,0 +1,20 @@
...
+# default values
+gpu_num=8
+gpu_per_model=1
+model_path=<your_model_path>
+base_port=8010
+
+touch .vllm_pid
+mkdir -p log
+
+for ((i=0; i < ${gpu_num}; i=i+{gpu_per_model})); do
```
garyzhang99, line 12: need a `$` here

docs/sphinx_doc/en/source/tutorial/208-distribute.md
```diff
@@ -2,426 +2,467 @@
...
+    def __init__(self, name):
+        super().__init__(name)
+
+    def get_answer(self, url: str, query: str):
+        """Simulate crawling the web and looking for answers"""
+        time.sleep(5)
+        return f"Answer from {self.name}"
+
+    def reply(self, x: dict = None) -> dict:
+        return Msg(
```
qbc2016, line 63: "role" is missing.

examples/environments/auction_simulation/listeners.py
```diff
@@ -0,0 +1,115 @@
+# -*- coding: utf-8 -*-
+"""Listerners for the auction simulation."""
```
qbc2016, line 2: typo "Listerners"

examples/environments/auction_simulation/env.py
```diff
@@ -0,0 +1,151 @@
...
+        if self.cur_bid_info is None:
+            self.fail()
+        else:
+            self.sold()
+
+    @event_func
+    def bid(self, bidder_name: str, item: Item, bid: int) -> bool:
+        """Bid for the auction.
+        Args:
+            bidder (`str`): The name of the bidder.
```
qbc2016, line 109: typo "bidder" -> "bidder_name"

src/agentscope/memory/temporary_memory.py
```diff
@@ -73,16 +73,14 @@ def add(
         else:
             record_memories = memories
 
+        # FIXME: a single message may be inserted multiple times
```
xieyxclack, line 76: shall we fix it in this pr?

src/agentscope/message/msg.py
```diff
@@ -228,6 +228,18 @@ def formatted_str(self, colored: bool = False) -> str:
                 colored_strs.append(f"{name}: {self.url}")
         return "\n".join(colored_strs)
 
+    def __eq__(self, value: object) -> bool:
```
xieyxclack, line 231: What if additional attributes are added to the message in the future?

src/agentscope/rpc/rpc_meta.py
```diff
@@ -0,0 +1,218 @@
...
+        func (`Callable`): The function to decorate.
+    """
+    func._is_sync = True  # pylint: disable=W0212
+    return func
+
+
+# TODO: add stream function decorator `stream_func`
+
+
+def generate_oid() -> str:
```
xieyxclack, line 50: A minor question: why is it named `oid`?

src/agentscope/server/launcher.py
```diff
@@ -135,13 +159,24 @@ async def _setup_agent_server_async(  # pylint: disable=R0912
...
-            Maximum time for reply messages to be cached in the server.
+        max_expire_time (`int`, defaults to `7200`):
+            Maximum time for async results to be cached in the server.
             Note that expired messages will be deleted.
+        max_timeout_seconds (`int`, defaults to `5`):
+            The maximum time (in seconds) that the server will wait for
+            the result of an async call.
         studio_url (`str`, defaults to `None`):
             URL of the AgentScope Studio.
         custom_agent_classes (`list`, defaults to `None`):
```
xieyxclack, line 182: `custom_agent_classes` -> `custom_classes`

xieyxclack (commented):
LGTM. Please refer to inline comments

pan-x-c:
To be discussed
1. the name of distributed functions. In the current version,  `async_func/sync_func` may be confused with `async` in Python `asyncio`. `unblocking_func / blocking_func` may be a better name.
2. the old  `agent_id` is renamed to `oid` (a short version of object_id), which may cause misunderstandings and incompatibility with legacy code. However, here we really need an `id` field for all objects, not just agents.

src/agentscope/memory/temporary_memory.py
pan-x-c: Fix it in another PR may be better

DavdGao (commented):
Some of my concerns in this PR:

1. The planned deprecation of placeholder may be too extensive. Consider this alternative: create a Mixin class that blocks certain remote function calls (e.g., a configurable reply). This way, the message class (and the env modules) can inherit from it, maintaining the functionality of placeholders without causing confusion in return types from an agent (`AsyncResult` or `Msg`). 

```python
class RemoteMixin:
    
    _enable: bool = False

    __remote_attrs__ = []

    def __getattr__(self, name):
        if _enable and name in self.__remote_attrs__:
            # ...
        else:
            super().__getattr__(name)

class Msg(RemoteMixin):
    __remote_attrs__ = ['reply', 'observe', 'xxx']
    """Functions/attrs need to be executed/obtained from remote."""

    # ...
```

2. If feasible, try not to use metaclasses for functionality implementation. They can be challenging for developers to understand.

3. I believe the agent should be the framework's core. Thus, the design of the environment module should align with the agent's settings, not vice versa. For instance, the agent's identity should be independent of the environment and distribution modules.

src/agentscope/message/msg.py
pan-x-c: If there are new attributes, this method must be modified at the same time

src/agentscope/rpc/rpc_meta.py
pan-x-c: short term of `object_id`

examples/paper_large_scale_simulation/README.md
```diff
@@ -0,0 +1,123 @@
...
+Moderator: The average value of round 3 is 48.32 [takes 1.134 s]
+Moderator: Save result to ./result/studio/random/1-2-100-1-0.667/2024-08-13-07:35:44
+```
+
+## References
+
+```
+@article{agentscope_simulation,
+      title={Very Large-Scale Multi-Agent Simulation in AgentScope},
+      author={Xuchen Pan and
```
xieyxclack, line 111: Please update the author list

src/agentscope/environment/env.py
```diff
@@ -0,0 +1,398 @@
...
+        return self.parent
+
+    def set_parent(self, parent: Env) -> None:
+        """Set the parent env of the current env.
+
+        Args:
+            parent (`Env`): The parent env.
+        """
+        if self.parent is not None:
+            self.parent.remove_child(self.name)
```
qbc2016, line 275: why "remove_child" here?

pan-x-c: To keep the environments in a tree structure. Without `remove_child` here, the previous parent env can access the child node even after `set_parent`.

qbc2016: aha, i was wrong about `parent` and  `self.parent`

xieyxclack (approved):

PR #379, opened 2024-08-02
Fix workstation react agent bug

qbc2016:
## Description

1. modify react agent and services in `workflow_node.py`
2. support partial for add function in ServiceToolkit

## Checklist

- [ ]  Code has passed all tests
- [ ]  Docstrings have been added/updated in Google Style


src/agentscope/service/service_toolkit.py
```diff
@@ -195,6 +195,10 @@ def bing_search(query: str, api_key: str, num_results=10):
 
         """
 
+        if isinstance(service_func, partial):
```
DavdGao, line 198: Why do we need to consider the partial function here?

qbc2016: For example, api_key should be passed in bing_serach.
I am not sure if there is another way to implement the following in `workflow_node.py`
 `service_toolkit.add(bing_search, api_key=BING_API_KEY, num_results=3)`

qbc2016: add TODO here

DavdGao (approved):
lgtm

PR #388, opened 2024-08-06
Support to serialize message objects in AgentScope and remove unused arguments.

DavdGao:
### About Serialization
1. Add a new serialize and deserialize function in AgentScope, which is responsible for serialization and deserialization in our library in the future;

### About Message **[IMPORTANT CHANGE]**
1. `Msg` class doesn't inherit `dict` any more. 
2. Remove `timestamp` and `name` variables from local attributes of placeholder class, since they are actually decided by the distributed agent. 
3. Remove `MessageBase` class

### Others
1. Remove unused parameter `memory_config` from the memory module; 
2. Add test for message class
3. Modify unit tests accordingly

## Note

In this PR, we only serialize and deserialize the message objects, leaving agent serialization in the future to keep this PR compact and simple.

## Checklist


src/agentscope/message/msg.py
```diff
@@ -1,168 +1,197 @@
...
 
-from ..utils.tools import _get_timestamp, _map_string_to_color_mark
+from ..serialize import is_serializable
+from ..utils.tools import (
+    _map_string_to_color_mark,
+    _get_timestamp,
+)
+
 
+class Msg:
```
ZiTao-Li, line 22: My concern is that making the `Msg`  no longer inheriting from `dict` in one PR may be too radical. Some existing implementations may be using `msg.get("content")` (like the [mention example](https://github.com/modelscope/agentscope/blob/21161fe9985ee2a2f617180b00a1424b81302d42/examples/conversation_with_mentions/main.py#L64)) and some developers may be used to this.

src/agentscope/agents/dialog_agent.py
```diff
@@ -16,7 +16,6 @@ def __init__(
         sys_prompt: str,
         model_config_name: str,
         use_memory: bool = True,
-        memory_config: Optional[dict] = None,
```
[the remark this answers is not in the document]
DavdGao: Done by adding kwargs accordingly.

src/agentscope/message/msg.py
```diff
@@ -1,168 +1,197 @@
...
+    the information of a message, including
 
-class MessageBase(dict):
-    """Base Message class, which is used to maintain information for dialog,
-    memory and used to construct prompt.
+    - id:           the identity of the message
+    - name:         who sends the message
+    - content:      the message content
+    - role:         the sender role chosen from 'system', 'user', 'assistant'
+    - url:          the url
```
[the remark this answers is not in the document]
DavdGao: Added more description.

src/agentscope/message/msg.py
DavdGao: Fix errors in the examples, and add a temporary solution, where Msg class supports `__getitem__` method with a warning.

src/agentscope/serialize.py
pan-x-c, line 1: Some concerns for the current serialization module:

1. `json` is not a good choice to serialize Python objects, `pickle` is better.
    1. need more code: users have to write serialization and deserialization methods for each customized class, not user-friendly.
    2. slow and bulky:  although `json` output is human readable, the output contains redundant characters, which increases the space usage and cost of network transmission.

2. The current version of `serialize` is more like `to_string`, `__string__` or `__repr__`, trying to interpret objects in a human-readable way. However, this goal can be achieved by recommending users to implement those methods and provide a default implementation like calling `__dict__`. And `serialize`/`deserialize` can be implemented by `pickle`, which is more efficient and compact, and need no extra code.

3. `Msg` is one of the most basic concepts in AgentScope. Its use frequency is even higher than `Agent`. Changing the interface of `Msg` even with a deprecated warning will cause a lot of code changes, which is extremely unfriendly from a user perspective.

src/agentscope/agents/dialog_agent.py
```diff
@@ -31,15 +33,16 @@ def __init__(
...
         super().__init__(
             name=name,
             sys_prompt=sys_prompt,
             model_config_name=model_config_name,
             use_memory=use_memory,
-            memory_config=memory_config,
+        )
+
+        logger.warning(
+            f"Unused keyword arguments is provided: {kwargs}",
```
qbc2016, line 45: should we check if kwargs exists here? Otherwise, it will always log out
"Unused keyword arguments is provided: {}"

src/agentscope/serialize.py
DavdGao: 1. The target of serialization is to support saving/exporting running status of applications/agents/messages, where the readability is more important.
2. The function `serialize` is renamed into `to_dict` in this PR.
4. Sooner or later, the `Msg` class will no longer inherit `dict`, and I think it's a good time to make this decision in our next version, `v1.0.0`.

src/agentscope/agents/dialog_agent.py
DavdGao: Changed in the new commit.

PR #505, opened 2025-01-04
Add support for Anthropic chat API in both streaming and non-streaming mode

DavdGao:
## Description

Add support for Anthropic chat API in both streaming and non-streaming mode

Already tested.

## Checklist


src/agentscope/models/post_model.py
```diff
@@ -76,13 +80,14 @@ def __init__(
...
-                json_args.get("model_name", None),
-            )
-        else:
-            model_name = None
+        if model_name is None:
+            if json_args is not None:
+                model_name = json_args.get(
+                    "model",
+                    json_args.get("model_name", None),
+                )
```
ZiTao-Li, line 88: Maybe we can give an warning and assign a default string for `model_name` here, because many of those self-host api may not need model name. Besides, many examples in `examples/model_configs_template/postapi_model_config_template.json` have no `model` or `model_name`

ZiTao-Li (commented):
Please refer to the inline comments

tests/model_test.py
```diff
@@ -26,6 +26,8 @@ def format(
         self,
         *args: Union[Msg, Sequence[Msg]],
     ) -> Union[List[dict], str]:
+        """Format the input for the model"""
+        print(*args)
```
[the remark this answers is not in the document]
DavdGao: This printing is to avoid pylint error report, since *args is not used actually.

src/agentscope/models/post_model.py
DavdGao: 1. Update the current model configurations by adding the `model_name` field.
2. When calling `super().__init__(config_name=config_name, model_name=model_name)`, if `model_name` is `None`, a ValueError  will br raised.

pan-x-c (approved):
LGTM

PR #534, opened 2025-02-16
Add support for deepseek 

PengHongyiNTU:
---
Add support for deepseek official api
---
## Description
This PR adds support for the Deepseek API. Since Deepseek uses an OpenAI-compatible API, the wrapper is almost the same as the OpenAI one. The key difference lies in the yield generator, which now needs to access both the `content` and `reasoning_content` fields from the response.

## issues
1. The Deepseek official URL is often unresponsive. For testing and debugging, use the Aliyun-hosted version as a fallback.

2. The streaming handling in logging.py causes bugs when displaying flags to indicate whether reasoning and final answer generation have started. Consider a better implementation.

3. Streaming content can be a bit chunky because responses arrive faster than they are consumed. A better implementation (e.g., buffering or fixed-rate throttling) is needed to handle streaming more effectively.

### DIFF
--- a/src/agentscope/models/_model_utils.py
+++ b/src/agentscope/models/_model_utils.py
@@ -27,7 +27,7 @@ def _verify_text_content_in_openai_delta_response(response: dict) -> bool:
 
 
 def _verify_text_content_in_openai_message_response(response: dict) -> bool:
-    """Verify if the text content exists in the openai streaming response
+    """Verify if the text content exists in the openai message response
 
     Args:
         response (`dict`):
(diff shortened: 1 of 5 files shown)


src/agentscope/models/deepseek_model.py
```diff
@@ -0,0 +1,320 @@
...
+
+class DeepSeekChatWrapper(DeepSeekWrapperBase):
+    """The model wrapper for DeepSeek Chat API."""
+
+    model_type: str = "deepseek_chat"
+    # Recommend use aliyun hosted deepseek
+    # then the model name is deepseek-r1 and deepseek-v3
+
+    # supported_models: list[str] = [
+    #     "deepseek-chat",
```
DavdGao, line 107: Since we don't hope developers to modify this file, it's suggested to add `deepseek-r1` and `deepseek-v3` into the supported models.

src/agentscope/models/deepseek_model.py
```diff
@@ -0,0 +1,320 @@
...
+    def __init__(
+        self,
+        config_name: str,
+        model_name: Optional[str] = None,
+        api_key: Optional[str] = None,
+        stream: bool = True,
+        generate_args: Optional[dict] = None,
+        **kwargs: Any,
+    ) -> None:
+        """Initalize the openai client.
```
DavdGao, line 126: Change the docstrings for deepseek.

src/agentscope/models/deepseek_model.py
```diff
@@ -0,0 +1,320 @@
...
+        """
+        self._save_model_invocation(
+            arguments=kwargs,
+            response=response,
+        )
+        usage = response.get("usage", None)
+        if usage is not None:
+            self.monitor.update_text_and_embedding_tokens(
+                model_name=self.model_name,
+                prompt_tokens=usage.get("prompt_tokens", 0),
```
DavdGao, line 318: What about reasoning tokens?

DavdGao (commented):
Please see inline comments and 

1. Add unittests for deepseek model wrapper.
2. Update doc, including tutorial and README.md

src/agentscope/models/_model_utils.py
```diff
@@ -48,3 +48,61 @@ def _verify_text_content_in_openai_message_response(response: dict) -> bool:
         return False
 
     return True
+
+
+def _verify_reasoning_content_in_openai_message_response(
```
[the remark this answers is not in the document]
PengHongyiNTU: Deepseek only, have moved this to the deepseek_model.py

src/agentscope/models/deepseek_model.py
```diff
@@ -0,0 +1,320 @@
...
+)
+
+
+class DeepSeekWrapperBase(ModelWrapperBase, ABC):
+    """The model wrapper for DeepSeek API."""
+
+    def __init__(
+        self,
+        config_name: str,
+        model_name: Optional[str] = None,
```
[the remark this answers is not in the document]
PengHongyiNTU: I follow the implementations of other models and it seems this model_name can be None. For example 
```python
class OpenAIWrapperBase(ModelWrapperBase, ABC):
  def __init__(
          self,
          config_name: str,
          model_name: str = None,
          api_key: str = None,
          organization: str = None,
          client_args: dict = None,
          generate_args: dict = None,
          **kwargs: Any,
      ) -> None:

```
and 
```python
 if model_name is None:
            model_name = config_name
            logger.warning("model_name is not set, use config_name instead.")
```

src/agentscope/models/deepseek_model.py
PengHongyiNTU: The API does not provide a separate breakdown of reasoning token usage. Instead, it returns the total aggregated token usage.

PR #541, opened 2025-02-26
Add dashscope application call

qbc2016:
## Description

As the title says.

## Checklist

- [ ]  Code has passed all tests
- [ ]  Docstrings have been added/updated in Google Style

### DIFF
--- a/src/agentscope/models/dashscope_model.py
+++ b/src/agentscope/models/dashscope_model.py
@@ -4,7 +4,7 @@
 import os
 from abc import ABC
 from http import HTTPStatus
-from typing import Any, Union, List, Sequence, Optional, Generator
+from typing import Any, Union, List, Sequence, Optional, Generator, Literal
 
 from loguru import logger
 
@@ -125,7 +125,8 @@ class DashScopeChatWrapper(DashScopeWrapperBase):
         self,
         config_name: str,
         model_name: str = None,
-        api_type: str = "Generation",
+        app_id: str = None,
+        api_type: Literal["generation", "application"] = "generation",
         api_key: str = None,
         stream: bool = False,
         generate_args: dict = None,
@@ -158,6 +159,7 @@ class DashScopeChatWrapper(DashScopeWrapperBase):
 
         self.stream = stream
         self.api_type = api_type.lower()
+        self.app_id = app_id
 
     def __call__(
         self,
@@ -238,11 +240,12 @@ class DashScopeChatWrapper(DashScopeWrapperBase):
                 # Set the result to be "message" format.
                 "result_format": "message",
                 "stream": stream,
-                "app_id"
-                if self.api_type == "application"
-                else "model": self.model_name,
             },
         )
+        if self.api_type == "application":
+            kwargs["app_id"] = self.app_id
+        else:
+            kwargs["model"] = self.model_name
 
         # Switch to the incremental_output mode
         if stream:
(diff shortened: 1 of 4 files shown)


rayrayraykk:
Please add a test in `TestDashScopeChatWrapper` and be careful with the dashscope version. Since the built bailian workflow handles `query` and `historyList` separately, we should not use the `model.format` to change the model input.

The other part looks good to me.

src/agentscope/models/dashscope_model.py
```diff
@@ -228,21 +231,28 @@ def __call__(
...
         kwargs.update(
             {
-                "model": self.model_name,
                 "messages": messages,
                 # Set the result to be "message" format.
                 "result_format": "message",
                 "stream": stream,
+                "app_id"
+                if self.api_type == "application"
+                else "model": self.model_name,
```
DavdGao, line 243: What's the value of `self.model_name` when `self.api_type == "application"`?

src/agentscope/models/dashscope_model.py
```diff
@@ -124,6 +125,7 @@ def __init__(
         self,
         config_name: str,
         model_name: str = None,
+        api_type: str = "Generation",
```
DavdGao, line 128: Use `typing.Literal` to list all candidates here.

src/agentscope/models/dashscope_model.py
```diff
@@ -294,9 +313,13 @@ def generator() -> Generator[str, None, None]:
                 kwargs,
                 response,
             )
+            if self.api_type == "application":
```
DavdGao, line 316: Similar as above, consider to use enum for `api_type`

DavdGao (commented):
Does application api support streaming mode here?

Others please see inline comments

src/agentscope/models/dashscope_model.py
DavdGao: Is `app_name` or `application_name` more appropriate when `api_type == "application"`? If so, consider to provide a new wrapper or add a new argument rather than directly using the argument `model_name`

qbc2016: it does not need `model_name`, `app_id` determines which application to call.

qbc2016: > Is `app_name` or `application_name` more appropriate when `api_type == "application"`? If so, consider to provide a new wrapper or add a new argument rather than directly using the argument `model_name`

ok, I'll change it to `app_id`.

src/agentscope/models/dashscope_model.py
qbc2016: Now, only `application` is used, maybe w'll use enum later when more APIs are supported.

qbc2016:
> Does application api support streaming mode here?
> 
> Others please see inline comments

yes, it supports streaming mode, and its activation is compatible with the parameter `"stream": True` in `model_configs`, I have modified the `generation()` function to support it.

src/agentscope/models/dashscope_model.py
```diff
@@ -124,6 +125,8 @@ def __init__(
         self,
         config_name: str,
         model_name: str = None,
+        app_id: str = None,
+        api_type: Literal["generation", "application"] = "generation",
```
DavdGao, line 129: Update the docstrings so that developers can know how to use this class as a model or application wrapper.

src/agentscope/models/dashscope_model.py
```diff
@@ -398,6 +428,32 @@ def format(
             `List[dict]`:
                 The formatted messages.
         """
+        if self.api_type == "application":
```
DavdGao, line 431: What's the requirement of the application api? Is it the same with the generation API?

DavdGao (approved):
please see inline comments

src/agentscope/models/dashscope_model.py
qbc2016: It requires the parameter `app_id` instead of `model`. Besides,  there are only minor differences, such as not supporting the `plugins` parameter, etc.

qbc2016:
see #601

PR #563, opened 2025-03-26
[WIP] close loop to avoid ignored exceptions in resouce recycle

rayrayraykk:
As the title says


DavdGao (commented):
Plz add test case in unittest to verifiy the mcp tool calling

PR #695, opened 2025-08-26
[Lightweight Bugfix] Fix when init Msg class, error calling function logger.chat

sairupe:
… logger.info

## AgentScope Version

AgentScope Version : v0

## Description

[Console Error Log]
  File "/Users/syrianazh/Desktop/pyWorkSpace/agentscope/src/agentscope/message/msg.py", line 175, in __init__
    logger.chat(self)
    ^^^^^^^^^^^
AttributeError: 'Logger' object has no attribute 'chat'

[Path]
https://github.com/agentscope-ai/agentscope/blob/v0/src/agentscope/message/msg.py#L175

[Cause]
I wanted to do some MCP test with logger print,  When I Init a ReActAgent with config parameter [verbose:true] and registry a [MCP service_toolkit], and then i calling agent(x) speak to agent. The calling step will execute to https://github.com/agentscope-ai/agentscope/blob/v0/src/agentscope/agents/_react_agent.py#L157 , in _reasoning(self) fuctions will new init a Msg use ReActAgent verbose param value assign to Msg echo param. And then when calling Msg init then logger.chat throw The Error. May be a careless coding.

[fix]
change logger.chat to logger.info

## Checklist

- [ ]  All tests are passing

### DIFF
--- a/src/agentscope/message/msg.py
+++ b/src/agentscope/message/msg.py
@@ -172,7 +172,7 @@ class Msg(BaseModel):
         )
 
         if echo:
-            logger.chat(self)
+            logger.info(self)
 
     def to_dict(self) -> dict:
         """Serialize the message into a dictionary, which can be
--- a/src/agentscope/message/msg.py
+++ b/src/agentscope/message/msg.py
@@ -172,7 +172,10 @@ class Msg(BaseModel):
         )
 
         if echo:
-            logger.info(self)
+            if hasattr(logger, "chat"):
+                logger.chat(self)
+            else:
+                logger.info(self)
 
     def to_dict(self) -> dict:
         """Serialize the message into a dictionary, which can be
--- a/tests/logger_test.py
+++ b/tests/logger_test.py
@@ -8,12 +8,13 @@ import unittest
 import datetime
 import uuid
 
+from unittest.mock import patch
 from loguru import logger
 
 from agentscope.logging import setup_logger
 from agentscope.manager import ASManager
 from agentscope.message import Msg
-from unittest.mock import patch
+
 
 
 class LoggerTest(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


src/agentscope/message/msg.py
```diff
@@ -172,7 +172,7 @@ def __init__(  # pylint: disable=too-many-branches
         )
 
         if echo:
-            logger.chat(self)
+            logger.info(self)
```
DavdGao: Consider adding a conditional check here?
```python
if echo:
    if hasattr(logger, "chat"):
        logger.chat(self)
    else:
        logger.info(self)
```

DavdGao (commented):
Please see inline comments, others look good to me

src/agentscope/message/msg.py
sairupe: The class Logger was import from loguru, an thrid-part library, a commonly used logging component, and the class isn't exist the 'chat' def fuction. Usually logger fuction name more likely use debug, info, error to discribe logging level. Will this way of writing cause confusion to others? My understanding is that your suggestions just for compatibility changes?

sairupe: OK, I found same code using in logging.py , there is assign log_msg refference to logger.chat, unified with studio and gradio. I will submit a new commit.
https://github.com/agentscope-ai/agentscope/blob/v0/src/agentscope/logging.py#L187

DavdGao: > The class Logger was import from loguru, an thrid-part library, a commonly used logging component, and the class isn't exist the 'chat' def fuction. Usually logger fuction name more likely use debug, info, error to discribe logging level. Will this way of writing cause confusion to others? My understanding is that your suggestions just for compatibility changes?

In version 0, the chat interface is implemented as a monkey patch to the loguru logger, automatically applied during ``agentscope.init`` initialization to handle the agent printing logic.

sairupe: I submit a new commit, fix with your opinion, add a condition check. And add some test code to coverage .Please check.

PR #808, opened 2025-09-24
fix deep research agent example condition

0xStorageRoot:
## AgentScope Version

1.0.4

## Description

from the deeper expansion's prompt I found instruction that:
1. **Find information with valuable, but insufficient or shallow content**: Carefully review the web search results to assess whether there is any snippet or web content that
    - could potentially help address checklist items or fulfill knowledge gaps of the task as the content increases
    - **but whose content is limited or only briefly mentioned**!
2. **Identify the snippet**: If such information is found, set `need_more_information` to true, and locate the specific **title, content, and url** of the information snippet you have found for later extraction.

While in the _follow_up method after calling llm with expansion prompt ,it only extracts more when the "need_more_information" is false, in opposite return a "sufficient_hint" response. 

## Checklist

- [ ]  All tests are passing

### DIFF
--- a/examples/agent_deep_research/deep_research_agent.py
+++ b/examples/agent_deep_research/deep_research_agent.py
@@ -599,7 +599,7 @@ class DeepResearchAgent(ReActAgent):
                 follow_up_subtask = {}
 
             #  Step #2: extract the url
-            if follow_up_subtask.get("need_more_information", False):
+            if follow_up_subtask.get("need_more_information", True):
                 expansion_response_msg = Msg(
                     "assistant",
                     follow_up_subtask.get(


Luohh5:
Thanks for your PR! I think there may be some misunderstandings about the code follow_up_subtask.get("need_more_information", False), which means retrieve the value of the key "need_more_information" from the dictionary follow_up_subtask, returning False if that key is missing. This default of False is intentional because the normal flow assumes we don’t need more information unless the flag is explicitly set. Changing it to True would treat every missing key as if more information were required, which could trigger unnecessary follow-ups and alter the logic in unexpected ways.

0xStorageRoot:
Got it,thanks for response !

PR #880, opened 2025-10-30
fix(agent): separate the logic of the arguments `enable_meta_tool` and `plan_notebook`

ceshihao:
## AgentScope Version

## Description

When enable_meta_tool is True, the plan tool group activation is controlled by agent.
Otherwise, the plan tool group is always active.

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/agent/_react_agent.py
+++ b/src/agentscope/agent/_react_agent.py
@@ -225,20 +225,24 @@ class ReActAgent(ReActAgentBase):
         self.plan_notebook = None
         if plan_notebook:
             self.plan_notebook = plan_notebook
-            # When enable_meta_tool is True, the plan tool group activation is
-            # controlled by agent.
-            # Otherwise, the plan tool group is always active.
-            tool_group_active = not enable_meta_tool
-            self.toolkit.create_tool_group(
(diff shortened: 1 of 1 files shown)


src/agentscope/agent/_react_agent.py
```diff
@@ -225,9 +225,14 @@ def __init__(
         self.plan_notebook = None
         if plan_notebook:
             self.plan_notebook = plan_notebook
+            # When enable_meta_tool is True, the plan tool group activation is
+            # controlled by agent.
+            # Otherwise, the plan tool group is always active.
+            tool_group_active = not enable_meta_tool
             self.toolkit.create_tool_group(
```
DavdGao: Thank you for your contribution. However, there appears to be a misunderstanding. When enable_meta_tool is set to False, tools should be added directly via register_tool_function (registering them into the "default" group), rather than creating a separate tool group.

ceshihao: Done.

Register plan tools in basic group if enable_meta_tool is not True

DavdGao (approved):
Thanks for your contribution, and it looks good to me.

PR #941, opened 2025-11-18
fix(toolkit): merge $defs when extending tool function schema with nested models

qbc2016:
## AgentScope Version

1.0.8dev

## Description

1. Merge $defs when extending tool function schema with nested models
2. Add unittest

## Checklist

- [ ]  All tests are passing

### DIFF
--- a/src/agentscope/tool/_registered_tool_function.py
+++ b/src/agentscope/tool/_registered_tool_function.py
@@ -93,18 +93,19 @@ class RegisteredToolFunction:
 
             # Check for conflicts and merge $defs
             for def_key, def_value in extended_schema["$defs"].items():
+                def_value_copy = deepcopy(def_value)
+                _remove_title_field(
+                    def_value_copy,
+                )  # pylint: disable=protected-access
+
                 if def_key in merged_params["$defs"]:
                     # Check if the two definitions are from the same BaseModel
                     # by comparing their content
                     # Create copies and remove title fields for comparison
-                    def_value_copy = deepcopy(def_value)
+
                     existing_def_copy = deepcopy(
                         merged_params["$defs"][def_key],
                     )
-
-                    _remove_title_field(
-                        def_value_copy,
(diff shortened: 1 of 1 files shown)


tests/toolkit_test.py
```diff
@@ -287,6 +301,81 @@ async def test_basic_functionalities(self) -> None:
                 chunk,
             )
 
+    async def test_extended_model_with_nested_models(self) -> None:
```
DavdGao: Please add a more conflict example: where the original python function has a nested BaseModel argument, and then we want to extended it by a new BaseModel class

```python
class MyBaseModel1(BaseModel):
    c: int

class MyBaseModel2(BaseModel):
    b: list[MyBaseModel1]

def func(a: MyBaseModel2):
    pass
```

And what about extending the `func` function with the class `MyBaseModel2`? It seems reasonable to reuse the same base model repeatedly.

src/agentscope/tool/_registered_tool_function.py
```diff
@@ -83,4 +84,25 @@ def extended_json_schema(self) -> dict:
...
+        # Merge $defs from extended schema to support nested models
+        if "$defs" in extended_schema:
+            merged_params = merged_schema["function"]["parameters"]
+            if "$defs" not in merged_params:
+                merged_params["$defs"] = {}
+
+            # Check for conflicts and merge $defs
+            for def_key, def_value in extended_schema["$defs"].items():
+                if def_key in merged_params["$defs"]:
+                    raise ValueError(
```
DavdGao: We shouldn't raise an error unconditionally here. It's valid for both the original function and the extended model to use the same BaseModel class.

Instead, we should first verify whether the two dictionaries originate from the same BaseModel instance. If they do, we should skip the validation. Otherwise, we should raise the error.

DavdGao:
Plz merge the main branch to pass the unittests

PR #986, opened 2025-12-01
fix(model): unfiorm kwargs in ollama and dashscope model

qbc2016:
## AgentScope Version

1.0.9dev

## Description

As the title says.

## Checklist

- [ ]  All tests are passing

### DIFF
--- a/src/agentscope/model/_ollama_model.py
+++ b/src/agentscope/model/_ollama_model.py
@@ -40,39 +40,40 @@ class OllamaChatModel(ChatModelBase):
         keep_alive: str = "5m",
         enable_thinking: bool | None = None,
         host: str | None = None,
-        client_kwargs: dict | None = None,
+        client_kwargs: dict[str, JSONSerializableObject] | None = None,
         generate_kwargs: dict[str, JSONSerializableObject] | None = None,
         **kwargs: Any,
     ) -> None:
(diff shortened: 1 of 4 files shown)


src/agentscope/model/_dashscope_model.py
```diff
@@ -55,6 +55,7 @@ def __init__(
         enable_thinking: bool | None = None,
         generate_kwargs: dict[str, JSONSerializableObject] | None = None,
         base_http_api_url: str | None = None,
+        **_kwargs: Any,
```
DavdGao, line 58: Why do we need the prefix "_"?

src/agentscope/model/_ollama_model.py
```diff
@@ -40,6 +40,8 @@ def __init__(
         keep_alive: str = "5m",
         enable_thinking: bool | None = None,
         host: str | None = None,
+        client_kwargs: dict | None = None,
```
DavdGao: Keep the typing of `client_kwargs` consistent with `generate_kwargs`

src/agentscope/model/_dashscope_model.py
qbc2016: Because the parameter is not used within the function. Without the underscore, pre-commit checks (like flake8) would raise a warning about unused parameters.

PR #1002, opened 2025-12-04
fix(repair json): hotfix for _json_loads_with_repair

qbc2016:
## AgentScope Version

1.0.9dev

## Description

1. Force dictionary return type for the function.
2. In case of non-dictionary output, truncate the preceding content and initiate retry logic to ensure a valid dictionary response.

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/_utils/_common.py
+++ b/src/agentscope/_utils/_common.py
@@ -37,6 +37,15 @@ def _json_loads_with_repair(
         This function is currently only used for parsing the streaming output
         of the argument field in `tool_use`, so the parsed result must be a
         dict.
+
+    Args:
+        json_str (`str`):
+            The JSON string to parse, which may be incomplete or malformed.
+
+    Returns:
+        `dict`:
+            A dictionary parsed from the JSON string after repair attempts.
+            Returns an empty dict if all repair attempts fail.
     """
     try:
         repaired = repair_json(json_str)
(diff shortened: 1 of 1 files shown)


src/agentscope/_utils/_common.py
```diff
@@ -29,23 +29,36 @@
...
         pass
 
-    try:
-        return json.loads(repaired)
-    except json.JSONDecodeError as e:
-        raise ValueError(
-            f"Failed to decode JSON string `{json_str}` after repairing it "
-            f"into `{repaired}`. Error: {e}",
-        ) from e
+    for i in range(len(json_str) - 1, 0, -1):
```
DavdGao, line 58: Please use the binary search method for the search.

DavdGao (commented):
plz see inline comments

src/agentscope/_utils/_common.py
qbc2016: This could be problematic because while the function might return a valid dictionary at length n, it may fail at length n+1. The binary search algorithm would then settle for a length less than n, which could lead to suboptimal or incorrect results.

DavdGao: Exactly, it looks good to me

src/agentscope/_utils/_common.py
```diff
@@ -29,23 +29,36 @@
...
+        try:
+            repaired = repair_json(json_str[:i])
+            result = json.loads(repaired)
+            if isinstance(result, dict):
+                return result
+        except Exception:
+            continue
+
+    raise ValueError(
```
DavdGao: Maybe print a warning and return an empty dict here?

PR #1024, opened 2025-12-10
raise exception on MCP disconnection

MengsD:
## AgentScope Version

## Description

[Please describe the background, purpose, changes made, and how to test this PR]

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/mcp/_stateful_client_base.py
+++ b/src/agentscope/mcp/_stateful_client_base.py
@@ -69,20 +69,30 @@ class StatefulClientBase(MCPClientBase, ABC):
             self.stack = None
             raise
 
-    async def close(self) -> None:
+    async def close(self, ignore_errors: bool = True) -> None:
         """Clean up the MCP client resources. You must call this method when
-        your application is done."""
+        your application is done.
+
+        Args:
+            ignore_errors (`bool`):
+                Whether to ignore errors during cleanup. Defaults to `True`.
+        """
         if not self.is_connected:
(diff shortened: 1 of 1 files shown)


src/agentscope/mcp/_stateful_client_base.py
```diff
@@ -83,6 +90,7 @@ async def close(self) -> None:
             logger.info("MCP client closed.")
         except Exception as e:
             logger.warning("Error during MCP client cleanup: %s", e)
+            raise e
```
DavdGao: The `except` statement seems meaningless if we raise the error here. 

So if we decide to raise it, perhaps we should remove the try except statements.

src/agentscope/mcp/_stateful_client_base.py
```diff
@@ -64,10 +64,17 @@ async def connect(self) -> None:
 
             self.is_connected = True
             logger.info("MCP client connected.")
-        except Exception:
-            await self.stack.aclose()
-            self.stack = None
-            raise
+        except Exception as e:
+            res_err = e
+            try:
```
DavdGao: (Similar as above) Perhaps remove the except statement in lines 71-72 since we decide to raise them

src/agentscope/mcp/_stateful_client_base.py
DavdGao, line 72: Maybe we should add an argument like `raise_exception: bool = False` or `ignore_exception: bool = True` here for compatibility

PR #1050, opened 2025-12-18
fix(gemini model): flatten json schema in tools for gemini chat model

qbc2016:
## AgentScope Version

1.0.10dev

## Description

Gemini API does not support `$def`, `$ref`, thus we should flatten the JSON schema, see #884.

## Checklist

- [x]  All tests are passing


src/agentscope/model/_gemini_model.py
DavdGao, line 487: Add note for this method about what we will do in this function

DavdGao (approved):

PR #1064, opened 2025-12-28
fix(model): Empty string output on GPT-5 truncation. Update: Ensure all OpenAI model responses return finish_reason consistently.

linuer:
## AgentScope Version
"1.0.10"

## Description
As the title says.

The result in AgentStudio after the fix.
as

## Checklist

- [ ]  All tests are passing

### DIFF
--- a/src/agentscope/message/_message_block.py
+++ b/src/agentscope/message/_message_block.py
@@ -14,6 +14,9 @@ class TextBlock(TypedDict, total=False):
     text: str
     """The text content"""
 
+    finish_reason: str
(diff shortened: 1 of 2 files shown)


DavdGao:
@linuer `TextBlock` is a critical foundational structure in AgentScope, so we must be extremely cautious when considering modifications to this data structure. 

Please further explain the necessity of modifying `TextBlock` here. If the goal is to expose the specific finish reason upstream for handling, please consider processing it at the model layer instead—for example, by throwing specific exception types based on the finish reason.

PR #1070, opened 2025-12-31
bugfix: parse anthropic response content block type logic error.

FFengIll:
## AgentScope Version

1.0.10 via `import agentscope; print(agentscope.__version__)`

## Description

This PR fixes type process error and inconsistent while parsing anthropic response which may return error result.

---

The buggy parser behave different for text block and may cause problem in many edge case,
ref to `agentscope.model._anthropic_model.AnthropicChatModel._parse_anthropic_stream_completion_response`

> Bad case example: anthropic return  blocks [text, tool_use], the parser return [text, text(empty)].

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/model/_anthropic_model.py
+++ b/src/agentscope/model/_anthropic_model.py
@@ -304,7 +304,10 @@ class AnthropicChatModel(ChatModelBase):
                     thinking_block["signature"] = content_block.signature
                     content_blocks.append(thinking_block)
 
-                elif hasattr(content_block, "text"):
+                elif (
+                    hasattr(content_block, "type")
+                    and content_block.type == "text"
+                ):
                     content_blocks.append(
                         TextBlock(
                             type="text",


qbc2016 (commented):
Thank you very much for your PR. I have a small question about the implementation. I'm trying to understand how this modification addresses the bad case mentioned in the PR description.

From my understanding, the logic for handling TextBlock and ToolUseBlock appears to remain the same before and after the change:

For TextBlock: both versions extract the text content
For ToolUseBlock: it wouldn't enter this branch in either case
Could you help clarify the connection? I might be missing something in my analysis.

FFengIll:
> Thank you very much for your PR. I have a small question about the implementation. I'm trying to understand how this modification addresses the bad case mentioned in the PR description.非常感谢你的公关。我有一个关于实现的小问题。我正在试图理解这个修改如何解决 PR 描述中提到的糟糕情况。
> 
> From my understanding, the logic for handling TextBlock and ToolUseBlock appears to remain the same before and after the change:据我了解，TextBlock 和 ToolUseBlock 的处理逻辑在更改前后似乎保持不变：
> 
> For TextBlock: both versions extract the text content对于 TextBlock：两个版本都提取文本内容 For ToolUseBlock: it wouldn't enter this branch in either case对于 ToolUseBlock：无论哪种情况，它都不会进入这个分支 Could you help clarify the connection? I might be missing something in my analysis.你能帮我澄清一下这个联系吗？我可能忽略了什么分析。

@qbc2016  Thanks for your response.

Yes, the logic appears identical when reviewing the old version statically

However, I encountered a runtime bug:
1. Old version: use tool enters as an empty text block.
2. New version: Uses strict determination to correctly render use tool as a tool block.

please check bellow debug snapshot for old version
- look at breakpoint
- current block is tool use
- but it entered into text block process

image


Furthermore, using new version - `type` for checking ensures more consistent, reliable, and safe behavior.

FFengIll:
With my experiment, I believe this bug is trigger by new anthropic module (agentscope use it for anthropic api).
In anthropic, pydantic and pythonic process make each block contains `text` attr, so we can not use `hasattr(x, 'text')` simply.
image

> This is a feature of pydantic

PR #1081, opened 2026-01-05
feat(memory): support database memory and memory compression

DavdGao:
## AgentScope Version

v1.0.11

## Description

- Add memory storage abstraction to support database for short-term memory
  - [x] in memory
  - [x] sqlalchemy
  - [x] redis

- [WIP] Support auto compression in the short-term memory module

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/memory/_working_memory/_in_memory_memory.py
+++ b/src/agentscope/memory/_working_memory/_in_memory_memory.py
@@ -119,7 +119,7 @@ class InMemoryMemory(MemoryBase):
             marks = []
         elif isinstance(marks, str):
             marks = [marks]
-        elif not isinstance(marks, list) and not all(
+        elif not isinstance(marks, list) or not all(
             isinstance(m, str) for m in marks
         ):
             raise TypeError(
--- a/src/agentscope/memory/_working_memory/_redis_memory.py
+++ b/src/agentscope/memory/_working_memory/_redis_memory.py
@@ -315,7 +315,7 @@ class RedisMemory(MemoryBase):
             # Record the message data
             await pipe.set(
                 self._get_message_key(m.id),
-                json.dumps(m.to_dict(), ensure_ascii=False, encodings="utf-8"),
+                json.dumps(m.to_dict(), ensure_ascii=False),
             )
 
             # Record the marks if provided
@@ -422,7 +422,7 @@ class RedisMemory(MemoryBase):
         return total_removed
 
     async def clear(self) -> None:
-        """Clear all messages belong to this section from the storage."""
+        """Clear all messages belong to this session from the storage."""
         msg_ids = await self._client.lrange(self._get_session_key(), 0, -1)
 
         # Get all mark keys using SCAN
--- a/src/agentscope/memory/_working_memory/_sqlalchemy_memory.py
+++ b/src/agentscope/memory/_working_memory/_sqlalchemy_memory.py
@@ -138,7 +138,8 @@ class AsyncSQLAlchemyMemory(MemoryBase):
         Raises:
             `ValueError`:
                 If the `engine` parameter is not an instance of
-                `sqlalchemy.ext.asyncio.AsyncEngine`.
+                `sqlalchemy.ext.asyncio.AsyncEngine` or `sqlalchemy.
+                ext.asyncio.AsyncSession`.
         """
         super().__init__()
 


src/agentscope/agent/_react_agent.py
```diff
@@ -45,6 +101,72 @@ class ReActAgent(ReActAgentBase):
...
+        """Allow arbitrary types in the pydantic model."""
+
+        enable: bool
+        """Whether to enable the auto compression feature."""
+
+        agent_token_counter: TokenCounterBase
+        """The token counter for the agent's model, which must be consistent
+        with the model used in the agent."""
+
+        trigger_threshold: int
```
Osier-Yi, line 120: In this configuration, compression appears to be triggered only when the token threshold is reached. I’m wondering whether users could customize the triggering condition—for example, to compress repetitive tool-use error messages even if the total token count hasn’t exceeded the limit.

src/agentscope/agent/_react_agent.py
```diff
@@ -37,6 +39,59 @@ class _QueryRewriteModel(BaseModel):
     )
 
 
+class SummarySchema(BaseModel):
```
rayrayraykk, line 43: `SummarySchema` and `MemoryMark` don’t seem to belong in the `ReActAgent` module.
They are memory-related data models and enums, so they should be moved to a dedicated `memory` or `schemas` module to keep this file focused on agent logic and avoid mixing concerns.

src/agentscope/agent/_react_agent.py
```diff
@@ -46,6 +101,72 @@ class ReActAgent(ReActAgentBase):
     - Structured output generation
     """
 
+    class CompressionConfig(BaseModel):
```
rayrayraykk, line 107: Same as above.

src/agentscope/memory/_working_memory/_base.py
```diff
@@ -0,0 +1,93 @@
...
+        """Update the compressed summary of the memory.
+
+        Args:
+            summary (`str`):
+                The new compressed summary.
+        """
+        self._compressed_summary = summary
+
+    @abstractmethod
+    async def add(
```
rayrayraykk, line 32: This change introduces a potential breaking change.
Previously, the second positional argument to `add()` was `allow_duplicates: bool`.
With the new signature, passing `MemoryMark.HINT` as a positional argument will be misinterpreted by older versions.

src/agentscope/agent/_react_agent.py
```diff
@@ -419,19 +544,17 @@ async def _reasoning(
             hint_msg = await self.plan_notebook.get_current_hint()
             if self.print_hint_msg and hint_msg:
                 await self.print(hint_msg)
-            await self._reasoning_hint_msgs.add(hint_msg)
+            await self.memory.add(hint_msg, MemoryMark.HINT)
```
rayrayraykk: To avoid silent misbehavior and improve readability, use a named argument for the new parameter (e.g., `mark=MemoryMark.HINT`).
Consider implementing backward-compatible handling in `add()` if older call styles need to be supported.

src/agentscope/agent/_react_agent.py
```diff
@@ -419,19 +544,17 @@ async def _reasoning(
...
             msgs=[
                 Msg("system", self.sys_prompt, "system"),
                 *await self.memory.get_memory(),
-                # The hint messages to guide the agent's behavior, maybe empty
-                *await self._reasoning_hint_msgs.get_memory(),
             ],
         )
         # Clear the hint messages after use
-        await self._reasoning_hint_msgs.clear()
+        await self.memory.delete_by_mark(MemoryMark.HINT)
```
rayrayraykk: Another breaking change here. Please see comment in memory module.

src/agentscope/memory/_working_memory/_base.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @abstractmethod
+    async def delete(
+        self,
+        msg_ids: list[str],
+        **kwargs: Any,
+    ) -> int:
+        """Delete items from the memory."""
+
+    async def delete_by_mark(
```
rayrayraykk, line 66: The new API call `delete_by_mark(MemoryMark.HINT)` is not compatible with the previous `delete(index)` method in `InMemoryMemory`.
This causes older code (and existing subclasses) to break if they don’t implement `delete_by_mark`.
To maintain backward compatibility, consider implementing `delete_by_mark` in `InMemoryMemory` as a wrapper that filters messages by mark and delegates to the existing `delete()` method.
Alternatively, accept `mark` as an optional keyword argument in `delete()` so both old and new call styles work:

```python
async def delete(self, index=None, *, mark=None):
    ...

```

src/agentscope/memory/_long_term_memory/_mem0/_mem0_long_term_memory.py
```diff
@@ -13,11 +13,13 @@
 
 from pydantic import field_validator
 
-from ..embedding import EmbeddingModelBase
-from ._long_term_memory_base import LongTermMemoryBase
-from ..message import Msg, TextBlock
-from ..model import ChatModelBase
-from ..tool import ToolResponse
+from agentscope.embedding import EmbeddingModelBase
```
rayrayraykk: Why not use relative import?

src/agentscope/memory/_long_term_memory/_mem0/_mem0_utils.py
```diff
@@ -13,8 +13,8 @@
 from mem0.llms.base import LLMBase
 
 
-from ..embedding import EmbeddingModelBase
-from ..model import ChatModelBase, ChatResponse
+from agentscope.embedding import EmbeddingModelBase
```
rayrayraykk: Same as above

src/agentscope/memory/_long_term_memory/_long_term_memory_base.py
```diff
@@ -3,9 +3,9 @@
 
 from typing import Any
 
-from ..message import Msg
-from ..module import StateModule
-from ..tool import ToolResponse
+from agentscope.message import Msg
```
rayrayraykk, line 6: Same as above.

src/agentscope/memory/_working_memory/_base.py
```diff
@@ -0,0 +1,93 @@
...
+"""The memory base class."""
+
+from abc import abstractmethod
+from typing import Any
+
+from ...message import Msg
+from ...module import StateModule
+
+
+class MemoryBase(StateModule):
```
rayrayraykk, line 11: Currently, the `Memory` classes (e.g., `InMemoryMemory`, `RedisStorageBase`) mix both **memory management algorithms** and **storage implementation** in a single class hierarchy.
This creates tight coupling between the logic (mark handling, summarization, filtering) and the persistence layer (in‑memory lists, Redis API calls).

I recommend refactoring to a `Memory` × `Storage` composition model:

- The `Memory` class would manage logic for message lifecycle, mark operations, summaries, etc.

- The `Storage` class would be an interchangeable persistence backend implementing basic CRUD for messages and marks (e.g., `InMemoryStorage`, `RedisStorage`).

- `Memory` would accept a `storage` instance via its constructor:

  ```python
  class Memory:
      def __init__(self, storage: StorageBase):
          self.storage = storage
  ```

  This allows swapping storage backends without changing the memory logic and makes testing easier.

src/agentscope/agent/_react_agent.py
```diff
@@ -864,3 +981,121 @@ async def _retrieve_from_knowledge(
...
+            logger.info(
+                "Memory compression is triggered (%d > "
+                "threshold %d) for agent %s.",
+                n_tokens,
+                self.compression_config.trigger_threshold,
+                self.name,
+            )
+
+            # Prepare the prompt used to compress the memories
+            compression_prompt = await self.formatter.format(
```
qbc2016: Should use `compression_formatter`?

DavdGao: fixed

src/agentscope/memory/_working_memory/_base.py
DavdGao: The algorithm logic is implemented in the `ReActAgent` class and the memory module only provides storage implementations

src/agentscope/memory/_long_term_memory/_long_term_memory_base.py
DavdGao: fixed

src/agentscope/memory/_long_term_memory/_mem0/_mem0_utils.py
DavdGao: fixed

src/agentscope/memory/_long_term_memory/_mem0/_mem0_long_term_memory.py
DavdGao: fixed

src/agentscope/agent/_react_agent.py
DavdGao: fixed

src/agentscope/memory/_working_memory/_base.py
DavdGao: add ``allow_duplicateds`` for the `InMemoryMemory` class

src/agentscope/agent/_react_agent.py
DavdGao: Actually, a single `CompressionConfig` class cannot accommodate all possible requirements. For highly specialized use cases, overriding the `_compress_memory_if_needed` method would be the recommended approach.

PR #1140, opened 2026-01-19
feat(session): add redis based session in agentscope

rayrayraykk:
## AgentScope Version

1.0.13dev

## Description

Add redis session 

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/session/_json_session.py
+++ b/src/agentscope/session/_json_session.py
@@ -46,7 +46,7 @@ class JSONSession(SessionBase):
     async def save_session_state(
         self,
         session_id: str,
-        user_id: str = "default_user",
+        user_id: str = "",
         **state_modules_mapping: StateModule,
     ) -> None:
         """Load the state dictionary from a JSON file.
@@ -54,7 +54,7 @@ class JSONSession(SessionBase):
         Args:
             session_id (`str`):
                 The session id.
-            user_id (`str`, default to `"default_user"`):
+            user_id (`str`, default to `""`):
                 The user ID for the storage.
             **state_modules_mapping (`dict[str, StateModule]`):
                 A dictionary mapping of state module names to their instances.
@@ -74,7 +74,7 @@ class JSONSession(SessionBase):
     async def load_session_state(
         self,
         session_id: str,
-        user_id: str = "default_user",
+        user_id: str = "",
         allow_not_exist: bool = True,
         **state_modules_mapping: StateModule,
     ) -> None:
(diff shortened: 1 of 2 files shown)


rayrayraykk:
Should I consider below features this PR? @DavdGao 

**Version rollback support**
`RedisSession` stores the whole session state as a single JSON blob under one Redis key (`{key_prefix}:session:{session_id}:state`). Each `save_session_state()` overwrites the previous value, so it does **not** support versioning / rollback (no `round_id` or history). If rollback is a requirement, consider storing multiple revisions (e.g., per-round keys, a Redis hash with `round_id` fields, or keeping a list/stream of snapshots) and adding an API to load a specific revision.

**Allow passing an external Redis client**
`RedisSession` currently always instantiates its own Redis client in `__init__` and only optionally accepts a `connection_pool`. It does **not** support injecting an externally managed `Redis` client instance (useful for shared pools, instrumentation, custom retry/backoff, lifecycle management, tests). Consider adding an optional `redis_client: Redis | None = None` parameter; if provided, use it directly and clarify ownership (e.g., only call `close()` when the client was created internally, or add a `close_on_exit` flag).

src/agentscope/session/_redis_session.py
```diff
@@ -0,0 +1,199 @@
...
+    from redis.asyncio import ConnectionPool, Redis
+else:
+    ConnectionPool = Any
+    Redis = Any
+
+
+class RedisSession(SessionBase):
+    """The Redis session class."""
+
+    SESSION_KEY = "agentscope:session:{session_id}:state"
```
rayrayraykk: Please double check the redis key @DavdGao , should we provide a {prefix} to custom this key?

DavdGao: That looks good to me. If we're going this route, I think we should also add the customized prefix to `RedisMemory` for consistency.

src/agentscope/session/_redis_session.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def _get_session_key(self, session_id: str) -> str:
+        """Get the Redis key to store a given session state."""
+        return self.SESSION_KEY.format(
+            session_id=session_id,
+        )
+
+    async def save_session_state(
+        self,
+        session_id: str,
```
DavdGao, line 103: Do we need to consider a user_id here, or is session_id sufficient? I'm not entirely sure about the best approach for this use case. However, we should consider to keep the key in `RedisMemory` and `RedisSession` consistent

PR #1210, opened 2026-02-07
fix(tracing): resolve async context loss in tracing decorator

### LINKED ISSUE #1208: [Bug]:trace_enabled context loss in async frameworks with uvicorn
**Describe the bug**
When using AgentScope with tracing_url configured in async frameworks like FastAPI, the trace_enabled flag stored in ContextVar may lose its value, causing the tracing
  functionality to fail silently.

  The issue occurs because uvicorn creates a new Context for each request, causing ContextVar values to reset to their default values. This means that even after
  agentscope.init() successfully sets trace_enabled = True, the value becomes False when processing HTTP requests, resulting in tracing being disabled.

**To Reproduce**

1. Initialize AgentScope with tracing configured:
```python
import agentscope
agentscope.init(
    project='test-project',
    studio_url='http://localhost:3000',
    tracing_url='http://localhost:3000/v1/traces'
)
```

2. Create a traced function:
```python
from agentscope.tracing import trace

@trace(name='test_function')
async def test_function():
    print('Traced function executed')
    return 'success'
```

3. Execute in FastAPI or similar async framework using `asyncio.create_task()`:
```python
import asyncio
from fastapi import FastAPI

app = FastAPI()

@app.get('/test')
async def test_endpoint():
    # In FastAPI, the request handler may run in a new task
    # where ContextVar context is lost
    task = asyncio.create_task(test_function())
    result = await task
    return {'result': result}
```

4. Check AgentScope Studio or tracing backend
   - Expected: Trace data should be recorded
   - Actual: No trace data is recorded

**Expected behavior**
When `tracing_url` is configured during `agentscope.init()`, the tracing functionality should work consistently across all async tasks and concurrency patterns, regardless of whether the `ContextVar` context is properly inherited or not.

**Error messages**
No explicit error messages are thrown. The issue manifests silently as missing trace data in the tracing backend (AgentScope Studio or external OpenTelemetry backend).

- AgentScope Version: 1.0.14 (via `import agentscope; print(agentscope.__version__)`)
- Python Version: 3.11.13
- OS: macOS (Darwin 25.1.0)
- Framework: FastAPI with uvicorn
- Uvicorn Configuration: uvicorn.run("server:app", host="0.0.0.0", port=8080) (default settings, single worker)

**Additional context**
Root cause: uvicorn creates new Context for each request
Through detailed investigation, I discovered that the issue is specifically caused by uvicorn's behavior:
 1. Module-level initialization sets trace_enabled = True
 2. uvicorn starts and reimports the module, setting trace_enabled = True
 3. lifespan runs successfully with trace_enabled = True
 4. uvicorn receives a request and creates a new Context
 5. ContextVar value is reset to default value False in the new Context
 6. Request processing runs with trace_enabled = False

**Suggested solution:**
Add a global flag to track whether tracing has been configured, and modify `_check_tracing_enabled()` to check this flag first:

```python

# In agentscope/tracing/_trace.py
def _check_tracing_enabled() -> bool:
    """Check if tracing is enabled"""

    # Check if a real TracerProvider is set (not the default proxy)
    tracer_provider = trace.get_tracer_provider()
    if isinstance(tracer_provider, TracerProvider):
        # A real TracerProvider exists, which means tracing was configured
        return True

    # Fall back to ContextVar for runtime control
    return _config.trace_enabled
```

This ensures that once tracing is configured via `init()`, it remains enabled regardless of ContextVar context propagation issues, while still allowing runtime control when tracing is not configured.

chunterg:
## AgentScope Version
1.0.15

## Description
This PR fixes issue #1208 where AgentScope's tracing decorator silently stops collecting spans when the decorated async function runs inside a task created with `asyncio.create_task()`.

**Background**: The issue occurs because ContextVar (used by `_config.trace_enabled`) is lost in async tasks, causing the tracing decorator to think tracing is disabled.

**Changes made**:
- Modified `_check_tracing_enabled()` to first check if a real OpenTelemetry TracerProvider is configured
- If TracerProvider exists, return True (tracing is enabled)
- Fall back to ContextVar only when no TracerProvider is set
- Fixed variable name conflict with pylint warning by using `trace_api` instead of `trace`

**How to test**: 
The fix can be verified by creating an async function decorated with `@trace`, then running it inside `asyncio.create_task()`. With this fix, tracing should continue to work even when ContextVar is lost.

## Checklist
- [x] Code has been formatted with `pre-commit run --all-files` command
- [x] All tests are passing (core tests: tracing, pipeline, react_agent, openai models/formatters)
- [ ] Related documentation has been updated (not needed for bugfix)

Fixes #1208

### DIFF
--- a/src/agentscope/tracing/_trace.py
+++ b/src/agentscope/tracing/_trace.py
@@ -69,20 +69,22 @@ T = TypeVar("T")
 def _check_tracing_enabled() -> bool:
     """Check if the OpenTelemetry tracer is initialized in AgentScope with an
     endpoint.
-    This approach handles async context
-    loss issues automatically, as it checks the actual OpenTelemetry state
-    rather than relying solely on ContextVar.
+
+    This approach handles async context loss issues automatically by checking
+    the actual OpenTelemetry state rather than relying solely on ContextVar.
+    A global_trace_enabled flag is also supported for manual control.
     """
-    from opentelemetry import trace as trace_api
-    from opentelemetry.sdk.trace import TracerProvider
+    # If global tracing is explicitly enabled, check TracerProvider state
+    if _config.global_trace_enabled:
+        from opentelemetry import trace as trace_api
+        from opentelemetry.sdk.trace import TracerProvider
 
-    # Check if a real TracerProvider is set (not the default proxy)
-    tracer_provider = trace_api.get_tracer_provider()
-    if isinstance(tracer_provider, TracerProvider):
-        # A real TracerProvider exists, which means tracing was configured
-        return True
+        tracer_provider = trace_api.get_tracer_provider()
+        if isinstance(tracer_provider, TracerProvider):
+            # A real TracerProvider exists, which means tracing was configured
+            return True
 
-    # Fall back to ContextVar for runtime control
+    # Default behavior: use ContextVar for runtime control
     return _config.trace_enabled
 
 
--- a/src/agentscope/tracing/_trace.py
+++ b/src/agentscope/tracing/_trace.py
@@ -69,11 +69,20 @@ T = TypeVar("T")
 def _check_tracing_enabled() -> bool:
     """Check if the OpenTelemetry tracer is initialized in AgentScope with an
     endpoint.
-
-    TODO: We expect an OpenTelemetry official interface to check if the
-     tracer is initialized. Leaving this function here as a temporary
-     solution.
+    This approach handles async context
+    loss issues automatically, as it checks the actual OpenTelemetry state
+    rather than relying solely on ContextVar.
     """
+    from opentelemetry import trace as trace_api
+    from opentelemetry.sdk.trace import TracerProvider
+
+    # Check if a real TracerProvider is set (not the default proxy)
+    tracer_provider = trace_api.get_tracer_provider()
+    if isinstance(tracer_provider, TracerProvider):
+        # A real TracerProvider exists, which means tracing was configured
+        return True
+
+    # Fall back to ContextVar for runtime control
     return _config.trace_enabled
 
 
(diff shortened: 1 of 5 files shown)


src/agentscope/tracing/_trace.py
```diff
@@ -69,11 +69,20 @@
...
     endpoint.
-
-    TODO: We expect an OpenTelemetry official interface to check if the
-     tracer is initialized. Leaving this function here as a temporary
-     solution.
+    This approach handles async context
+    loss issues automatically, as it checks the actual OpenTelemetry state
+    rather than relying solely on ContextVar.
     """
+    from opentelemetry import trace as trace_api
```
zhijianma: Once a TracerProvider is configured, the function will always return True based on the isinstance check, effectively ignoring the trace_enabled configuration. This means users lose the ability to manually disable tracing through the configuration setting. We should modify the logic to respect both the TracerProvider state and the user's configuration choice, ensuring that manual control over tracing remains possible even when a TracerProvider is present.

chunterg: Thank you for the valuable feedback! I've submitted a new fix.
                                                                                                                                                                       
  **Key changes:**
  1. Added `global_trace_enabled` parameter (defaults to False for backward compatibility)
  2. When `global_trace_enabled=False` (default): Uses ContextVar for runtime control, ignoring TracerProvider state
  3. When `global_trace_enabled=True`: Checks TracerProvider state to fix async context loss

  This ensures users retain full manual control while providing an opt-in solution for async context loss issues.

  **Usage:**
  ```python
  # Default behavior (backward compatible)
  agentscope.init(tracing_url="http://localhost:8080/v1/traces")

  # Enable global tracing to fix async context loss
  agentscope.init(
      tracing_url="http://localhost:8080/v1/traces",
      global_trace_enabled=True
  )

  The implementation respects both TracerProvider state and user configuration choice. Would appreciate your review!

alexvanc:
Actually project, name, run_id, these parameters in agentscope.init are also lost in FastAPI apps

DavdGao:

PR #1246, opened 2026-02-18
fix(config): windows temp file lock when loading default tuner config

dipeshbabu:
AgentScope version
0.1.5

Description
On Windows, NamedTemporaryFile keeps the file handle open and can lock the file. The default config fallback in src/agentscope/tuner/_config.py writes a YAML template to a temporary file and then calls load_config on the same path while the handle is still open. This can fail with permission denied or file in use errors on Windows.

Purpose
Make the default config fallback reliable on Windows while keeping behavior unchanged on Linux and macOS.

Changes made
Create the temporary YAML file with delete set to False so it can be reopened by path on Windows. Ensure the temporary file is closed before calling load_config. Remove the temporary file afterward in a finally block using best effort cleanup.

How to test
On Windows, run the code path that triggers the default config fallback by calling _load_config_from_path_or_default with config_path set to None. Confirm it loads successfully without file lock errors. Also run the same path on Linux or macOS to confirm no behavior change.

Checklist

*  [x] Code has been formatted with pre commit run all files command
*  [x] All tests are passing
*  [x] Docstrings are in Google style
*  [] Related documentation has been updated (e.g. links, examples, etc.)
*  [x] Code is ready for review

### DIFF
--- a/src/agentscope/tuner/_config.py
+++ b/src/agentscope/tuner/_config.py
@@ -181,7 +181,9 @@ def _load_config_from_path_or_default(
             ) as tmp:
                 tmp_path = tmp.name
                 yaml.dump(default_config, tmp)
-                tmp.flush()
+
+            # After the `with` block, the file handle is closed.
+            # Windows can safely reopen it by path.
             config = load_config(tmp_path)
         finally:
             if tmp_path is not None:
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -22,9 +22,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: '3.10'
     - name: Install dependencies
--- a/.github/workflows/sphinx_docs.yml
+++ b/.github/workflows/sphinx_docs.yml
@@ -17,9 +17,9 @@ jobs:
       OS: ${{ matrix.os }}
       PYTHON: '3.10'
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
       - name: Setup Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v6
         with:
           python-version: '3.10'
       - name: Update setuptools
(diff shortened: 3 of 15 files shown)


src/agentscope/tuner/_config.py
```diff
@@ -1,14 +1,16 @@
 # -*- coding: utf-8 -*-
 """Configuration conversion for tuner."""
-from typing import Any, Callable, List, Tuple
-from datetime import datetime
+import contextlib
 import inspect
+import os
```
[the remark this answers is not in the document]
dipeshbabu: Thanks. Moved contextlib and os imports into _load_config_from_path_or_default where they are used to match the lazy loading style guidance, while keeping the Windows temp file locking fix unchanged.

dipeshbabu:
@DavdGao could you review this as well?

DavdGao:
@pan-x-c Plz take a look at this PR

pan-x-c:
It looks good, but the `tuner` module requires a Linux system to run, and I'm not sure if it needs to be fixed.

src/agentscope/tuner/_config.py
```diff
@@ -169,10 +171,22 @@ def _load_config_from_path_or_default(
...
+                encoding="utf-8",
+            ) as tmp:
+                tmp_path = tmp.name
+                yaml.dump(default_config, tmp)
+                tmp.flush()
+            config = load_config(tmp_path)
+        finally:
+            if tmp_path is not None:
+                with contextlib.suppress(OSError):
+                    os.unlink(tmp_path)
```
pan-x-c, line 191:
```suggestion
        tmp = tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=False)
        try:
            yaml.dump(default_config, tmp)
            tmp.close()
            config = load_config(tmp.name)
        finally:
            os.unlink(tmp.name)
```
This approach might be simpler.

dipeshbabu: > This approach might be simpler.

I implemented the same simplified flow, but kept a `with NamedTemporaryFile(...) as tmp:` context manager to satisfy the repo’s pylint rule (R1732). The file handle is still closed before `load_config(tmp_path)` so behavior remains the same and stays Windows-safe, and cleanup remains best-effort.

dipeshbabu: @pan-x-c could you review the pr now?

PR #1251, opened 2026-02-19
fix(gemini): handle None usage in streaming response parsing

shlokmestry:
## AgentScope Version

Current main branch (latest)

---

## Description

This PR fixes a `NoneType` error that occurs during Gemini streaming response parsing when the `usage` field is `None`.

### Problem

In certain streaming scenarios, the Gemini API may return a response chunk where `usage` is `None`.  
The existing implementation assumed `usage` was always present, which led to:
AttributeError: 'NoneType' object has no attribute ...

This caused runtime crashes during streaming generation.

### Solution

Added a defensive check to ensure `usage` is not `None` before accessing its attributes in `_parse_gemini_stream_generation_response`.

The fix ensures:

- No crash when `usage` is `None`
- Existing behavior remains unchanged when `usage` is available
- Streaming continues safely even if partial metadata is returned

---

## Changes Made

- Added `None` guard before accessing `usage`
- Improved robustness of Gemini streaming parser
- No breaking API changes
- No changes to external interfaces

---

## How to Test

Run:

```bash
pytest tests/model_gemini_test.py -q


shlokmestry:
Fixed in the latest commit : refactored the usage parsing into a _extract_usage() helper method, which eliminates the indentation issue and also resolves the too-many-branches pylint warning. The inline blocks have been replaced with a clean single line call.

shlokmestry:
Done! Ran pre-commit run --all-files and committed the black formatting fix. Ready for re-review when you get a chance

DavdGao (approved):
LGTM, thanks for your contribution for AgentScope

PR #1254, opened 2026-02-24
feat(memory): support offloading compressed memory to searchable storage

ascl1u:
## AgentScope Version

1.0.16

## Description

Implements memory offloading for `ReActAgent`, allowing original messages to be stored in searchable storage before they are compressed out of the context window.

Offloading Interface: added `MemoryOffloadingBase` and a default `InMemorySearchableStorage`.
Agent Integration: updated `ReActAgent` to automatically offload messages during compression if a storage backend is configured.
Recall Tool: added `search_offloaded_memory` tool to allow agents to proactively search and retrieve detailed history.
Testing: logic verified via `tests/memory_offloading_test.py`.

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/agent/_react_agent.py
+++ b/src/agentscope/agent/_react_agent.py
@@ -14,6 +14,7 @@ from ._react_agent_base import ReActAgentBase
 from .._logging import logger
 from ..formatter import FormatterBase
 from ..memory import MemoryBase, LongTermMemoryBase, InMemoryMemory
+from ..memory._offloading import MemoryOffloadingBase
 from ..message import (
     Msg,
     ToolUseBlock,
@@ -170,6 +171,13 @@ class ReActAgent(ReActAgentBase):
         `compression_model` is provided, the `compression_formatter` must also
         be provided."""
 
+        offload_storage: MemoryOffloadingBase | None = None
+        """The storage backend to offload compressed messages to. If
+        provided, compressed messages will be stored in this backend
+        and searchable via the `search_offloaded_memory` tool.
+        When not provided, compressed messages are only summarized
+        and not offloaded."""
+
     finish_function_name: str = "generate_response"
     """The name of the function used to generate structured output. Only
     registered when structured output model is provided in the reply call."""
@@ -310,6 +318,15 @@ class ReActAgent(ReActAgentBase):
         if enable_meta_tool:
             self.toolkit.register_tool_function(
                 self.toolkit.reset_equipped_tools,
+            )
+
+        # Register the search tool for offloaded compressed memory
+        if (
+            compression_config
+            and compression_config.offload_storage is not None
+        ):
+            self.toolkit.register_tool_function(
+                self.search_offloaded_memory,
             )
 
         self.parallel_tool_calls = parallel_tool_calls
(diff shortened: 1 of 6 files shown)


src/agentscope/agent/_react_agent.py
```diff
@@ -1125,9 +1142,98 @@ async def _compress_memory_if_needed(self) -> None:
...
+                    )
+
             else:
                 logger.warning(
                     "Failed to obtain compression summary from the model "
                     "structured output in agent %s.",
                     self.name,
                 )
+
+    async def search_offloaded_memory(
```
DavdGao: Since we have already created a search function within the `MemoryOffloadingBase` class, we can register it directly rather than re-creating a new one within the agent. The current implementation in the agent adds no agent-specific logic, making it redundant.

src/agentscope/memory/_offloading/_offloading_base.py
```diff
@@ -0,0 +1,94 @@
...
+            msgs (`list[Msg]`):
+                The original messages that were compressed.
+            summary (`str`):
+                The LLM-generated summary of the compressed messages.
+            **kwargs (`Any`):
+                Additional keyword arguments for the storage operation.
+        """
+
+    @abstractmethod
+    async def search(
```
DavdGao: Consider introducing a `list_tools` abstract method (Just like MCP server does) instead of fixing the abstract method `search` logic here. This would give subclasses more flexibility in how they expose their tools, rather than relying on a specific search/reflection mechanism.

For example, a local file offloading class doesn't necessarily need a dedicated search tool — existing tools like view_text_file and shell tools may be sufficient. Forcing a search method on all subclasses would be overly restrictive in such cases.

src/agentscope/memory/_offloading/_in_memory_offloading.py
```diff
@@ -0,0 +1,170 @@
...
+    """The original messages that were compressed."""
+
+    timestamp: str
+    """ISO-format timestamp when the chunk was offloaded."""
+
+    keywords: list[str] = field(default_factory=list)
+    """Extracted keywords for search (auto-generated from summary)."""
+
+
+class InMemorySearchableStorage(MemoryOffloadingBase):
```
DavdGao: Consider following a consistent naming pattern for this class, similar to the base class convention — e.g., XxxMemoryOffloading — to keep the codebase naming uniform.

src/agentscope/memory/_offloading/_offloading_base.py
```diff
@@ -0,0 +1,94 @@
...
+        await storage.store(messages, summary="User discussed travel plans")
+        results = await storage.search("travel", limit=3)
+"""
+from abc import ABC, abstractmethod
+from typing import Any
+
+from ...message import Msg
+
+
+class MemoryOffloadingBase(ABC):
```
DavdGao: Should MemoryOffloadingBase inherit from agentscope.module.StateModule? Also, I'm curious how the memory offloading abstraction is expected to support serialization and deserialization within the agent — should we register it as part of the agent state within the `ReActAgent` class?

DavdGao (commented):
@ascl1u Thanks, that's excellent work in this pull request! Please see the inline comments for several issues:

- Serialization and deserialization
- Abstract methods
- Class naming

src/agentscope/memory/_offloading/_offloading_base.py
ascl1u: Registered it as a direct attribute in `ReActAgent` so its state is automatically captured during the agent's serialization and deserialization process.

PR #1266, opened 2026-03-01
fix(tool): prevent deadlock in execute_shell_command with large output (#1255)

### LINKED ISSUE #1255: [Bug]:`execute_shell_command` deadlocks when subprocess output exceeds pipe buffer
**Describe the bug**

`agentscope.tool.execute_shell_command` can **deadlock** when the child process writes more to stdout/stderr than the OS pipe buffer (often ~64KB). The parent waits for the process to exit with `proc.wait()` while not reading the pipes; the child blocks on writing once the buffer is full. Neither side progresses.

**To Reproduce**

1. **Code** (async):

```python
import asyncio
from agentscope.tool import execute_shell_command

async def main():
    # GitHub releases API returns a large JSON (often >64KB)
    resp = await execute_shell_command(
        command='curl -s -H "Accept: application/vnd.github+json" '
                '"https://api.github.com/repos/agentscope-ai/agentscope/releases"',
        timeout=60,
    )
    print(resp)

asyncio.run(main())
```

2. **Run:** `python repro.py`

3. **Observe:** The script hangs and never prints. The same `curl` command runs to completion in a normal terminal.

**Expected behavior**

The tool should return with `<returncode>0</returncode>`, `<stdout>...</stdout>` containing the API response, and no hang.

**Error messages**

There is no exception or error message; the call simply blocks indefinitely (until timeout, if any).

**Environment**

- AgentScope Version: 1.0.13
- Python Version: 3.12
- OS: macOS (darwin)

**Additional context**

**Root cause:** In `_shell.py`, the implementation does:

```python
await asyncio.wait_for(proc.wait(), timeout=timeout)
stdout, stderr = await proc.communicate()
```

Waiting for the process with `proc.wait()` without reading stdout/stderr allows the pipe buffers to fill. Once they are full, the child blocks on `write()` and never exits, so `proc.wait()` never returns → deadlock.

lailoo:
## Summary

- **Bug**: `execute_shell_command` deadlocks when subprocess output exceeds pipe buffer (~64KB-256KB)
- **Root cause**: Original code calls `proc.wait()` before `proc.communicate()`, causing pipe buffer to fill and block child process
- **Fix**: Use `asyncio.wait()` with `communicate()` task to avoid cancellation and preserve buffered data

Fixes #1255

## Problem

When a subprocess writes more than the OS pipe buffer size (typically 64KB-256KB) to stdout/stderr, the following deadlock occurs:

1. Child process fills pipe buffer and blocks on write()
2. Parent waits for process exit with `proc.wait()` without reading pipes
3. Neither side can progress → deadlock

**Before fix (original code in main):**
```python
# Original implementation (lines 40-47 in main branch)
try:
    await asyncio.wait_for(proc.wait(), timeout=timeout)  # ❌ Waits without reading pipes
    stdout, stderr = await proc.communicate()              # Never reached if deadlock
    stdout_str = stdout.decode("utf-8")
    stderr_str = stderr.decode("utf-8")
    returncode = proc.returncode
```

**Reproduction:**
```python
# Command generating ~1MB output (exceeds pipe buffer)
cmd = 'python -c "[sys.stdout.write(str(i)*1000+chr(10)) for i in range(1000)]"'
res = await execute_shell_command(cmd, timeout=10)
# Result: TimeoutError after 10s, returncode=-1, only first ~200KB captured
```

## Changes

**File: `src/agentscope/tool/_coding/_shell.py`**

Replace `wait() + communicate()` pattern with `asyncio.wait()` to avoid pipe deadlock:

```python
# New implementation
communicate_task = asyncio.create_task(proc.communicate())
done, pending = await asyncio.wait([communicate_task], timeout=timeout)

if communicate_task in done:
    # Process completed within timeout
    stdout, stderr = communicate_task.result()
    # ... decode and return
else:
    # Timeout: terminate and collect remaining output
    proc.terminate()
    stdout, stderr = await communicate_task  # Still collects buffered data
    # ... decode and return with timeout error
```

**Key improvements:**
- `communicate()` starts reading pipes immediately, preventing buffer fill
- `asyncio.wait()` doesn't cancel the task on timeout, preserving pipe readers
- Timeout path still collects partial output after termination

**File: `tests/tool_test.py`**

Add regression test `test_execute_shell_command_large_output` covering ~1MB stdout.

## Test plan

- [x] **Reproduced deadlock on main branch**: Test with 1MB output times out after 10s, returns `returncode=-1` with only ~200KB captured
- [x] **Verified fix resolves deadlock**: Same test passes in <2s with full output captured
- [x] **New regression test**: `test_execute_shell_command_large_output` covers large stdout scenario
- [x] **All 6 existing tool tests pass**: Including timeout behavior (partial output collection)
- [x] **All 60 tool/toolkit tests pass**: No regressions introduced

## Effect on User Experience

**Before:** Commands producing large output (e.g., `curl` fetching large JSON, `git log`, data processing scripts) hang indefinitely until timeout, losing most output.

**After:** All commands complete normally regardless of output size, with full output captured.

### DIFF
--- a/src/agentscope/tool/_coding/_shell.py
+++ b/src/agentscope/tool/_coding/_shell.py
@@ -40,7 +40,7 @@ async def execute_shell_command(
     # Use asyncio.wait instead of wait_for to avoid cancelling communicate()
     # which would close the pipe readers and lose buffered data.
     communicate_task = asyncio.create_task(proc.communicate())
-    done, pending = await asyncio.wait(
+    done, _ = await asyncio.wait(
         [communicate_task],
         timeout=timeout,
     )


DavdGao:
@lailoo Plz format the code by executing `pre-commit run --all-files`

lailoo:
@DavdGao  Thanks for the review!   I've fixed the pylint warning about the unused variable and just pushed the update. The CI should pass now. Please let me know if there's anything else I should adjust\! 🙏

PR #1268, opened 2026-03-01
fix(deps): resolve uv sync numpy dependency conflict

### LINKED ISSUE #1267: [Bug]:uv sync fails due to numpy version conflict between qdrant-client and py-openjudge
## Description

Running `uv sync` (or `uv pip install -e ".[dev]"`) fails with a dependency resolution error because `qdrant-client==1.15.1` and `py-openjudge` have incompatible `numpy` version requirements:

- `qdrant-client==1.15.1` requires `numpy>=2.1.0` (on Python >= 3.13)
- `py-openjudge` (all versions) requires `numpy>=1.22.0,<2.0.0`

Both packages are included transitively through the `dev` optional-dependency group:
- `dev` → `full` → `rag` → `vdbs` → `qdrant` → `qdrant-client==1.15.1`
- `dev` → `py-openjudge`

Note: `uv pip install -e .` (core dependencies only) works fine since neither package is in the core `dependencies`.

## Environment

- Python: 3.13.12
- uv: latest
- OS: macOS

## Error Message

```
$ uv sync
  × No solution found when resolving dependencies for split (markers: python_full_version >= '3.14'
  │ and sys_platform == 'win32'):
  ╰─▶ Because qdrant-client==1.15.1 depends on numpy{python_full_version >= '3.13'}>=2.1.0 and all
      versions of py-openjudge depend on numpy>=1.22.0,<2.0.0, we can conclude that all versions of
      py-openjudge and qdrant-client==1.15.1 are incompatible.
      And because agentscope[dev] depends on py-openjudge, we can conclude that agentscope[dev] and
      qdrant-client==1.15.1 are incompatible.
      And because agentscope[dev] depends on qdrant-client==1.15.1 and your project requires
      agentscope[dev], we can conclude that your project's requirements are unsatisfiable.
```

Even `uv sync --no-dev` fails because `uv` resolves **all** dependency groups for satisfiability, not just the ones being installed.

## Root Cause

`py-openjudge` pins `numpy<2.0.0`, but `qdrant-client>=1.13.0` requires `numpy>=2.1.0` on Python >= 3.13 (via environment marker `python_full_version >= '3.13'`). These two constraints are mutually exclusive on modern Python.

## Solution

Add `override-dependencies` in `[tool.uv]` to remove the numpy upper bound from `py-openjudge`:

```toml
[tool.uv]
# Override py-openjudge's numpy<2.0.0 constraint to resolve conflict
# with qdrant-client's numpy>=2.1.0 requirement on Python>=3.13
override-dependencies = [
    "numpy>=1.22.0",
]
```

This allows `uv` to resolve `numpy>=2.1.0` for Python >= 3.13 (satisfying `qdrant-client`) while keeping `numpy>=1.22.0` as the lower bound.

### Approaches that did NOT work

| Approach | Why it failed |
|----------|--------------|
| Relaxing `qdrant-client` version range (`>=1.13.0,<1.16.0`) | All versions in this range have the same `numpy>=2.1.0` requirement for Python >= 3.13 |
| Adding `python_version < '3.13'` marker to `py-openjudge` | `uv` still resolves cross-environment satisfiability and finds the conflict |
| Limiting `requires-python` to `<3.14` | Conflict still exists for the Python 3.13 split |
| Declaring `qdrant` and `openjudge` as `[tool.uv] conflicts` | `dev` → `full` → `vdbs` → `qdrant` transitively pulls in `qdrant-client`, so the conflict declaration can't separate them |

### ⚠️ Note

`py-openjudge` may not be fully compatible with `numpy>=2.0.0` at runtime. If issues arise, consider removing `py-openjudge` from `dev` dependencies or replacing it with a numpy 2.x-compatible alternative.

rouxiaomin:
## Summary
- add `[tool.uv].override-dependencies` for `numpy>=1.22.0`
- remove the transitive `numpy<2.0.0` restriction coming from `py-openjudge`
- allow `uv sync` resolution to proceed with `qdrant-client` on Python >= 3.13

Closes #1267

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -189,8 +189,8 @@ build-backend = "setuptools.build_meta"
 version = {attr = "agentscope._version.__version__"}
 
 [tool.uv]
-# Override py-openjudge's numpy<2.0.0 constraint to resolve conflict
-# with qdrant-client's numpy>=2.1.0 requirement on Python>=3.13
+# Override py-openjudge's numpy<2.0.0 constraint to resolve conflicts
+# with dependencies requiring newer numpy (e.g., qdrant-client on Python>=3.13).
 override-dependencies = [
     "numpy>=1.22.0",
 ]


Kanaricc:
Thanks a lot for your contribution!

OpenJudge recently updated its `numpy` dependency (see this commit: https://github.com/agentscope-ai/OpenJudge/commit/25a885b7654a0660a8a78dda0b2c0515e047d3bb
). We recommend using the latest version.

rouxiaomin:
@Kanaricc Thank you very much. I just saw that version 0.2.3 of openjudge has been released and it fixes this issue. I will close this issue.

PR #1271, opened 2026-03-02
fix(openai-model): support newer vllm reasoning field parsing

WeiSibo:
## Summary

This PR adds compatibility for newer vLLM (v0.16.0+) OpenAI-compatible responses that return
`reasoning` instead of `reasoning_content`.

AgentScope currently reads `reasoning_content` when constructing thinking blocks.
With newer vLLM versions, thinking output may be missed.

## Changes

- keep backward compatibility with `reasoning_content`
- add fallback to `reasoning` for streaming responses
- add fallback to `reasoning` for non-streaming responses

## Motivation

Upstream vLLM has migrated reasoning output from `reasoning_content` to `reasoning`.
In v0.16.0, vLLM explicitly removed the deprecated `reasoning_content` message field.
This patch allows AgentScope to support both old and new vLLM versions without
breaking existing integrations.

## References

- [vLLM latest docs: Reasoning Outputs](https://docs.vllm.ai/en/latest/features/reasoning_outputs/)
- [vLLM RFC: `reasoning_content` -> `reasoning`](https://github.com/vllm-project/vllm/issues/27755)
- [vLLM release v0.16.0: deprecated `reasoning_content` field removed](https://github.com/vllm-project/vllm/releases/tag/v0.16.0)
- [Older vLLM docs: `reasoning_content` (for comparison)](https://docs.vllm.ai/en/v0.9.1/features/reasoning_outputs.html)

This change is backward compatible and only extends the existing parsing logic.

### DIFF
--- a/src/agentscope/model/_openai_model.py
+++ b/src/agentscope/model/_openai_model.py
@@ -375,7 +375,9 @@ class OpenAIChatModel(ChatModelBase):
                 choice = chunk.choices[0]
 
                 delta_reasoning = getattr(
-                    choice.delta, "reasoning_content", None
+                    choice.delta,
+                    "reasoning_content",
+                    None,
                 )
                 if not isinstance(delta_reasoning, str):
                     delta_reasoning = getattr(choice.delta, "reasoning", None)
--- a/src/agentscope/model/_openai_model.py
+++ b/src/agentscope/model/_openai_model.py
@@ -374,11 +374,13 @@ class OpenAIChatModel(ChatModelBase):
 
                 choice = chunk.choices[0]
 
-                thinking += (
-                    getattr(choice.delta, "reasoning_content", None) 
-                    or getattr(choice.delta, "reasoning", None)
-                    or ""
-                )
+                delta_reasoning = getattr(choice.delta, "reasoning_content", None)
+                if not isinstance(delta_reasoning, str):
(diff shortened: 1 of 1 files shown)


WeiSibo:
I pushed a follow-up fix for the failing OpenAI model tests.

I verified it locally with:

python -m pytest tests/model_openai_test.py -q

and all tests in tests/model_openai_test.py pass now. @DavdGao

DavdGao:
@WeiSibo Thanks for you contribution, and it looks good to me. Plz run `pre-commit run --all-files` to format the code first

DavdGao (approved):
LGTM

lessmore991:
\agentscope-core\src\main\java\io\agentscope\core\formatter\openai\dto\OpenAIMessage.java lines 150–152:   
`public void setReasoningContent(String reasoningContent) {
        this.reasoningContent = reasoningContent;
    }`
The code was not updated accordingly, making it impossible to pass reasoning content to the vLLM model service API. This is because the vLLM API now only accepts the reasoning field and no longer recognizes reasoning_content.

References：
[vLLM only accepts the reasoning field. - vllm/entrypoints/chat_utils.py#L297](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/chat_utils.py#L297)

Will this issue be addressed in a future version to ensure compatibility?

PR #1298, opened 2026-03-09
fix(hooks): prevent Agent crash when Studio disconnects

### LINKED ISSUE #1297: fix(hooks): studio_pre_print_hook crashes Agent when Studio disconnects
### Problem Description

The `as_studio_forward_message_pre_print_hook` function raises an exception `raise e from None` after 3 retry attempts when the Studio session is disconnected or unreachable, causing the Agent process to crash.

### Steps to Reproduce

1. Start an Agent with Studio hook enabled
2. Simulate Studio disconnection (e.g., stop Studio service or connect to a non-existent address)
3. Agent sends a message triggering the `pre_print` hook
4. Hook retries 3 times then throws an exception, causing Agent to crash

Minimal reproduction code:

```python
import asyncio
from agentscope.agent import AgentBase
from agentscope.message import Msg, TextBlock
from agentscope.hooks._studio_hooks import as_studio_forward_message_pre_print_hook

class TestAgent(AgentBase):
    def __init__(self, name: str):
        super().__init__()
        self.name = name
        self._disable_console_output = False
    
    async def reply(self, msg: Msg) -> Msg:
        await self.print(msg)
        return msg
    
    async def observe(self, msg: Msg) -> None:
        pass
    
    async def handle_interrupt(self, *args, **kwargs) -> Msg:
        return Msg("test", "Interrupt handled", "assistant")

# Register hook with non-existent Studio URL
TestAgent.register_class_hook(
    "pre_print",
    "studio_forward",
    lambda self, kwargs: as_studio_forward_message_pre_print_hook(
        self, kwargs, studio_url="http://127.0.0.1:9999", run_id="test"
    ),
)

agent = TestAgent(name="TestAgent")

async def main():
    msg = Msg("user", [TextBlock(type="text", text="Hello")], "user")
    await agent.reply(msg)  # Raises ConnectionError exception

asyncio.run(main())
```

### Expected Behavior

When Studio is disconnected, the Agent should continue running normally instead of crashing.

### Actual Behavior

Raises `ConnectionError` exception, causing Agent operation to fail.

### Proposed Fix

Replace `raise e from None` with graceful degradation (log warning and return):

```python
# Before fix
raise e from None

# After fix
logger.warning(
    "Failed to forward message to Studio after %d retries: %s. "
    "Agent will continue running without Studio forwarding.",
    n_retry,
    e,
)
return
```

### Environment

- AgentScope version: 1.0.16
- Python version: 3.13.2
- OS: Linux/macOS/Windows

YingchaoX:
## Description

This PR fixes the issue where `as_studio_forward_message_pre_print_hook` crashes the Agent when Studio is disconnected or unreachable.

## Changes

- Modified `src/agentscope/hooks/_studio_hooks.py`
- Replaced `raise e from None` with graceful degradation (log warning and return)
- Added import for `logger` from `.._logging`

## Behavior Change

**Before:** When Studio is unreachable after 3 retries, an exception is raised causing Agent to crash.

**After:** When Studio is unreachable after 3 retries, a warning is logged and the function returns gracefully, allowing Agent to continue running.

## Testing

Tested with a minimal reproduction script that simulates Studio disconnection:
- Agent no longer crashes when Studio is unreachable
- Warning message is properly logged
- Agent continues to operate normally

Fixes #1297

### DIFF
--- a/src/agentscope/hooks/_studio_hooks.py
+++ b/src/agentscope/hooks/_studio_hooks.py
@@ -51,10 +51,10 @@ def as_studio_forward_message_pre_print_hook(
                 n_retry += 1
                 continue
 
-            # Graceful degradation: log warning and return to avoid crashing the Agent
+            # Graceful degradation: log warning and return to avoid crashing
             logger.warning(
                 "Failed to forward message to Studio after %d retries: %s. "
-                "Agent will continue running without Studio forwarding.",
+                "Agent will continue without Studio forwarding.",
                 n_retry,
                 e,
             )


src/agentscope/hooks/_studio_hooks.py
```diff
@@ -50,4 +51,11 @@ def as_studio_forward_message_pre_print_hook(
                 n_retry += 1
                 continue
 
-            raise e from None
+            # 静默降级：打印警告并返回，避免 Agent 进程崩溃
```
DavdGao: Please use English annotations.

YingchaoX: Fixed. PTAL, thanks!

DavdGao (approved):
LGTM

PR #1299, opened 2026-03-09
fix(examples): await register_mcp_client in deep_research_agent

YingchaoX:
## Description

Fix missing `await` in `register_mcp_client()` call in `deep_research_agent.py`.

## Problem

The original code used `asyncio.get_running_loop().create_task()` without properly awaiting the coroutine:

```python
asyncio.get_running_loop().create_task(
    self.toolkit.register_mcp_client(search_mcp_client),
)
```

This caused:
- MCP tools registered only in schema but not actually callable
- Race conditions between tool registration and usage
- Tools appearing available but failing when called

## Solution

Added `_ensure_mcp_initialized()` async method that properly awaits MCP client registration on first `reply()` call:

```python
async def _ensure_mcp_initialized(self) -> None:
    if not self._mcp_initialized:
        await self.toolkit.register_mcp_client(self._search_mcp_client)
        self._mcp_initialized = True
```

## Changes

- Removed unused `asyncio` import
- Store `search_mcp_client` reference in `__init__`
- Add `_ensure_mcp_initialized()` method
- Call initialization in `reply()` before processing

## Checklist

- [x] pre-commit checks passed
- [x] Code follows project style guidelines
- [x] Changes are minimal and focused

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -22,9 +22,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: '3.10'
     - name: Install dependencies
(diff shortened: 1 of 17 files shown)


DavdGao:
@luohaohao plz check this pull request

Luohh5:
LGTM. Thanks for your contribution!

PR #1302, opened 2026-03-09
feat(models): add MiniMax model support

ximiximi423:
## Summary
- Add support for **MiniMax-M2.5** and **MiniMax-M2.5-highspeed** chat models via OpenAI-compatible API
- Add support for **MiniMax TTS** models (speech-2.8-hd, speech-2.8-turbo, etc.) via T2A v2 HTTP API
- 17 comprehensive unit tests (9 chat model + 8 TTS)

### MiniMax-M2.5 Model Features
- **OpenAI-compatible API**: Uses the `openai` SDK with MiniMax base URL
- **Interleaved Thinking**: Parses `<think>...</think>` tags into `ThinkingBlock` objects
- **Tool Calling**: Full support for function calling with `tools` and `tool_choice`
- **Streaming**: Both streaming and non-streaming modes
- **Structured Output**: Pydantic `BaseModel` support

### Usage Example
```python
from agentscope.model import MiniMaxChatModel
model = MiniMaxChatModel(model_name="MiniMax-M2.5", api_key="your-key")
```

### MiniMax TTS Features
- **T2A v2 HTTP API**: Direct HTTP calls via `httpx`
- **Multiple Models**: speech-2.8-hd (default), speech-2.8-turbo, etc.
- **Streaming**: Newline-delimited JSON streaming
- **Voice Settings**: Configurable voice_id, speed, volume, pitch, emotion
- **Audio Formats**: mp3, pcm, flac, wav

### Usage Example
```python
from agentscope.tts import MiniMaxTTSModel
tts = MiniMaxTTSModel(api_key="your-key", model_name="speech-2.8-hd")
```

## Test plan
- [x] All 17 new unit tests pass
- [x] All 373 total tests pass (0 failures)

### DIFF
--- a/src/agentscope/model/_minimax_model.py
+++ b/src/agentscope/model/_minimax_model.py
@@ -1,0 +1,389 @@
+# -*- coding: utf-8 -*-
+# pylint: disable=too-many-branches
+"""MiniMax Chat model class."""
+import copy
+import json
+import re
+from datetime import datetime
+from typing import (
+    Any,
+    TYPE_CHECKING,
+    List,
+    AsyncGenerator,
+    Literal,
+    Type,
+)
+from collections import OrderedDict
+
+from pydantic import BaseModel
+
+from . import ChatResponse
+from ._model_base import ChatModelBase
+from ._model_usage import ChatUsage
+from .._logging import logger
+from .._utils._common import _json_loads_with_repair
+from ..message import (
+    ToolUseBlock,
+    TextBlock,
+    ThinkingBlock,
+)
+from ..tracing import trace_llm
+from ..types import JSONSerializableObject
+
+if TYPE_CHECKING:
+    from openai.types.chat import ChatCompletion
+    from openai import AsyncStream
+else:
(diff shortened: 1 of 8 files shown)


src/agentscope/model/_minimax_model.py
```diff
@@ -0,0 +1,389 @@
...
+
+        super().__init__(model_name, stream)
+
+        import openai
+
+        client_kwargs = client_kwargs or {}
+        if "base_url" not in client_kwargs:
+            client_kwargs["base_url"] = _DEFAULT_MINIMAX_BASE_URL
+
+        self.client = openai.AsyncClient(
```
qbc2016: In the MiniMax official documentation, it recommends using the Anthropic-compatible API, so I wonder if we could just use AnthropicChatModel and AnthropicFormatter instead?

qbc2016:
Thank you for the contribution!

The MiniMax official documentation recommends using the Anthropic-compatible API, so I wonder if we could just use AnthropicChatModel and AnthropicFormatter instead?

DavdGao:

Issue #1149, opened 2026-01-21
[Bug]:set_console_output_enabled(False) unexpectedly disables AgentScope Studio tracing/output


WuJiaoJue:
### Description

In recent versions of AgentScope (main branch), calling:

```python
agent.set_console_output_enabled(enabled=False)
```

does not only disable console output, but also **silently disables message forwarding to AgentScope Studio**, including message display and tracing.

This behavior is **not documented** and is **inconsistent with the method name and its docstring**, which only mention disabling console output in production environments.

------

### Expected Behavior

- `set_console_output_enabled(False)` should **only affect console printing**
- AgentScope Studio should continue to receive:
  - messages
  - metadata
  - traces / spans

This was the behavior in earlier versions, where console output and Studio tracing were independent.

------

### Actual Behavior

When `set_console_output_enabled(False)` is set:

- Console output is disabled ✅
- **Studio message display and tracing are also disabled ❌**
- No warning or documentation indicates this global side effect

Root cause (from source code):

In
 `src/agentscope/hooks/_studio_hooks.py`

```python
if self._disable_console_output:
    return
```

Studio hooks directly depend on the console output flag, effectively treating Studio as a “visual console” rather than an independent output channel.

------

### Why This Is a Problem

1. **Semantic mismatch**

   - Method name and docstring only reference console output:

     > “in a production environment, you may want to disable the console output to avoid messy logs”

   - There is no mention that this disables Studio / tracing.

2. **Breaks production workflows**

   - A common use case is:
     - disable console logs in production
     - keep Studio / tracing enabled for observability
   - This is currently impossible without modifying source code.

3. **Backward incompatibility**

   - Previous AgentScope versions allowed disabling console output while keeping tracing intact.
   - The behavior change is undocumented.

------

### Suggested Improvements

One of the following would resolve the issue:

1. **Decouple Studio output from console output**

   - Introduce a separate flag, e.g.:

     ```
     set_studio_output_enabled(True | False)
     ```

2. **At minimum, update documentation**

   - Clearly state that `set_console_output_enabled(False)`:
     - disables console output
     - disables Studio message display
     - disables tracing

3. **Emit a warning**

   - When disabling console output while Studio is enabled, warn users about the side effect.

------

### Temporary Workaround

Manually comment out or modify the following line:

https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/hooks/_studio_hooks.py#L19

```
if self._disable_console_output:
    return
```

[...]

PR #1328, opened 2026-03-15
fix: invoke print hooks before console output check

KJyang-0114:
## Summary

This fixes issue #1149 where `set_console_output_enabled(False)` also disabled AgentScope Studio forwarding.

## Root Cause

The `print()` method in `AgentBase` checked `_disable_console_output` and returned early **before** any hooks were invoked. This meant the Studio pre_print hook never ran when console output was disabled.

## Fix

Invoke `pre_print` and `post_print` hooks **before** the console output check. This decouples Studio/tracing from console output:

- `set_console_output_enabled(False)` now only disables console printing
- Studio continues to receive messages, metadata, and traces

## Changes

- `src/agentscope/agent/_agent_base.py`:
  - Added invocation of `pre_print` hooks before the early return
  - Added invocation of `post_print` hooks after console output (but before cleanup)

## Testing

- Syntax validation passed
- Manual test: Enable Studio, disable console output, verify messages still appear in Studio

## Risk

- Low: Only adds hook invocations, no behavior changes to existing console output logic

## Rollback

Simply revert this commit to restore previous behavior.

### DIFF
--- a/src/agentscope/agent/_agent_base.py
+++ b/src/agentscope/agent/_agent_base.py
@@ -227,6 +227,15 @@ class AgentBase(StateModule, metaclass=_AgentMeta):
             # from monopolizing the event loop.
             await asyncio.sleep(0)
 
+        # Invoke pre_print hooks (e.g., Studio forwarding) BEFORE the console
+        # output check, so that Studio receives messages even when console
+        # output is disabled.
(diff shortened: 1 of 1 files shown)


DavdGao:

PR #1350, opened 2026-03-19
fix(common): more robust JSON string loading

opposj:
## AgentScope Version

1.0.17

## Description

I encountered a JSON parsing issue while working with a locally deployed vLLM model fine-tuned on the Qwen-3 series. With tool calls resolved by Hermes, I obtained the following function argument (take an example):

```py
args = '"{\\"query\\": \\"weak value amplification second order expansion average meter observable\\"}"' 
```

Currently, the JSON parsing in https://github.com/agentscope-ai/agentscope/blob/c19b8bc5723ec562fff77d3386f10de3e426f8da/src/agentscope/_utils/_common.py#L31 acts in a once-to-pass scheme, resulting in:

```py
json.loads('"{\\"query\\": \\"weak value amplification second order expansion average meter observable\\"}"')
# -> '{"query": "weak value amplification second order expansion average meter observable"}'
```

Still a string and fails the `isinstance` check. However, it can actually be parsed via `json.loads` again:

```py
json.loads('{"query": "weak value amplification second order expansion average meter observable"}')
# -> {'query': 'weak value amplification second order expansion average meter observable'}
```

Based on this observation, it is more robust to iteratively parse the raw string until a `dict` check is satisfied, and that's all about this commit. This commit is fully backward compatible and does not affect any other components, just providing a more robust JSON string parsing.

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/_utils/_common.py
+++ b/src/agentscope/_utils/_common.py
@@ -49,10 +49,10 @@ def _json_loads_with_repair(
             Returns an empty dict if all repair attempts fail.
     """
     try:
-        repaired = repair_json(json_str, stream_stable=True)
-        result = json.loads(repaired)
-        if isinstance(result, dict):
-            return result
+        result = repair_json(json_str, stream_stable=True)
+        while not isinstance(result := json.loads(result), dict):
+            ...
+        return result
 
     except Exception:
         if len(json_str) > 100:
--- a/src/agentscope/_utils/_common.py
+++ b/src/agentscope/_utils/_common.py
@@ -50,6 +50,7 @@ def _json_loads_with_repair(
     """
     try:
         result = repair_json(json_str, stream_stable=True)
+        # For nested JSON strings like '"{\\"key\\": \\"value\\"}"'
         while not isinstance(result := json.loads(result), dict):
             ...
         return result


src/agentscope/_utils/_common.py
```diff
@@ -49,10 +49,10 @@ def _json_loads_with_repair(
             Returns an empty dict if all repair attempts fail.
     """
     try:
-        repaired = repair_json(json_str, stream_stable=True)
-        result = json.loads(repaired)
-        if isinstance(result, dict):
-            return result
+        result = repair_json(json_str, stream_stable=True)
+        while not isinstance(result := json.loads(result), dict):
```
DavdGao, line 54: @opposj This introduces an infinite loop — the loop body is ... (no-op), so if json.loads(result) never returns a dict, it will loop forever. What's the intention here?

DavdGao (changes requested):

src/agentscope/_utils/_common.py
opposj: Actually, an infinite loop is technically impossible due to the "reduction" behavior of `json.loads`. Given any object, `json.loads` will never return itself unchanged. A possible type from `dict[str, Any] | list[Any] | str | float | int | bool | None` is always guaranteed, or an exception is raised. For any return type that is not a `dict` or `str`, when fed into `json.loads` again, an exception occurs, and we have captured it correctly. For `str`, as mentioned before, `json.loads` will never infinitely resolve a string; every time a string is fed, the output is shortened until an exception is hit. Overall, as our goal is to return a valid `dict`, iteratively checking for a potentially "nested" string is a more robust method with no risk introduced :)

I have added a simple example that demonstrates this "nested" case in the comment of the changed code.

DavdGao:

PR #1369, opened 2026-03-26
fix(tool): avoid shell deadlock on large output

### LINKED ISSUE #1255: [Bug]:`execute_shell_command` deadlocks when subprocess output exceeds pipe buffer

sztoplover-bit:
## AgentScope Version

1.0.17

## Description

Fixes #1255.

`execute_shell_command` previously waited for the subprocess to exit before reading stdout/stderr. If the child process produced more data than the OS pipe buffer, the child could block on write while the parent was still waiting on `proc.wait()`, causing a deadlock.

This PR switches the implementation to start `proc.communicate()` immediately and wait on that task with a timeout. On timeout, the process is terminated and the same `communicate()` task is awaited so partial stdout/stderr are still preserved.

## Changes

- use a background `proc.communicate()` task instead of `proc.wait()` followed by `communicate()`
- shield the communicate task while waiting with timeout
- keep partial stdout behavior for timeout cases
- add a regression test that emits 128KB to stdout and verifies the shell tool returns successfully

## How to Test

- `./.venv/bin/pytest tests/tool_test.py -q -k execute_shell_command`
- direct manual check with a large stdout payload via `execute_shell_command`

## Checklist

- [ ] Code has been formatted with `pre-commit run --all-files` command
- [x] All tests are passing
- [ ] Related documentation has been updated (e.g. links, examples, etc.)

### DIFF
--- a/src/agentscope/tool/_coding/_shell.py
+++ b/src/agentscope/tool/_coding/_shell.py
@@ -37,9 +37,13 @@ async def execute_shell_command(
         bufsize=0,
     )
 
+    communicate_task = asyncio.create_task(proc.communicate())
+
     try:
-        await asyncio.wait_for(proc.wait(), timeout=timeout)
-        stdout, stderr = await proc.communicate()
+        stdout, stderr = await asyncio.wait_for(
+            asyncio.shield(communicate_task),
+            timeout=timeout,
+        )
         stdout_str = stdout.decode("utf-8")
         stderr_str = stderr.decode("utf-8")
         returncode = proc.returncode
@@ -52,7 +56,7 @@ async def execute_shell_command(
         returncode = -1
         try:
             proc.terminate()
-            stdout, stderr = await proc.communicate()
+            stdout, stderr = await communicate_task
             stdout_str = stdout.decode("utf-8")
             stderr_str = stderr.decode("utf-8")
             if stderr_str:
--- a/tests/tool_test.py
+++ b/tests/tool_test.py
@@ -33,9 +33,7 @@ class ToolTest(IsolatedAsyncioTestCase):
         # empty output
         res = await execute_python_code(code="a = 1 + 1")
         self.assertEqual(
-            "<returncode>0</returncode>"
-            "<stdout></stdout>"
-            "<stderr></stderr>",
+            "<returncode>0</returncode>" "<stdout></stdout>" "<stderr></stderr>",
             res.content[0]["text"],
         )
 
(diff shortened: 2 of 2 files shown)


src/agentscope/tool/_coding/_shell.py
```diff
@@ -37,9 +37,13 @@ async def execute_shell_command(
         bufsize=0,
     )
 
+    communicate_task = asyncio.create_task(proc.communicate())
+
     try:
-        await asyncio.wait_for(proc.wait(), timeout=timeout)
-        stdout, stderr = await proc.communicate()
+        stdout, stderr = await asyncio.wait_for(
+            asyncio.shield(communicate_task),
```
qbc2016, line 44: As far as I know, due to shield protection, even after a timeout occurs, the process continues running in the background, consuming system resources.

DavdGao:

PR #1469, opened 2026-04-13
fix(memory): flush session before inserting mark records to fix FK constraint error

### LINKED ISSUE #1380: [Bug]:Foreign Key Constraint Error when using structured_model in AsyncSQLAlchemyMemory
### Bug Description

When using the structured_model parameter in AgentScope agents (e.g., ReActAgent), which internally uses AsyncSQLAlchemyMemory to store messages with marks, a foreign key constraint error occurs:

```
sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (pymysql.err.IntegrityError) (1452, 'Cannot add or update a child row: a foreign key constraint fails (`test`.`message_mark`, CONSTRAINT `1` FOREIGN KEY (`msg_id`) REFERENCES `message` (`id`) ON DELETE CASCADE)')
[SQL: INSERT INTO message_mark (msg_id, mark) VALUES (%s, %s)]
[parameters: ('user_913e9801-user_913e9801_itinerary-UqwuZZTuLozNPZrHM7fv3X', <_MemoryMark.HINT: 'hint'>)]
```

### Root Cause Analysis
The issue is in `AsyncSQLAlchemyMemory.add()` method (line 444-489 in _sqlalchemy_memory.py):

```python
# Add messages to message table
for i, m in enumerate(messages_to_add):
    message_record = self.MessageTable(...)
    self.session.add(message_record)  # ← Only marks as pending insert

# Create mark records if marks are provided
if marks:
    mark_records = [...]
    if mark_records:
        await self.session.run_sync(
            lambda session: session.bulk_insert_mappings(
                self.MessageMarkTable,
                mark_records,  # ← Executes SQL immediately
            ),
        )

await self.session.commit()  # ← Commits at the end
```

### The Problem:

- `session.add(message_record)` only marks the message object for insertion (in-memory)
- `bulk_insert_mappings()` executes SQL immediately
- At this point, the message record doesn't exist in the database yet
- The foreign key constraint fails because msg_id references a non-existent message

### Suggested Fix
Add await self.session.flush() before bulk_insert_mappings() to ensure messages are written to the database first

octo-patch:
Fixes #1380

## Problem

When using `structured_model` with marks in `AsyncSQLAlchemyMemory.add()`, a foreign key constraint violation occurs:

```
sqlalchemy.exc.PendingRollbackError: ... (pymysql.err.IntegrityError) (1452, 'Cannot add or update a child row: a foreign key constraint fails (message_mark, CONSTRAINT 1 FOREIGN KEY (msg_id) REFERENCES message (id) ON DELETE CASCADE)')
```

The root cause is a timing issue in `add()`:
1. `session.add(message_record)` only stages the message row in SQLAlchemy's unit-of-work (in-memory), without issuing an INSERT yet.
2. `session.bulk_insert_mappings()` immediately executes SQL to insert mark rows, referencing `msg_id` values that do not yet exist in the database.
3. The foreign key constraint fails at step 2 because the parent rows were never flushed.

## Solution

Add `await self.session.flush()` after staging the message records and before inserting the mark records. `flush()` sends the pending `INSERT` statements for the message rows to the database within the current transaction, making them visible to the subsequent `bulk_insert_mappings()` call without committing the transaction.

## Testing

The fix satisfies the foreign key constraint by ensuring parent rows exist in the database before child rows reference them. The existing `commit()` at the end of the method atomically finalizes both the messages and their marks.

### DIFF
--- a/src/agentscope/memory/_working_memory/_sqlalchemy_memory.py
+++ b/src/agentscope/memory/_working_memory/_sqlalchemy_memory.py
@@ -451,6 +451,11 @@ class AsyncSQLAlchemyMemory(MemoryBase):
             )
             self.session.add(message_record)
 
+        # Flush pending message inserts so foreign key constraints are
+        # satisfied when mark records reference them below.
+        if marks:
+            await self.session.flush()
+
         # Create mark records if marks are provided (use bulk insert)
         if marks:
             mark_records = [


qbc2016:
Thank you for this. However, we already have a similar [one](#1396). Closing now, feel free to discuss there.

PR #1479, opened 2026-04-14
feat(tool): add max_calls limit for tool invocations

gaozexun:
## AgentScope Version
1.0.19dev

## Description
Background:
Some tools may be called repeatedly in ReAct loops, causing potential infinite loops and unnecessary resource consumption.

Purpose:
Add `max_calls` control to limit tool invocation count per registered tool.

Changes:
- Added `max_calls: int | None = None` and `call_count: int = 0` to `RegisteredToolFunction`.
- Extended `Toolkit.register_tool_function(...)` with `max_calls` argument.
- Added validation: negative `max_calls` raises `ValueError`.
- In `Toolkit.call_tool_function(...)`, enforce call limit:
  - if limit exceeded, return `ToolCallLimitError`
  - otherwise increment `call_count` before execution.
- Added tests in `tests/tool_max_calls_test.py`:
  - basic limit behavior
  - negative validation
  - zero limit
  - unlimited mode
  - call count increment

How to test:
```bash
cd /Users/gaozexun/project/idea/agentscope
git switch codex/issue-1362-max-calls
.venv/bin/python -m pytest tests/tool_max_calls_test.py tests/toolkit_basic_test.py -q
```

Checklist
 Code has been formatted with pre-commit run --all-files command
 All tests are passing
 Docstrings are in Google style
 Related documentation has been updated (e.g. links, examples, etc.)
 Code is ready for review


gaozexun:
截屏2026-04-14 22 56 50

贴一张图，能够正确执行。超过调用次数就不再执行工具了

0xbrainkid:
The `max_calls` limit is a practical safety net for ReAct loops — capping tool invocations prevents both infinite loops and runaway resource consumption.

One extension worth adding to the call count tracking: **call pattern as a behavioral trust signal.** An agent that consistently hits `max_calls` limits across sessions is exhibiting a behavioral pattern distinct from an agent that completes tasks efficiently. Tracking per-tool call efficiency (tasks_completed / tool_calls) as a behavioral metric gives orchestrators an early signal that an agent is over-relying on a particular tool — which can indicate poor task decomposition, prompt drift, or adversarial manipulation.

A minimal addition to the `RegisteredToolFunction` tracking:

```python
@dataclass
class RegisteredToolFunction:
    # Existing
    func: Callable
    max_calls: int | None = None
    call_count: int = 0
    
    # Behavioral signal additions:
    sessions_hit_limit: int = 0      # how often this tool has been capped
    limit_hit_rate: float = 0.0      # sessions_hit_limit / total_sessions
```

`limit_hit_rate` above a threshold (e.g., >30% of sessions hitting the cap) is a behavioral anomaly worth flagging — either the `max_calls` is too low for legitimate use, or the agent is behaving unexpectedly. This data could feed into an AgentFolio trust score update without requiring any external API call — the local count accumulates, and periodic sync to the registry propagates it.

DavdGao:
Thanks for your contribution! I have some concerns regarding the lifecycle and design of `max_calls`:

1. **Lifecycle clarity**: In deployment scenarios such as a FastAPI service, since `max_calls` is not part of the agent's persistent state, the call count will only be effective within a single `reply` invocation and will be reset on the next one. However, the error message returned to the agent when the limit is exceeded doesn't make this clear — it reads more like the tool has been permanently and globally disabled, even for subsequent `reply` calls.

2. **Semantics of `max_calls`**: Should `max_calls` represent the maximum number of *consecutive* calls to a tool, rather than the total number of calls within a session? 

3. **Usability concern**: As an agent application developer, I'm honestly not sure which tools this parameter should be applied to. For example, calling `read_file` multiple times — or even consecutively — within a single `reply` is perfectly reasonable. It seems difficult to determine upfront which tools are likely to be over-invoked by the agent, making it hard to know where to set this limit proactively.

4. **Overlapping functionality**: What is the expected agent behavior once `max_calls` is exceeded — should it fall back to other tools, or respond directly to the user? From this perspective, does `max_calls` overlap in functionality with the `max_iters` parameter already available in `ReActAgent`? If so, it might be worth clarifying how these two parameters are intended to complement each other.

gaozexun:
> Thanks for your contribution! I have some concerns regarding the lifecycle and design of `max_calls`:
> 
> 1. **Lifecycle clarity**: In deployment scenarios such as a FastAPI service, since `max_calls` is not part of the agent's persistent state, the call count will only be effective within a single `reply` invocation and will be reset on the next one. However, the error message returned to the agent when the limit is exceeded doesn't make this clear — it reads more like the tool has been permanently and globally disabled, even for subsequent `reply` calls.
> 2. **Semantics of `max_calls`**: Should `max_calls` represent the maximum number of _consecutive_ calls to a tool, rather than the total number of calls within a session?
> 3. **Usability concern**: As an agent application developer, I'm honestly not sure which tools this parameter should be applied to. For example, calling `read_file` multiple times — or even consecutively — within a single `reply` is perfectly reasonable. It seems difficult to determine upfront which tools are likely to be over-invoked by the agent, making it hard to know where to set this limit proactively.
> 4. **Overlapping functionality**: What is the expected agent behavior once `max_calls` is exceeded — should it fall back to other tools, or respond directly to the user? From this perspective, does `max_calls` overlap in functionality with the `max_iters` parameter already available in `ReActAgent`? If so, it might be worth clarifying how these two parameters are intended to complement each other.

Hi @author,

Thank you for your valuable feedback! I really appreciate your insights. After reviewing your comments, I realize this issue is more complex than I initially thought, and my current PR does have some gaps.

Let me share my thoughts and see if we can align on the approach. I'm happy to revise my PR based on our discussion.

## Context: Why This PR?

This PR addresses Issue #1362, which resonates with my own experience at work.

### The Problem I Faced

In production environments, I've observed that tool calls can sometimes enter long loops, causing extreme latency spikes in our monitoring dashboards. This is particularly problematic for:

1. **Query tools** (like search, database queries) - These don't typically need to be called repeatedly in a single conversation, but without limits, ReAct might keep calling them with adjusted parameters, creating latency spikes.

2. **Action tools** - Without execution limits, ReAct might continuously retry with different parameters (potentially even entering infinite loops in edge cases), which poses a risk in production scenarios.

### My Use Case

From a production stability and latency perspective, I believe there should be a parameter to limit tool call counts within a single conversation turn. This would:
- Prevent runaway loops
- Reduce unexpected latency spikes
- Provide a safety net for production deployments

Just as an interface needs a threshold to prevent frequent calls, an Agent also needs stability. Therefore, both the query tool and the action tool should have a threshold.

gaozexun:
1 and 2. I think  the maximum number of executions for a tool is appropriate. 

3. Most tools don't need to consider this. However, in a production environment, business-related tools and agents need to consider maximum call limits to avoid duplicate calls or creating a loop of activity. Duplicate calls to action tools may cause stability risks.

5. I think there is overlap:

4.1 The necessity of `max_call` is to control the number of times the tool executes. If the number is exceeded, it should not exit, but ReActAgent should be allowed to continue executing other tools.

4.2 `max_iters` prevents the entire Agent from looping, causing it to fail quickly.

4.3 In other words, we can set `max_iters` to prevent ReAct loops, and we can also use `max_call` to combine fine-grained control of tool stability while ensuring the stability of ReAct!

PR #1520, opened 2026-04-22
fix(memory): add mem0ai 2.0 API compatibility for Mem0LongTermMemory

### LINKED ISSUE #1511: [Bug] Mem0 2.0 API breaking change causes Mem0LongTermMemory to fail in AgentScope
# [Bug] Mem0AI 2.0 API breaking change causes Mem0LongTermMemory to fail in AgentScope

## 🐛 Description

After upgrading `mem0ai` from 1.x to 2.0, the `Mem0LongTermMemory` class in AgentScope no longer works properly. The root cause is that `mem0ai` 2.0 introduced a breaking change to the `search()` method: it no longer supports passing `user_id`, `agent_id`, `run_id` as top-level parameters. Instead, these must be provided within a `filters` dictionary.

## 🔍 Steps to Reproduce

1. Install `mem0ai` 2.0 or later:
   ```bash
   pip install mem0ai>=2.0.0
   ```

2. Use `Mem0LongTermMemory` in AgentScope:
   ```python
   from agentscope.memory import Mem0LongTermMemory
   
   memory = Mem0LongTermMemory(
       user_name="test_user",
       agent_name="test_agent",
       # ... other configs
   )
   
   # Execute a retrieval operation
   await memory.retrieve(query_msg)
   ```

3. Observe the error:
   ```
   "text": "Error retrieving memory: Top-level entity parameters frozenset({'user_id', 'run_id', 'agent_id'}) are not supported in search(). Use filters={'user_id': '...'} instead."
   ```

## 📌 Expected Behavior

`Mem0LongTermMemory` should work seamlessly with `mem0ai` 2.0's updated API and successfully execute retrieval operations.

## 🛠️ Root Cause Analysis

According to `mem0ai` 2.0 documentation, the API has changed as follows:

| Version | Call Style |
|---------|-------------|
| v1.x | `client.search(query, user_id="alice")` |
| v2.0 | `client.search(query, filters={"user_id": "alice"})` |

For scenarios with multiple entity parameters, v2.0 requires using `AND`/`OR` logical operators:

```python
filters = {
    "AND": [
        {"user_id": "traveler_cam"},
        {"agent_id": "travel_planner"},
        {"run_id": "tokyo-2025-weekend"}
    ]
}
```

AgentScope's `Mem0LongTermMemory` implementation currently uses the v1.x parameter passing style and needs to be adapted.

## 📁 Affected Files

Based on AgentScope's code structure, the issue likely exists in the `retrieve()` method of the `Mem0LongTermMemory` class, where it internally calls `mem0ai`'s `search()` operation.

## 💡 Proposed Fix

Modify the `search()` call in `Mem0LongTermMemory` to convert top-level parameters into a `filters` dictionary:

**Before (v1.x style):**
```python
# Current implementation likely resembles
result = self.mem0_client.search(query, user_id=user_id, agent_id=agent_id, run_id=run_id)
```

**After (v2.0 style):**
```python
# Build filters dictionary
filters_conditions = []
if user_id:
    filters_conditions.append({"user_id": user_id})
if agent_id:
    filters_conditions.append({"agent_id": agent_id})
if run_id:
    filters_conditions.append({"run_id": run_id})

if len(filters_conditions) == 1:
    filters = filters_conditions[0]
elif len(filters_conditions) > 1:
    filters = {"AND": filters_conditions}
else:
    filters = None

result = self.mem0_client.search(query, filters=filters)
```

## 🔄 Compatibility Consideration

Since some users may still be using `mem0ai` 1.x, consider adding version detection for backward compatibility:

```python
from packaging import version
import mem0ai

if version.parse(mem0ai.__version__) >= version.parse("2.0.0"):
    # Use new filters approach
    ...
else:
    # Use old top-level parameters approach
    ...
```

## 📋 Environment

- AgentScope version: latest
- `mem0ai` version: 2.0.0+
- Python version: 3.x

## 🔗 References

- Mem0AI Entity Partitioning Documentation: [Partition Memories by Entity](https://docs.mem0.ai/cookbooks/essentials/entity-partitioning-playbook)
- AgentScope Mem0LongTermMemory Integration: [Long-Term Memory Integration](https://deepwiki.com/agentscope-ai/agentscope/5.5-embedding-system)

octo-patch:
Fixes #1511

## Problem

mem0ai 2.0 introduced a breaking change to both `search()` and `add()`: entity identifiers (`user_id`, `agent_id`, `run_id`) are no longer accepted as top-level parameters. They must now be passed inside a `filters` dict:

| Version | Call Style |
|---------|------------|
| v1.x | `client.search(query, user_id="alice")` |
| v2.0 | `client.search(query, filters={"user_id": "alice"})` |

This caused `Mem0LongTermMemory` to raise:
```
Error retrieving memory: Top-level entity parameters frozenset({'user_id', 'run_id', 'agent_id'}) are not supported in search(). Use filters={'user_id': '...'} instead.
```

## Solution

- Add `_detect_mem0_v2()` static method to detect the installed mem0ai version.
- Add `_build_entity_filters()` instance method that returns the correct kwargs format:
  - mem0ai >= 2.0.0: `{"filters": {"user_id": ..., ...}}` (with `AND` for multiple identifiers)
  - mem0ai < 2.0.0: `{"user_id": ..., "agent_id": ..., "run_id": ...}`
- Update all three call sites (`retrieve()`, `retrieve_from_memory()`, `_mem0_record()`) to use `_build_entity_filters()`.

This preserves full backward compatibility with older mem0ai versions.

## Testing

Verified the logic manually against the mem0ai 2.0 documentation. The version check uses the existing `packaging` dependency already imported in this module.


DavdGao:
Closing for 2.0 branch migration. Feel free to create a new pr on the v1 branch if needed

PR #1554, opened 2026-05-01
fix: prevent double end signal in stream_printing_messages and add NotImplementedError in UserAgent.observe

Ricardo-M-L:
## Summary

**Bug 1 - `stream_printing_messages` double end signal:**
When `task.done()` was already true, the code put the end signal then registered a done_callback that also fired immediately — causing a double put. Fix: only register callback when task is not done, then unconditionally put end_signal.

**Bug 2 - `UserAgent.observe` empty implementation:**
Method declared `async` with no body, silently doing nothing. Unlike `handle_interrupt` which raises `NotImplementedError`, this made silent failures possible. Fix: raise `NotImplementedError`.

## Test plan
- [ ] Verify stream_printing_messages only emits one end signal
- [ ] Verify calling observe() on UserAgent raises NotImplementedError


DavdGao:
Closing for 2.0 branch migration. Feel free to create a new pr on the v1 branch if needed

PR #1566, opened 2026-05-06
feat(sandbox): Add the sandbox module in agentscope 2.0

cuiyuebing:
## PR Title Format

## AgentScope Version

## Description

Add sandbox

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/sandbox/__init__.py
+++ b/src/agentscope/sandbox/__init__.py
@@ -5,7 +5,8 @@ Three layers in one package:
 
 Layer 1 (bottom):  ``SandboxConnection`` — backend primitives
                    (exec, read, write, destroy, ...)
-                   + module-level ``create_connection`` registry for dispatch.
+                   + module-level ``create_sandbox_connection`` registry
+                   for dispatch.
 Layer 2 (middle):  ``Sandbox`` — agent-facing proxy
                    (tools, skills, MCP, file facade).
 Layer 3 (top):     ``SandboxManager`` — session lifecycle, pooling.
@@ -14,40 +15,42 @@ Layer 3 (top):     ``SandboxManager`` — session lifecycle, pooling.
 # --- Layer 1: backend primitives ---
 from .connection import (
     SandboxConnection,
-    create_connection,
-    get_connection_class,
-    register_connection_class,
+    create_sandbox_connection,
+    get_sandbox_connection_type,
+    register_sandbox_connection_type,
 )
 from .exceptions import CapabilityError, SandboxError, UnsupportedOperation
 from .types import (
-    ExecResult,
-    ExposedPortEndpoint,
+    SandboxExecutionResult,
+    SandboxInternalEndpoint,
     SandboxConnectionCapabilities,
-    SandboxCreateOptions,
+    SandboxInitializationConfig,
     SerializedSandboxState,
 )
 
 # --- Config ---
-from .config import (
+from .backend_config import (
     BackendParams,
     DockerBackendParams,
     E2BBackendParams,
     LocalBackendParams,
-    McpGatewayConfig,
-    McpServerConfig,
-    SandboxConfig,
-    SkillConfig,
-    ToolDef,
 )
+from .config import (
+    MCPServerConfig,
+    SandboxConfig,
+    SkillsConfig,
+    ToolDefinition,
+)
+from .mcp_gateway import MCPGatewayConfig
 
 # --- Layer 2: agent-side proxy ---
 from .sandbox import FileAccessor, Sandbox
 
 # --- Layer 2.5: MCP gateway ---
-from .gateway import MCPGateway
+from .mcp_gateway import MCPGateway
 
 # --- Layer 3: manager ---
-from .manager import SandboxManager
+from .sandbox_manager import SandboxManager
 
 # --- Backends (register themselves on import) ---
 from . import local_temp as _local_temp  # noqa: F401
@@ -68,26 +71,27 @@ __all__ = [
     "SandboxError",
     "UnsupportedOperation",
     # types
-    "ExecResult",
-    "ExposedPortEndpoint",
+    "SandboxExecutionResult",
+    "SandboxInternalEndpoint",
     "SandboxConnectionCapabilities",
-    "SandboxCreateOptions",
+    "SandboxInitializationConfig",
     "SerializedSandboxState",
     # layer 1
     "SandboxConnection",
-    "create_connection",
-    "get_connection_class",
-    "register_connection_class",
-    # config
+    "create_sandbox_connection",
+    "get_sandbox_connection_type",
+    "register_sandbox_connection_type",
+    # backend config
     "BackendParams",
     "DockerBackendParams",
     "E2BBackendParams",
     "LocalBackendParams",
-    "McpGatewayConfig",
-    "McpServerConfig",
+    # sandbox config
+    "MCPGatewayConfig",
+    "MCPServerConfig",
     "SandboxConfig",
-    "SkillConfig",
-    "ToolDef",
+    "SkillsConfig",
+    "ToolDefinition",
     # layer 2
     "FileAccessor",
     "Sandbox",
--- a/src/agentscope/sandbox/__init__.py
+++ b/src/agentscope/sandbox/__init__.py
@@ -8,8 +8,7 @@ Layer 1 (bottom):  ``SandboxConnection`` — backend primitives
                    + module-level ``create_connection`` registry for dispatch.
 Layer 2 (middle):  ``Sandbox`` — agent-facing proxy
                    (tools, skills, MCP, file facade).
-Layer 3 (top):     ``SandboxManager`` + ``SandboxPool`` — session lifecycle
-                   & pooling.
+Layer 3 (top):     ``SandboxManager`` — session lifecycle, pooling.
 """
 
 # --- Layer 1: backend primitives ---
@@ -47,8 +46,8 @@ from .sandbox import FileAccessor, Sandbox
 # --- Layer 2.5: MCP gateway ---
 from .gateway import MCPGateway
 
-# --- Layer 3: manager & pool ---
-from .manager import SandboxManager, SandboxPool
+# --- Layer 3: manager ---
+from .manager import SandboxManager
 
 # --- Backends (register themselves on import) ---
 from . import local_temp as _local_temp  # noqa: F401
@@ -57,6 +56,11 @@ try:
     from .docker import DockerSandboxConnection  # noqa: F401
 except ImportError:
     DockerSandboxConnection = None  # type: ignore[assignment,misc]
+
+try:
+    from .e2b import E2BSandboxConnection  # noqa: F401
+except ImportError:
+    E2BSandboxConnection = None  # type: ignore[assignment,misc]
 
 __all__ = [
     # exceptions
@@ -91,7 +95,7 @@ __all__ = [
     "MCPGateway",
     # layer 3
     "SandboxManager",
-    "SandboxPool",
     # backends
     "DockerSandboxConnection",
+    "E2BSandboxConnection",
 ]
--- a/src/agentscope/sandbox/__init__.py
+++ b/src/agentscope/sandbox/__init__.py
@@ -13,20 +13,8 @@ Layer 3 (top):     ``SandboxManager`` — session lifecycle, pooling.
 """
 
 # --- Layer 1: backend primitives ---
-from .connection import (
-    SandboxConnection,
-    create_sandbox_connection,
-    get_sandbox_connection_type,
-    register_sandbox_connection_type,
-)
-from .exceptions import CapabilityError, SandboxError, UnsupportedOperation
-from .types import (
-    SandboxExecutionResult,
-    SandboxInternalEndpoint,
-    SandboxConnectionCapabilities,
-    SandboxInitializationConfig,
-    SerializedSandboxState,
-)
+# --- Backends (register themselves on import) ---
+from . import local_temp as _local_temp  # noqa: F401
 
 # --- Config ---
 from .backend_config import (
@@ -41,19 +29,29 @@ from .config import (
     SkillsConfig,
     ToolDefinition,
 )
-from .mcp_gateway import MCPGatewayConfig
+from .connection import (
+    SandboxConnection,
+    create_sandbox_connection,
+    get_sandbox_connection_type,
+    register_sandbox_connection_type,
+)
+from .exceptions import CapabilityError, SandboxError, UnsupportedOperation
+
+# --- Layer 2.5: MCP gateway ---
+from .mcp_gateway import MCPGateway, MCPGatewayConfig
 
 # --- Layer 2: agent-side proxy ---
 from .sandbox import FileAccessor, Sandbox
 
-# --- Layer 2.5: MCP gateway ---
-from .mcp_gateway import MCPGateway
-
 # --- Layer 3: manager ---
 from .sandbox_manager import SandboxManager
-
-# --- Backends (register themselves on import) ---
-from . import local_temp as _local_temp  # noqa: F401
+from .types import (
+    SandboxConnectionCapabilities,
+    SandboxExecutionResult,
+    SandboxInitializationConfig,
+    SandboxInternalEndpoint,
+    SerializedSandboxState,
+)
 
 try:
     from .docker import DockerSandboxConnection  # noqa: F401
--- a/src/agentscope/sandbox/__init__.py
+++ b/src/agentscope/sandbox/__init__.py
@@ -1,0 +1,96 @@
+"""Sandbox layer for AgentScope.
+
+Three layers in one package:
+
+Layer 1 (bottom):  ``SandboxConnection`` — backend primitives
+                   (exec, read, write, destroy, ...)
+                   + module-level ``create_connection`` registry for dispatch.
+Layer 2 (middle):  ``Sandbox`` — agent-facing proxy
+                   (tools, skills, MCP, file facade).
+Layer 3 (top):     ``SandboxManager`` + ``SandboxPool`` — session lifecycle
+                   & pooling.
+"""
+
+# --- Layer 1: backend primitives ---
+from .connection import (
+    SandboxConnection,
+    create_connection,
+    get_connection_class,
+    register_connection_class,
+)
+from .exceptions import CapabilityError, SandboxError, UnsupportedOperation
+from .types import (
+    ExecResult,
+    ExposedPortEndpoint,
+    SandboxConnectionCapabilities,
+    SandboxCreateOptions,
+    SerializedSandboxState,
+)
+
+# --- Config ---
+from .config import (
+    BackendParams,
+    DockerBackendParams,
+    E2BBackendParams,
+    LocalBackendParams,
+    McpGatewayConfig,
+    McpServerConfig,
+    SandboxConfig,
+    SkillConfig,
+    ToolDef,
+)
+
+# --- Layer 2: agent-side proxy ---
+from .sandbox import FileAccessor, Sandbox
+
+# --- Layer 2.5: MCP gateway ---
+from .gateway import MCPGateway
+
+# --- Layer 3: manager & pool ---
+from .manager import SandboxManager, SandboxPool
+
+# --- Backends (register themselves on import) ---
+from . import local_temp as _local_temp  # noqa: F401
+
+try:
+    from .docker import DockerSandboxConnection  # noqa: F401
+except ImportError:
+    DockerSandboxConnection = None  # type: ignore[assignment,misc]
+
+__all__ = [
+    # exceptions
+    "CapabilityError",
+    "SandboxError",
+    "UnsupportedOperation",
+    # types
+    "ExecResult",
+    "ExposedPortEndpoint",
+    "SandboxConnectionCapabilities",
+    "SandboxCreateOptions",
+    "SerializedSandboxState",
+    # layer 1
+    "SandboxConnection",
+    "create_connection",
+    "get_connection_class",
+    "register_connection_class",
+    # config
+    "BackendParams",
+    "DockerBackendParams",
+    "E2BBackendParams",
+    "LocalBackendParams",
+    "McpGatewayConfig",
+    "McpServerConfig",
+    "SandboxConfig",
+    "SkillConfig",
+    "ToolDef",
+    # layer 2
+    "FileAccessor",
+    "Sandbox",
+    # gateway
+    "MCPGateway",
+    # layer 3
+    "SandboxManager",
+    "SandboxPool",
+    # backends
+    "DockerSandboxConnection",
+]
--- a/src/agentscope/sandbox/__init__.py
+++ b/src/agentscope/sandbox/__init__.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """Sandbox layer for AgentScope.
 
 Three layers in one package:
--- a/src/agentscope/sandbox/config.py
+++ b/src/agentscope/sandbox/config.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """Single-source-of-truth configuration for one Sandbox instance.
 
 Users write *one* ``SandboxConfig``; the Sandbox layer internally merges
@@ -9,10 +10,10 @@ from __future__ import annotations
 from dataclasses import dataclass, field
 from typing import Any
 
-
 # ---------------------------------------------------------------------------
 # Backend parameter sets
 # ---------------------------------------------------------------------------
+
 
 @dataclass(slots=True)
 class BackendParams:
@@ -24,12 +25,16 @@ class BackendParams:
 
 @dataclass(slots=True)
 class DockerBackendParams(BackendParams):
+    """Parameters for the Docker ``SandboxConnection`` backend."""
+
     type: str = "docker"
     image: str = "ubuntu:22.04"
 
 
 @dataclass(slots=True)
 class E2BBackendParams(BackendParams):
+    """Placeholder backend params for future E2B integration."""
+
     type: str = "e2b"
     template: str = "base"
     api_key: str = ""
@@ -37,6 +42,8 @@ class E2BBackendParams(BackendParams):
 
 @dataclass(slots=True)
 class LocalBackendParams(BackendParams):
+    """Parameters for the local temp-dir ``SandboxConnection`` backend."""
+
     type: str = "local_temp"
     base_dir: str = "/tmp"
 
@@ -44,6 +51,7 @@ class LocalBackendParams(BackendParams):
 # ---------------------------------------------------------------------------
 # MCP / Skills / Tools
 # ---------------------------------------------------------------------------
+
 
 @dataclass(slots=True)
 class McpServerConfig:
@@ -57,7 +65,7 @@ class McpServerConfig:
 
 @dataclass(slots=True)
 class McpGatewayConfig:
-    """Aggregate MCP servers behind one port; auto-merged into exposed_ports."""
+    """Gateway settings (listen port merged into ``exposed_ports``)."""
 
     enabled: bool = False
     port: int = 5600
@@ -66,6 +74,8 @@ class McpGatewayConfig:
 
 @dataclass(slots=True)
 class SkillConfig:
+    """Where skills live in the sandbox and optional host bind-mount."""
+
     skills_dir: str = "/root/skills"
     persist: bool = False
     host_dir: str | None = None
@@ -84,6 +94,7 @@ class ToolDef:
 # ---------------------------------------------------------------------------
 # Top-level config
 # ---------------------------------------------------------------------------
+
 
 @dataclass(slots=True)
 class SandboxConfig:
--- a/src/agentscope/sandbox/config.py
+++ b/src/agentscope/sandbox/config.py
@@ -5,8 +5,6 @@ Users write *one* ``SandboxConfig``; the Sandbox layer internally merges
 implied ports / volumes / env before handing them to the Connection.
 """
 
-from __future__ import annotations
-
 from dataclasses import dataclass, field
 from typing import Any
 
@@ -17,7 +15,14 @@ from typing import Any
 
 @dataclass(slots=True)
 class BackendParams:
-    """Base for all backend parameter sets."""
+    """User-facing backend config — typed fields for each vendor.
+
+    ``Sandbox._merge_infra_requirements()`` flattens these into a
+    vendor-neutral ``SandboxCreateOptions`` (with an opaque ``extra``
+    dict) before passing them to ``SandboxConnection.create()``.
+    This keeps the user config statically typed while the connection
+    layer stays backend-agnostic.
+    """
 
     type: str
     extra: dict[str, Any] = field(default_factory=dict)
@@ -33,11 +38,15 @@ class DockerBackendParams(BackendParams):
 
 @dataclass(slots=True)
 class E2BBackendParams(BackendParams):
-    """Placeholder backend params for future E2B integration."""
+    """Parameters for the E2B ``SandboxConnection`` backend."""
 
     type: str = "e2b"
     template: str = "base"
     api_key: str = ""
+    domain: str = ""
+    timeout: int = 300
+    envs: dict[str, str] = field(default_factory=dict)
+    metadata: dict[str, str] = field(default_factory=dict)
 
 
 @dataclass(slots=True)
@@ -83,7 +92,14 @@ class SkillConfig:
 
 @dataclass(slots=True)
 class ToolDef:
-    """Static tool definition that gets registered at sandbox start."""
+    """Static tool registered at sandbox start.
+
+    ``handler`` is a shell command executed inside the sandbox when
+    :meth:`Sandbox.call_tool` is invoked. The tool arguments are
+    serialized as JSON and appended to the command, e.g.
+    ``echo '{"msg": "hi"}'``.  When ``handler`` is ``None`` the tool
+    is metadata-only and cannot be called.
+    """
 
     name: str
     description: str = ""
--- a/src/agentscope/sandbox/config.py
+++ b/src/agentscope/sandbox/config.py
@@ -52,13 +52,34 @@ class ToolDefinition:
     shell_cmd: str | None = None
 
 
+_DEFAULT_SANDBOX_INSTRUCTIONS = """<workspace>
+You have access to a sandboxed execution environment (backend: {backend_type}).
+
+All commands run **inside the sandbox** — not on the host. Use the tools
+and ``exec`` interface to interact with the sandbox filesystem and processes.
+
+### Filesystem layout
+```
+/root/
+├── skills/      # reusable skills (when configured)
+└── sessions/    # offloaded context and tool results
+```
+
+### Key capabilities
+- Execute shell commands via ``exec``.
+- Read / write files via ``sandbox.file.read`` / ``sandbox.file.write``.
+- Tools registered at sandbox start are callable via ``call_tool``.
+- MCP servers (if configured) are aggregated by the gateway.
+</workspace>"""
+
+
 @dataclass(slots=True)
 class SandboxConfig:
     """One config drives everything: backend + tool/skill/MCP + env.
 
     ``backend.type`` selects the registered ``SandboxConnection`` subclass.
     Ports, volumes, and env that are *implied* by MCP gateway / skills are
-    merged automatically in ``Sandbox.start()``.
+    merged automatically in ``Sandbox.initialize()``.
 
     ``endpoint`` is optional (remote tunnel / control-plane URL); merged into
     ``SandboxInitializationConfig.extra`` for providers that need it.
@@ -67,6 +88,7 @@ class SandboxConfig:
     backend: BackendParams
 
     endpoint: str | None = None
+    instructions: str = _DEFAULT_SANDBOX_INSTRUCTIONS
 
     exposed_ports: list[int] = field(default_factory=list)
     volumes: dict[str, str] = field(default_factory=dict)
--- a/src/agentscope/sandbox/config.py
+++ b/src/agentscope/sandbox/config.py
@@ -1,0 +1,113 @@
+"""Single-source-of-truth configuration for one Sandbox instance.
+
+Users write *one* ``SandboxConfig``; the Sandbox layer internally merges
+implied ports / volumes / env before handing them to the Connection.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any
+
+
+# ---------------------------------------------------------------------------
+# Backend parameter sets
+# ---------------------------------------------------------------------------
+
+@dataclass(slots=True)
+class BackendParams:
+    """Base for all backend parameter sets."""
+
+    type: str
+    extra: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass(slots=True)
+class DockerBackendParams(BackendParams):
+    type: str = "docker"
+    image: str = "ubuntu:22.04"
+
+
+@dataclass(slots=True)
+class E2BBackendParams(BackendParams):
+    type: str = "e2b"
+    template: str = "base"
+    api_key: str = ""
+
+
+@dataclass(slots=True)
+class LocalBackendParams(BackendParams):
+    type: str = "local_temp"
+    base_dir: str = "/tmp"
+
+
+# ---------------------------------------------------------------------------
+# MCP / Skills / Tools
+# ---------------------------------------------------------------------------
+
+@dataclass(slots=True)
+class McpServerConfig:
+    """One MCP server to start inside the sandbox."""
+
+    name: str
+    command: str
+    args: list[str] = field(default_factory=list)
+    env: dict[str, str] = field(default_factory=dict)
+
+
+@dataclass(slots=True)
+class McpGatewayConfig:
+    """Aggregate MCP servers behind one port; auto-merged into exposed_ports."""
+
+    enabled: bool = False
+    port: int = 5600
+    mcp_name: str = "sandbox"
+
+
+@dataclass(slots=True)
+class SkillConfig:
+    skills_dir: str = "/root/skills"
+    persist: bool = False
+    host_dir: str | None = None
+
+
+@dataclass(slots=True)
+class ToolDef:
+    """Static tool definition that gets registered at sandbox start."""
+
+    name: str
+    description: str = ""
+    parameters: dict[str, Any] = field(default_factory=dict)
+    handler: str | None = None
+
+
+# ---------------------------------------------------------------------------
+# Top-level config
+# ---------------------------------------------------------------------------
+
+@dataclass(slots=True)
+class SandboxConfig:
+    """One config drives everything: backend + tool/skill/MCP + env.
+
+    ``backend.type`` selects the registered ``SandboxConnection`` subclass.
+    Ports, volumes, and env that are *implied* by MCP gateway / skills are
+    merged automatically in ``Sandbox.start()``.
+
+    ``endpoint`` is optional (remote tunnel / control-plane URL); merged into
+    ``SandboxCreateOptions.extra`` for providers that need it.
+    """
+
+    backend: BackendParams
+
+    endpoint: str | None = None
+
+    exposed_ports: list[int] = field(default_factory=list)
+    volumes: dict[str, str] = field(default_factory=dict)
+    env: dict[str, str] = field(default_factory=dict)
+    startup_commands: list[str] = field(default_factory=list)
+
+    mcp_servers: list[McpServerConfig] = field(default_factory=list)
+    mcp_gateway: McpGatewayConfig = field(default_factory=McpGatewayConfig)
+
+    skills: SkillConfig | None = None
+    tools: list[ToolDef] = field(default_factory=list)
--- a/src/agentscope/sandbox/config.py
+++ b/src/agentscope/sandbox/config.py
@@ -13,12 +13,29 @@ from .mcp_gateway import MCPGatewayConfig
 
 @dataclass(slots=True)
 class MCPServerConfig:
-    """One MCP server to start inside the sandbox."""
+    """One MCP server to start inside the sandbox.
+
+    Supports two transport types:
+    - ``"stdio"`` (default): spawns a local process via command + args.
+    - ``"http"``: connects to a remote MCP server via HTTP (SSE or
+      StreamableHTTP).
+
+    For stdio: ``command`` and ``args`` are required.
+    For http: ``url`` is required; ``headers`` and ``timeout`` are optional.
+    """
 
     name: str
-    command: str
+    transport: str = "stdio"  # "stdio" | "http"
+
+    # stdio fields
+    command: str = ""
     args: list[str] = field(default_factory=list)
     env: dict[str, str] = field(default_factory=dict)
+
+    # http fields
+    url: str = ""
+    headers: dict[str, str] = field(default_factory=dict)
+    timeout: float = 30.0
 
 
 @dataclass(slots=True)
--- a/src/agentscope/sandbox/config.py
+++ b/src/agentscope/sandbox/config.py
@@ -1,69 +1,18 @@
 # -*- coding: utf-8 -*-
-"""Single-source-of-truth configuration for one Sandbox instance.
+"""Configuration for the logical Sandbox layer.
 
-Users write *one* ``SandboxConfig``; the Sandbox layer internally merges
-implied ports / volumes / env before handing them to the Connection.
+Backend-specific parameters live in :mod:`.backend_config`.
 """
 
 from dataclasses import dataclass, field
 from typing import Any
 
-# ---------------------------------------------------------------------------
-# Backend parameter sets
-# ---------------------------------------------------------------------------
+from .backend_config import BackendParams
+from .mcp_gateway import MCPGatewayConfig
 
 
 @dataclass(slots=True)
-class BackendParams:
-    """User-facing backend config — typed fields for each vendor.
-
-    ``Sandbox._merge_infra_requirements()`` flattens these into a
-    vendor-neutral ``SandboxCreateOptions`` (with an opaque ``extra``
-    dict) before passing them to ``SandboxConnection.create()``.
-    This keeps the user config statically typed while the connection
-    layer stays backend-agnostic.
-    """
-
-    type: str
-    extra: dict[str, Any] = field(default_factory=dict)
-
-
-@dataclass(slots=True)
-class DockerBackendParams(BackendParams):
-    """Parameters for the Docker ``SandboxConnection`` backend."""
-
-    type: str = "docker"
-    image: str = "ubuntu:22.04"
-
-
-@dataclass(slots=True)
-class E2BBackendParams(BackendParams):
-    """Parameters for the E2B ``SandboxConnection`` backend."""
-
-    type: str = "e2b"
-    template: str = "base"
-    api_key: str = ""
-    domain: str = ""
-    timeout: int = 300
-    envs: dict[str, str] = field(default_factory=dict)
-    metadata: dict[str, str] = field(default_factory=dict)
-
-
-@dataclass(slots=True)
-class LocalBackendParams(BackendParams):
-    """Parameters for the local temp-dir ``SandboxConnection`` backend."""
-
-    type: str = "local_temp"
-    base_dir: str = "/tmp"
-
-
-# ---------------------------------------------------------------------------
-# MCP / Skills / Tools
-# ---------------------------------------------------------------------------
-
-
-@dataclass(slots=True)
-class McpServerConfig:
+class MCPServerConfig:
     """One MCP server to start inside the sandbox."""
 
     name: str
@@ -73,16 +22,7 @@ class McpServerConfig:
 
 
 @dataclass(slots=True)
-class McpGatewayConfig:
-    """Gateway settings (listen port merged into ``exposed_ports``)."""
-
-    enabled: bool = False
-    port: int = 5600
-    mcp_name: str = "sandbox"
-
-
-@dataclass(slots=True)
-class SkillConfig:
+class SkillsConfig:
     """Where skills live in the sandbox and optional host bind-mount."""
 
     skills_dir: str = "/root/skills"
@@ -91,25 +31,25 @@ class SkillConfig:
 
 
 @dataclass(slots=True)
-class ToolDef:
+class ToolDefinition:
     """Static tool registered at sandbox start.
 
-    ``handler`` is a shell command executed inside the sandbox when
-    :meth:`Sandbox.call_tool` is invoked. The tool arguments are
-    serialized as JSON and appended to the command, e.g.
-    ``echo '{"msg": "hi"}'``.  When ``handler`` is ``None`` the tool
-    is metadata-only and cannot be called.
+    Attributes:
+        name: Unique tool name.
+        description: Human-readable description of what the tool does.
+        parameters: JSON Schema describing the tool's input arguments,
+            used for validation and for exposing the tool via MCP.
+        shell_cmd: Shell command executed inside the sandbox when
+            :meth:`Sandbox.call_tool` is invoked. The tool arguments are
+            serialized as JSON and appended to the command, e.g.
+            ``echo '{"msg": "hi"}'``.  When ``shell_cmd`` is ``None``
+            the tool is metadata-only and cannot be called.
     """
 
     name: str
     description: str = ""
     parameters: dict[str, Any] = field(default_factory=dict)
-    handler: str | None = None
-
-
-# ---------------------------------------------------------------------------
-# Top-level config
-# ---------------------------------------------------------------------------
+    shell_cmd: str | None = None
 
 
 @dataclass(slots=True)
@@ -121,7 +61,7 @@ class SandboxConfig:
     merged automatically in ``Sandbox.start()``.
 
     ``endpoint`` is optional (remote tunnel / control-plane URL); merged into
-    ``SandboxCreateOptions.extra`` for providers that need it.
+    ``SandboxInitializationConfig.extra`` for providers that need it.
     """
 
     backend: BackendParams
@@ -133,8 +73,8 @@ class SandboxConfig:
     env: dict[str, str] = field(default_factory=dict)
     startup_commands: list[str] = field(default_factory=list)
 
-    mcp_servers: list[McpServerConfig] = field(default_factory=list)
-    mcp_gateway: McpGatewayConfig = field(default_factory=McpGatewayConfig)
+    mcp_servers: list[MCPServerConfig] = field(default_factory=list)
+    mcp_gateway: MCPGatewayConfig = field(default_factory=MCPGatewayConfig)
 
-    skills: SkillConfig | None = None
-    tools: list[ToolDef] = field(default_factory=list)
+    skills: SkillsConfig | None = None
+    tools: list[ToolDefinition] = field(default_factory=list)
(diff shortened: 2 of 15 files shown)


src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,425 @@
...
+import asyncio
+import io
+import posixpath
+import tarfile
+import uuid
+from concurrent.futures import ThreadPoolExecutor
+from pathlib import Path, PurePosixPath
+from typing import Any
+
+import docker as docker_sdk
```
DavdGao: 1. Move the docker import to a lazy import to avoid loading it unless required
2. Introduce a sandbox optional dependency group in project.toml and include docker as part of it

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,425 @@
...
+
+    def __init__(
+        self,
+        client: docker_sdk.DockerClient,
+        container: Container,
+        *,
+        instance_id: str,
+        workspace: str = _DEFAULT_WORKSPACE,
+        exposed_ports: dict[int, int] | None = None,
+    ) -> None:
```
DavdGao, line 93: - The __init__ method is missing a docstring, and 
- What does `workspace` represent, and how does it differ from other directory-related concepts?
- Since `exposed_ports` also appears in `SandboxCreateOptions`, please clarify the difference between using the constructor directly vs. the create factory method. If they serve the same purpose, consider consolidating into a single entry point to avoid confusion.

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,549 @@
+# -*- coding: utf-8 -*-
+"""Sandbox — the single agent-facing proxy class.
+
+Lifecycle: ``start()`` → use → ``close()``.  Use as async context manager.
+``close()`` calls ``connection.destroy()`` — full resource cleanup.
+"""
+
+from __future__ import annotations
```
DavdGao: Since this project targets Python 3.11+, `from __future__ import annotations` is unnecessary here and can be removed.

Similar for the other files.

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,549 @@
...
+import uuid
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any
+
+import frontmatter
+import mcp.types as mtypes
+from mcp.types import Tool as McpToolSchema
+
+from agentscope.skill import Skill
```
DavdGao: For consistency, please use relative imports (e.g., from .skill import Skill) for internal dependencies, as seen on lines 24–27.

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,549 @@
...
+from agentscope.tool import MCPTool
+
+from .config import SandboxConfig, ToolDef
+from .connection import SandboxConnection, create_connection
+from .gateway import MCPGateway
+from .types import SandboxCreateOptions
+
+logger = logging.getLogger(__name__)
```
DavdGao: Use the unified logger via `from .._logging import logger`

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,124 @@
...
+@dataclass(slots=True)
+class BackendParams:
+    """Base for all backend parameter sets."""
+
+    type: str
+    extra: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass(slots=True)
+class DockerBackendParams(BackendParams):
```
DavdGao: How do the *Params classes relate to / integrate with `SandboxCreateOptions`?

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@dataclass(slots=True)
+class ToolDef:
+    """Static tool definition that gets registered at sandbox start."""
+
+    name: str
+    description: str = ""
+    parameters: dict[str, Any] = field(default_factory=dict)
+    handler: str | None = None
```
DavdGao: What's the use of this `ToolDef` class and the attribute `handler`?

src/agentscope/sandbox/local_temp.py
```diff
@@ -0,0 +1,173 @@
...
+from .connection import SandboxConnection, register_connection_class
+from .exceptions import UnsupportedOperation
+from .types import (
+    ExecResult,
+    SandboxCreateOptions,
+    SerializedSandboxState,
+)
+
+
+class LocalTempSandboxConnection(SandboxConnection):
```
DavdGao, line 31: What's the intended use case for LocalTempSandboxConnection? If it simply wraps a local directory + async execution, why not skip the sandbox abstraction altogether instead of introducing a dedicated class?

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,197 @@
...
+        await asyncio.gather(
+            *(self.destroy(sid) for sid in ids),
+            return_exceptions=True,
+        )
+
+    # ─── Pool support ─────────────────────────────────────────
+
+    def enable_pool(self, *, warm_size: int = 4) -> SandboxPool:
+        """Create and attach a ``SandboxPool`` to this manager."""
+        self._pool = SandboxPool(manager=self, warm_size=warm_size)
```
DavdGao: The manager and pool classes hold references to each other, which creates a circular dependency. Consider merging them into a single class to simplify the design.

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,425 @@
...
+                _DOCKER_EXECUTOR,
+                self._container.reload,
+            )
+            return self._container.status == "running"
+        except docker.errors.APIError:
+            return False
+
+    # ─── capabilities ─────────────────────────────────────────
+
+    def supports_exposed_ports(self) -> bool:
```
DavdGao: Consider use class attribute directly

DavdGao (changes requested):
Plz see inline comments

src/agentscope/sandbox/local_temp.py
cuiyuebing: In this backend, commands run directly on the host with
full filesystem access inside the temp directory.  This backend exists
to:

1. Exercise the full ``SandboxConnection`` API without requiring
   Docker or E2B, making it useful for unit tests and CI.
2. Serve as a minimal reference implementation for new backend
   authors: every abstract method is implemented in < 10 lines.
3. Support lightweight local workflows (e.g. skill prototyping)
   where container overhead is unnecessary.

src/agentscope/sandbox/config.py
cuiyuebing: ToolDef: The static tool registered

handler: A shell command executed inside the sandbox when
    :meth:`Sandbox.call_tool` is invoked. The tool arguments are
    serialized as JSON and appended to the command, e.g.
    ``echo '{"msg": "hi"}'``.  When ``handler`` is ``None`` the tool
    is metadata-only and cannot be called.

src/agentscope/sandbox/config.py
cuiyuebing: *Params classes provide typed fields for each vendor. ``Sandbox._merge_infra_requirements()`` flattens these into a vendor-neutral ``SandboxCreateOptions`` (with an opaque ``extra`` dict) before passing them to ``SandboxConnection.create()``. This keeps the user config statically typed while the connection layer stays backend-agnostic.

src/agentscope/sandbox/docker.py
cuiyuebing: ```workspace``` is the root working directory for all sandbox operations. ```exec``` defaults ```cwd``` to here; ```read(path)``` and ```write(path)``` resolve relative paths against this directory. 

As for the two exposed_ports, the comparison is as follows:

| | `SandboxCreateOptions.exposed_ports` | `DockerSandboxConnection.__init__` `exposed_ports` |
|---|---|---|
| Type | `list[int]` | `dict[int, int]` |
| When | Before the container is created | After the container is started |
| Represent | "I want these container ports to be exposed" (a wish list) | "Container port X is mapped to host port Y" (the resolved result) |
|Who sets it | The user, via `SandboxConfig.exposed_ports` | The `create()` factory, by inspecting Docker's port assignment |

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+"""Shared value types for sandbox connection + factory layers."""
+
+from dataclasses import dataclass, field
+from typing import Any
+
+
+@dataclass(frozen=True, slots=True)
+class ExecResult:
```
weizhang25: 1. Use complete words in class names. exec -> execution
2. ExecResult is too general. Not self explainable. Maybe SandboxExecutionResult

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
...
+    """Result of executing a command inside a sandbox."""
+
+    exit_code: int
+    stdout: bytes
+    stderr: bytes
+
+    def ok(self) -> bool:
```
weizhang25: nitpick. boolean function naming pattern is more like is_***, is_ok( )

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
...
+        """Return ``True`` if ``exit_code == 0``."""
+        return self.exit_code == 0
+
+
+@dataclass(slots=True)
+class SandboxCreateOptions:
```
weizhang25: nitpick. class SandboxInitializationConfig

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
...
+    """Serializable snapshot for resume / reconnect.
+
+    Attributes:
+        backend: Must match the connection class's ``backend_id`` so
+            ``resume(state)`` can dispatch to the right factory.
+        payload: Opaque to the core layer; typically holds vendor ids
+            (e.g. E2B ``sandbox_id``), workspace root path, tokens, etc.
+    """
+
+    backend: str
```
weizhang25: why not calling it backend_id too, in-line with the name in the connection class?

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@dataclass(slots=True)
+class SerializedSandboxState:
+    """Serializable snapshot for resume / reconnect.
+
+    Attributes:
+        backend: Must match the connection class's ``backend_id`` so
+            ``resume(state)`` can dispatch to the right factory.
+        payload: Opaque to the core layer; typically holds vendor ids
```
weizhang25: Opaque: did you mean "transparent"?

"core layer": whoever calls sandbox.export_state() and keeps such a SSS object.

payload: Transparent to anyone outside of a sandbox instance. Typically ....

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@dataclass(frozen=True, slots=True)
+class SandboxConnectionCapabilities:
+    """Optional features a backend may support.
+
+    Check before calling PTY / networking / snapshot APIs.
+    """
+
+    pty: bool = False
```
weizhang25: Again. Please make these boolean names better.
is/has_pty
exposing_ports
is/has_snapshot

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
...
+    snapshot: bool = False
+
+
+@dataclass(frozen=True, slots=True)
+class ExposedPortEndpoint:
+    """Host endpoint for a logical container port when port mapping exists."""
+
+    host: str
+    port: int
+    tls: bool = False
```
weizhang25: tls : have a better boolean variable name pattern. Plus some description.

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
...
+
+    type: str
+    extra: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass(slots=True)
+class DockerBackendParams(BackendParams):
+    """Parameters for the Docker ``SandboxConnection`` backend."""
+
+    type: str = "docker"
```
weizhang25: nitpick. All the type attributes should be read-only property.

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
...
+    ``handler`` is a shell command executed inside the sandbox when
+    :meth:`Sandbox.call_tool` is invoked. The tool arguments are
+    serialized as JSON and appended to the command, e.g.
+    ``echo '{"msg": "hi"}'``.  When ``handler`` is ``None`` the tool
+    is metadata-only and cannot be called.
+    """
+
+    name: str
+    description: str = ""
+    parameters: dict[str, Any] = field(default_factory=dict)
```
weizhang25, line 68: what is parameters for? Add doc.

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
...
+class SkillConfig:
+    """Where skills live in the sandbox and optional host bind-mount."""
+
+    skills_dir: str = "/root/skills"
+    persist: bool = False
+    host_dir: str | None = None
+
+
+@dataclass(slots=True)
+class ToolDef:
```
weizhang25: Def -> Definition

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
...
+@dataclass(slots=True)
+class E2BBackendParams(BackendParams):
+    """Parameters for the E2B ``SandboxConnection`` backend."""
+
+    type: str = "e2b"
+    template: str = "base"
+    api_key: str = ""
+    domain: str = ""
+    timeout: int = 300
+    envs: dict[str, str] = field(default_factory=dict)
```
weizhang25: Both env and envs are used in these classes. Can we consolidate them into one name pattern?

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
...
+class McpGatewayConfig:
+    """Gateway settings (listen port merged into ``exposed_ports``)."""
+
+    enabled: bool = False
+    port: int = 5600
+    mcp_name: str = "sandbox"
+
+
+@dataclass(slots=True)
+class SkillConfig:
```
weizhang25: SkillsConfig to match the name skills in sandboxconfig

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,209 @@
...
+    async def close(self) -> None:
+        """Soft cleanup: release local handles only.
+
+        Default delegates to ``destroy()``; override if you want a lighter
+        teardown that keeps the remote sandbox alive (e.g. for pool reuse).
+        """
+        await self.destroy()
+
+    @abstractmethod
+    async def running(self) -> bool:
```
weizhang25: def is_running or is_alive

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,464 @@
...
+_DEFAULT_IMAGE = "ubuntu:22.04"
+_DEFAULT_WORKSPACE = "/workspace"
+
+_DOCKER_EXECUTOR = ThreadPoolExecutor(
+    max_workers=8,
+    thread_name_prefix="agentscope-docker-sandbox",
+)
+
+
+class DockerSandboxConnection(SandboxConnection):
```
weizhang25, line 43: Refer to my "connection" comment in connection.py

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,464 @@
...
+    _supports_snapshot = True
+
+    def __init__(
+        self,
+        client: Any,
+        container: Any,
+        *,
+        instance_id: str,
+        workspace: str = _DEFAULT_WORKSPACE,
+        exposed_ports: dict[int, int] | None = None,
```
weizhang25: This means container_port_to_host_port_mapping, give it a better name

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
...
+
+@dataclass(slots=True)
+class SandboxCreateOptions:
+    """Normalized inputs for ``SandboxConnection.create``.
+
+    ``extra`` holds backend-specific flags (E2B template, Docker image, etc.)
+    without forcing the core layer to know each vendor's schema.
+    """
+
+    backend: str
```
weizhang25: backend_id ?

src/agentscope/sandbox/__init__.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
```
weizhang25, line 1: can we name the py file as mcp_gateway.py, to make it clear, since we only talk about mcp gateway in this file.

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
...
+    Check before calling PTY / networking / snapshot APIs.
+    """
+
+    pty: bool = False
+    exposed_ports: bool = False
+    snapshot: bool = False
+
+
+@dataclass(frozen=True, slots=True)
+class ExposedPortEndpoint:
```
weizhang25: class name is confusing. Endpoint == host:port. Also this represents host:port inside a box. It is not "exposed" to outside.

SandboxInternalEndpoint , SandboxInnerEndpoint?

src/agentscope/sandbox/types.py
```diff
@@ -0,0 +1,70 @@
...
+    env: dict[str, str] = field(default_factory=dict)
+    exposed_ports: list[int] = field(default_factory=list)
+    volumes: dict[str, str] = field(default_factory=dict)
+    startup_commands: list[str] = field(default_factory=list)
+    extra: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass(slots=True)
+class SerializedSandboxState:
+    """Serializable snapshot for resume / reconnect.
```
weizhang25, line 41: Add doc saying that an instance of this SSS class is produced by sandbox.export_state( ) method. This make things much easier to understand here.

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,209 @@
...
+    ) -> "SandboxConnection":
+        """Reattach to an existing sandbox from serialized state (optional)."""
+        raise UnsupportedOperation(
+            "resume not implemented for this backend",
+        )
+
+    # ─── execution ────────────────────────────────────────────
+
+    @abstractmethod
+    async def exec(
```
weizhang25, line 79: can we just name it as def run_shell_command( ) to be clear?

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,209 @@
...
+    async def write(self, path: str, data: bytes) -> None:
+        """Write bytes to a sandbox-relative path."""
+
+    # ─── lifecycle ────────────────────────────────────────────
+
+    @abstractmethod
+    async def destroy(self) -> None:
+        """Hard cleanup: release **all** backend resources.
+
+        Idempotent. After ``destroy()`` the connection is unusable.
```
weizhang25: After destroy the *sandbox instance* is unusable.

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,209 @@
...
+        """Enter context manager (identity)."""
+        return self
+
+    async def __aexit__(self, *exc: object) -> None:
+        """Leave context manager — destroy connection."""
+        await self.destroy()
+
+    # ─── capabilities ─────────────────────────────────────────
+
+    def capabilities(self) -> SandboxConnectionCapabilities:
```
weizhang25: def get_cap... getter method name starts with get_

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
```
weizhang25, line 1: In general I'd prefer to have subfolders under src/agentsscope/sandbox/
at least sandbox_connection/ for physical box and sandbox/ for logic.

Split this config to move classes into these 2 subfolders.

Files should be organized by functional modules, not by class types. Putting all classes into one file just b/c they are all config classes does not make much sense.

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
...
+    """User-facing backend config — typed fields for each vendor.
+
+    ``Sandbox._merge_infra_requirements()`` flattens these into a
+    vendor-neutral ``SandboxCreateOptions`` (with an opaque ``extra``
+    dict) before passing them to ``SandboxConnection.create()``.
+    This keeps the user config statically typed while the connection
+    layer stays backend-agnostic.
+    """
+
+    type: str
```
weizhang25: Should type be a class attribute, not an instance attribute?

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
...
+    base_dir: str = "/tmp"
+
+
+# ---------------------------------------------------------------------------
+# MCP / Skills / Tools
+# ---------------------------------------------------------------------------
+
+
+@dataclass(slots=True)
+class McpServerConfig:
```
weizhang25: split config.py. Some are with SandboxConnection. Some are with Sandbox (the logic sandbox)

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Global registry: backend_id → Connection class
+# ---------------------------------------------------------------------------
+
+_registry: dict[str, type[SandboxConnection]] = {}
+
+
+def register_connection_class(cls: type[SandboxConnection]) -> None:
```
weizhang25: def register_sandbox_connection_type(cls:...)

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,209 @@
...
+        raise TypeError(
+            f"Cannot read backend_id from {cls.__name__}; "
+            "ensure backend_id is a concrete @property on the class.",
+        )
+    if bid in _registry:
+        raise ValueError(f"SandboxConnection already registered for {bid!r}")
+    _registry[bid] = cls
+
+
+def get_connection_class(backend_id: str) -> type[SandboxConnection]:
```
weizhang25: def get_sandbox_connection_type

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,209 @@
...
+        return _registry[backend_id]
+    except KeyError as e:
+        available = list(_registry.keys())
+        raise KeyError(
+            f"No SandboxConnection registered for {backend_id!r}. "
+            f"Available backends: {available}",
+        ) from e
+
+
+async def create_connection(
```
weizhang25: def create_sandbox_connection

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,464 @@
...
+from .exceptions import UnsupportedOperation
+from .types import (
+    ExecResult,
+    ExposedPortEndpoint,
+    SandboxCreateOptions,
+    SerializedSandboxState,
+)
+
+
+def _import_docker() -> tuple[Any, Any, Any]:
```
weizhang25: Personally I'd put _import_docker() and the _CONSTANTS below into the class as static function and class constants. Just for better encapsulation.

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,464 @@
...
+            },
+        )
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _ensure_image(client: Any, image: str) -> None:
```
weizhang25: put this into create as an inline function

src/agentscope/sandbox/mcp_gateway.py
```diff
@@ -0,0 +1,202 @@
+# -*- coding: utf-8 -*-
```
weizhang25, line 1: name the file as mcp_gateway.py

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
...
+    """One MCP server to start inside the sandbox."""
+
+    name: str
+    command: str
+    args: list[str] = field(default_factory=list)
+    env: dict[str, str] = field(default_factory=dict)
+
+
+@dataclass(slots=True)
+class McpGatewayConfig:
```
weizhang25: Move this to gateway.py

src/agentscope/sandbox/gateway.py
```diff
@@ -0,0 +1,202 @@
...
+        tools = await gw.list_tools()
+        result = await gw.call_tool("read_file", {"path": "a.txt"})
+        await gw.close()
+    """
+
+    def __init__(self, config: McpGatewayConfig) -> None:
+        self._config = config
+        self._clients: list[StdIOStatefulClient] = []
+        self._routes: dict[str, _ToolRoute] = {}
+        self._started = False
```
weizhang25: _is_started

src/agentscope/sandbox/gateway.py
```diff
@@ -0,0 +1,202 @@
...
+
+    @property
+    def started(self) -> bool:
```
weizhang25: is_started

src/agentscope/sandbox/gateway.py
```diff
@@ -0,0 +1,202 @@
...
+        await gw.start(mcp_configs, cwd="/sandbox/root")
+        tools = await gw.list_tools()
+        result = await gw.call_tool("read_file", {"path": "a.txt"})
+        await gw.close()
+    """
+
+    def __init__(self, config: McpGatewayConfig) -> None:
+        self._config = config
+        self._clients: list[StdIOStatefulClient] = []
+        self._routes: dict[str, _ToolRoute] = {}
```
weizhang25: self._routes -> self._tool_routes

src/agentscope/sandbox/gateway.py
```diff
@@ -0,0 +1,202 @@
...
+        route = self._routes.get(exposed_name)
+        return route.mcp_name if route else None
+
+
+# ---------------------------------------------------------------------------
+# Route builder (module-level for testability)
+# ---------------------------------------------------------------------------
+
+
+def _build_routes(
```
weizhang25: define a staticmethod in class for encapsulation.

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,464 @@
...
+
+    async def exec(
+        self,
+        command: str,
+        *,
+        timeout: float | None = None,
+        cwd: str | None = None,
+        env: dict[str, str] | None = None,
+    ) -> ExecResult:
+        workdir = self._workspace if cwd is None else self._resolve(cwd)
```
weizhang25: nitpick. Foo is None -> not Foo

src/agentscope/sandbox/gateway.py
```diff
@@ -0,0 +1,202 @@
...
+            raise KeyError(
+                f"Tool {name!r} not found in gateway. Available: {available}",
+            )
+        return await route.client.session.call_tool(
+            route.original_name,
+            arguments=args or {},
+        )
+
+    def has_tool(self, name: str) -> bool:
+        """Return whether ``name`` is a routed (possibly prefixed) tool."""
```
weizhang25: prefixed?

src/agentscope/sandbox/gateway.py
```diff
@@ -0,0 +1,202 @@
...
+    # ─── info ─────────────────────────────────────────────────
+
+    def list_servers(self) -> list[dict[str, Any]]:
+        """Metadata rows for :meth:`Sandbox.list_mcps`."""
+        return [
+            {"name": c.name, "command": "stdio", "connected": c.is_connected}
+            for c in self._clients
+        ]
+
+    def tool_origin(self, exposed_name: str) -> str | None:
```
weizhang25: def get_mcp_name(self, exposed_tool_name: str)

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,205 @@
...
+        self._pool_warm_size: int = 0
+        self._pool_free: asyncio.Queue[str] = asyncio.Queue()
+        self._pool_in_use: set[str] = set()
+        self._pool_config: SandboxConfig | None = None
+        self._pool_lock = asyncio.Lock()
+        self._pool_enabled = False
+
+    # ─── core CRUD ────────────────────────────────────────────
+
+    async def create(
```
weizhang25: create( ) is confusing and unclear. def create_sandbox( )

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,205 @@
...
+        self,
+        config: SandboxConfig,
+        *,
+        endpoint: str | None = None,
+    ) -> str:
```
weizhang25: create( ) should return the sandbox object directly. Returning an id and call get( ) to get the box is not straitforward for external user.

Although create is also called in pool logic, it's internal logic, you can always get box id by box_obj.get_sid

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,205 @@
...
+        """Create & start a sandbox, return its sandbox_id."""
+        if endpoint is not None:
```
weizhang25: if endpoint:

src/agentscope/sandbox/sandbox_manager.py
```diff
@@ -0,0 +1,205 @@
+# -*- coding: utf-8 -*-
```
weizhang25, line 1: manayger.py -> sandbox_manager.py

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,205 @@
...
+        self._pool_warm_size = warm_size
+        self._pool_enabled = True
+        return self
+
+    @property
+    def pool_enabled(self) -> bool:
+        """Whether :meth:`enable_pool` has been called."""
+        return self._pool_enabled
+
+    async def pool_warm(self, config: SandboxConfig) -> None:
```
weizhang25: def warm_up_pool( )

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,205 @@
...
+                    sid = await self.create(self._pool_config)
+                    await self._pool_free.put(sid)
+            elif delta < 0:
+                for _ in range(-delta):
+                    if not self._pool_free.empty():
+                        sid = await self._pool_free.get()
+                        await self.destroy(sid)
+
+    @property
+    def pool_stats(self) -> dict[str, int]:
```
weizhang25: def get_pool_statistics( ) 
def get_pool_information( )
def get_pool_state( )

src/agentscope/sandbox/sandbox_manager.py
```diff
@@ -0,0 +1,205 @@
...
+        try:
+            sid = await asyncio.wait_for(
+                self._pool_free.get(),
+                timeout=timeout,
+            )
+        except asyncio.TimeoutError:
+            raise RuntimeError(
+                "pool_acquire timed out — no free sandbox",
+            ) from None
+        async with self._pool_lock:
```
weizhang25, line 171: lock should cover both pool_free.get and pool_in_use.add. treat get and add as an atomic operation.

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,205 @@
...
+        )
+
+    # ─── Pool (integrated to avoid circular references) ───────
+
+    def enable_pool(self, *, warm_size: int = 4) -> "SandboxManager":
+        """Enable warm-pool mode with ``warm_size`` pre-created sandboxes.
+
+        Returns ``self`` for chaining.
+        """
+        self._pool_warm_size = warm_size
```
weizhang25: warm_size -> capacity

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,563 @@
...
+        a minimal ``Skill`` is built from scan / import metadata.
+        """
+        return [
+            await self._skill_entry_to_skill(s) for s in self._skills.values()
+        ]
+
+    def _skill_dir_resolved(self, entry: _SkillEntry) -> str:
+        """Host path to the skill dir when ``workspace_root`` is available."""
+        root = getattr(self._conn, "workspace_root", None)
+        if root is not None:
```
weizhang25: if root:

Maybe replace all these in this pr?
if x is not None -> if x

if x is None -> if not x

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,563 @@
...
+
+        When ``SKILL.md`` exists under the skill directory, it is parsed with
+        ``python-frontmatter`` (same as :class:`LocalSkillLoader`). Otherwise
+        a minimal ``Skill`` is built from scan / import metadata.
+        """
+        return [
+            await self._skill_entry_to_skill(s) for s in self._skills.values()
+        ]
+
+    def _skill_dir_resolved(self, entry: _SkillEntry) -> str:
```
weizhang25: def _resolve_skill_dir( )

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,563 @@
...
+        self._conn: SandboxConnection | None = None
+        self._id: str = uuid.uuid4().hex[:12]
+        self._started = False
+
+        self._tools: dict[str, _ToolEntry] = {}
+        self._mcp_servers: dict[str, _McpServerHandle] = {}
+        self._skills: dict[str, _SkillEntry] = {}
+
+        self._gateway: MCPGateway | None = None
+        self.file: FileAccessor | None = None
```
weizhang25: self.file is defined but not actually used?

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,205 @@
...
+            self._pool_config = config
+            for _ in range(self._pool_warm_size):
+                sid = await self.create(config)
+                await self._pool_free.put(sid)
+            logger.info(
+                "Pool: warmed %d sandboxes",
+                self._pool_warm_size,
+            )
+
+    async def pool_acquire(
```
weizhang25: def acquire_from_pool

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,205 @@
...
+            )
+        except asyncio.TimeoutError:
+            raise RuntimeError(
+                "pool_acquire timed out — no free sandbox",
+            ) from None
+        async with self._pool_lock:
+            self._pool_in_use.add(sid)
+        return self.get(sid)
+
+    async def pool_release(self, sandbox: Sandbox) -> None:
```
weizhang25: def release_from_pool

src/agentscope/sandbox/manager.py
```diff
@@ -0,0 +1,205 @@
...
+            if await sandbox.connection.running():
+                await self._pool_free.put(sid)
+            else:
+                await self.destroy(sid)
+                if self._pool_config:
+                    new_sid = await self.create(self._pool_config)
+                    await self._pool_free.put(new_sid)
+                logger.info("Pool: replaced dead sandbox %s", sid)
+
+    async def pool_resize(self, new_size: int) -> None:
```
weizhang25: def resize_pool

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,563 @@
...
+                pass
+        self._mcp_servers[name] = handle
+        logger.info(
+            "Started MCP server %r in sandbox %s (pid=%s)",
+            name,
+            self._id,
+            handle.pid,
+        )
+
+    async def _scan_and_register_skills(self) -> None:
```
weizhang25: def register_skills

src/agentscope/sandbox/config.py
```diff
@@ -0,0 +1,140 @@
...
+    :meth:`Sandbox.call_tool` is invoked. The tool arguments are
+    serialized as JSON and appended to the command, e.g.
+    ``echo '{"msg": "hi"}'``.  When ``handler`` is ``None`` the tool
+    is metadata-only and cannot be called.
+    """
+
+    name: str
+    description: str = ""
+    parameters: dict[str, Any] = field(default_factory=dict)
+    handler: str | None = None
```
weizhang25: Can we just name handler as shell_cmd? handler usually means something else, not a cmd

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,563 @@
...
+            return
+        for name in r.stdout.decode(errors="replace").strip().split("\n"):
+            name = name.strip()
+            if name:
+                self._skills[name] = _SkillEntry(
+                    name=name,
+                    path=f"{skills_dir}/{name}",
+                )
+
+    async def _run_tool_handler(
```
weizhang25: def _run_tool_via_connection

handler (aka shell cmd) is a concept inside of connection. Do not use it in this function name.

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,563 @@
...
+    """
+
+    def __init__(self, config: SandboxConfig) -> None:
+        self._config = config
+        self._conn: SandboxConnection | None = None
+        self._id: str = uuid.uuid4().hex[:12]
+        self._started = False
+
+        self._tools: dict[str, _ToolEntry] = {}
+        self._mcp_servers: dict[str, _McpServerHandle] = {}
```
weizhang25: _mcp_servers -> _non_gateway_mcp_servers ?

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,563 @@
...
+        args: list[str],
+        env: dict[str, str],
+    ) -> None:
+        """Start an MCP server via exec (non-gateway fallback)."""
+        env_prefix = " ".join(f"{k}={v}" for k, v in env.items())
+        args_str = " ".join(args)
+        full_cmd = f"{env_prefix} nohup {command} {args_str} &".strip()
+        r = await self.connection.exec(full_cmd, timeout=30)
+        handle = _McpServerHandle(name=name, command=command)
+        if r.ok():
```
weizhang25: The handle.pid will not be set if r is not OK. then what? will the handle work properly?

src/agentscope/sandbox/connection.py
cuiyuebing: exec() should be more commonly used

weizhang25: Tried to resolve this comment but failed. Guess I am not a collaborator so I don't have the permission.

src/agentscope/sandbox/config.py
weizhang25: Tried to resolve this comment but failed. Guess I am not a collaborator so I don't have the permission.

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,577 @@
...
+                            session=session,
+                        ),
+                    )
+                    seen.add(gw_tool.name)
+
+        return tools
+
+    async def call_tool(
+        self,
+        name: str,
```
weizhang25, line 244: 貌似call_tool和list_tools有配合上的bug。
list_tools返回的MCPTool对象，里面存的是tool的original name，不是exposed name。
但是call_tool按exposed name找tool，所以如果exposed和original name不一致，且用户只通过list tools结果解析tool name，那么call tool就不可能成功。

cuiyuebing: 我这边会报warning但没有raise error，call tool也是成功的，不确定是不是fastapi版本不一样的原因。但我按你的error修了一下，可以试试现在的代码

src/agentscope/sandbox/local_temp.py
```diff
@@ -0,0 +1,179 @@
...
+        options: SandboxInitializationConfig,
+    ) -> "LocalTempSandboxConnection":
+        """Allocate a temp directory and run ``startup_commands``."""
+        if options.backend_id != "local_temp":
+            msg = f"expected backend 'local_temp', got {options.backend_id!r}"
+            raise ValueError(msg)
+        base = Path(options.extra.get("base_dir", "/tmp"))
+        base.mkdir(parents=True, exist_ok=True)
+        prefix = options.extra.get("prefix", "ws")
+        root = base / f"as_sandbox_{prefix}_{uuid.uuid4().hex[:12]}"
```
weizhang25: nitpick. Is this "as" for agentscope? If so better use the full name, b/c "as" itself is a word. This may look confusing. This comment applies to several similar scenarios.

src/agentscope/sandbox/e2b.py
```diff
@@ -0,0 +1,576 @@
...
+from .exceptions import UnsupportedOperation
+from .types import (
+    SandboxExecutionResult,
+    SandboxInitializationConfig,
+    SandboxInternalEndpoint,
+    SerializedSandboxState,
+)
+
+
+def _import_e2b() -> tuple[Any, Any]:
```
weizhang25: nitpick. Put this function into the class as a static method, for better encapsulation.

src/agentscope/sandbox/e2b.py
```diff
@@ -0,0 +1,576 @@
...
+        from e2b import PtySize as _PtySize
+    except ImportError as exc:
+        raise ImportError(
+            "E2BSandboxConnection requires the `e2b` "
+            "package. Install with: pip install agentscope[sandbox]",
+        ) from exc
+    return _AsyncSandbox, _PtySize
+
+
+_DEFAULT_TEMPLATE = "base"
```
weizhang25: All these constants. Move them into the class as class constant properties.

src/agentscope/sandbox/e2b.py
```diff
@@ -0,0 +1,576 @@
...
+
+    @property
+    def workspace_root(self) -> Path:
+        """Logical workspace root inside the sandbox."""
+        return Path(self._workspace)
+
+    @property
+    def sandbox_id(self) -> str:
+        """The E2B-assigned sandbox identifier."""
+        return self._sandbox.sandbox_id
```
weizhang25, line 124: The type of the sandbox argument is Any in __init__. How do you make sure the sandbox object has a sandbox_id attribute?

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,221 @@
...
+    async def resolve_exposed_port(
+        self,
+        port: int,
+    ) -> SandboxInternalEndpoint:
+        """Map logical container port to host endpoint."""
+        raise CapabilityError("exposed_ports", backend=self.backend_id)
+
+    # ─── optional: persistence ────────────────────────────────
+
+    async def export_state(self) -> SerializedSandboxState:
```
weizhang25, line 183: You may need to add more doc string to export_state( ) and snapshot_workspace( ), to explain the relationship between state and workspace. e.g. workspace, if exists, is part of state, etc.

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,221 @@
...
+class SandboxConnection(ABC):
+    """Handle to one running sandbox instance.
+
+    Required: exec + read/write + destroy + close + is_running.
+    Optional: PTY, ports, snapshot, resume (gate on ``get_capabilities()``).
+    """
+
+    @property
+    @abstractmethod
+    def backend_id(self) -> str:
```
weizhang25: This is more like backend_type, id usually refers to the id of an instance, an individual object.
This "id" is used in the get...type, register...type methods, clearly it means "type"

src/agentscope/sandbox/connection.py
```diff
@@ -0,0 +1,221 @@
...
+        return self._supports_exposed_ports
+
+    def supports_snapshot(self) -> bool:
+        """Whether workspace snapshot/restore is implemented."""
+        return self._supports_snapshot
+
+    # ─── optional: PTY ────────────────────────────────────────
+
+    async def pty_start(self, command: str, **kwargs: object) -> int:
+        """PTY attach — unsupported unless overridden."""
```
weizhang25: Add return type description to doc string. Session id?

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,453 @@
...
+        return Path(self._workspace)
+
+    # ─── factory ──────────────────────────────────────────────
+
+    @classmethod
+    async def create(
+        cls,
+        options: SandboxInitializationConfig,
+    ) -> "DockerSandboxConnection":
+        if options.backend_id != "docker":
```
weizhang25: Do we really need this id/type check? It does not matter what backend id is, if this create call finishes successfully. As long as the config argument provides all the necessary data that the create( ) function needs.

src/agentscope/sandbox/e2b.py
```diff
@@ -0,0 +1,576 @@
...
+
+    # ─── factory ──────────────────────────────────────────────
+
+    @classmethod
+    async def create(
+        cls,
+        options: SandboxInitializationConfig,
+    ) -> "E2BSandboxConnection":
+        """Provision a new E2B sandbox and return a connected instance."""
+        if options.backend_id != "e2b":
```
weizhang25: Do we really need this id/type check? It does not matter what backend id is, if this create call finishes successfully. As long as the config argument provides all the necessary data that the create( ) function needs.

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,453 @@
...
+
+        return conn
+
+    @classmethod
+    async def resume(
+        cls,
+        state: SerializedSandboxState,
+    ) -> "DockerSandboxConnection":
+        docker_sdk, docker_errors, _ = cls._import_docker()
+        if state.backend_id != "docker":
```
weizhang25: See my other comment about backend_id check. The state input is valid as long as the input meets all the operations of the function body, even w/o the id check.
I'd suggest that we remove all the backend_id check everywhere.

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,453 @@
...
+            raise ValueError("backend mismatch for resume")
+
+        container_id = state.payload.get("container_id")
+        if not isinstance(container_id, str):
```
weizhang25: if not container_id or not isinstance(....)
empty string is still a string

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,453 @@
...
+                cls._EXECUTOR,
+                lambda: client.containers.get(container_id),
+            )
+        except docker_errors.NotFound as e:
+            client.close()
+            raise UnsupportedOperation(
+                f"container {container_id} no longer exists",
+            ) from e
+
+        await loop.run_in_executor(cls._EXECUTOR, container.reload)
```
weizhang25: Need to put loop.run_in_executor in a try block? This comment also applies to other similar calls that are not protected by try.

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,453 @@
...
+            )
+            stdout, stderr = result.output
+            code = result.exit_code if result.exit_code is not None else -1
+            return SandboxExecutionResult(
+                exit_code=code,
+                stdout=stdout or b"",
+                stderr=stderr or b"",
+            )
+
+        if not timeout:
```
weizhang25: put run_in_executor in a try block, assuming sandbox.exec( ) should never throw?

src/agentscope/sandbox/e2b.py
```diff
@@ -0,0 +1,576 @@
...
+from .types import (
+    SandboxExecutionResult,
+    SandboxInitializationConfig,
+    SandboxInternalEndpoint,
+    SerializedSandboxState,
+)
+
+
+def _import_e2b() -> tuple[Any, Any]:
+    """Lazy-import ``e2b`` SDK; raises ``ImportError`` with guidance."""
```
weizhang25: Do we really need lazy import here? To make an e2b box work, user must call create or resume anyway. The only case you don't need to import e2b is where the user creates an e2b box instance but never call its create/resume function. Will this really happen?

src/agentscope/sandbox/e2b.py
```diff
@@ -0,0 +1,576 @@
...
+            host=host,
+            port=443,
+            is_tls_enabled=True,
+        )
+
+    # ─── capabilities: snapshot ───────────────────────────────
+
+    async def snapshot_workspace(self) -> bytes:
+        """Export the workspace directory as a tar archive."""
+        result = await self._sandbox.commands.run(
```
weizhang25: The whole tar-read-rm operation chain should be put into a lock. AKA make this snapshot_ws function thread safe.

src/agentscope/sandbox/e2b.py
```diff
@@ -0,0 +1,576 @@
...
+            format="bytes",
+        )
+        # Clean up temp file.
+        await self._sandbox.commands.run(
+            "rm -f /tmp/_ws_snapshot.tar",
+            timeout=10,
+        )
+        return bytes(data)
+
+    async def restore_workspace(self, data: bytes) -> None:
```
weizhang25: snapshot_ws and restore_ws should be mutually exclusive. These 2 should share a common lock.

src/agentscope/sandbox/e2b.py
```diff
@@ -0,0 +1,576 @@
...
+    routing) and snapshot/restore (via the pause/connect API).
+    """
+
+    _supports_exposed_ports = True
+    _supports_snapshot = True
+    _supports_pty = True
+
+    def __init__(
+        self,
+        sandbox: Any,
```
weizhang25: Why still annotate sandbox as Any, when the doc string clearly has the real data type?

src/agentscope/sandbox/e2b.py
```diff
@@ -0,0 +1,576 @@
...
+                "domain": self._domain or "",
+            },
+        )
+
+
+register_sandbox_connection_type(E2BSandboxConnection)
+
+
+async def _run_command_with_retry(
+    sandbox: Any,
```
weizhang25: sandbox: one of e2b native box types? Please make it clear. The word "sandbox" is flying around everywhere. It is heavily overloaded.

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,453 @@
...
+    _EXECUTOR = ThreadPoolExecutor(
+        max_workers=8,
+        thread_name_prefix="agentscope-docker-sandbox",
+    )
+
+    _supports_exposed_ports = True
+    _supports_snapshot = True
+
+    @staticmethod
+    def _import_docker() -> tuple[Any, Any, Any]:
```
weizhang25: Is lazy import necessary? All actual functionalities of this docker sandbox requires docker interaction. The only case where docker imports are not needed is when you create a DockerSandbox instance and let it do nothing. Will this happen?

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,453 @@
...
+        except ImportError as exc:
+            raise ImportError(
+                "DockerSandboxConnection requires the `docker` "
+                "package. Install with: pip install agentscope[sandbox]",
+            ) from exc
+        return _docker_sdk, _docker_errors, _Container
+
+    def __init__(
+        self,
+        client: Any,
```
weizhang25: Why still annotate client and container as Any, when the doc string clearly has the real data type?

src/agentscope/sandbox/mcp_gateway.py
```diff
@@ -0,0 +1,228 @@
...
+        """Connect MCP servers and build the routing table."""
+        if self._is_started:
+            return
+
+        raw: list[tuple[str, MCPClient, list[mtypes.Tool]]] = []
+        for cfg in mcp_configs:
+            client = MCPClient(
+                name=cfg.name,
+                is_stateful=True,
+                mcp_config=StdioMCPConfig(
```
weizhang25: Why hard code StdioMCPConfig as the config type? What about HttpMCPConfig?
In sandbox.py _start_gateway(self), _gateway.start with the workspace root of self._conn, which can be a docker or e2b box connection, not necessarily a local temp box connection.

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,754 @@
...
+    async def list_mcps(self) -> list[dict[str, Any]]:
+        """List managed MCP servers (gateway-managed or exec-started)."""
+        if self._gateway:
+            return self._gateway.list_servers()
+        return [
+            {"name": h.name, "command": h.command, "pid": h.pid}
+            for h in self._exec_mcp_servers.values()
+        ]
+
+    async def add_mcp(
```
weizhang25, line 510: add_mcp( ) does not update self._tools. Then list_tools( ) won't show the tools in the new mcp server.

src/agentscope/sandbox/sandbox.py
```diff
@@ -0,0 +1,754 @@
...
+        if self._gateway:
+            raise RuntimeError(
+                "Dynamic remove_mcp is not supported when MCPGateway "
+                "is enabled (one-period limitation)",
+            )
+        handle = self._exec_mcp_servers.pop(name, None)
+        if handle and handle.pid:
+            kill_cmd = f"kill {handle.pid} 2>/dev/null || true"
+            await self.connection.exec(kill_cmd, timeout=5)
+        self._tools = {
```
weizhang25: Why updating self.tools here, if add_mcp does not add any new entry to self._tools?

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,498 @@
...
+                f"container {container_id} could not be inspected: {e}",
+            ) from e
+
+        if container.status != "running":
+            try:
+                await loop.run_in_executor(cls._EXECUTOR, container.start)
+            except docker_errors.APIError as e:
+                client.close()
+                raise UnsupportedOperation(
+                    f"container {container_id} could not be started: {e}",
```
weizhang25, line 274: no need to include {e} in the text b/c this raises from e

src/agentscope/sandbox/docker.py
```diff
@@ -0,0 +1,498 @@
...
+            raise UnsupportedOperation(
+                f"container {container_id} no longer exists",
+            ) from e
+
+        try:
+            await loop.run_in_executor(cls._EXECUTOR, container.reload)
+        except docker_errors.APIError as e:
+            client.close()
+            raise UnsupportedOperation(
+                f"container {container_id} could not be inspected: {e}",
```
weizhang25, line 265: no need to include {e} in the text b/c this raises from e

src/agentscope/sandbox/mcp_gateway.py
```diff
@@ -0,0 +1,348 @@
...
+
+if TYPE_CHECKING:
+    from .config import MCPServerConfig
+
+
+@dataclass(slots=True)
+class MCPGatewayConfig:
+    """Gateway settings (listen port merged into ``exposed_ports``).
+
+    The gateway is always enabled — the ``enabled`` field is kept only for
```
weizhang25, line 34: nitpick. Remove doc string about "enabled" field?

src/agentscope/sandbox/mcp_gateway.py
```diff
@@ -0,0 +1,348 @@
...
+        """Create an MCPClient for a config and connect it."""
+        cwd = cwd_override or self._cwd
+
+        if cfg.transport == "http":
+            mcp_config = HttpMCPConfig(
+                url=cfg.url,
+                headers=cfg.headers or None,
+                timeout=cfg.timeout,
+            )
+            client = MCPClient(
```
weizhang25, line 270: nitpick. 270 ~ 274 and 283 ~ 287 are identical. just keep one copy and move it to line 288

src/agentscope/sandbox/mcp_gateway.py
```diff
@@ -0,0 +1,348 @@
...
+        effective_cwd = cwd or self._cwd
+        await self._connect_server(mcp_config, cwd_override=effective_cwd)
+        self._rebuild_tool_routes()
+        logger.info(
+            "MCPGateway: dynamically added server %r (%d total tools)",
+            mcp_config.name,
+            len(self._tool_routes),
+        )
+
+    async def remove_server(self, name: str) -> None:
```
weizhang25, line 160: Should add_server() and remove_server() be mutually exclusive? or make _rebuild_tool_routes( ) concurrency-safe.

src/agentscope/sandbox/sandbox_manager.py
```diff
@@ -0,0 +1,219 @@
...
+                )
+
+    # --- pool (no circular refs) ---
+
+    def enable_pool(self, *, capacity: int = 4) -> "SandboxManager":
+        """Enable warm-pool mode with ``capacity`` pre-created sandboxes.
+
+        Returns ``self`` for chaining.
+        """
+        self._pool_capacity = capacity
```
weizhang25, line 138: need to ensure capacity is positive

src/agentscope/sandbox/sandbox_manager.py
```diff
@@ -0,0 +1,219 @@
...
+            self._pool_in_use.add(sid)
+        return self.get_sandbox(sid)
+
+    async def release_to_pool(self, sandbox: Sandbox) -> None:
+        """Return a sandbox to the pool; replace it if dead."""
+        sid = sandbox.sandbox_id
+        async with self._pool_lock:
+            self._pool_in_use.discard(sid)
+            if await sandbox.connection.is_running():
+                await self._pool_free.put(sid)
```
weizhang25, line 190: Sandbox has been used by the current user. It may store data related to the user. Do we need to reset a box to its initial "clean" state before giving it to the next user? New user using a "used" box may have unexpected consequence.

For example, a tool in a mcp server in a sandbox stores user 1's api key. User 1 released this box to queue. User 2 acquired the box and called the tool directly, using user 1's key to access model. While actually user 2 should expected an "unused" tool that is waiting for his own api key.

src/agentscope/sandbox/sandbox_manager.py
```diff
@@ -0,0 +1,219 @@
...
+        return self
+
+    @property
+    def pool_enabled(self) -> bool:
+        """Whether :meth:`enable_pool` has been called."""
+        return self._pool_enabled
+
+    async def warm_up_pool(self, config: SandboxConfig) -> None:
+        """Pre-create ``capacity`` sandboxes and fill the free queue."""
+        if not self._pool_enabled:
```
weizhang25, line 149: What if I call warm_up_pool N times? The actual queue length will be N*capacity....

src/agentscope/sandbox/sandbox_manager.py
```diff
@@ -0,0 +1,219 @@
...
+        async with self._pool_lock:
+            try:
+                sid = await asyncio.wait_for(
+                    self._pool_free.get(),
+                    timeout=timeout,
+                )
+            except asyncio.TimeoutError:
+                raise RuntimeError(
+                    "acquire_from_pool timed out — no free sandbox",
+                ) from None
```
weizhang25, line 180: Nitpick. Is "from None" necessary?

src/agentscope/sandbox/sandbox_manager.py
```diff
@@ -0,0 +1,219 @@
...
+            self._pool_in_use.add(sid)
+        return self.get_sandbox(sid)
+
+    async def release_to_pool(self, sandbox: Sandbox) -> None:
+        """Return a sandbox to the pool; replace it if dead."""
+        sid = sandbox.sandbox_id
+        async with self._pool_lock:
+            self._pool_in_use.discard(sid)
+            if await sandbox.connection.is_running():
```
weizhang25, line 189: box.connection.is_running -> box.started

src/agentscope/sandbox/sandbox_manager.py
```diff
@@ -0,0 +1,219 @@
...
+        self._pool_capacity = capacity
+        self._pool_enabled = True
+        return self
+
+    @property
+    def pool_enabled(self) -> bool:
+        """Whether :meth:`enable_pool` has been called."""
+        return self._pool_enabled
+
+    async def warm_up_pool(self, config: SandboxConfig) -> None:
```
weizhang25, line 147: warm_up should be protected by lock too.

PR #1595, opened 2026-05-12
fix(mcp): preserve $defs in MCPTool input schema and strip titles recursively

qbc2016:
## Problem

Two related issues caused broken schemas when MCP tools with Pydantic sub-model parameters were exposed to LLMs:

1. **`$defs` silently stripped** — `MCPTool.__init__` manually constructed `input_schema` by copying only `properties` and `required` from `tool.inputSchema`.  Any `$defs` block (and other advanced keywords like `anyOf`, `oneOf`) was dropped.  The LLM would receive a schema where a `$ref: "#/$defs/SomeModel"` pointer pointed to nothing, causing tool-call failures.

2. **`_remove_title_field` did not recurse into `$defs`** — Pydantic auto-generates `"title"` fields for every sub-model inside `$defs`.  These noise fields were not being removed, which conflicts with strict LLM API modes (e.g. OpenAI Structured Outputs).

### `src/agentscope/tool/_adapters.py`
Replaced the manual schema construction in `MCPTool.__init__` with a shallow copy of the full `inputSchema`, using `setdefault` to fill in required defaults (`type`, `properties`, `required`).  All other fields — `$defs`, `anyOf`, `oneOf`, etc. — are now preserved.

### `src/agentscope/tool/_utils.py`
Added `$defs` recursion to `_remove_title_field` so that `title` fields inside referenced sub-schemas are consistently stripped, matching the existing behaviour for top-level and property-level titles.

## Tests

- `tests/toolkit_test.py` — `RemoveTitleFieldTest`: four unit tests covering top-level, property-level, `$defs`-level, and boolean-schema edge cases.
- `tests/mcp_sse_client_test.py` — `SseSchemaDefsPreservationTest`: end-to-end integration test using a real FastMCP server with a Pydantic sub-model parameter.  Validates that `MCPTool.input_schema` retains `$defs` and that `Toolkit.get_function_schemas()` output has those titles stripped.

### DIFF
--- a/tests/mcp_sse_client_test.py
+++ b/tests/mcp_sse_client_test.py
@@ -6,6 +6,7 @@ from multiprocessing import Process
 from unittest.async_case import IsolatedAsyncioTestCase
 
 from mcp.server import FastMCP
+from pydantic import BaseModel
 
 from agentscope.mcp import MCPClient, HttpMCPConfig
 from agentscope.message import ToolCallBlock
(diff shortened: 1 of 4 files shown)


tests/mcp_sse_client_test.py
```diff
@@ -246,3 +277,88 @@ async def test_stateful_client(self) -> None:
...
+            0,
+            "At least one $defs entry expected",
+        )
+
+        # 2. get_function_schemas() must preserve $defs and strip titles
+        toolkit = Toolkit(tools=[mcp_tool])
+        schemas = toolkit.get_function_schemas()
+        self.assertEqual(len(schemas), 1)
+
+        params = schemas[0]["function"]["parameters"]
```
DavdGao: Assert against the complete JSON schema rather than just a few specific fields.

tests/toolkit_test.py
```diff
@@ -1027,3 +1028,67 @@ async def test_meta_tool(self) -> None:
...
+            },
+        }
+        self.fn(schema)
+
+        self.assertNotIn(
+            "title",
+            schema["$defs"]["MyModel"],
+            "Top-level title inside $defs must be removed",
+        )
+        self.assertNotIn(
```
DavdGao: Similar as above

DavdGao (approved):

PR #1601, opened 2026-05-14
fix: uniform multimodal models and non-moltimodal models

### LINKED ISSUE #1575: fix(dashscope model): uniform multimodal models and non-moltimodal models
Now we use a `multimodality` parameter to separate, and need to find a better way.

jiangxiaobin96:
fix [#1575](https://github.com/agentscope-ai/agentscope/issues/1575)
deprecat `multimodality` and get multimodality information from message.
```python
def _contains_multimodal_content(
        messages: list[dict[str, Any]],
    ) -> bool:
        """Return whether the formatted messages contain media blocks."""
        media_keys = {"image", "audio", "video"}
        for message in messages:
            content = message.get("content")
            if not isinstance(content, list):
                continue

            for item in content:
                if (
                    isinstance(item, dict)
                    and media_keys.intersection(item.keys())
                ):
                    return True
        return False
```

### DIFF
--- a/src/agentscope/model/_dashscope_model.py
+++ b/src/agentscope/model/_dashscope_model.py
@@ -54,16 +54,14 @@ class DashScopeChatModel(ChatModelBase):
 
     This class provides a unified interface for DashScope API by automatically
     selecting between text-only (Generation API) and multimodal
-    (MultiModalConversation API) endpoints. The `multimodality` parameter
-    allows explicit control over API selection:
+    (MultiModalConversation API) endpoints. Endpoint selection is handled
+    internally based on the formatted message content and model family:
 
-    - When `multimodality=True`: Forces use of MultiModalConversation API
-      for handling images, videos, and other multimodal inputs
-    - When `multimodality=False`: Forces use of Generation API for
-      text-only processing
-    - When `multimodality=None` (default): Automatically selects the API
-      based on model name (e.g., models with "-vl" suffix or starting
-      with "qvq" will use MultiModalConversation API)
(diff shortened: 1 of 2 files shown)


jiangxiaobin96:
Could you have a look? @qbc2016

qbc2016:
> Could you have a look? @qbc2016

Thanks for the PR! I appreciate the effort, but I have some concerns about this approach.

The main issue is that DashScope uses two different APIs for non-multimodal and multimodal models, and they're incompatible with each other. Additionally, multimodal models can also handle plain text conversations—they accept text-only input as well.

So, determining which API to use based on whether messages contain multimodal content might not work well in all cases.

jiangxiaobin96:
message+model name(qvq/-vl), Does it make sense?
```
def _should_use_multimodal_api(
        self,
        messages: list[dict[str, Any]],
    ) -> bool:
        """Select the DashScope endpoint from message content and model."""
        if self._contains_multimodal_content(messages):
            return True

        return self.model_name.startswith("qvq") or "-vl" in self.model_name
```

qbc2016:
Not really. For example, qwen3.5-plus, qwen3.6-plus, etc., are also multimodal models, but they don't follow that naming pattern. Future models might not either, so name-based detection isn't reliable. Additionally, Bailian hosts many third-party models—some multimodal, some not—which makes it nearly impossible to enumerate them all.

jiangxiaobin96:
It is indeed difficult to distinguish, and it seems that the multimodal parameter is still the best way at the moment.

PR #1611, opened 2026-05-15
feature(workspace/sandbox): refactor workspace/sandbox in agentscope 2.0

cuiyuebing:
## PR Title Format

## AgentScope Version

## Description

refactor workspace/sandbox in agentscope 2.0

## Checklist

- [x]  All tests are passing


src/agentscope/workspace/workspace_manager_base.py
```diff
@@ -0,0 +1,385 @@
...
+        self._pool_workspaces: dict[str, WorkspaceBase] = {}
+
+    # ── abstract: subclass-provided ───────────────────────────────
+
+    @abstractmethod
+    async def initialize(self) -> None:
+        """Initialize the manager (connect clients, warm pools, etc.)."""
+
+    @abstractmethod
+    async def close(self) -> None:
```
weizhang25: @abstractmethod
def _do_specific_close_action( )

def close():
    self._do_specific_close_action( )
    # the following 2 are mandatory.
    # don't expect a developer will actually do what you said in doc string.
    self._close_pool( )
    self._close_all_workspaces()

src/agentscope/workspace/workspace_manager_base.py
```diff
@@ -0,0 +1,385 @@
...
+                "Pool: warmed %d workspaces",
+                self._pool_capacity,
+            )
+
+    async def acquire_from_pool(
+        self,
+        *,
+        timeout: float | None = None,
+    ) -> WorkspaceBase:
+        """Acquire a free workspace from the pool.
```
weizhang25, line 264: add lock

src/agentscope/workspace/workspace_manager_base.py
```diff
@@ -0,0 +1,385 @@
...
+    def pool_enabled(self) -> bool:
+        """Whether :meth:`enable_pool` has been called."""
+        return self._pool_enabled
+
+    # ── pool: lifecycle ───────────────────────────────────────────
+
+    async def warm_up_pool(self) -> None:
+        """Pre-create ``capacity`` workspaces and fill the free queue.
+
+        Safe to call only once; raises if the pool already has workspaces.
```
weizhang25: warm-up should be called inside of enable-pool. Need to set a flag when it is called, so that warm-up only executes once in enable-pool.
def enable_pool( )
    lock
    if self.enabled:
      return
    self.warm-up-pool
    self.enabled = True

src/agentscope/workspace/workspace_manager_base.py
```diff
@@ -0,0 +1,385 @@
...
+
+    async def _create_for_pool(self) -> WorkspaceBase:
+        """Create a workspace for pool use.
+
+        Defaults to calling :meth:`_do_create` with no overrides.
+        Override in subclasses if pool workspaces need different config.
+        """
+        return await self._do_create()
+
+    async def _reset_for_pool(self, workspace: WorkspaceBase) -> None:
```
weizhang25: Sounds like the WorkspaceBase need to have a reset function. Manager does not know the details of a workspace, so manager cannot "reset" a workspace.
_reset_for_pool should just call workspace.reset( )

src/agentscope/workspace/workspace_manager_base.py
```diff
@@ -0,0 +1,385 @@
...
+
+        The default implementation is a no-op.  Subclasses that manage
+        sensitive per-user state **should** override this.
+        """
+
+    async def _close_pool(self) -> None:
+        """Close all pool workspaces. Call from ``close()``."""
+        if not self._pool_workspaces:
+            return
+        tasks = [ws.close() for ws in self._pool_workspaces.values()]
```
weizhang25: add lock

src/agentscope/workspace/workspace_manager_base.py
```diff
@@ -0,0 +1,385 @@
...
+        if capacity <= 0:
+            raise ValueError(
+                f"Pool capacity must be positive, got {capacity}",
+            )
+        self._pool_capacity = capacity
+        self._pool_enabled = True
+        return self
+
+    @property
+    def pool_enabled(self) -> bool:
```
weizhang25: Do we really need this flag? Just let user call the pool functions. The function should throw with error msg.

src/agentscope/workspace/types.py
```diff
@@ -0,0 +1,48 @@
...
+    stdout: bytes
+    stderr: bytes
+
+    def is_ok(self) -> bool:
+        """Return ``True`` if ``exit_code == 0``."""
+        return self.exit_code == 0
+
+
+@dataclass(frozen=True, slots=True)
+class InternalEndpoint:
```
weizhang25: Is this only for docker-workspace? if so move it to docker-ws.py

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+
+
+class _SkillEntry(TypedDict):
+    hash: str
+    skill_name: str
+
+
+class _SkillsFile(TypedDict):
+    skills_dir_mtime: float
+    skills: dict[str, _SkillEntry]
```
weizhang25: Add doc string to explain the meaning of the key

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+
+    async def list_mcps(self) -> list[MCPClient]:
+        return list(self._mcps)
+
+    # ── skill discovery ────────────────────────────────────────────
+
+    async def list_skills(self) -> list[Skill]:
+        skills_dir = os.path.join(self._workdir, "skills")
+        if not await aiofiles.ospath.isdir(skills_dir):
+            return []
```
weizhang25, line 377: Shouldn't line 179 throw?

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+
+    # ── offload ────────────────────────────────────────────────────
+
+    async def offload_context(
+        self,
+        session_id: str,
+        msgs: list[Msg],
+        **kwargs: Any,
+    ) -> str:
+        path = os.path.join(
```
weizhang25, line 416: This offload function is not thread safe, intended?

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+                return
+        raise KeyError(
+            f"MCP {name!r} not found. "
+            f"Available: {[m.name for m in self._mcps]}",
+        )
+
+    # ── dynamic skill management ───────────────────────────────────
+
+    async def add_skill(self, skill_path: str) -> None:
+        skill_path = os.path.abspath(skill_path)
```
weizhang25, line 524: should add_skill be thread safe?

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+        )
+        await self._save_skills_file(skills_dir, skills_file)
+        logger.info(
+            "LocalWorkspace: added skill %r as %r",
+            raw_name,
+            agent_name,
+        )
+
+    async def remove_skill(self, name: str) -> None:
+        skills_dir = os.path.join(self._workdir, "skills")
```
weizhang25: Same here. When you have a resource and IO operation, always think about thread safety. Use lock if needed.

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+        skill_paths: list[str] | None = None,
+        mcps: list[MCPClient] | None = None,
+        instructions: str = _DEFAULT_INSTRUCTIONS,
+    ) -> None:
+        self._id = uuid.uuid4().hex[:12]
+        self._workdir = os.path.abspath(workdir)
+        self._skill_paths = list(
+            dict.fromkeys(os.path.abspath(p) for p in (skill_paths or [])),
+        )
+        self._instructions = instructions
```
weizhang25, line 158: Can we just do self._inst = inst.format(workdir=self._workdir) so get_inst( ) does not have to repeat this everytime?

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+    async def list_skills(self) -> list[Skill]:
+        skills_dir = os.path.join(self._workdir, "skills")
+        if not await aiofiles.ospath.isdir(skills_dir):
+            return []
+
+        skills_file = await self._load_skills_file(skills_dir)
+        current_mtime = await aiofiles.ospath.getmtime(skills_dir)
+
+        if current_mtime != skills_file["skills_dir_mtime"]:
+            skills_file = await self._reconcile_skills_dir(
```
weizhang25: Should reconcile be protected by a lock?

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+                return None
+            async with aiofiles.open(skill_md, "r", encoding="utf-8") as f:
+                content_str = await f.read()
+            content = frontmatter.loads(content_str)
+            name = content.get("name")
+            desc = content.get("description")
+            if not name or not desc:
+                return None
+            h = hashlib.sha256(content_str.encode("utf-8")).hexdigest()
+            return skill_path, str(name), h
```
weizhang25: returned skill_path is never used. plus it is from the input. Remove it from return data.

src/agentscope/workspace/exceptions.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+"""Workspace exception hierarchy."""
+
+
+class WorkspaceError(Exception):
```
weizhang25: This PR does not raise any of these exception types. Why are they still needed?

src/agentscope/workspace/workspace_base.py
```diff
@@ -0,0 +1,142 @@
...
+    async def close(self) -> None:
+        """Release all resources and connections."""
+
+    async def is_alive(self) -> bool:
+        """Check if the workspace is still operational.
+
+        Override in container/sandbox backends to perform a real
+        liveness check. Defaults to ``True`` for local workspaces.
+        """
+        return True
```
weizhang25, line 75: This should default to False. Even local workspace is not functional if it has no skills/mcps/tools etc. Local workspace also need a real check.

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+    # ── lifecycle ──────────────────────────────────────────────────
+
+    async def initialize(self) -> None:
+        for mcp in self._mcps:
+            if (
+                mcp.is_stateful or mcp.mcp_config.type == "stdio_mcp"
+            ) and not mcp.is_connected:
+                await mcp.connect()
+
+        skills_dir = os.path.join(self._workdir, "skills")
```
weizhang25: move 150 and 151 into _install_initial_skills( )

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+            skills_dir,
+        )
+        await self._save_skills_file(skills_dir, skills_file)
+        logger.info("LocalWorkspace: removed skill %r", name)
+
+    # ── internal: initial skill install ────────────────────────────
+
+    async def _install_initial_skills(self, skills_dir: str) -> None:
+        skills_file = await self._load_skills_file(skills_dir)
+        existing = skills_file["skills"]
```
weizhang25: Do we expect a newly created local workspace has pre-installed skills under skills folder?
Shouldn't a new workspace have an empty skills folder?

src/agentscope/workspace/gateway/_server.py
```diff
@@ -0,0 +1,447 @@
...
+                pass
+            self._stack = None
+            self.session = None
+
+
+# ── route table ───────────────────────────────────────────────────
+
+
+class _ToolRoute:
+    __slots__ = ("client", "original_name")
```
weizhang25: What's the point of using __slots__ here?

src/agentscope/workspace/gateway/_server.py
```diff
@@ -0,0 +1,447 @@
...
+
+        app.add_middleware(_TokenAuth)
+
+    async def _health(
+        request: Request,  # pylint: disable=unused-argument
+    ) -> PlainTextResponse:
+        """Liveness probe endpoint."""
+        return PlainTextResponse("ok")
+
+    async def _admin_add(request: Request) -> JSONResponse:
```
weizhang25: Just name it as _add_mcp. admin is confusing and has no actual meaning here, b/c there is no admin authentication check here.

src/agentscope/workspace/local_workspace.py
```diff
@@ -0,0 +1,647 @@
...
+            skills_file["skills"] = existing
+            skills_file["skills_dir_mtime"] = await aiofiles.ospath.getmtime(
+                skills_dir,
+            )
+            await self._save_skills_file(skills_dir, skills_file)
+
+    # ── internal: skill file I/O ───────────────────────────────────
+
+    async def _load_skills_file(self, skills_dir: str) -> _SkillsFile:
+        path = os.path.join(skills_dir, ".skills")
```
weizhang25: As we discussed offline. This .skills metadata layer can be removed. All skill operation should just be applied to the folders directly.

src/agentscope/workspace/e2b_workspace_manager.py
```diff
@@ -0,0 +1,150 @@
...
+        gateway_port: int = E2BWorkspace.GATEWAY_PORT,
+        default_env: dict[str, str] | None = None,
+        default_metadata: dict[str, str] | None = None,
+        default_startup_commands: list[str] | None = None,
+    ) -> None:
+        super().__init__()
+        self._template = template
+        self._api_key = api_key
+        self._domain = domain
+        self._timeout = timeout
```
weizhang25: _timeout_seconds just to be precise? No unit and no doc string. People has to guess or read code to figure it out.

src/agentscope/workspace/gateway/_mixin.py
```diff
@@ -0,0 +1,421 @@
...
+            log_text = gw_log.stdout.decode(errors="replace").strip()
+            logger.error(
+                "Gateway failed to start. /tmp/gw.log:\n%s",
+                log_text or "(empty)",
+            )
+            raise
+
+        gw_headers = {**self._gw_platform_headers()}
+        if self._gateway_token:
+            gw_headers["Authorization"] = f"Bearer {self._gateway_token}"
```
weizhang25: Add doc string to explain the magic word Bearer.

src/agentscope/workspace/gateway/_mixin.py
```diff
@@ -0,0 +1,421 @@
...
+
+        self._gateway_mcpc = _RestGatewayClient(
```
weizhang25: nitpick. mcpc -> mcp_client. mcpc lookds confusing.

src/agentscope/workspace/e2b_workspace.py
```diff
@@ -0,0 +1,440 @@
...
+            raise ValueError(
+                f"Invalid skill at {skill_path}: SKILL.md not found",
+            )
+
+        dir_name = os.path.basename(os.path.abspath(skill_path))
+        await self._exec(f"mkdir -p {self.SKILLS_DIR}")
+
+        for root, _dirs, files in os.walk(skill_path):
+            for fname in files:
+                local = os.path.join(root, fname)
```
weizhang25, line 367: Add a sample skill_path, local, rel, remote here for easy understanding.

src/agentscope/workspace/e2b_workspace.py
```diff
@@ -0,0 +1,440 @@
...
+                rel = os.path.relpath(local, skill_path)
+                remote = f"{self.SKILLS_DIR}/{dir_name}/{rel}"
+                with open(local, "rb") as f:
+                    data = f.read()
+                await self._sandbox.files.write(remote, data)
+
+        logger.info("E2BWorkspace: added skill %r", dir_name)
+
+    async def remove_skill(self, name: str) -> None:
+        dest = f"{self.SKILLS_DIR}/{shlex.quote(name)}"
```
weizhang25: line 324 has a different path pattern than line 332 ?

src/agentscope/workspace/e2b_workspace.py
```diff
@@ -0,0 +1,440 @@
...
+                            stdout=(getattr(e, "stdout", "") or "").encode(
+                                "utf-8",
+                            ),
+                            stderr=(getattr(e, "stderr", "") or "").encode(
+                                "utf-8",
+                            ),
+                        )
+                    raise
+                await asyncio.sleep(delay)
+                delay = min(delay * 1.5, 5.0)
```
weizhang25: Why increase sleep interval after each exception?  Also 1.5^4 is 5.06, this min( ) make no sense. Just use a fixed delay time.

src/agentscope/workspace/e2b_workspace.py
```diff
@@ -0,0 +1,440 @@
...
+        skill_md = os.path.join(skill_path, "SKILL.md")
+        if not os.path.isfile(skill_md):
+            raise ValueError(
+                f"Invalid skill at {skill_path}: SKILL.md not found",
+            )
+
+        dir_name = os.path.basename(os.path.abspath(skill_path))
+        await self._exec(f"mkdir -p {self.SKILLS_DIR}")
+
+        for root, _dirs, files in os.walk(skill_path):
```
weizhang25, line 365: skill_path contains files and folders other than SKILL.md? Shouldn't a skill_path refer to one skill, and should only have a SKILL.md file? Why copying all other non-skill files?

src/agentscope/workspace/e2b_workspace.py
```diff
@@ -0,0 +1,440 @@
...
+        metadata: dict[str, str] | None = None,
+        startup_commands: list[str] | None = None,
+        instructions: str = _DEFAULT_INSTRUCTIONS,
+    ) -> None:
+        self._template = template
+        self._api_key = api_key
+        self._domain = domain
+        self._timeout = timeout
+        self._working_dir = working_dir
+        self._mcp_servers = list(mcp_servers or [])
```
weizhang25: = mcp_servers or []

Just let __init__ fail if mcp_servers is not a list. Do not try to fix callers' data problem

src/agentscope/workspace/e2b_workspace.py
```diff
@@ -0,0 +1,440 @@
...
+        self._gateway_port = gateway_port
+        self._env = dict(env or {})
```
weizhang25: = env or {}

Actually please fix all datatype(foo or bar) syntax.

src/agentscope/workspace/e2b_workspace.py
```diff
@@ -0,0 +1,440 @@
...
+                    f"stderr: {r.stderr.decode(errors='replace')}\n"
+                    f"stdout: {r.stdout.decode(errors='replace')}",
+                )
+
+        if self._mcp_servers:
+            await self._start_gateway()
+
+        self._started = True
+
+    async def is_alive(self) -> bool:
```
weizhang25, line 179: What's the expectation of being alive? is box-is-running enough? What if I removed all mcp's and skills from the box, making the workspace unfunctional, will it still be considered as alive?

src/agentscope/workspace/docker_workspace.py
```diff
@@ -0,0 +1,555 @@
...
+        gateway_port: int = GATEWAY_PORT,
+        exposed_ports: list[int] | None = None,
+        volumes: dict[str, str] | None = None,
+        env: dict[str, str] | None = None,
+        startup_commands: list[str] | None = None,
+        instructions: str = _DEFAULT_INSTRUCTIONS,
+    ) -> None:
+        self._image = image
+        self._working_dir = working_dir
+        self._mcp_servers = list(mcp_servers or [])
```
weizhang25: list(foo or []) -> foo or []
This also applies to dict

src/agentscope/mcp/_mcp_client.py
```diff
@@ -132,20 +132,42 @@ def _create_http_client(
...
+        def _client_factory(
+            headers: dict[str, str] | None = None,
+            timeout: httpx.Timeout | None = None,
+            auth: httpx.Auth | None = None,
+        ) -> httpx.AsyncClient:
+            kwargs: dict[str, Any] = {
+                "follow_redirects": True,
+                "verify": config.verify,
+            }
+            if timeout is not None:
```
weizhang25: if foo is not None -> if foo

src/agentscope/workspace/docker_workspace.py
```diff
@@ -0,0 +1,555 @@
...
+                        block.source,
+                        Base64Source,
+                    ):
+                        block = await self._offload_data(block)
+                    content.append(block)
+                msg.content = content
+            lines.append(msg.model_dump_json())
+
+        payload = "\n".join(lines) + "\n"
+        await self._exec(f"mkdir -p {base}")
```
weizhang25, line 385: Is base a relative path?

src/agentscope/workspace/docker_workspace.py
```diff
@@ -0,0 +1,555 @@
...
+                        d = await self._offload_data(block)
+                        url = str(d.source.url)
+                    else:
+                        url = str(block.source.url)
+                    parts.append(
+                        f"<data url='{url}' name='{block.name}' "
+                        f"media_type='{block.source.media_type}'/>",
+                    )
+
+        await self._exec(f"mkdir -p {base}")
```
weizhang25, line 423: Is base a relative path?

src/agentscope/workspace/docker_workspace.py
```diff
@@ -0,0 +1,555 @@
...
+        import os
+
+        skill_md = os.path.join(skill_path, "SKILL.md")
+        if not os.path.isfile(skill_md):
+            raise ValueError(
+                f"Invalid skill at {skill_path}: SKILL.md not found",
+            )
+
+        dir_name = os.path.basename(os.path.abspath(skill_path))
+        dest = f"{self.SKILLS_DIR}/{dir_name}"
```
weizhang25: Is dest just logging? if so just put the f-str in the info.

src/agentscope/workspace/mcp_enhanced_workspace.py
```diff
@@ -0,0 +1,421 @@
...
+                for t in tools
+            ]
+        except ImportError:
+            return list(tools)
+
+    def invalidate_cache(self) -> None:
+        """Clear the cached tool list so the next query re-fetches."""
+        self._cached_tools = None
+
+    async def get_tool(self, name: str) -> _RestToolProxy:
```
weizhang25: Is get_tool actually used?

src/agentscope/workspace/gateway/_mixin.py
```diff
@@ -0,0 +1,421 @@
...
+        """Whether the client considers itself connected."""
+        return self._connected
+
+    async def connect(self) -> None:
+        """Mark the client as connected."""
+        self._connected = True
+
+    async def close(  # pylint: disable=unused-argument
+        self,
+        ignore_errors: bool = True,
```
weizhang25: unused ignore_errors

src/agentscope/workspace/gateway/_mixin.py
```diff
@@ -0,0 +1,421 @@
...
+    """
+
+    def __init__(
+        self,
+        base_url: str,
+        headers: dict[str, str],
+    ) -> None:
+        self._base_url = base_url
+        self._headers = headers
+        self._connected = True
```
weizhang25: self._connected: A client can close then connect again?
Just remove this connected concept. If list_tools works then it works.

src/agentscope/workspace/gateway/_mixin.py
```diff
@@ -0,0 +1,421 @@
...
+                    name=t["name"],
+                    description=t.get("description", ""),
+                    inputSchema=t.get("inputSchema", {}),
+                )
+                for t in tools
+            ]
+        except ImportError:
+            return list(tools)
+
+    def invalidate_cache(self) -> None:
```
weizhang25: invalidate_cache is not needed. list_tools refreshes tool cache everytime.

Also self._cached_tools is not needed if get_tool is gone.

src/agentscope/workspace/gateway/_mixin.py
```diff
@@ -0,0 +1,421 @@
...
+    * ``_mcp_servers: list[MCPServerConfig]``
+    * ``_gateway_port: int``
+    """
+
+    # These attributes are initialised by the concrete workspace's
+    # ``__init__`` — the mixin only reads/writes them.
+    _gateway_token: str
+    _gateway_mcpc: "MCPClient | _RestGatewayClient | None"
+    _gateway_base_url: str
+    _mcp_servers: list[Any]
```
weizhang25: Any -> MCPServerConfig ?

src/agentscope/workspace/gateway/_mixin.py
```diff
@@ -0,0 +1,421 @@
...
+        """
+        return {}
+
+    async def _exec(
+        self,
+        command: str,
+        *,
+        timeout: float | None = None,
+    ) -> "ExecutionResult":
+        """Execute a command inside the workspace (provided by subclass)."""
```
weizhang25: This _exec sounds like a workspace function, not a gateway function.  But gateway needs it....
Defining it here looks strange. Need to figure out a better design. Just leave this as is for now.

src/agentscope/workspace/gateway/_server.py
```diff
@@ -0,0 +1,447 @@
...
+from mcp.server.fastmcp import FastMCP
+import mcp.types as mtypes
+
+TOOL_NAME_SEPARATOR = "___"
+
+
+# ── upstream MCP client ───────────────────────────────────────────
+
+
+class _UpstreamClient:
```
weizhang25: Can this be named _MCPServerClient? The meaning of Upstream is vague

src/agentscope/workspace/mcp_gateway_app.py
```diff
@@ -0,0 +1,447 @@
...
+
+# ── route table ───────────────────────────────────────────────────
+
+
+class _ToolRoute:
+    __slots__ = ("client", "original_name")
+
+    def __init__(self, client: _UpstreamClient, original_name: str) -> None:
+        self.client = client
+        self.original_name = original_name
```
weizhang25, line 123: Add doc string to explain original name.

src/agentscope/workspace/gateway/_server.py
```diff
@@ -0,0 +1,447 @@
...
+
+class _ToolRoute:
+    __slots__ = ("client", "original_name")
+
+    def __init__(self, client: _UpstreamClient, original_name: str) -> None:
+        self.client = client
+        self.original_name = original_name
+
+
+def _build_routes(
```
weizhang25: Can we just name all tools as server_name__tool_name, even when there is only one such tool name? So all the exposed tool names have the same format

src/agentscope/workspace/gateway/_mixin.py
weizhang25: 1611里，本质上应该在workspace里定义的exec方法，现在被定义在了gatewayMixin里，虽然是gateway里要调用它，但逻辑上看起来有些别扭。
""Execute a command inside the workspace (provided by subclass)."""

我建议这样改一下：

class workspace_base(ABC): 
   @abstractmethod
   def _exec:
        

class local_workspace(ws_base):
    def _exec():
        """implementation"""

class workspace_with_mcp_gateway(ws_base):
        """ws with mcp enhancement"""
        def _exec():
            raise

class docker_ws(ws_w_mcp_gtwy):
        def _exec():
        """implementation"""

class e2b_ws(ws_w_mcp_gtwy):
    def _exec():
        """implementation"""

src/agentscope/workspace/mcp_enhanced_workspace.py
cuiyuebing: It is used by ``Toolkit._load_mcp_tools`` to obtain an invocable handle for each tool discovered via :meth:`list_tools`.

src/agentscope/workspace/gateway/_mixin.py
weizhang25: if _exec is only for gateway mixin, just update the doc string by removing the word workspace. It's confusing.

src/agentscope/workspace/e2b_workspace.py
cuiyuebing: Yep, a skill can contain other types of files.

src/agentscope/workspace/docker_workspace.py
cuiyuebing: It is an absolute path

src/agentscope/workspace/docker_workspace.py
cuiyuebing: It is an absolute path

src/agentscope/workspace/mcp_enhanced_workspace.py
```diff
@@ -0,0 +1,461 @@
...
+    * ``_mcp_servers: list[MCPServerConfig]``
+    * ``_gateway_port: int``
+    * ``_gateway_token: str``
+    * ``_gateway_mcp_client``
+    * ``_gateway_base_url: str``
+    """
+
+    # These attributes are initialised by the concrete workspace's
+    # ``__init__`` — this class only reads/writes them.
+    _gateway_token: str
```
weizhang25, line 197: 174 ~ 178: Why are these class variables not instance variables?

src/agentscope/workspace/gateway/_mixin.py
```diff
@@ -0,0 +1,461 @@
...
+            name: Name of the tool on the gateway.
+            base_url: Gateway base URL (e.g.
+                ``http://127.0.0.1:5600``).
+            headers: HTTP headers (auth token, etc.).
+        """
+        self._name = name
+        self._base_url = base_url
+        self._headers = headers
+
+    async def __call__(self, **kwargs: Any) -> str:
```
weizhang25: nitpick. Will line 72 always return data of str type?

src/agentscope/workspace/mcp_enhanced_workspace.py
weizhang25: The toolkit in v2_dev branch does not have _load_mcp_tools. Could you double check?

src/agentscope/workspace/gateway/_mixin.py
```diff
@@ -0,0 +1,461 @@
...
+            if t_name == name:
+                return _RestToolProxy(
+                    name=name,
+                    base_url=self._base_url,
+                    headers=self._headers,
+                )
+        raise ValueError(f"Tool {name!r} not found")
+
+
+class GatewayMixin(WorkspaceBase):
```
weizhang25: This is not a Mixin pattern anymore. It's a subclass of ws. Rename it to class WorkspaceWithMCP or something similar indicating the idea that this is ws base + mcp.
Also rename this _mixin.py file accordingly.
And remove all the gatewaymixin text from this PR.

src/agentscope/workspace/mcp_enhanced_workspace.py
```diff
@@ -0,0 +1,461 @@
...
+    _gateway_mcp_client: "MCPClient | _RestGatewayClient | None"
+    _gateway_base_url: str
+    _mcp_servers: "list[MCPServerConfig]"
+    _gateway_port: int
+
+    # ── hooks for subclasses ──────────────────────────────────────
+
+    @abstractmethod
+    async def _gw_write_remote(self, path: str, data: bytes) -> None:
+        """Write *data* to *path* inside the container / sandbox."""
```
weizhang25: This is not a mixin pattern anymore. These 3 def _gw_*** methods: We can remove the _gw prefix. All subclasses of this ws_with_mcp class have mcp gateway ability. You don't need to emphasize on the concept of "gateway" anymore.

src/agentscope/workspace/mcp_enhanced_workspace.py
```diff
@@ -0,0 +1,461 @@
...
+            )
+            if resp.status_code >= 400:
+                raise RuntimeError(
+                    f"tool call {self._name!r} failed "
+                    f"({resp.status_code}): {resp.text}",
+                )
+            return resp.json().get("result", "")
+
+
+class _RestGatewayClient:
```
weizhang25: Looks like this client class can be a subclass of MCPClient. The ws_with_mcp init method can just take a MCPClient type parameter.

src/agentscope/workspace/mcp_enhanced_workspace.py
weizhang25: Seems they should be instance variables. The constructors of docker ws and e2b ws should have a super().__init__( ) to use these attributes directly

src/agentscope/workspace/mcp_gateway_app.py
```diff
@@ -0,0 +1,452 @@
...
+from mcp.client.streamable_http import streamable_http_client
+from mcp.server.fastmcp import FastMCP
+
+TOOL_NAME_SEPARATOR = "___"
+
+
+# ── MCP server client ─────────────────────────────────────────────
+
+
+class _MCPServerClient:
```
weizhang25, line 50: This looks very similar to MCPClient. Why not use MCPClient but defining this new class?

src/agentscope/workspace/gateway/_server.py
```diff
@@ -0,0 +1,452 @@
...
+    def __init__(self, client: _MCPServerClient, original_name: str) -> None:
+        self.client = client
+        #: The tool's upstream name before gateway-level prefixing.
+        #: When the host calls ``server___tool``, the gateway uses
+        #: ``original_name`` (i.e. just ``tool``) to invoke the real
+        #: tool on the upstream MCP server via ``session.call_tool``.
+        self.original_name = original_name
+
+
+def _build_routes(
```
weizhang25: Can we move _build_routes( ) into _GatewayState.rebuild( )?

src/agentscope/workspace/mcp_gateway_app.py
```diff
@@ -0,0 +1,452 @@
...
+                break
+        if target is None:
+            return JSONResponse(
+                {"error": f"{name!r} not found"},
+                status_code=404,
+            )
+        _state.clients.remove(target)
+        await target.close()
+        _state.rebuild()
+        _register_proxy_tools(server, _state)
```
weizhang25, line 278: _register_proxy_tools call in remove_mcp looks suspicious.

in add_mcp:
Say server got tools a, b, c from add_mcp(x), and got tools d, e from add_mcp(y). _state has reference to a, b, c, d, e.
after _register_proxy_tools in add_mcp( ): server(a, b, c, d, e) 

Now remove_mcp(x) will remove the client x. _state is rebuilt with reference to d, e only.

Can _register_proxy_tools(server, _state) really remove a b c from server?

src/agentscope/workspace/gateway/_server.py
```diff
@@ -0,0 +1,452 @@
...
+            )
+
+    app.routes.insert(0, Route("/health", _health))
+    app.routes.insert(1, Route("/mcp/add", _add_mcp, methods=["POST"]))
+    app.routes.insert(
+        2,
+        Route("/mcp/remove", _remove_mcp, methods=["POST"]),
+    )
+    app.routes.insert(3, Route("/mcp/list", _list_mcp))
+    app.routes.insert(4, Route("/api/tools", _api_tools))
```
weizhang25: nitpick. (4, ('/tools/list', _list_tools))   (5, ('/tools/call', _call_tool))

src/agentscope/workspace/mcp_enhanced_workspace.py
```diff
@@ -0,0 +1,461 @@
...
+
+        # Refresh tool list so removed tools are no longer visible
+        if self._gateway_mcp_client:
+            await self._gateway_mcp_client.list_tools()
+
+        logger.info("%s: removed MCP %r", type(self).__name__, name)
+
+    # ── gateway lifecycle (called from workspace.initialize) ──────
+
+    async def _start_gateway(self) -> None:
```
weizhang25, line 334: implement workspace_base. initialize( ) like this:

class gateway_ws
async def initialize(self):
    """ implement the abstract function of ws_base"""
     await self._start_gateway( )


class docker_ws(gateway_ws)

async def initialize( ):
    await super().initialize()
    # docker specific init logic goes here
    ....
    

_start_gateway is a gateway-specific function. It should not be exposed in subclass of gateway ws. gateway ws is not a mixin anymore. Use parent-child pattern to build out the call structure.

src/agentscope/workspace/mcp_enhanced_workspace.py
```diff
@@ -0,0 +1,461 @@
+# -*- coding: utf-8 -*-
```
weizhang25, line 1: Seems this _mixin.py should be renamed to gateway_ws.py or ws_with_gateway.py, and be moved into workspace folder.

It is not a fully function ws. so it does not need a manager.py

src/agentscope/workspace/mcp_gateway_app.py
```diff
@@ -0,0 +1,452 @@
+# -*- coding: utf-8 -*-
```
weizhang25, line 1: Rename this file to something similar to the new file name of _mixin.py. Put both into workspace folder directly.

e.g. mcp_enhanced_workspace.py and mcp_http_frontend_app.py

src/agentscope/workspace/mcp_gateway_app.py
cuiyuebing: This intentionally duplicates a subset of ``agentscope.mcp.MCPClient`` because this script is copied into the container as a standalone file and ``agentscope`` is not installed there.

src/agentscope/workspace/mcp_gateway_app.py
```diff
@@ -0,0 +1,452 @@
...
+"""In-container MCP Gateway — runs *inside* a Docker / E2B container.
```
weizhang25: Line 2: an mcp-enhanced ws will copy this file into the ws environment, and execute this file from there.
Do not mention the concrete subclass name/type. This is the parent. the logic should work in any child class.

src/agentscope/workspace/mcp_gateway_app.py
cuiyuebing: Uses a clear-and-rebuild strategy: **all previously registered tools are removed first**, then only the tools present in ``state.schemas`` (which reflects the current set of connected clients) are re-registered.  This guarantees that tools belonging to a removed upstream server are no longer exposed.

src/agentscope/workspace/mcp_enhanced_workspace.py
cuiyuebing: Oh sorry, it should be Toolkit.register_mcp

src/agentscope/workspace/mcp_enhanced_workspace.py
cuiyuebing: ``MCPClient`` extends ``pydantic.BaseModel`` with required fields (``name``, ``is_stateful``, ``mcp_config``) and runs validation in ``model_post_init``.  The REST transport here is different (it forwards calls to an already-running gateway over HTTP) and does not fit the ``StdioMCPConfig | HttpMCPConfig`` discriminated union.

src/agentscope/workspace/mcp_enhanced_workspace.py
weizhang25: The registoer_mcp function expects an MCPClient type, and does a pretty strict type check. It also calls client.is_stateful, which is not defined here.

That's why I suggest this class be extended from MCPClient. and you override all functions with the logic here. So this type of client can be seamlessly accepted everywhere an MCPClient is expected.

src/agentscope/workspace/mcp_enhanced_workspace.py
weizhang25: uh e2b and docker ws still call this _start_gateway function, but not using the super().initialize() approach?

DavdGao:
implemented in #1650

Issue #1608, opened 2026-05-15
[Bug]: tool_choice参数支持的问题


shanghai-Jerry:
**<u>AgentScope is an open-source project. To involve a broader community, we recommend asking your questions in English.</u>**

**Describe the bug**

 tool_choice在传递structure model时，在很多模型上都会出现不支持的报错

shanghai-Jerry:
## Summary of findings from reproducing this issue with DeepSeek V4 reasoning models

### Environment
- AgentScope: 1.0.19.post1
- Python: 3.14
- Model: `deepseek-v4-pro` via `https://api.deepseek.com` (OpenAI-compatible endpoint)
- Provider: DeepSeek official API

### Root cause analysis

**1. ReActAgent always passes tool_choice parameter**

When `reply(msg, structured_model=SomeModel)` is called:
- Line 424 in `_react_agent.py`: `tool_choice = "required"`
- Line 572: `self.model(prompt, tools=self.toolkit.get_json_schemas(), tool_choice=tool_choice)`

The `OpenAIChatModel.__call__` then passes this directly to the API. Many model APIs (especially reasoning/thinking models) reject this parameter.

**2. DeepSeek V4 thinking mode does NOT accept tool_choice**

From DeepSeek thinking mode docs:
- V3.2+ supports tool calls in thinking mode (model can autonomously call tools)
- But the tool_choice parameter itself is not accepted for reasoning models
- Oh My Pi also lists `supportsToolChoice: false` for deepseek-v4-pro and deepseek-v4-flash

Error: `{"error": {"message": "deepseek-reasoner does not support this tool_choice", ...}}`

This applies even when thinking mode is explicitly enabled via `extra_body={"thinking": {"type": "enabled"}}`.

**3. OpenAIChatFormatter strips ThinkingBlock**

`OpenAIChatModel._parse_openai_stream_response` correctly converts reasoning_content to ThinkingBlock, but `OpenAIChatFormatter.supported_blocks` does NOT include ThinkingBlock. The `else` clause silently skips it, meaning reasoning_content is lost in multi-turn conversations.

DeepSeek requires reasoning_content to be preserved in assistant messages when tool calls are involved (otherwise returns 400).

**4. _structured_via_tool_call also sets tool_choice**

The fallback path in `OpenAIChatModel.__call__` also sends tool_choice as a specific function name, with the same issue.

### Impact of removing tool_choice

[...]

PR #1613, opened 2026-05-15
fix(model): skip forced tool_choice for providers that don't support it

FallingSnowFlake:
## AgentScope Version
1.0.19.post1

## Description

Models exposed by DeepSeek / GLM / Kimi, etc., through OpenAI-compatible interfaces have incomplete support for tool_choice="required" or tool_choice specifying a specific function name. Directly passing these parameters will be rejected by the server and throw a BadRequestError.

about issue : #1608

Changes in this update:

- Added a helper function _supports_structured_tool_choice(model_name) that determines, based on the model name, whether the provider supports forced tool_choice.
- In OpenAIChatModel._structured_via_tool_call, when the provider does not support it, tool_choice is dropped. During the response phase, it verifies that the model has indeed invoked the schema tool (covers both streaming and non‑streaming). If not, a ValueError is raised to clearly inform the caller.
- In the two places within ReActAgent.reply where tool_choice="required" is set, provider capability checks have been added to prevent _reasoning from passing unsupported parameters to the model.
- Added a  model_openai_test  to verify that unsupported providers do not include tool_choice in the request.

Relevant model parameter notes:

Kimi doc: https://platform.kimi.com/docs/guide/migrating-from-openai-to-kimi#%E5%85%B3%E4%BA%8E-tool_choice
GLM: https://docs.bigmodel.cn/cn/guide/capabilities/function-calling

## Checklist

- [x]  All tests are passing


src/agentscope/_utils/_common.py
```diff
@@ -27,6 +27,8 @@
 else:
     Tool = "mcp.types.Tool"
 
+_STRUCTURED_TOOL_CHOICE_UNSUPPORTED_PROVIDERS = ("deepseek", "glm", "kimi")
```
shanghai-Jerry, line 30: 这个实现是不是太定制化

FallingSnowFlake: Because the three models currently in main use domestically are these, and none of them have their own SDK support. Secondly, there’s no way to handle this via a formatter.

From a framework practicality perspective, I think support should be provided for some of the commonly used domestic models. I’m not sure if the project maintainers have any different views on this.

If Minimax also doesn’t support it, we can just add it directly, so constants are used here to manage them.

However, there’s one issue to note: DeepSeek on the Alibaba Cloud platform does support OpenAI’s parse method, so when returning structured data, it may not go through the `via_tool...` method internally.

Of course, my development experience might be somewhat limited, so perhaps there’s a better approach?

shanghai-Jerry: https://github.com/agentscope-ai/agentscope/issues/1608， 我这个issue里面其实写了我怎么解决的。通过重写openmodel和formatter，可以让开发者自己选择

DavdGao:

PR #1477, opened 2026-04-14
fix(agent): add missing await on memory.add in DeepResearchAgent._summarizing

Ricardo-M-L:
## Description

In `DeepResearchAgent._summarizing()`, the call to `self.memory.add()` at line 973 is missing the `await` keyword. All other calls to `self.memory.add()` in this file correctly use `await` (lines 229, 273, 288, 358, 404, 422).

Without `await`, the coroutine is never executed, causing:
- `RuntimeWarning: coroutine 'MemoryBase.add' was never awaited`
- The summarization result is not stored in memory

## Changes

Add `await` to the `self.memory.add(summarize_result)` call in `_summarizing()`.

## Checklist

- [x] Pre-commit hooks pass
- [x] Verified all other memory.add calls in the file use await

PR #1654, opened 2026-05-25
fix(realtime): retain strong reference to fire-and-forget tool tasks

Ricardo-M-L:
## Summary

`RealtimeAgent._model_response_loop` schedules each tool call via `asyncio.create_task(self._acting(...))` but discards the returned `Task`. Per the [asyncio docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task):

> Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks.

Without a strong reference, the GC can collect the task before the event loop runs it to completion — the tool call then silently never executes, leaving the model waiting for a result that never arrives and dropping the corresponding side effects.

## Why this catches my eye

The class already uses the \"assign to `self._...`\" pattern for its two long-lived loops:

```python
# _realtime_agent.py L117, L122
self._external_event_handling_task = asyncio.create_task(self._forward_loop(),)
self._model_response_handling_task = asyncio.create_task(self._model_response_loop(outgoing_queue),)
```

so the codebase clearly knows the rule. The tool-execution `create_task` at L290 (the bare fire-and-forget) is the one site that missed it. Same flavour as the missing `await` you accepted in #1477.

## Fix shape

Standard \"keep a set, auto-discard on done\" idiom that the Python docs recommend for fire-and-forget tasks:

```python
self._tool_execution_tasks: set[asyncio.Task] = set()  # in __init__

# at the call site
task = asyncio.create_task(self._acting(...))
self._tool_execution_tasks.add(task)
task.add_done_callback(self._tool_execution_tasks.discard)
```

This:
- Keeps a strong reference until the task completes (so GC can't reclaim it)
- Auto-removes the entry on completion (so the set doesn't grow unbounded)
- Is the pattern recommended in the asyncio docs for exactly this case

## Test plan

- [ ] Existing `RealtimeAgent` tests should remain green — semantics unchanged for tools that complete normally
- [ ] Manual: long-running session with many tool calls under memory pressure — without the fix, some tool calls were observed never returning; with the fix, all complete


DavdGao:

PR #1701, opened 2026-05-27
feat(omni): streaming audio + live captions for DashScope/OpenAI omni

qbc2016:
## AgentScope Version

2.0.0

## Description

End-to-end streaming audio for omni-style chat models (Qwen omni, gpt-audio-mini), playable in the web UI with live captions. Both providers deliver raw PCM upstream; we wrap it as WAV (RIFF header on the first chunk, raw PCM thereafter, fixed-size WAV on the final chunk) so the frontend can start playback immediately while the model keeps streaming.

## What changed

- **OpenAI streaming audio**: force `audio.format=pcm16` on the wire (OpenAI rejects other formats with `stream=true`) and re-wrap as WAV downstream. Mirrors the DashScope path.
- **Live captions**: fold `delta.audio.transcript` into `delta_text` so per-delta TextBlocks emit alongside audio chunks. Without this, the agent's streaming pipeline (which only routes `is_last=False` chunks through `_convert_chat_response_to_event`) never produced TextBlockDeltaEvents for omni transcripts and the UI showed audio
  but no text.
- **Shared WAV header builder**: extracted `_streaming_wav_header` (was duplicated across both models) into `agentscope/_utils/_audio.py::build_streaming_wav_header`.
- **Model card schema**: `from_yaml` now resolves Pydantic `$ref`s and collapses `Optional[X]` (`anyOf` + `null`) so nested fields like `audio.voice.enum` can be overridden per-model. Override application uses deep-merge so sibling keys survive. The `audio` parameter is auto-filtered out for models that don't declare an `audio/*` output type.
- **Per-model voice/format**: `gpt-audio-mini.yaml` and `qwen3.5-omni-plus.yaml` carry their authoritative voice enums and lock `format.enum: [pcm16]` (verified empirically — DashScope upstream is raw PCM regardless of the requested format). Pydantic defaults aligned: `voice` → `alloy`/`Tina`, `format` → `pcm16`.
- **Formatter cleanup**: removed the dead `role` parameter from `_format_openai_data_block` / `_format_response_data_block` and all call sites in the moonshot and openai-response formatters.


src/agentscope/message/_base.py
```diff
@@ -283,7 +284,18 @@ def append_event(self, event: AgentEvent) -> Self:
                         event.block_id,
                     )
                 elif event.data:
-                    block.source.data += event.data
```
DavdGao, line 286: We have similar logic in agentscope/agentscope-typescript, which handles frontend rendering. Please create a PR in that repo to keep the concatenation logic consistent.

src/agentscope/_utils/_audio.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
+"""Audio utilities shared across model providers."""
+import struct
+
+
+def build_streaming_wav_header(
```
DavdGao: Prefix with “_”

src/agentscope/message/_base.py
qbc2016: https://github.com/agentscope-ai/agentscope-typescript/pull/31

src/agentscope/model/_dashscope/_model.py
```diff
@@ -393,11 +443,25 @@ async def _parse_stream_response(
                 ),
             )
         if acc_audio_data:
+            # PCM bytes were already streamed incrementally above (first
```
DavdGao, line 446: Why here we turn the pcm data into wav format rather than directly return the accumulated pcm data

qbc2016: Three reasons in order of importance:

1. **Direct (non-Agent) callers need a self-contained block.**
When you go through `Agent` → SSE → frontend, the `WavStreamPlayer` consumes the per-delta chunks live and the final `is_last` block actually gets filtered out of the message history (audio blocks aren't fed back into the model's context next turn). But if someone uses the model directly:

```python
async for resp in await model([...]):
    pass
final = resp  # only the last one
```
…they only have the final `ChatResponse` in hand. The streaming chunks are gone. They need one block that fully represents the audio, ready to serialize / save to disk / render in `<audio controls>`.

**2. Raw PCM has no metadata.**
PCM bytes alone don't carry sample rate / channels / bit depth — a downstream consumer has no way to decode them without out-of-band info. WAV's RIFF header carries all of that in a standard way, so the block is self-describing.

**3. The streaming chunks aren't a "proper" WAV.**
The first streaming chunk uses `_build_streaming_wav_header()` which sets the RIFF chunk size and `data` chunk size to `0xFFFFFFFF` (placeholder for "unknown length yet"). That's fine for live decoders that read until EOS, but it's not a valid WAV file you can save to `audio.wav` and double-click — many strict decoders (and serialization tools) want real lengths. So we re-assemble with `wave.open()`, which writes the actual sizes into the header.

Bonus: `media_type` stays consistent (`audio/wav` for both the streaming deltas and the final block), so any consumer can treat them uniformly.

examples/web_ui/frontend/src/components/chat/MessageBubble.tsx
```diff
@@ -176,6 +219,9 @@ function renderBlock(
 
 		case 'data': {
 			const dataType = block.source.media_type.split('/')[0];
+			if (dataType === 'audio') {
```
DavdGao: What's the difference here

examples/web_ui/frontend/package.json
```diff
@@ -11,7 +11,7 @@
 		"preview": "vite preview"
 	},
 	"dependencies": {
-		"@agentscope-ai/agentscope": "^0.0.6",
+		"@agentscope-ai/agentscope": "^0.0.5",
```
DavdGao: rollback the change here

examples/web_ui/frontend/package.json
```diff
@@ -44,7 +43,6 @@
 	},
 	"devDependencies": {
 		"@eslint/js": "^10.0.1",
-		"@tailwindcss/typography": "^0.5.19",
```
DavdGao, line 47: why remove the typography here

examples/web_ui/frontend/package.json
```diff
@@ -35,7 +35,6 @@
 		"react-router-dom": "^7.15.1",
 		"remark-gfm": "^4.0.1",
 		"shadcn": "^4.7.0",
-		"shadcn-prose": "^1.2.1",
```
DavdGao, line 38: similarly

examples/web_ui/frontend/src/components/chat/MessageBubble.tsx
qbc2016: putting it inside `switch (dataType)` would still run `let data = \data:...;base64,${block.source.data}`first. For a streaming audio block, `block.source.data` keeps growing per delta, so we'd rebuild a multi-MB base64 string on every render for nothing. Early-returning skips that work.

DavdGao (changes requested):
1. Clicking "Replay Audio" on multiple bubbles simultaneously causes overlapping playback. Please ensure that only one audio plays at a time — starting a new one should automatically stop the current one.

2. Please unify the voice selection UI by replacing the current selector with a ShadCN DropdownMenu component to maintain a consistent style.

Image

3. The parameter settings panel appears misaligned after this PR's changes. Please refer to the current main branch where labels and inputs are properly aligned in separate columns, and restore the correct layout accordingly.

Image

DavdGao (commented):
lgtm

DavdGao (approved):

PR #1703, opened 2026-05-27
fix(tool): support plain return values in FunctionTool

### LINKED ISSUE #1702: [Bug]: FunctionTool fails when the documented plain function tool returns a string
## Describe the bug

The documentation shows that a plain Python function can be wrapped as a tool:

https://docs.agentscope.io/v2/building-blocks/tool#wrap-function-as-tool

In that section, the `get_weather` example returns a plain `str`:

```python
def get_weather(location: str) -> str:
    """Get weather information.

    Args:
        location (`str`):
            The location to get weather for.
    """
    return f"The weather in {location} is sunny."
```

This suggests that a normal Python function returning `str` can be wrapped and used as a tool.

However, when such a function is wrapped with `FunctionTool`, the tool call fails at runtime. `FunctionTool.__call__()` currently returns the wrapped function result directly, while `Toolkit.call_tool()` only accepts tool results in AgentScope's internal tool result format.

As a result, the documented `get_weather(...) -> str` style function does not work unless users manually return `ToolChunk`.

## To Reproduce

```python
import asyncio
import json

from agentscope.tool import FunctionTool, Toolkit


def get_weather(location: str) -> str:
    """Get weather information.

    Args:
        location (`str`):
            The location to get weather for.
    """
    return f"The weather in {location} is sunny."


async def main() -> None:
    toolkit = Toolkit(tools=[FunctionTool(get_weather)])

    tool_call = ToolCallBlock(
        id="call_get_weather",
        name="get_weather",
        input=json.dumps({"location": "Chengdu"}),
    )

    async for item in toolkit.call_tool(tool_call, AgentState()):
        print(item)


asyncio.run(main())
```

Run:

```bash
python reproduce_function_tool_plain_return.py
```

## Error messages

The call fails because the wrapped function returns `str`:

```text
DeveloperOrientedException: The tool function must return a ToolChunk object, or an AsyncGenerator/Generator of ToolChunk objects, but got <class 'str'>.
```

## Expected behavior

The documented function should work when wrapped as a tool:

```python
def get_weather(location: str) -> str:
    return f"The weather in {location} is sunny."
```

It should produce a successful text tool result instead of raising a `DeveloperOrientedException`.

If `FunctionTool` intentionally requires wrapped functions to return `ToolChunk`, then the documentation should state that explicitly and update the `get_weather` example accordingly.

## Environment

- AgentScope Version: 2.0.0, source checkout from `main` at commit `fca8f160`
- Python Version: 3.11.14
- OS: macOS 26.5 arm64

xunx911:
## AgentScope Version

2.0.0

## Description

Fixes #1702.

The `Wrap function as tool` documentation shows a plain Python function returning `str`:

https://docs.agentscope.io/v2/building-blocks/tool#wrap-function-as-tool

However, `FunctionTool.__call__()` previously returned the wrapped function result directly. When the wrapped function returned a plain value such as `str` or `dict`, `Toolkit.call_tool()` treated it as an invalid tool result and raised `DeveloperOrientedException`.

This PR normalizes plain Python return values in `FunctionTool` into `ToolChunk` text results.

The PR is split into two commits intentionally:

1. `fix(tool): support plain return values in FunctionTool`

   This is the minimal fix for #1702. It normalizes non-streaming plain return values while preserving existing `ToolChunk` behavior.

2. `feat(tool): normalize plain yielded values in FunctionTool streams`

   This applies the same normalization logic to values yielded by sync and async generator tools. If maintainers prefer this PR to only include the minimal bug fix, I can drop this second commit.

Validation:

```bash
pre-commit run --all-files
pytest tests
```

I also verified the documented `get_weather(...) -> str` style tool call with a local fork-branch installation. It now produces a successful tool result instead of raising `DeveloperOrientedException`.

## Checklist

- [x]  All tests are passing

### DIFF
--- a/src/agentscope/tool/_adapters.py
+++ b/src/agentscope/tool/_adapters.py
@@ -116,8 +116,27 @@ class FunctionTool(ToolBase):
         else:
             result = self._func(**kwargs)
 
-        if isinstance(result, (AsyncGenerator, Generator)):
-            return result
+        if isinstance(result, AsyncGenerator):
+
+            async def _stream() -> AsyncGenerator[ToolChunk, None]:
+                async for chunk in result:
+                    if isinstance(chunk, ToolChunk):
+                        yield chunk
+                    else:
+                        yield self._convert_func_result_to_chunk(chunk)
+
+            return _stream()
+
+        if isinstance(result, Generator):
+
+            async def _stream() -> AsyncGenerator[ToolChunk, None]:
+                for chunk in result:
+                    if isinstance(chunk, ToolChunk):
+                        yield chunk
+                    else:
+                        yield self._convert_func_result_to_chunk(chunk)
+
+            return _stream()
 
         return self._convert_func_result_to_chunk(result)
 
--- a/src/agentscope/tool/_adapters.py
+++ b/src/agentscope/tool/_adapters.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 """Adapters to convert functions and MCP tools to ToolProtocol."""
 import inspect
+import json
 from contextlib import _AsyncGeneratorContextManager
 from datetime import timedelta
-from typing import Callable, Any, AsyncGenerator
+from typing import Callable, Any, AsyncGenerator, Generator
 
 from mcp import ClientSession
 import mcp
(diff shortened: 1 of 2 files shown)


DavdGao (commented):
@xunx911 Plz add unittests for the plain return values, and others lgtm

xunx911:
Added unit tests for plain FunctionTool return values, including sync string returns and async JSON-serializable returns. I also added tests for the optional streaming normalization commit to cover sync and async generators yielding plain values.

Validation:

```bash
./.venv/bin/python -m py_compile src/agentscope/tool/_adapters.py
./.venv/bin/python -m pytest tests/toolkit_test.py -q
```

Result: `15 passed`.

DavdGao (approved):
lgtm

PR #1705, opened 2026-05-28
fix(webui): fix frontend build with current dependencies

yuanchangsai77:
## Summary

- Update the calendar classNames mapping for the installed react-day-picker version by using month_grid instead of the removed table key.
- Switch the skills hook from the missing skillApi export to the existing workspaceApi.skill client.
- Add TypeScript 6 deprecation handling so the frontend project can type-check with the current compiler.

## Why

The Web UI frontend currently fails type-checking with the resolved dependency versions because a react-day-picker classNames key is no longer valid, skillApi is not exported from the API module, and TypeScript 6 reports the existing baseUrl configuration as a blocking deprecation.

## Validation

- ./node_modules/.bin/tsc -b
- ./node_modules/.bin/vite build


yuanchangsai77:
Closing in favor of #1708, which uses the contribution-guide style branch name and links the tracking issue.

PR #1710, opened 2026-05-28
fix(workspace): add locks to mcp and skill operations in LocalWorkspace.

weizhang25:
## Description

The MCP add/remove operations, and skills list/add/remove operations in LocalWorkspace are not protected with lock. This will result in corrupted data.

Each skill operation involve reading and writing of the single .skills index file, and each individual skill folder and the underlying SKILL.md. Too much work to apply lock at the actual location where file IO is executed. So just simply apply the lock to the top level add/remove/list functions.

## Checklist

- [ ]  All tests are passing


DavdGao (commented):
Thanks for the contribution! A few points:

**1. `_skill_lock` doesn't actually protect `sessions/` and `data/` in `reset()`.**
`reset()` holds `_skill_lock` while `rmtree`-ing `skills/`, `sessions/`, and `data/`. But the writers of those last two directories — `offload_context()`, `offload_tool_result()`, `_offload_data_block()` — don't acquire any lock. So `reset()` can still race with concurrent offloads (rmtree may fail mid-traversal, or an offload may write into a directory that's just been deleted). The race isn't introduced by this PR, but adding a `_skill_lock` block around those rmtree calls reads as if they're protected when they aren't.

Two ways to make the lock honest:
- **(a)** Rename `_skill_lock` → `_workspace_lock` and have `offload_context`/`offload_tool_result`/`_offload_data_block` also acquire it. The PR then actually closes the race.
- **(b)** Move the `sessions/` and `data/` rmtree calls *outside* `_skill_lock` and treat that race as a separate issue. The remaining lock then matches its name and scope.

I'd lean toward (b) since this PR's scope is "mcp and skill operations" and tackling sessions/data is scope creep — but happy with either as long as the lock's coverage matches its name.

DavdGao (approved):
lgtm

PR #1716, opened 2026-05-28
fix: wrap string function tool results

### LINKED ISSUE #1702: [Bug]: FunctionTool fails when the documented plain function tool returns a string

he-yufeng:
﻿## Summary
- convert plain string results from FunctionTool into text ToolChunk output
- keep existing ToolChunk and streaming result handling unchanged
- add regression coverage for the documented get_weather-style function tool

Fixes #1702.

## To verify
- python -m py_compile src\agentscope\tool\_adapters.py tests\toolkit_test.py
- python -m pytest tests\toolkit_test.py -k "sync_function_returning_string or sync_non_streaming_function" -q
- git diff --check

### DIFF
--- a/src/agentscope/tool/_adapters.py
+++ b/src/agentscope/tool/_adapters.py
@@ -115,6 +115,9 @@ class FunctionTool(ToolBase):
         else:
             result = self._func(**kwargs)
 
+        if isinstance(result, str):
+            return ToolChunk(content=[TextBlock(text=result)])
+
         return result
 
 
--- a/tests/toolkit_test.py
+++ b/tests/toolkit_test.py
@@ -523,6 +523,39 @@ class RegisterFunctionTest(IsolatedAsyncioTestCase):
             },
         )
 
+    async def test_sync_function_returning_string(self) -> None:
+        """Test wrapping a plain function that returns text."""
+
+        def get_weather(location: str) -> str:
+            """Get weather information.
+
+            Args:
+                location: The location to get weather for
+            """
+            return f"The weather in {location} is sunny."
+
+        toolkit = Toolkit(tools=[FunctionTool(get_weather)])
+
+        state = AgentState()
+        tool_call = ToolCallBlock(
+            id="test_weather",
+            name="get_weather",
+            input=json.dumps({"location": "Chengdu"}),
+        )
+
+        chunks = []
+        response = None
+        async for result in toolkit.call_tool(tool_call, state):
+            if isinstance(result, ToolChunk):
+                chunks.append(result)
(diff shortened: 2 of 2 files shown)


DavdGao:
Thanks for your interest in contributing to AgentScope. Before submitting, please check whether the issue has already been assigned. This issue is already being tracked and addressed in issue #1702 and PR #1703. Closing this for now.

PR #1728, opened 2026-05-29
security(mcp): add command allowlist and shell metacharacter validation

dfgvaetyj3456356-hash:
MCP command allowlist + shell metacharacter validation security fix.

### DIFF
--- a/src/agentscope/mcp/_config.py
+++ b/src/agentscope/mcp/_config.py
@@ -1,15 +1,46 @@
 # -*- coding: utf-8 -*-
 """The MCP configurations."""
+import re
+import shutil
 from pathlib import Path
 from typing import Literal
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, ValidationInfo, field_validator
+
+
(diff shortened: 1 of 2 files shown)


dfgvaetyj3456356-hash:
Hi maintainers, friendly follow-up on this focused MCP command validation change. The goal is to keep MCP config execution paths explicit and reject shell metacharacter/unsafe command inputs before they reach execution. Happy to adjust the allowlist shape or tests if you prefer a different compatibility tradeoff. Thanks for taking a look.

DavdGao:

> Hi maintainers, friendly follow-up on this focused MCP command validation change. The goal is to keep MCP config execution paths explicit and reject shell metacharacter/unsafe command inputs before they reach execution. Happy to adjust the allowlist shape or tests if you prefer a different compatibility tradeoff. Thanks for taking a look.

I'm curious about when the allowed commands validation actually kicks in. Since both command and allowed_commands are defined together on the same config class, it seems like the security check could be trivially bypassed — whoever sets the command can just whitelist it at the same time, which would render the restriction meaningless.

PR #1729, opened 2026-05-29
security(utils): prevent SSRF by blocking private/internal URLs

dfgvaetyj3456356-hash:
SSRF IP pinning + timeout security fix.

### DIFF
--- a/src/agentscope/_utils/_common.py
+++ b/src/agentscope/_utils/_common.py
@@ -4,12 +4,15 @@ import asyncio
 import base64
 import functools
 import inspect
+import ipaddress
 import json
 import os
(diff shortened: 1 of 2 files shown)


dfgvaetyj3456356-hash:
Hi maintainers, friendly follow-up on the SSRF hardening PR. The change blocks private/internal URL targets before fetching remote bytes while preserving normal public URL use. Happy to tune the validation behavior or add more cases if there is a preferred project pattern. Thanks.

DavdGao:
I'm not sure if we should restrict downloads from local URLs. @qbc2016, what do you think?

PR #1734, opened 2026-05-29
fix(storage): expire Redis message lists

### LINKED ISSUE #1704: [Bug]: RedisMemory.add() leaks msg keys with TTL=-1 due to pipeline SCAN race
**Describe the bug**

Due to the usage of redis pipeline usage. When RedisMemory.add() stores a new message, the last msg key created in each add() call never gets a TTL set. Over multiple conversation turns, this causes orphan keys to accumulate in Redis and never expire.

**Root Cause**
In _working_memory/_redis_memory.py, add() (line ~540-554) uses a pipeline to batch Redis operations:

pipe = self._client.pipeline()

for m in messages_to_add:
    await pipe.set(                                    # ① SET queued, NOT written yet
        self._get_message_key(m.id),
        json.dumps(m.to_dict(), ensure_ascii=False),
    )

await self._refresh_session_ttl(pipe=pipe)            # ② SCAN Redis → add expire for found keys
await pipe.execute()                                   # ③ SET + EXPIRE committed together
_refresh_session_ttl() (line ~277-313) uses SCAN to discover all session keys, then adds EXPIRE commands to the same pipeline:

async def _refresh_session_ttl(self, pipe=None):
    ...
    cursor, keys = await self._client.scan(            # SCAN runs IMMEDIATELY
        cursor, match=self._get_session_pattern(),
    )
    for key in keys:
        await pipe.expire(key, self.key_ttl)           # EXPIRE queued in pipe
The problem: At step ②, SCAN runs immediately against Redis's current state. But the pipe.set() at step ① has not been committed yet — it's still queued in the pipeline. So SCAN cannot see the new msg key, and no EXPIRE command is queued for it. When pipe.execute() runs at step ③, the new msg key is created but has no TTL (TTL=-1).

The next add() call WILL find this key via SCAN and fix its TTL, but the last msg key in the current call remains uncovered.

**To Reproduce**
import asyncio
from agentscope.memory import RedisMemory

async def test():
    mem = RedisMemory(
        session_id="test-session",
        user_id="test-user",
        host="127.0.0.1",
        port=6379,
        password="...",
        key_prefix="test:",
        key_ttl=10,                     # TTL 10 seconds
    )
    from agentscope.message import Msg
    await mem.add(Msg("user", "hello", "user"))
    # Wait a few seconds, then check Redis:
    #   test:user_id:test-user:session:test-session:messages  → TTL ~10s ✅
    #   test:user_id:test-user:session:test-session:msg:xxx   → TTL=-1  ❌

**Expected behavior**
All keys created by add() should have TTL set, so they expire automatically when the session expires.

**Error messages**
None

- AgentScope Version: 1.0.19.post1
- Python Version: 3.11.15
- OS: windows

**Suggested Fix**
Call pipe.expire() immediately after pipe.set() for the new message key:

for m in messages_to_add:
    await pipe.set(
        self._get_message_key(m.id),
        json.dumps(m.to_dict(), ensure_ascii=False),
    )
    if self.key_ttl is not None:
        await pipe.expire(self._get_message_key(m.id), self.key_ttl)  # ← add this
The same should apply to pipe.rpush() for the session key (line ~532) and mark-related keys (lines ~547-549) if they are also new.

he-yufeng:
## Summary
- refresh TTL on Redis message list writes, including streaming replacements
- keep message history lists from surviving after the configured storage TTL
- add fakeredis coverage for append and replace paths

Fixes #1704

## To verify
- python -m pytest tests\storage_redis_test.py -q
- python -m black --check --line-length=79 src\agentscope\app\storage\_redis_storage.py tests\storage_redis_test.py
- python -m flake8 --extend-ignore=E203 src\agentscope\app\storage\_redis_storage.py tests\storage_redis_test.py
- python -m py_compile src\agentscope\app\storage\_redis_storage.py tests\storage_redis_test.py
- git diff --check

### DIFF
--- a/src/agentscope/app/storage/_redis_storage.py
+++ b/src/agentscope/app/storage/_redis_storage.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # pylint: disable=too-many-public-methods
 """The Redis storage implementation."""
+
 from datetime import datetime
 from typing import Any, TYPE_CHECKING, Self
 
@@ -114,6 +115,10 @@ class RedisStorage(StorageBase):
     async def _set_with_ttl(self, key: str, value: str) -> None:
         """SET a key and optionally apply the sliding TTL."""
         await self._client.set(key, value)
+        await self._refresh_key_ttl(key)
+
+    async def _refresh_key_ttl(self, key: str) -> None:
+        """Apply the sliding TTL to a key, if configured."""
         if self.key_ttl is not None:
             await self._client.expire(key, self.key_ttl)
 
@@ -816,8 +821,10 @@ class RedisStorage(StorageBase):
             last_msg = Msg.model_validate_json(last_raw)
             if last_msg.id == msg.id:
                 await self._client.lset(key, -1, msg.model_dump_json())
+                await self._refresh_key_ttl(key)
                 return
         await self._client.rpush(key, msg.model_dump_json())
+        await self._refresh_key_ttl(key)
 
     async def get_message(
         self,
--- a/tests/storage_redis_test.py
+++ b/tests/storage_redis_test.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """Unit tests for RedisStorage using fakeredis."""
+
 from unittest.async_case import IsolatedAsyncioTestCase
 
 import fakeredis.aioredis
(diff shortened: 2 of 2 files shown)


DavdGao (commented):
Thanks for the contribution! Two minor cleanups:

1. The `_refresh_key_ttl` docstring says "for non-string Redis structures", but `EXPIRE` is type-agnostic — the helper isn't tied to any value type. Suggest dropping that qualifier, e.g. "Apply the sliding TTL to a key, if configured."
2. `_set_with_ttl` could reuse this helper instead of repeating the `if self.key_ttl is not None: await self._client.expire(...)` branch inline.

he-yufeng:
Updated the review cleanup in `9dfa04eb`:

- `_set_with_ttl()` now reuses `_refresh_key_ttl()` instead of duplicating the `expire()` call.
- `_refresh_key_ttl()` no longer implies it is only for non-string structures.

Validation:
- `python -m ruff check src\agentscope\app\storage\_redis_storage.py tests\storage_redis_test.py`
- `.\.venv\Scripts\python.exe -m pytest tests\storage_redis_test.py -q` -> 34 passed
- `git diff --check -- src\agentscope\app\storage\_redis_storage.py tests\storage_redis_test.py`

DavdGao (commented):
Plz run `pre-commit run --all-files` to format the code. others lgtm

DavdGao (approved):

PR #1735, opened 2026-05-30
fix(tool): clean the file cache of the builtin Read tool

### LINKED ISSUE #1723: fix(tool): drop the read cache when a tool call block is evicted during context compression
**Background**

Currently, `Write` and `Edit` tools are coupled with `Read` — a file must be read before it can be modified. This read result is cached on the corresponding `Read` tool call block to facilitate the validation.

However, during context compression, when a `Read` tool call block is evicted from the context, its associated cache is not cleared. Over time, this causes the cache to grow unboundedly, leading to unnecessary memory consumption.

**Changes**

- During context compression, explicitly identify evicted `Read` tool call blocks and clear their associated cache
- This ensures the cache lifecycle is tied to the corresponding tool call block's presence in the context, preventing unbounded cache growth

googs1025:
## Summary
- derive Read cache retention from file paths referenced by reserved context
- clear evicted Read caches during context compression
- add regression tests for evicted, reserved, and unreferenced Read caches

Fixes #1723

## Tests
- uv run pytest tests/builtin_file_cache_test.py tests/compress_context_test.py tests/builtin_read_test.py tests/builtin_write_test.py tests/builtin_edit_test.py

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -436,7 +436,7 @@ class Agent:
                 f"'{path}', you can refer to it when needed.</system-reminder>"
             )
 
-        self._clear_unreserved_read_cache(msgs_to_reserve)
+        await self._clear_unreserved_read_cache(msgs_to_reserve)
 
         # Update the context
         self.state.context = msgs_to_reserve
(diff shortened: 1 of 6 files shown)


googs1025:
实现逻辑更新：Read cache 本身按 file path 维护，一份缓存可能被多个 Read tool call 共享，所以不再给 cache entry 绑定 tool_call_id。上下文压缩完成后，只从保留的消息中收集仍然存在的 Read file path，并让 ToolContext 只保留这些路径对应的缓存；没有被 reserved Read 引用的 read cache 会被清理。这样既能在 Read block 被压缩后释放无引用缓存，也能保留同一文件被再次 Read 且仍在上下文中的缓存。已补充对应回归测试。

DavdGao (changes requested):
Thanks for the contribution!

I'd push back on the `tool_call_id` injection approach though. The Read cache is keyed by `file_path` and holds one entry per file (a later Read on the same file overwrites the earlier entry in `cache_file`), while many Read tool calls can share that single entry. Stamping one `tool_call_id` onto a path-keyed entry encodes a one-to-one relationship that doesn't actually exist, and it leads to a real bug: if `tool_call_2` overwrites the cache for file A and is later evicted while `tool_call_1` is reserved, the entry gets dropped even though the reserved call still depends on it.

Since `file_path` already identifies the cache, eviction can be derived purely from the reserved messages: collect the `file_path`s referenced by Read tool calls in `msgs_to_reserve`, and drop every cache entry whose path is not in that set. No need to inspect the compressed messages at all. That removes `ReadCacheEntry.tool_call_id`, the `_tool_call_id` parameter on `Read.__call__`, and the `inspect.signature` injection in `Toolkit.call_tool` — only `_agent.py` and `ToolContext` need to change.

googs1025:
已按你的建议调整为基于 `file_path` 的清理逻辑。现在会从压缩后保留的上下文里收集还存在的 `Read` 调用路径，然后只保留这些路径对应的 read cache。

之前那版 `tool_call_id` 的绑定已经去掉了，也补了几个回归测试，覆盖被压缩掉、仍被保留、以及没有引用时的 cache 清理场景。

测试已跑：
`uv run pytest tests/builtin_file_cache_test.py tests/compress_context_test.py tests/builtin_read_test.py tests/builtin_write_test.py tests/builtin_edit_test.py`

DavdGao (approved):
lgtm

PR #1738, opened 2026-05-31
feat(middleware): add BudgetControlMiddleware for token budget enforcement

### LINKED ISSUE #1727: feat(middleware): add a budget middleware to control the total token usage
**Background**

In scheduled tasks, users may not be present to monitor agent execution. To prevent agents from running indefinitely, it is necessary to enforce a maximum token budget. Once the budget is reached, a hint message should be injected before the next reasoning step, prompting the agent to exit the ReAct loop and produce a concluding response (i.e., without invoking any tools).

In AgentScope 2.0, this can be elegantly handled via middleware. Since the `Msg` class already tracks token usage statistics, we only need to accumulate the token consumption of the current reply before each reasoning step. When the usage exceeds the predefined token budget, a hint message is injected prior to reasoning, instructing the agent to wrap up immediately.

**Changes**

- Add a new middleware subclass under `agentscope/middleware/`:

```python
class BudgetControlMiddleware(MiddlewareBase):
    def __init__(
        self, 
        max_tokens: int, 
        hint_message:str = "You have reached the maximum token budget. "
                                          "Please wrap up immediately and provide a final "
                                          "concluding response without invoking any tools.",
    ) -> None:
        self.max_tokens = max_tokens

    def on_reasoning(self, ...) -> ...:
        ...
```

MannXo:
## Summary

Implements `BudgetControlMiddleware` as requested in #1727.

- Hooks `on_reasoning` to intercept each reasoning step in the ReAct loop
- Tracks cumulative token usage (`input_tokens + output_tokens`) per reply via `ModelCallEndEvent`, keyed by `agent.state.reply_id` so parallel replies on the same instance don't interfere
- Once `used >= max_tokens`, injects a configurable hint message into `agent.state.context` and forces `tool_choice=ToolChoice(mode="none")` before the next reasoning step, prompting the agent to wrap up without further tool calls
- Exports `BudgetControlMiddleware` from `agentscope.middleware`

## Usage

```python
from agentscope.middleware import BudgetControlMiddleware

agent = Agent(
    ...,
    middlewares=[BudgetControlMiddleware(max_tokens=10000)],
)
```

## Test plan

- [x] `test_under_budget_no_hint_injected` — hint not added when usage stays below threshold
- [x] `test_budget_exceeded_injects_hint` — hint message appears in context when budget is hit
- [x] `test_budget_exceeded_forces_tool_choice_none` — `tool_choice` is overridden to `none` when budget is hit
- [x] `test_token_accumulation_across_steps` — `input + output` tokens from `ModelCallEndEvent` accumulate correctly
- [x] `pre-commit run --all-files` — all hooks pass
- [x] `pytest tests` — 473 passed, 147 skipped, 0 failures

Closes #1727


src/agentscope/middleware/_budget.py
```diff
@@ -0,0 +1,101 @@
...
+                Maximum total tokens (input + output) allowed per reply.
+                Once this threshold is reached, the agent is instructed to
+                wrap up without calling any more tools.
+            hint_message (`str`, optional):
+                The message injected into the agent's context when the
+                budget is exceeded. Defaults to a built-in wrap-up prompt.
+        """
+        self.max_tokens = max_tokens
+        self.hint_message = hint_message
+        self._used_tokens: dict[str, int] = {}
```
DavdGao: Leaks one entry per reply — the dict only ever grows. Hook `on_reply` and pop the entry when `ReplyEndEvent` fires (with a `finally:` cleanup as a safety net for error paths).

src/agentscope/middleware/_budget.py
```diff
@@ -0,0 +1,101 @@
...
+        Yields:
+            Events from the reasoning process.
+        """
+        reply_id = agent.state.reply_id
+        used = self._used_tokens.get(reply_id, 0)
+        tool_choice = input_kwargs.get("tool_choice")
+
+        if used >= self.max_tokens:
+            agent.state.context.append(
+                UserMsg(name="user", content=self.hint_message),
```
DavdGao: Don't fabricate a `UserMsg` — it misrepresents the conversation history (the user didn't say this) and breaks "one user input ↔ one assistant reply." Use `HintBlock` instead (designed for exactly this, converted to a user message at the LLM boundary; see `_tool_offload_middleware.py` for an example). Also: today this fires on **every** reasoning step after the budget is hit, polluting context with duplicates. Gate it so the hint is injected at most once per reply (track per `reply_id`, clear alongside `_used_tokens`).

src/agentscope/middleware/_budget.py
```diff
@@ -0,0 +1,101 @@
...
+            )
+            tool_choice = ToolChoice(mode="none")
+
+        async for event in next_handler(tool_choice=tool_choice):
```
DavdGao: Forward all `input_kwargs`, not just `tool_choice`:
```python
input_kwargs["tool_choice"] = tool_choice
async for event in next_handler(**input_kwargs):
```
Today `input_kwargs` only carries `tool_choice`, but if the schema grows this middleware will silently drop new fields.

tests/middleware_budget_test.py
```diff
@@ -0,0 +1,166 @@
...
+        # At least one call must have received tool_choice with mode="none"
+        self.assertTrue(
+            any(
+                getattr(tc, "mode", None) == "none"
+                for tc in received_tool_choices
+                if tc is not None
+            ),
+        )
+
+    async def test_token_accumulation_across_steps(self) -> None:
```
DavdGao, line 211: Only one response is mocked, so this tests "one event gets recorded," not "tokens accumulate across steps." Mock several responses (each below the budget) whose **cumulative** total crosses the threshold; assert (a) the running sum and (b) the hint fires only after the cumulative crossing.

DavdGao (changes requested):
Thanks for the contribution! Inline comments cover the changes I'd like before this merges.

src/agentscope/middleware/_budget.py
DavdGao: We should clarify the scope of the token budget — whether it applies across the entire execution (spanning multiple replies) or just a single reply. This distinction matters significantly in the agent service.

Currently, middlewares are stateless and not persisted. In a stateless chat endpoint where the agent and middlewares are assembled fresh on each request, used_tokens is never saved. This becomes especially problematic in human-in-the-loop scenarios:

- First reply: yields a `RequireUserConfirmationEvent` and the request ends — middleware state is dropped
- User returns `UserConfirmationResultEvent`: the reply resumes, but the newly assembled middleware has lost all previously accumulated state.

src/agentscope/middleware/_budget.py
MannXo: fixed. hint is now injected as `AssistantMsg` containing a `HintBlock`, matching the `ToolOffloadMiddleware` pattern. Injection is also gated to once per reply via `_hint_injected`.

src/agentscope/middleware/_budget.py
MannXo: fixed

src/agentscope/middleware/_budget.py
MannXo: Added `on_reply` that captures `reply_id` from `ReplyStartEvent` and cleans up both `_used_tokens` and `_hint_injected` in a finally block.

tests/middleware_budget_test.py
MannXo: Rewrote the test as a genuine multi-step scenario. step 1 is a tool call (200+100=300 tokens, budget=300), step 2 verifies that the accumulated count triggered enforcement by asserting a `HintBlock` was injected before the final reasoning call.

DavdGao (approved):
@MannXo Thanks for the contribution! I made a small change to persist the middleware context across multiple reply calls.

PR #1748, opened 2026-06-01
feat: add MemoryTool for explicit agent memory management

Libres-coder:
## Summary

Add `MemoryTool` — a builtin tool that allows agents to explicitly store, retrieve, update, and delete named memories that persist within a session and survive context compression.

Inspired by Claude Code's memory system.

## Motivation

AgentScope 2.0 already has automatic context compression (`compress_context`), but agents lack fine-grained control over what specific information to remember. `MemoryTool` complements automatic compression by giving agents explicit memory management capabilities.

### New files
- `src/agentscope/tool/_builtin/_memory.py` — MemoryTool implementation
- `tests/memory_tool_test.py` — 15 test cases

### Modified files
- `src/agentscope/state/_state.py` — Added `MemoryEntry` and `MemoryContext` classes with LRU eviction (max 50 entries); `AgentState` now includes `memory_context`
- `src/agentscope/agent/_agent.py` — `_prepare_model_input()` auto-injects stored memories as a `<agent-memories>` UserMsg block
- `src/agentscope/tool/_builtin/__init__.py` — Export `MemoryTool`
- `src/agentscope/tool/__init__.py` — Export `MemoryTool`

## Usage

```python
from agentscope.tool import MemoryTool, Toolkit

toolkit = Toolkit(tools=[MemoryTool()])
```

The agent can then use three actions:
- `save` — Store or update a memory by key
- `retrieve` — Get a specific memory or list all
- `delete` — Remove a memory by key

Stored memories are automatically injected into every model call, so the agent always sees them without needing to manually retrieve.

## Design decisions

- **UserMsg injection** (not system prompt): Keeps system prompt static, allows memories to be safely compressed, consistent with `state.summary` pattern
- **LRU eviction**: Default max 50 entries prevents unbounded memory growth
- **Opt-in**: MemoryTool is added explicitly via `Toolkit(tools=[...])`, not forced on all agents
- **Full values**: No truncation in auto-injected view; LRU eviction controls total size

## Tests

28/28 passing (15 memory tool + 5 compression + 8 tool result compression)

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -1982,6 +1982,24 @@ class Agent:
         messages = [
             SystemMsg(name="system", content=await self._get_system_prompt()),
         ]
+        # The agent's explicit memories
+        if self.state.memory_context.entries:
+            mem_lines = [
+                f"- {e.key}: {e.value}"
+                for e in sorted(
+                    self.state.memory_context.entries.values(),
+                    key=lambda e: e.created_at,
+                    reverse=True,
+                )
+            ]
+            messages.append(
+                UserMsg(
+                    name="user",
+                    content="<agent-memories>\n"
+                    + "\n".join(mem_lines)
+                    + "\n</agent-memories>",
+                ),
+            )
         # The compressed summary
         if self.state.summary:
             messages.append(
--- a/src/agentscope/state/_state.py
+++ b/src/agentscope/state/_state.py
@@ -130,6 +130,66 @@ class ToolContext(BaseModel):
         ]
 
 
+class MemoryEntry(BaseModel):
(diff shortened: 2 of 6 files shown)


Libres-coder:
ptal,thx!

DavdGao:
Sorry for the delayed clarification. Regarding long-term memory, we are not currently considering implementing it as part of the AgentState class (i.e., adding a memory_context field to AgentState). Since agent state is inherently scoped to a single session, a memory_context field would still essentially be extracting contextual information within that session — which is no different in nature from the summary produced during context compression.

For long-term memory, we are planning to implement it via an Agent Middleware layer, backed by the file system or a third-party database, enabling cross-session memory migration and persistence.

PR #1320, opened 2026-03-12
feat: add MiniMax chat model support

octo-patch:
## Summary

Adds `MiniMaxChatModel` as a dedicated model class for MiniMax's LLM API, enabling first-class support for MiniMax models in AgentScope.

### Changes

- **`src/agentscope/model/_minimax_model.py`**: New model class extending `OpenAIChatModel` with MiniMax-specific defaults (base URL, API key env var, model names)
- **`src/agentscope/model/__init__.py`**: Added `MiniMaxChatModel` to module exports

### Usage

```python
from agentscope.model import MiniMaxChatModel

model = MiniMaxChatModel(
    model_name="MiniMax-M2.5",
    api_key="your-api-key",  # or set MINIMAX_API_KEY env var
)
```

### Key Details

- Extends `OpenAIChatModel` since MiniMax API is OpenAI-compatible
- Default base URL: `https://api.minimax.io/v1`
- Default model: `MiniMax-M2.5` (204K context window)
- Also supports `MiniMax-M2.5-highspeed` for speed-optimized inference
- Reads API key from `MINIMAX_API_KEY` environment variable
- Temperature note: MiniMax does not accept exactly 0; use small positive values

## Test Plan

- [ ] Import `MiniMaxChatModel` from `agentscope.model`
- [ ] Verify default base URL is set to MiniMax endpoint
- [ ] Test chat completion with MiniMax-M2.5
- [ ] Test streaming mode
- [ ] Verify `MINIMAX_API_KEY` env var is read when api_key not provided

PR #1764, opened 2026-06-03
feat(models): add MiniMax chat model support (M3 default)

octo-patch:
## Summary
Add support for **MiniMax** chat models via the OpenAI-compatible API, with **MiniMax-M3** as the default model. This PR introduces the full MiniMax provider integration including credential, chat model, formatter, and per-model configuration.

## Models

| Model | Context | Max Output | Image Input |
|-------|---------|------------|-------------|
| `MiniMax-M3` (default) | 512K | 128K | Yes |
| `MiniMax-M2.7` | 200K | 32K | No |
| `MiniMax-M2.7-highspeed` | 200K | 32K | No |

## Changes

- Add `MiniMaxCredential` for API key + base URL management (`https://api.minimax.io/v1`)
- Add `MiniMaxChatModel` — OpenAI-compatible chat client with:
  - Streaming and non-streaming modes
  - Interleaved thinking (`<think>...</think>` → `ThinkingBlock`)
  - Tool calling with `tools` and `tool_choice`
  - Structured output via Pydantic
  - Automatic retries on transient errors
- Add `MiniMaxChatFormatter` and `MiniMaxMultiAgentFormatter` (advertise `text/plain` and `image/*` inputs)
- Add per-model YAML configuration (model card discovery via `list_models()`)
- Register provider in `credential/__init__.py`, `credential/_factory.py`, `formatter/__init__.py`, `model/__init__.py`
- 9 unit tests covering init, regular call, thinking content, tool calls, streaming, think-tag parsing, structured output, and model card listing

## Why
MiniMax-M3 is the latest generation model with significant improvements over the M2.5 series (which is what the earlier closed PR #1320 / open PR #1302 was based on): 512K context window, 128K max output, image input support, and stronger reasoning.

## Test plan
- [x] All 9 new unit tests pass
- [x] Model card listing returns M3, M2.7, M2.7-highspeed with correct context/output sizes
- [x] M3 is marked active and advertises `image/*` input

### DIFF
--- a/src/agentscope/formatter/_minimax_formatter.py
+++ b/src/agentscope/formatter/_minimax_formatter.py
@@ -1,0 +1,308 @@
+# -*- coding: utf-8 -*-
+"""The MiniMax formatter for agentscope."""
+from typing import Any
+
+from pydantic import Field
+
+from ._openai_formatter import _OpenAIFormatterBase
+from .._logging import logger
+from ..message import (
+    Msg,
+    TextBlock,
+    DataBlock,
+    HintBlock,
+    ToolCallBlock,
+    ToolResultBlock,
+    ThinkingBlock,
+)
+
+
+class MiniMaxChatFormatter(_OpenAIFormatterBase):
+    """The MiniMax formatter for chatbot scenario.
+
+    MiniMax exposes an OpenAI-compatible chat API. The M3 model supports
+    image input (image/* media types) but does not support audio input.
+    This formatter is structurally identical to :class:`OpenAIChatFormatter`
+    but advertises only image and text input types.
+    """
+
+    input_types: list[str] = Field(
+        default_factory=lambda: ["text/plain", "image/*"],
+        description=(
+            "The supported input types. "
+            'Defaults to ``["text/plain", "image/*"]``. '
+            "MiniMax does not support audio input."
+        ),
+    )
+
+    async def format(
+        self,
+        msgs: list[Msg],
+    ) -> list[dict[str, Any]]:
+        """Format message objects into MiniMax API required format.
+
+        Args:
+            msgs (`list[Msg]`):
+                The list of message objects to format.
+
+        Returns:
+            `list[dict[str, Any]]`:
+                A list of dictionaries, where each dictionary has
+                ``role`` and ``content`` keys.
+        """
+        self.assert_list_of_msgs(msgs)
+
+        messages: list[dict] = []
+        i = 0
+        while i < len(msgs):
+            msg = msgs[i]
+            content_blocks: list[dict] = []
+            tool_calls: list[dict] = []
+
+            for block in msg.get_content_blocks():
+                if isinstance(block, TextBlock):
+                    content_blocks.append(
+                        {"type": "text", "text": block.text},
+                    )
+
+                elif isinstance(block, DataBlock):
+                    formatted = self._format_openai_data_block(
+                        block,
+                        role=msg.role,
+                    )
+                    if formatted is not None:
+                        content_blocks.append(formatted)
+
+                elif isinstance(block, HintBlock):
+                    if content_blocks or tool_calls:
+                        msg_minimax = {
+                            "role": msg.role,
+                            "name": msg.name,
+                            "content": content_blocks or None,
+                        }
+                        if tool_calls:
+                            msg_minimax["tool_calls"] = tool_calls
+                        messages.append(msg_minimax)
+                        content_blocks = []
+                        tool_calls = []
+
+                    messages.append(
+                        {
+                            "role": "user",
+                            "content": [
+                                {"type": "text", "text": block.hint},
+                            ],
+                        },
+                    )
+
+                elif isinstance(block, ToolCallBlock):
+                    tool_calls.append(
+                        {
+                            "id": block.id,
+                            "type": "function",
+                            "function": {
+                                "name": block.name,
+                                "arguments": block.input,
+                            },
+                        },
+                    )
+
+                elif isinstance(block, ToolResultBlock):
+                    if content_blocks or tool_calls:
+                        msg_flush = {
+                            "role": msg.role,
+                            "name": msg.name,
+                            "content": content_blocks or None,
+                        }
+                        if tool_calls:
+                            msg_flush["tool_calls"] = tool_calls
+                        messages.append(msg_flush)
+                        content_blocks = []
+                        tool_calls = []
+
+                    (
+                        textual_output,
+                        multimodal_data,
+                    ) = self.convert_tool_result_to_string(block.output)
+
+                    messages.append(
+                        {
+                            "role": "tool",
+                            "tool_call_id": block.id,
+                            "content": textual_output,
+                            "name": block.name,
+                        },
+                    )
+
+                    if multimodal_data:
+                        promo_content = []
+                        for item in multimodal_data:
+                            if isinstance(item, TextBlock):
+                                promo_content.append(
+                                    {"type": "text", "text": item.text},
+                                )
+                            elif isinstance(item, DataBlock):
+                                fmt_item = self._format_openai_data_block(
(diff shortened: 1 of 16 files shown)


DavdGao:
@qbc2016 Plz review this pull request

src/agentscope/model/_minimax/_model.py
```diff
@@ -0,0 +1,476 @@
...
+from .._base import ChatModelBase, _TOOL_CHOICE_LITERAL_MODES
+from .._model_response import ChatResponse, StructuredResponse
+from .._model_usage import ChatUsage
+from ...credential import MiniMaxCredential
+from ...formatter import FormatterBase, MiniMaxChatFormatter
+from ...message import Msg, ThinkingBlock, ToolCallBlock, TextBlock
+from ...tool import ToolChoice
+
+if TYPE_CHECKING:
+    from openai.types.chat import ChatCompletion
```
qbc2016: Minimax officially recommends using the Anthropic‑compatible API for calls, so it’s suggested to use Anthropic here.

src/agentscope/model/_minimax/_model.py
```diff
@@ -0,0 +1,476 @@
...
+
+class MiniMaxChatModel(ChatModelBase):
+    """The MiniMax chat model.
+
+    MiniMax provides an OpenAI-compatible chat API. The model supports
+    image input (multimodal) and tool calls. Temperature must be in
+    ``(0.0, 1.0]`` and the default is ``1.0``.
+    """
+
+    class Parameters(BaseModel):
```
qbc2016, line 48: Please add the parameter `thinking_enable: bool` as other chat models do

src/agentscope/formatter/_minimax_formatter.py
```diff
@@ -0,0 +1,308 @@
...
+                        if promo_content:
+                            messages.append(
+                                {
+                                    "role": "user",
+                                    "name": "system-reminder",
+                                    "content": promo_content,
+                                },
+                            )
+
+                elif isinstance(block, ThinkingBlock):
```
qbc2016: Shall we pass the thinking block back?

src/agentscope/formatter/_minimax_formatter.py
```diff
@@ -0,0 +1,308 @@
...
+    ToolCallBlock,
+    ToolResultBlock,
+    ThinkingBlock,
+)
+
+
+class MiniMaxChatFormatter(_OpenAIFormatterBase):
+    """The MiniMax formatter for chatbot scenario.
+
+    MiniMax exposes an OpenAI-compatible chat API. The M3 model supports
```
qbc2016: Please modify the formatter after changing to Anthropic-cimpatible api.

qbc2016 (commented):
And please also add corresponding scripts in [scrips/model_examples](https://github.com/agentscope-ai/agentscope/tree/main/scripts/model_examples)

src/agentscope/model/_minimax/_model.py
octo-patch: Done. Rewrote the model on the Anthropic Python SDK and switched the credential base URL to `https://api.minimax.io/anthropic`. `MiniMaxChatModel` now mirrors `AnthropicChatModel` end-to-end (streaming events, system-message extraction, tool schema mapping, retry exceptions).

src/agentscope/model/_minimax/_model.py
octo-patch: Done. `Parameters` now exposes `thinking_enable: bool = False` and `thinking_budget: int | None` (validated to be positive). When `thinking_enable` is true the model sends `thinking={'type':'enabled','budget_tokens':...}` to the Anthropic-compatible endpoint and `max_tokens` is auto-expanded if it's smaller than the budget, matching how other chat models in this repo wire up extended thinking.

src/agentscope/formatter/_minimax_formatter.py
octo-patch: Yes — now that the formatter subclasses `AnthropicChatFormatter` / `AnthropicMultiAgentFormatter` it inherits the [documented thinking-block round-trip behaviour](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#preserving-thinking-blocks) verbatim, so thinking blocks are preserved across turns for reasoning continuity.

src/agentscope/formatter/_minimax_formatter.py
octo-patch: Done. The formatter dropped from ~300 lines to a thin wrapper: `MiniMaxChatFormatter` extends `AnthropicChatFormatter` and `MiniMaxMultiAgentFormatter` extends `AnthropicMultiAgentFormatter`, only overriding `input_types` (text + image, since the MiniMax M-series chat endpoint doesn't accept audio) and the conversation-history prompt. All Anthropic-compatible request shaping is reused directly.

octo-patch:
Thanks for the review! I've pushed the requested changes:

- **Anthropic-compatible API**: `MiniMaxCredential` default base URL is now `https://api.minimax.io/anthropic` and `MiniMaxChatModel` is rebuilt on the Anthropic Python SDK, mirroring `AnthropicChatModel` (streaming, tool schema, retries).
- **`thinking_enable` parameter**: `Parameters` exposes `thinking_enable: bool = False` and `thinking_budget: int | None`, gated the same way other chat models in this repo wire up extended thinking.
- **Formatter slimmed down**: `MiniMaxChatFormatter` / `MiniMaxMultiAgentFormatter` now just subclass the Anthropic formatters, so thinking-block round-trip preservation is inherited verbatim.
- **Example scripts added** under `scripts/model_examples/`, mirroring the `anthropic_*.py` layout:
  - `minimax_call.py` — simple call, tool call, structured output
  - `minimax_multimodal.py` — image URL / local `file://` / explicit base64
  - `minimax_multiagent.py` — `MiniMaxMultiAgentFormatter` end-to-end
  - `minimax_multiagent_multimodal.py` — multi-agent + image input

All 19 unit tests pass and ruff lint/format is clean. PTAL when you have a moment.

src/agentscope/model/_minimax/_model.py
```diff
@@ -0,0 +1,551 @@
...
+            gt=0,
+        )
+
+        thinking_enable: bool = Field(
+            default=False,
+            title="Thinking",
+            description="The thinking enable for the LLM output.",
+        )
+
+        thinking_budget: int | None = Field(
```
qbc2016, line 67: Does minimax support `thinking_budget`? Since I did not see it in its official document.

octo-patch:
Pushed a formatting-only commit to fix the pre-commit failure — ran black at line-length 79 and wrapped a few long string literals that were tripping flake8 E501. No logic changes.

src/agentscope/model/_minimax/_model.py
```diff
@@ -0,0 +1,557 @@
...
+
+if TYPE_CHECKING:
+    from anthropic.types.message import Message
+    from anthropic import AsyncStream
+else:
+    Message = Any
+    AsyncStream = Any
+
+
+class MiniMaxChatModel(ChatModelBase):
```
qbc2016, line 35: Should `MiniMaxChatModel` inherit from `AnthropicChatModel`? That might help maintain better consistency.

qbc2016 (commented):
Please also modify `/scripts/model_examples/run_tests.py` and `/scripts/model_examples/README.md`, correspondingly.

octo-patch:
Good catches, and sorry for the slow turnaround here. You're right that MiniMax's official API doesn't document a `thinking_budget` parameter — I'll drop it and keep just `thinking_enable`. Having `MiniMaxChatModel` inherit from `AnthropicChatModel` makes sense now that we're on the Anthropic-compatible endpoint, so I'll refactor in that direction and pass the thinking block back through the formatter. I'll push the reworked version and re-request your review.

PR #1772, opened 2026-06-03
fix(webui): make the chat session sidebar a mobile overlay drawer

### LINKED ISSUE #1677: perf(webui): improve the usability and user experience of the current web ui
**Background**

AgentScope currently provides a Web UI in [examples/web_ui](https://github.com/agentscope-ai/agentscope/tree/main/examples/web_ui) for the agent service, enabling users to interact with and manage their agents visually. The current Web UI includes the following pages:

- **Chat**: Agent conversation and workspace management
- **Schedule**: Scheduled task management
- **Credential**: Authentication and credential management

However, the current implementation has room for further improvement in terms of usability, performance, and overall user experience.

We welcome any developers interested in frontend development to share their ideas or contribute directly to this effort.

fancyboi999:
## AgentScope Version

2.0.0

## Description

**Problem.** On mobile (`< 768px`) the chat page's session sidebar was a fixed 320px column (`<Sidebar collapsible="none" className="w-80">`, always open) sitting next to `<main className="size-full">` (which forces `width: 100%`). Combined with the persistent icon rail, the sidebar pushed the chat content past the viewport: the top toolbar and message text were clipped off the right edge and the page overflowed horizontally.

**Solution.**

- **Fix the overflow at the root**: `main` changes from `size-full` to `h-full min-w-0 flex-1`. `min-w-0` lets the content column shrink below its content's intrinsic width instead of forcing 100% and overflowing.
- **Mobile = overlay drawer**: on `< 768px` the session sidebar renders as a `fixed` overlay (off the layout flow) with a backdrop, so it no longer steals layout width. It starts collapsed (initialised synchronously from `window.innerWidth` to avoid a first-paint flash) and reopens on desktop. Its width is clamped with `max-w-[calc(100vw - var(--sidebar-width-icon) - 1px)]`, so it never overflows even at 320px.
- **Dismissal**: backdrop click, the `Escape` key, or selecting a session all close the mobile drawer.
- **Desktop unchanged**: every mobile branch is gated on `useIsMobile`, so on desktop the sidebar renders exactly as before (`collapsible="none"`, in-flow, `w-80`).

`MOBILE_BREAKPOINT` is now exported from `hooks/use-mobile.ts` and reused by the synchronous initialiser, so the breakpoint isn't a duplicated magic number.

**Scope.** Two files: `pages/chat/index.tsx` and `hooks/use-mobile.ts` (exporting one constant).

**Known limitation.** The mobile overlay is a non-modal panel: it does not implement a full focus trap / `aria-modal`. Reusing a full `Sheet`/`Dialog` would require extracting the ~155-line sidebar body, which is out of scope for this fix. Keyboard dismissal still works via `Escape` and the focusable backdrop button.

## How to test

- `tsc -b`, `eslint`, and `prettier --check` pass; `pre-commit` trailing-whitespace passes. The only eslint warnings are 4 pre-existing `exhaustive-deps` ones unrelated to this change.
- Desktop (`>= 768px`): the sidebar renders in-flow at `w-80`, no horizontal overflow (verified `document.documentElement.scrollWidth === window.innerWidth`).
- Mobile: the sidebar collapses by default and the content fills the viewport without clipping; opening it shows an overlay drawer + backdrop that does not push content. The `max-w` clamp keeps the drawer within the viewport down to 320px (`left = 49px`, `max-w = 100vw - 49px`, right edge `= 100vw`). Browser screenshots were captured at 390/500px during development.

Refs #1677

## Checklist

- [x] Code has been formatted with project linters (`tsc -b`, `eslint`, `prettier`, `pre-commit` all pass)
- [x] All tests are passing (no test suite for this example frontend; verified via type-check, lint, and browser inspection)
- [x] Docstrings are in Google style (N/A — inline comments only, consistent with the file)
- [x] Related documentation has been updated (N/A — bugfix, no doc/API change)

### DIFF
--- a/examples/web_ui/frontend/src/pages/chat/ChatViewport.tsx
+++ b/examples/web_ui/frontend/src/pages/chat/ChatViewport.tsx
@@ -12,6 +12,7 @@ import { ModelParametersPopover } from '@/components/popover/ModelParametersPopo
 import { LlmSelect } from '@/components/select/LlmSelect';
 import { PermissionModeSelect } from '@/components/select/PermissionModeSelect.tsx';
 import { Button } from '@/components/ui/button';
+import { SidebarTrigger } from '@/components/ui/sidebar';
 import { useAvailableModels } from '@/hooks/useAvailableModels';
 import { useMessages } from '@/hooks/useMessages';
 import { useSessions } from '@/hooks/useSessions';
@@ -299,6 +300,7 @@ export function ChatViewport({ agentId, sessionId, onTeamUpdated }: ChatViewport
 				<div className="flex flex-col flex-1 min-h-0 p-2">
 					<div className="flex flex-row gap-x-2 justify-between">
 						<div id="tour-llm-select" className="flex flex-row items-center gap-x-1">
+							<SidebarTrigger className="md:hidden" />
 							<LlmSelect
 								value={selectedModel}
 								onChange={handleLlmChange}
--- a/examples/web_ui/frontend/src/pages/chat/index.tsx
+++ b/examples/web_ui/frontend/src/pages/chat/index.tsx
@@ -54,6 +54,8 @@ import {
 	SidebarMenuAction,
 	SidebarMenuButton,
 	SidebarMenuItem,
+	SidebarProvider,
+	useSidebar,
 } from '@/components/ui/sidebar';
 import { AudioProvider } from '@/context/AudioContext';
 import { useAgents } from '@/hooks/useAgents';
@@ -102,7 +104,7 @@ const ChatPageInner = () => {
 		remove: removeSession,
 	} = useSessions(urlAgentId ?? null);
 
-	const [sidebarOpen, setSidebarOpen] = useState(true);
+	const { isMobile, setOpen, setOpenMobile } = useSidebar();
 	const [editOpen, setEditOpen] = useState(false);
 	const [deleteOpen, setDeleteOpen] = useState(false);
 	const [renameOpen, setRenameOpen] = useState(false);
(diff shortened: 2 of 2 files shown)


DavdGao (changes requested):
On the mobile drawer: I'd suggest dropping `collapsible="none"` and using shadcn's built-in mobile handling. The Sidebar already branches on `useIsMobile` internally and renders a Sheet (Radix Dialog) on mobile — which gives you focus trap + aria-modal for free (resolves the "Known limitation"), plus Escape, backdrop, and Cmd/Ctrl+B.

Concretely:

- Wrap the page in `<SidebarProvider defaultOpen>`.
- Drop `collapsible="none"` (default `offcanvas`); let the Provider own open/close.
- Replace the hand-rolled hamburger with `<SidebarTrigger className="md:hidden" />`.
- For "tap a session closes on mobile," call `setOpenMobile(false)` from `useSidebar()`.

That removes the backdrop `<button>`, the Escape `useEffect`, the breakpoint-sync `useEffect`, the window.innerWidth initialiser, the `fixed inset-y-0 …` block, the `MOBILE_BREAKPOINT` export, the `onOpenSidebar` prop, and the new i18n key — ~50 lines out for ~5 in, and the a11y limitation goes away.

One behavior change: `offcanvas` lets desktop users collapse too. If you want "desktop = always open," `md:hidden` on the trigger keeps it mobile-only.

fancyboi999:
Switched to the framework's built-in mobile handling as suggested — the hand-rolled backdrop, Escape handler, breakpoint-sync effect, `window.innerWidth` init, `MOBILE_BREAKPOINT` export, `onOpenSidebar` prop, and the i18n key are all gone (~50 lines removed). The a11y "Known limitation" is resolved too, since the Sheet brings focus trap + aria-modal for free.

One deviation from the exact suggestion, with the reason. A plain `collapsible="offcanvas"` doesn't hold up here because of the app rail (`AppSidebar`): offcanvas renders a `fixed left-0` (z-10) container that paints over the rail.

**`collapsible="offcanvas"` — the app rail disappears (covered by the fixed container):**

offcanvas covers the app rail

Measured in the page: app rail is `left:0 width:49 position:static`, the offcanvas container is `left:0 width:256 position:fixed z-index:10` — it sits right on top of the 49px rail.

So instead of dropping `collapsible` entirely, the session `Sidebar` switches it by breakpoint: `none` on desktop (stays in normal flow beside the rail), `offcanvas` on mobile (which hits the framework's `isMobile` → Sheet branch). That keeps your Sheet-for-mobile point while preserving the rail.

**Desktop — rail preserved, sidebar inline (unchanged from main):**

desktop, rail preserved

**Mobile — default closed, `md:hidden` hamburger in the toolbar:**

mobile, drawer closed

**Mobile — drawer open (Sheet + backdrop):**

mobile, drawer open

`ChatPageInner` is wrapped in its own `SidebarProvider` (nested under AppLayout's), `setOpenMobile(false)` closes on session tap, and Escape / backdrop come from the Sheet. Verified in the browser at both breakpoints; `tsc` / `eslint` / `prettier` clean.

If you'd rather the rail also fold into the drawer on mobile (a single offcanvas for both), I'm happy to look — but that means restructuring AppLayout's provider, so I kept this PR scoped to the session sidebar.

fancyboi999:
Follow-up: verified the close interactions against **real sessions** (backend up, 4 sessions from Redis), not just the empty state.

- Tapping a *different* session both navigates and closes the drawer (URL switches + `setOpenMobile(false)`).
- Tapping the current session, Escape, and the backdrop each close it as well.

mobile drawer with real sessions, active row highlighted

DavdGao (approved):
LGTM

PR #1774, opened 2026-06-03
fix(model): forward Qwen thinking toggle

### LINKED ISSUE #1765: [Bug]: thinking_enable parameter not passed to extra_body.enable_thinking in OpenAI-compatible Qwen model
**Describe the bug**
When using Qwen models in thinking mode with AgentScope's OpenAI-compatible wrapper, the `thinking_enable` parameter is not properly forwarded to `extra_body.enable_thinking`. This prevents users from disabling the thinking mode, which is problematic because Qwen's thinking mode does not support structured output (JSON schema enforcement). As a result, when structured output is required, users cannot disable thinking mode through AgentScope's API.

**To Reproduce**

1. Configure an OpenAI-compatible model wrapper for Qwen with thinking enabled:
```python
from agentscope.model import OpenAIChatModel

model = OpenAIChatModel(
    ...,
    thinking_enable=True  # or False, but it doesn't get passed through
)
```

2. Attempt to use structured output (e.g., with response_format or JSON schema):
```python
response = model(messages,structured_model=...)
```

3. Observe that the thinking mode remains active (or cannot be disabled), causing structured output to fail or produce invalid JSON.

**Expected behavior**
The `thinking_enable` parameter should be properly passed through to the API request's `extra_body.enable_thinking` field, allowing users to:
- Disable thinking mode when structured output is required
- Have explicit control over the thinking mode toggle in OpenAI-compatible mode

Example error:
```
openai.BadRequestError: Error code: 400 - {'error': {'message': '<400> InternalError.Algo.InvalidParameter: The tool_choice parameter does not support being set to required or object in thinking mode', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_parameter_error'}, 'id': 'chatcmpl-be58834c-e4a4-93c3-9c76-f0c53e458bf6', 'request_id': 'be58834c-e4a4-93c3-9c76-f0c53e458bf6'}
```

- AgentScope Version: [2.0.0]
- Python Version: [3.12]
- Model API: Qwen via DashScope OpenAI-compatible endpoint

**Additional context**
This issue affects workflows that require both:
1. Using Qwen models through the OpenAI-compatible API
2. Structured output capabilities (function calling, JSON schema enforcement)

Since Qwen's thinking mode is incompatible with structured output, having the ability to toggle `enable_thinking` via `extra_body` is essential. The fix would involve ensuring the `thinking_enable` parameter in the model configuration is correctly mapped to `extra_body={"enable_thinking": thinking_enable}` in the API request payload.

he-yufeng:
Fixes #1765.

OpenAI-compatible Qwen endpoints use the OpenAI Chat Completions surface, but the thinking toggle has to be sent through `extra_body.enable_thinking`. AgentScope stores `thinking_enable` on `OpenAIChatModel.Parameters`, but the OpenAI-compatible request path never forwarded it, so users could not explicitly disable Qwen thinking mode before using structured output.

This patch forwards `thinking_enable` into `extra_body.enable_thinking` only when the parameter was explicitly set. That keeps the default OpenAIChatModel request clean for normal OpenAI models, while allowing `OpenAIChatModel.Parameters(thinking_enable=False)` to send the provider-specific toggle needed by Qwen-compatible endpoints.

Validation:
- .venv\Scripts\python.exe -m pytest tests\model_openai_chat_test.py -q -k "thinking_enable or text_response"
- .venv\Scripts\python.exe -m pytest tests\model_openai_chat_test.py -q
- .venv\Scripts\python.exe -m py_compile src\agentscope\model\_openai_chat\_model.py tests\model_openai_chat_test.py
- .venv\Scripts\python.exe -m ruff check src\agentscope\model\_openai_chat\_model.py tests\model_openai_chat_test.py
- git diff --check

Note: `ruff format --check ...` would reformat pre-existing lines in these files, so I did not run whole-file formatting for this scoped fix.

### DIFF
--- a/src/agentscope/model/_openai_chat/_model.py
+++ b/src/agentscope/model/_openai_chat/_model.py
@@ -122,6 +122,7 @@ class OpenAIChatModel(ChatModelBase):
         context_size: int = 128000,
         formatter: FormatterBase | None = None,
         client_kwargs: dict[str, Any] | None = None,
+        extra_body: dict[str, Any] | None = None,
     ) -> None:
         """Initialize the OpenAI chat model.
 
@@ -149,6 +150,9 @@ class OpenAIChatModel(ChatModelBase):
             client_kwargs (`dict[str, Any] | None`, defaults to `None`):
                 Extra keyword arguments forwarded to ``openai.AsyncClient``
                 (e.g. ``timeout``, ``default_headers``, ``http_client``).
+            extra_body (`dict[str, Any] | None`, defaults to `None`):
+                Additional request body fields forwarded to
+                OpenAI-compatible APIs.
         """
         super().__init__(
             credential=credential,
@@ -161,6 +165,7 @@ class OpenAIChatModel(ChatModelBase):
         )
         self.formatter = formatter or OpenAIChatFormatter()
         self.client_kwargs = client_kwargs or {}
+        self.extra_body = dict(extra_body) if extra_body is not None else None
 
     @classmethod
     def _get_retryable_exceptions(cls) -> tuple[Type[Exception], ...]:
@@ -247,16 +252,10 @@ class OpenAIChatModel(ChatModelBase):
             }
             kwargs["modalities"] = ["text", "audio"]
 
-        kwargs.update(generate_kwargs)
+        if self.extra_body is not None:
+            kwargs["extra_body"] = dict(self.extra_body)
 
-        if (
-            "thinking_enable" in self.parameters.model_fields_set
-            and self._uses_qwen_extra_body(model_name)
-        ):
-            kwargs.setdefault("extra_body", {})
-            kwargs["extra_body"][
-                "enable_thinking"
-            ] = self.parameters.thinking_enable
+        kwargs.update(generate_kwargs)
 
         fmt_tools, fmt_tool_choice = self._format_tools(tools, tool_choice)
 
@@ -683,12 +682,3 @@ class OpenAIChatModel(ChatModelBase):
             return tools, {"type": "function", "function": {"name": mode}}
 
         return tools, mode
-
-    def _uses_qwen_extra_body(self, model_name: str) -> bool:
-        base_url = str(self.credential.base_url or "").lower()
-        model = model_name.lower()
-        return (
-            "qwen" in model
-            or "dashscope" in base_url
-            or "aliyuncs.com" in base_url
-        )
--- a/src/agentscope/model/_openai_chat/_model.py
+++ b/src/agentscope/model/_openai_chat/_model.py
@@ -249,6 +249,12 @@ class OpenAIChatModel(ChatModelBase):
 
         kwargs.update(generate_kwargs)
 
+        if "thinking_enable" in self.parameters.model_fields_set:
+            kwargs.setdefault("extra_body", {})
+            kwargs["extra_body"][
+                "enable_thinking"
+            ] = self.parameters.thinking_enable
+
         fmt_tools, fmt_tool_choice = self._format_tools(tools, tool_choice)
 
         if fmt_tools:
--- a/src/agentscope/model/_openai_chat/_model.py
+++ b/src/agentscope/model/_openai_chat/_model.py
@@ -249,7 +249,10 @@ class OpenAIChatModel(ChatModelBase):
 
         kwargs.update(generate_kwargs)
 
-        if "thinking_enable" in self.parameters.model_fields_set:
+        if (
+            "thinking_enable" in self.parameters.model_fields_set
+            and self._uses_qwen_extra_body(model_name)
+        ):
             kwargs.setdefault("extra_body", {})
             kwargs["extra_body"][
                 "enable_thinking"
@@ -680,3 +683,12 @@ class OpenAIChatModel(ChatModelBase):
             return tools, {"type": "function", "function": {"name": mode}}
 
         return tools, mode
+
+    def _uses_qwen_extra_body(self, model_name: str) -> bool:
+        base_url = str(self.credential.base_url or "").lower()
+        model = model_name.lower()
+        return (
+            "qwen" in model
+            or "dashscope" in base_url
+            or "aliyuncs.com" in base_url
+        )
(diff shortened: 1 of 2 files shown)


qbc2016:
Thank you for the contribution! 

We did some testing and found that when `enable_thinking` is enabled and passed through `extra_body`, it causes an error with the official OpenAI models:
`openai.BadRequestError: Error code: 400 - {'error': {'message': "Unknown parameter: 'enable_thinking'.", 'type': 'invalid_request_error', 'param': 'enable_thinking', 'code': 'unknown_parameter'}}`

So this approach might not be the most ideal solution, as it could potentially break compatibility with the official OpenAI API.

he-yufeng:
Thanks for testing this. I agree that forwarding enable_thinking to the official OpenAI API is not acceptable.

I updated the branch so the OpenAI chat model only adds the Qwen-compatible extra_body flag when the model/base_url looks like Qwen/DashScope. Official OpenAI models such as gpt-4.1 now keep thinking_enable local and do not receive enable_thinking in extra_body.

Validation run locally against this checkout:

- PYTHONPATH=src python -m pytest tests/model_openai_chat_test.py -q: 20 passed
- python -m py_compile src/agentscope/model/_openai_chat/_model.py tests/model_openai_chat_test.py: passed
- python -m ruff check src/agentscope/model/_openai_chat/_model.py tests/model_openai_chat_test.py: passed
- git diff --check: passed

qbc2016:
Sorry for the late reply. In my opinion, I think that if we aim to support compatibility, it would be better to make the implementation more robust. I’ve also noticed feedback from other users—for example, using `OpenAIChatModel` to call DeepSeek or models deployed with vLLM. To support these scenarios, relying solely on URL-based detection may not be sufficient.

Our current thought is to add an `extra_body` parameter to `Parameters`, allowing users with different needs to pass custom arguments as a dictionary, which would then be forwarded directly to the model call. This might provide a more flexible and general solution.

What do you think about this approach? If it’s convenient for you, you could update it directly in this PR. Otherwise, we’d be happy to open a separate PR ourselves. Looking forward to your thoughts.

he-yufeng:
Updated the branch to the generic approach you suggested in b5f0da73. `OpenAIChatModel.Parameters` now accepts an `extra_body` dictionary and forwards it to any OpenAI-compatible endpoint without model-name or URL detection. Per-call `extra_body` still takes precedence over model defaults.

The Qwen-specific automatic forwarding and provider detection have been removed. Users can now pass `extra_body={"enable_thinking": false}` for Qwen, or arbitrary provider-specific fields for DeepSeek, vLLM, and other compatible endpoints.

Validation: `tests/model_openai_chat_test.py` (20 passed), ruff check, py_compile, and `git diff --check`.

he-yufeng:
Rebased the generic extra_body implementation onto current main and resolved the conflict with the newly added audio/voice request path. The final diff remains limited to generic extra_body forwarding and focused tests; 20 OpenAI chat model tests pass locally.

qbc2016 (commented):
LGTM.

he-yufeng:
Thanks again for the review and LGTM. The generic xtra_body implementation is still green across the CI matrix. Please let me know if you want any naming or documentation tweak before merge.

src/agentscope/model/_openai_chat/_model.py
```diff
@@ -87,6 +87,14 @@ class Parameters(BaseModel):
             le=1,
         )
 
+        extra_body: dict[str, Any] | None = Field(
```
DavdGao: The `Parameters` class is responsible for frontend rendering, so plz move the argument `extra_body` to the constructor instead, rather than within the `Parameters` class

DavdGao (changes requested):
@he-yufeng Plz see inline comments and move the argument `extra_body` to the constructor instead.

he-yufeng:
Updated the PR to move the generic `extra_body` default onto the `OpenAIChatModel` constructor instead of `OpenAIChatModel.Parameters`.

Current behavior:
- `OpenAIChatModel(..., extra_body={...})` sets default OpenAI-compatible request body fields for this model instance.
- per-call `model(..., extra_body={...})` still overrides the constructor default.
- default OpenAI usage still sends no provider-specific `extra_body`.

Validated locally:

```text
python -m py_compile src\agentscope\model\_openai_chat\_model.py tests\model_openai_chat_test.py
PYTHONPATH=src python -m pytest tests\model_openai_chat_test.py -q --basetemp .tmp\pytest-agentscope-1774
# 20 passed
python -m ruff check src\agentscope\model\_openai_chat\_model.py tests\model_openai_chat_test.py
git diff --check
```

DavdGao:
> Updated the PR to move the generic `extra_body` default onto the `OpenAIChatModel` constructor instead of `OpenAIChatModel.Parameters`.
> 
> Current behavior:
> 
> * `OpenAIChatModel(..., extra_body={...})` sets default OpenAI-compatible request body fields for this model instance.
> * per-call `model(..., extra_body={...})` still overrides the constructor default.
> * default OpenAI usage still sends no provider-specific `extra_body`.
> 
> Validated locally:
> 
> ```
> python -m py_compile src\agentscope\model\_openai_chat\_model.py tests\model_openai_chat_test.py
> PYTHONPATH=src python -m pytest tests\model_openai_chat_test.py -q --basetemp .tmp\pytest-agentscope-1774
> # 20 passed
> python -m ruff check src\agentscope\model\_openai_chat\_model.py tests\model_openai_chat_test.py
> git diff --check
> ```

@he-yufeng Plz run pre-commit commands to format the code, others lgtm

he-yufeng:
Fixed the pre-commit failure from the latest run. The issue was one test docstring at 80 columns under flake8's 79-column limit; the branch now shortens that line.

Revalidated locally:

```text
python -m py_compile src\agentscope\model\_openai_chat\_model.py tests\model_openai_chat_test.py
PYTHONPATH=src python -m pytest tests\model_openai_chat_test.py -q --basetemp .tmp\pytest-agentscope-1774
# 20 passed
python -m ruff check src\agentscope\model\_openai_chat\_model.py tests\model_openai_chat_test.py
python -m flake8 tests\model_openai_chat_test.py --select=E501 --max-line-length=79
git diff --check
```

DavdGao (approved):
lgtm

PR #1784, opened 2026-06-04
fix(model): honor explicit thinking_enable=False on Ollama and Gemini

### LINKED ISSUE #1781: Bug: OllamaChatModel cannot disable 'think' mode via kwargs update
## Description

In the `OllamaChatModel` class, the logic for controlling the `think` mode is currently as follows:

```python
    if self.parameters.thinking_enable:
        kwargs["think"] = True

    kwargs.update(generate_kwargs)
```

This implementation makes it impossible to disable the `think` mode, because even if you want to set it to `False` through `generate_kwargs`, the assignment above always overwrites the value to `True` if `self.parameters.thinking_enable` is set. As a result, there's no way to disable `think` mode through runtime parameters or user intent.

## Expected Behavior
- It should be possible to disable `think` mode by overriding with `generate_kwargs` or by correctly setting parameters.
- The logic should not overwrite `think=False` when provided in `generate_kwargs`.

## Steps to Reproduce
1. Attempt to invoke the `OllamaChatModel` with `think=False` in `generate_kwargs`.
2. Observe that `think` is always set to `True` if `self.parameters.thinking_enable` is enabled.

## Suggestion
Update the order of `kwargs` updates or add logic to respect the user input from `generate_kwargs`.

---

- [MUST] Respect explicit `think` value passed via `generate_kwargs`.
- [SHOULD] Add related test coverage for this scenario.

## Environment
- agentscope version: (please specify)
- Python version: (please specify)

qbc2016:
## AgentScope Version

2.0.0

## Description

- Switch `thinking_enable` from `bool` (default `False`) to `bool | None` (default `None` = model default) and explicitly forward `True`/`False` to the API. This aligns with the symmetric handling in DashScope, DeepSeek and Moonshot.
- Gemini's `False` path sends `thinking_budget=0` to actually stop thinking on Flash variants. Pro models (which cannot disable thinking) will now surface a clear API error instead of silently continuing to think.
- Adds the missing `thinking_enable=True` flag to three example scripts (`openai_chat_call`, `openai_response_call`, `xai_call`) where `reasoning_effort` was set without it and was being silently dropped by the wrapper's `if thinking_enable and reasoning_effort` gate.

## Checklist

- [ ]  All tests are passing


src/agentscope/model/_ollama/_model.py
```diff
@@ -35,11 +35,13 @@ class Parameters(BaseModel):
             gt=0,
         )
 
-        thinking_enable: bool = Field(
-            default=False,
+        thinking_enable: bool | None = Field(
```
DavdGao: ==> None

src/agentscope/model/_gemini/_model.py
```diff
@@ -89,10 +89,13 @@ class Parameters(BaseModel):
             gt=0,
         )
 
-        thinking_enable: bool = Field(
-            default=False,
+        thinking_enable: bool | None = Field(
```
DavdGao: ==> bool

DavdGao (approved):

PR #1814, opened 2026-06-07
fix(service): trust local workspace root

### LINKED ISSUE #1791: fix(service): set the workspace directory as working_directories in `PermissionContext`
**Background**

AgentScope 2.0 currently supports agents operating within a workspace, but the workspace root directory is not included in the `working_directories` of `PermissionContext`. As a result, some file operations within the workspace still require user confirmation.

**Changes**

When assembling the agent in the chat service, check whether the workspace root directory is already present in `PermissionContext.working_directories`. If yes, skip; if no, append it automatically.

Maybe the workspace class need to expose an interface to retrieve the base working directory.

he-yufeng:
## Summary

Fixes #1791.

This adds the resolved local workspace root to the session permission context before the chat service assembles the agent. That gives `ACCEPT_EDITS` the same trusted root that the workspace tools are already operating inside, so ordinary workspace file operations do not keep asking for confirmation.

The change keeps the boundary narrow:

- `WorkspaceBase.working_directory` defaults to `None`
- `LocalWorkspace` returns its absolute `workdir`
- the chat service skips existing entries instead of overwriting user/session-supplied ones

## To verify

- `PYTHONPATH=src PYTHONUTF8=1 PYTHONIOENCODING=utf-8 python -m pytest tests/service_toolkit_test.py -q`
- `python -m ruff check src/agentscope/app/_service/_chat.py src/agentscope/workspace/_base.py src/agentscope/workspace/_local_workspace.py tests/service_toolkit_test.py`
- `PYTHONPATH=src PYTHONUTF8=1 PYTHONIOENCODING=utf-8 python -m py_compile src/agentscope/app/_service/_chat.py src/agentscope/workspace/_base.py src/agentscope/workspace/_local_workspace.py tests/service_toolkit_test.py`
- `git diff --check`

`ruff format --check` on the touched files reports existing formatter churn on current `main`, so I did not run formatter over the full files for this small service fix.

### DIFF
--- a/src/agentscope/app/_service/_chat.py
+++ b/src/agentscope/app/_service/_chat.py
@@ -34,6 +34,36 @@ from ...event import (
     ExternalExecutionResultEvent,
 )
 from ...message import AssistantMsg, Msg, ToolCallState
+from ...permission import AdditionalWorkingDirectory
+from ...state import AgentState
+from ...workspace import WorkspaceBase
+
+
+def _with_workspace_working_directory(
+    agent_state: AgentState,
+    workspace: WorkspaceBase,
+) -> AgentState:
+    """Include the workspace root in the session permission context."""
+    workdir = workspace.working_directory
+    if workdir is None:
+        return agent_state
+
+    permission_context = agent_state.permission_context
+    if workdir in permission_context.working_directories:
+        return agent_state
+
+    working_directories = dict(permission_context.working_directories)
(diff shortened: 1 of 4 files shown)


DavdGao:
Closing in favor of #1823, which covers all workspace implementations (Docker, E2B, Local) and marks `working_directory` as abstract to enforce subclass compliance. Thanks for the contribution!

PR #1815, opened 2026-06-07
fix(team): inherit leader permission rules

### LINKED ISSUE #1793: feat(agent-team): copy the permission rules when creating agent team members
**Background**

Within the agent service (`agentscope.app`), an agent can create team members via the `AgentCreate` tool. However, the sub-agent's permission context is currently empty by default. It should inherit the permission rules from the team leader to avoid redundant user confirmations.

**Changes**

- When creating a sub-agent via `AgentCreate`, copy the permission rules from the team leader and apply it to the newly created sub-agent.
- A possible approach: set `is_state_injected=True` in the `AgentCreate` tool so that it has access to the leader's state, then copy the permission rules from it.

he-yufeng:
## Summary

Fixes #1793.

`AgentCreate` currently creates worker sessions with a fresh `PermissionContext`, so a worker loses the leader session's confirmed working directories and allow/deny/ask rules. The only thing carried over is the model config.

This PR makes `AgentCreate` state-injected and builds the worker state from the leader's live permission context:

- copy working directories and permission rules with `model_copy(deep=True)`
- keep the requested `permission_mode` as the worker's mode
- fall back to the stored leader session state for direct/tool-unit calls without injected state

That keeps user-confirmed permissions available to worker agents while still letting the leader choose `default`, `explore`, `accept_edits`, etc. per worker.

## To verify

- `PYTHONPATH=src PYTHONUTF8=1 PYTHONIOENCODING=utf-8 python -m pytest tests/service_team_tools_test.py -q`
- `python -m ruff check src/agentscope/app/_tools/_agent_create.py tests/service_team_tools_test.py`
- `python -m py_compile src/agentscope/app/_tools/_agent_create.py tests/service_team_tools_test.py`
- `git diff --check`

### DIFF
--- a/src/agentscope/app/_tools/_agent_create.py
+++ b/src/agentscope/app/_tools/_agent_create.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import copy
 import json
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 from pydantic import Field
 
@@ -12,12 +12,18 @@ from ._team_tool_base import _TeamToolBase
 from .._types import SubAgentTemplate
 from ..storage import AgentData, AgentRecord, SessionConfig
 from ...message import HintBlock, TextBlock, ToolResultState
+from ...permission import PermissionMode
 from ...state import AgentState
 from ...tool import ToolChunk, ParamsBase
 
 if TYPE_CHECKING:
     from ..message_bus import MessageBus
     from ..storage import StorageBase
+
+
+_PERMISSION_MODE_BY_VALUE: dict[str, PermissionMode] = {
+    mode.value: mode for mode in PermissionMode
+}
 
 
 _DEFAULT_SYSTEM_PROMPT_TEMPLATE = (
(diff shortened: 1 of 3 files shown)


DavdGao (commented):
@he-yufeng Plz solve the conflicts first

he-yufeng:
Resolved the rebase conflict against current main. The updated patch keeps the new sub-agent template flow while preserving leader permission-rule inheritance, and copies the template task context for the worker session. Local verification: python -m py_compile src/agentscope/app/_tools/_agent_create.py tests/service_team_tools_test.py; python -m ruff check src/agentscope/app/_tools/_agent_create.py tests/service_team_tools_test.py; PYTHONPATH=src PYTHONUTF8=1 PYTHONIOENCODING=utf-8 python -m pytest tests/service_team_tools_test.py -q (21 passed).

he-yufeng:
Friendly ping after the conflict-resolution update: this is now rebased on current main, all GitHub checks are green, and the focused local service-team tests passed. Please let me know if there is anything else you want changed before review.

src/agentscope/app/_tools/_agent_create.py
```diff
@@ -190,6 +231,8 @@ async def __call__(
         description: str,
         prompt: str,
         subagent_type: str = "default",
+        permission_mode: str = "default",
```
DavdGao: Similar as above

src/agentscope/app/_tools/_agent_create.py
```diff
@@ -74,12 +80,47 @@ class _AgentCreateParams(ParamsBase):
             "deadlines the member needs."
         ),
     )
+    permission_mode: Literal[
```
DavdGao: The permission mode arguments is already removed from the AgentCreate tool and set by subagent template. Plz remove this argument and focus on inherit the permission rules from the main agent

DavdGao (approved):
lgtm

PR #1816, opened 2026-06-07
fix(agent): hint after repeated tool errors

### LINKED ISSUE #1795: feat(agent): support consecutive tool failure detection and fallback hint in ReAct agent
**Background**

When an agent repeatedly calls the same tool and keeps getting errors, it may get stuck in an infinite retry loop. A fallback mechanism should be triggered in such cases to insert a `HintBlock` reminding the agent to stop retrying or try an alternative approach.

**Changes**

- Add two parameters to `ReActConfig`:
  - `max_tool_retries` (int): the maximum number of times the same tool is allowed to fail consecutively before triggering the fallback hint. Default could be `3`.
  - `tool_error_hint` (str): the hint message to inject as a `HintBlock` when the threshold is reached.
- Before each reasoning step inside the agent, check whether the same tool has failed consecutively for `max_tool_retries` times. If so, inject the `HintBlock` with `tool_error_hint` to guide the agent toward an alternative path.
- A helper function within `Agent` class that checks, starting from the last acting step, whether the same tool has been called consecutively more than max_tool_retries times with errors.

he-yufeng:
## Summary

Fixes #1795.

This adds a small ReAct-level guard for retry loops where the agent keeps calling the same tool and keeps getting error results.

Changes:

- add `ReActConfig.max_tool_retries` with default `3`
- add `ReActConfig.tool_error_hint`
- before each reasoning step, inspect the trailing same-tool `ToolResultState.ERROR` streak
- when the streak crosses the threshold, append a `HintBlock` telling the agent to stop retrying the same call as-is and try another approach

Denied tool calls are not counted as tool failures. Mixed-tool failures also do not trigger the hint.

## To verify

- `PYTHONPATH=src PYTHONUTF8=1 PYTHONIOENCODING=utf-8 python -m pytest tests/agent_basic_test.py -q`
- `python -m ruff check src/agentscope/agent/_config.py src/agentscope/agent/_agent.py tests/agent_basic_test.py`
- `python -m py_compile src/agentscope/agent/_config.py src/agentscope/agent/_agent.py tests/agent_basic_test.py`
- `git diff --check`


YingchaoX:
I was looking through this implementation and noticed one possible edge case.

`_get_repeated_tool_error()` currently counts trailing `ToolResultBlock`s with the same tool name. This may treat a single reasoning step that fans out multiple calls to the same tool as a retry loop. For example, if the model emits three `Read` calls with different `file_path`s in one turn and all three fail, the current logic would inject the retry hint, even though the agent has not retried the same call across reasoning steps.

Maybe this guard should count failed attempts across reasoning/acting rounds rather than raw trailing result blocks. It may also be safer to key the retry streak by something like `(tool_name, normalized_input)` instead of only `tool_name`.

A regression test for “multiple same-tool calls with different inputs in one model response all fail” could help clarify the intended behavior here.

One smaller ordering question: the hint is injected before `compress_context()`. If compression triggers, could the newly inserted hint be compressed or transformed before the next model call? Injecting after compression but immediately before reasoning might make the behavior more predictable.

he-yufeng:
Addressed the fan-out and ordering concerns in 24346ddf. Repeated failures are now counted by reasoning/action batch and normalized (tool name, input) signature, so multiple same-tool calls with different inputs in one model response do not count as retries. The hint is injected after context compression. Added a regression test for same-turn Read fan-out. Validation: agent_basic_test.py 8 passed, ruff check, py_compile, and git diff --check.

he-yufeng:
Pushed a narrow follow-up that fixes the pre-commit Black/flake8 failure. Repository-pinned Black and flake8 now pass, and the focused agent_basic_test suite passes (8 tests).

he-yufeng:
Third nudge. @YingchaoX's fan-out edge case was addressed back in June (retries are counted by reasoning/action batch with a normalized tool+input signature, plus a regression test for same-turn Read fan-out), so the review point should be settled. The change itself is unchanged since: after N repeated tool errors with the same signature, the agent gets a hint to stop and reassess instead of looping. CI green; anything else you'd like adjusted?

REISEN2330:
Nice implementation — I checked out this branch and verified it locally:
`pytest tests/agent_basic_test.py` → 8 passed (including the fan-out
regression test, which addresses the earlier edge-case review exactly).

One thing is blocking this PR: it now conflicts with `main` (mergeable:
false) in `src/agentscope/agent/_agent.py`,
`src/agentscope/agent/_config.py`, and `tests/agent_basic_test.py`.
A rebase onto latest main should resolve them — the hunks look mechanical.

Minor design note (matches the issue spec, so not a blocker): the hint is
injected before every reasoning step while the streak persists, so a
stubborn model accumulates one HintBlock per round. Worth keeping an eye
on context growth.

@DavdGao, once rebased this looks ready for maintainer review.

DavdGao (approved):
lgtm

PR #1818, opened 2026-06-07
fix(service): add workspace root to permission context

### LINKED ISSUE #1791: fix(service): set the workspace directory as working_directories in `PermissionContext`

Premsenareddy:
## Summary
- add a workspace-visible root path interface to WorkspaceBase and concrete workspace backends
- include the resolved workspace root in PermissionContext.working_directories during ChatService agent assembly
- preserve existing working directory entries when already configured

Fixes #1791

## Testing
- .venv/bin/python -m pytest -p no:cacheprovider tests/chat_service_test.py tests/service_toolkit_test.py tests/permission_engine_test.py tests/permission_mode_test.py -q
- .venv/bin/python -m pytest -p no:cacheprovider tests/builtin_read_test.py tests/builtin_write_test.py tests/builtin_glob_test.py tests/builtin_grep_test.py tests/workspace_local_test.py tests/builtin_bash_test.py -q
- git diff --check
- py_compile on touched files

### DIFF
--- a/src/agentscope/app/_service/_chat.py
+++ b/src/agentscope/app/_service/_chat.py
@@ -34,6 +34,26 @@ from ...event import (
     ExternalExecutionResultEvent,
 )
 from ...message import AssistantMsg, Msg, ToolCallState
+from ...permission import AdditionalWorkingDirectory
+from ...state import AgentState
+from ...workspace import WorkspaceBase
+
+
+def _ensure_workspace_working_directory(
+    agent_state: AgentState,
+    workspace: WorkspaceBase,
+) -> None:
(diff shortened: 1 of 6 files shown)


DavdGao:
Fixed in #1823

Issue #1800, opened 2026-06-05
feat(webui): display background tool execution status in the side panel


DavdGao:
**Background**

Some tools in AgentScope can run in the background (e.g., long-running tasks). Currently, there is no way for users to know whether any tools are running in the background for the current session, or which ones they are. Adding visibility into background tool execution would improve transparency and user experience.

**Changes**

- Frontend: after the right side panel is introduced in #1799, add a new section to it displaying the background tool execution status for the current session — including whether any tools are running and the details of each.

**Dependencies**
- #1799: right side panel support in the chat page
- #1720: backend API for listing background running tools

Issue #1663, opened 2026-05-26
feat(rag): migrate the previous RAG and long-term memory implementation within a unified abstraction


DavdGao:
## Background

In version 1.0, we introduced RAG and long-term memory modules, both of which have received significant contributions from the developer community. In the upcoming version, we are revisiting these two modules and, given their similar interfaces and overlapping functionality, we have decided to consolidate them into a single unified abstraction.

## Changes

- Introduces a new `rag` module (tentative name) that unifies RAG and long-term memory under a single abstraction
- The proposed unified interface is as follows:

```python
class XxxBase:

    def list_tools() -> list[ToolBase]:
        """The provided agentic tools for the agent to use for retrieval."""
        return []

    @abstractmethod
    def record(content: str) -> None:
        """Record the content into the rag."""

    @abstractmethod
    def search(keyword: str) -> str:
        """Search the keyword in the storage."""

    def get_instructions() -> str | None:
        """The instructions to attach to the system prompt."""
        return None
```

## Notes
- The class name `XxxBase` is a placeholder and subject to change
- Feedback on the interface design and naming is welcome

DavdGao:
Actually, we find that RAG & long-term memory can be handled entirely by middlewares — no extra abstractions or RAG-related arguments in the Agent constructor are needed.

We simply:

- Add an on_compress_context hook to `MiddlewareBase`
- Add an optional list_tools interface to `MiddlewareBase`

PR #1824, opened 2026-06-07
feat(webui): add session context side panel

### LINKED ISSUE #1799: feat(webui): support to display the current permission rules within the chat page
**Background**

The frontend can already retrieve the `PermissionContext` for the current active session. We should display the associated permission rules in the chat page to improve visibility and transparency.

**Changes**

- Add a collapsible side panel on the right side of the chat page, similar to the sidebar in a coding IDE — hidden by default and expandable on click.
- This panel will consolidate the current workspace info and the new permission context display in one place. Note that the existing workspace display uses a drawer, but the new implementation should use a persistent side panel that allows users to interact with the chat and other parts of the UI while the panel is open.

**Note**
To maintain consistency with the existing codebase, the side panel should be implemented using shadcn and TailwindCSS components.

Premsenareddy:
## Summary

- Replace the chat workspace drawer with a persistent right-side session context panel.
- Add a Permissions tab that displays the current permission mode, working directories, and allow/ask/deny rules.
- Keep the existing MCP and Skill workspace management controls in the new panel.
- Sync permission context from session snapshots and live `state_updated` events.

Fixes #1799

## Validation

- `corepack pnpm exec prettier --check src/pages/chat/ChatViewport.tsx src/components/chat/WorkspacePanel.tsx src/api/types.ts`
- `corepack pnpm --filter frontend lint`
- `corepack pnpm --filter frontend build`
- `git diff --check`
- Automated browser smoke test against a temporary fakeredis-backed AgentScope service:
  - opened a seeded chat session
  - opened and closed the right-side panel
  - verified Workspace and Permissions tabs
  - verified permission mode/rule sections render
  - verified chat input remains present while the panel is open


DavdGao:
@Premsenareddy Thanks for your interest in contributing to AgentScope! Please leave a comment on the corresponding issue #1799  so we can mark it as assigned and avoid duplicated effort.

DavdGao (commented):
@Premsenareddy Thanks for the contribution! However, the current implementation has some issues: MCP, Skill, and Permissions are crammed into a single fixed‑width panel via nested tabs, so only one group is visible at a time, the width can't be adjusted, and there's no clean place to add Background Tasks later.

Suggested changes:

1. Put a single button at the top‑right that opens a dropdown with one checkbox per panel (Permissions, MCPs, Skills, and Background Tasks later). The checkbox state controls whether that panel is visible — multiple can be checked at once and shown side by side.
2. Split the current monolith into independent panels (PermissionsPanel, McpPanel, SkillsPanel), each owning its own header, search, and primary action (e.g. Add MCP belongs to McpPanel only). Adding a new panel later is just a new component plus one entry in the dropdown.
3. Compose the visible panels with shadcn Resizable so each panel can be dragged independently. Persist the visibility and widths to localStorage so the layout sticks across sessions.

An example is the right panels of claude code:

Image

DavdGao:
If this turns out to be a larger change than you expected and you'd like some help along the way, just let us know

Premsenareddy:
Thanks for the detailed direction. I pushed a follow-up that reshapes the panel implementation around that model:

- replaced the nested tab layout with a single session-panels menu using checkbox items for Permissions, MCPs, Skills, plus a disabled Background Tasks placeholder
- split the monolith into independent PermissionsPanel, McpPanel, and SkillsPanel sections, each with its own header and panel-specific search/action
- composed visible panels side by side with a shadcn-style Resizable wrapper around react-resizable-panels
- persisted panel visibility and panel widths in localStorage
- merged current upstream main to resolve the PR conflict

Local validation passed: prettier check, frontend build, and frontend lint. Lint still reports the same existing warnings in unrelated files.

DavdGao:
Hi @Premsenareddy, thank you for your contribution and the effort you put into this PR!

After careful review, we've decided to close this PR for the following reasons:

1. **The current implementation does not fully meet our requirements** for the session context panel. The design and interaction model we're targeting differs from what's implemented here, and the gap would require significant rework rather than incremental fixes.

2. **This PR is blocking other in-flight features** that depend on the same area of the codebase — specifically the background tasks display (#1800) and RAG configuration (#1663). In order to unblock those and move forward, we've implemented the panel functionality in a new PR #1945 , which better aligns with the overall architecture we have in mind.

We truly appreciate your interest in AgentScope and the time you invested in this. The exploration you did here was helpful as a reference. We hope this doesn't discourage you from contributing — there are plenty of other issues open and we'd love to see you involved!

Thanks again 🙏

PR #1832, opened 2026-06-09
feat(tts): add TTS module with DashScope backend and streaming middleware

### LINKED ISSUE #1678: feat(tts): add the basic abstraction for the new TTS module
**Background**

AgentScope has supported multiple TTS models in previous versions, including both non-realtime and realtime (streaming-input) modes. As we move toward 2.0, we plan to re-integrate them back while also taking this opportunity to improve the overall design of the TTS module.

**Proposed Design**

Our current proposal is to integrate the TTS module with the new **credential module**, and expose the model information to the frontend so that users can browse and select from candidate model cards. A single base class covers both non-realtime and realtime TTS via the `supports_streaming_input` flag and an async context manager lifecycle.

The proposed base class is as follows:

```python
from abc import ABC, abstractmethod
from typing import Any, AsyncGenerator

from agentscope.credential import CredentialBase
from agentscope.message import Msg
from agentscope.tts import TTSResponse


class TTSModelBase(ABC):
    credential: CredentialBase
    model: str
    stream: bool
    supports_streaming_input: bool = False

    def __init__(
        self,
        credential: CredentialBase,
        model: str,
        stream: bool,
    ) -> None:
        self.credential = credential
        self.model = model
        self.stream = stream

    async def __aenter__(self) -> "TTSModelBase": ...
    async def __aexit__(self, exc_type, exc_value, tb) -> None: ...

    async def connect(self) -> None:
        """Open the realtime connection. Realtime models only."""

    async def close(self) -> None:
        """Close the realtime connection. Realtime models only."""

    async def push(self, msg: Msg, **kwargs: Any) -> TTSResponse:
        """Non-blocking push of a text chunk; returns whatever audio is
        currently available (may be empty). Realtime models only."""

    @abstractmethod
    async def synthesize(
        self,
        msg: Msg | None = None,
        **kwargs: Any,
    ) -> TTSResponse | AsyncGenerator[TTSResponse, None]:
        """Synthesize speech. Blocks until synthesis is complete; returns a
        single TTSResponse when `stream=False`, or an `AsyncGenerator` of
        TTSResponse chunks when `stream=True`."""
```

And the unified response object:

```python
from dataclasses import dataclass, field
from typing import Literal
from agentscope.message import DataBlock

@dataclass
class TTSResponse:
    content: DataBlock | None
    """The audio chunk produced by the TTS model. The audio format is
    indicated by `content.source.media_type` (e.g. "audio/pcm",
    "audio/mpeg")."""
    id: str
    created_at: str
    type: Literal["tts"] = "tts"
    usage: TTSUsage | None = None
    metadata: dict | None = None
    is_last: bool = True
```

**Key Design Decisions**

- **Credential integration**: Each TTS model is bound to a `CredentialBase` instance, keeping authentication logic centralized and reusable.
- **Model metadata**: `model` is a first-class attribute so the frontend can enumerate and display available model cards.
- **Unified non-realtime / realtime API**: `supports_streaming_input` plus the `__aenter__` / `__aexit__` / `connect` / `close` / `push` lifecycle lets one base class cover both modes; non-realtime models only implement `synthesize`.
- **Streaming output**: `synthesize` may return either a single `TTSResponse` or an `AsyncGenerator[TTSResponse, None]`, governed by the `stream` flag.
- **Returns `TTSResponse`**: TTS is a model-level abstraction invoked inside an agent. The agent that consumes a `TTSResponse` is responsible for wrapping the audio payload into `DATA_BLOCK_*` agent events. This keeps TTS aligned with `EmbeddingModelBase` / chat models, while `RealtimeModelBase` remains the event-emitting abstraction for full-session use cases.
- **Use `DataBlock` for audio payloads**: AgentScope 2.0 unified binary content under `DataBlock` (discriminated by `source.media_type`); there is no separate `AudioBlock`. TTS implementations should populate `content = DataBlock(source=Base64Source(data=..., media_type="audio/..."))` so the agent layer can wrap it directly into `DATA_BLOCK_*` events without any conversion.

We welcome any thoughts and discussion on this design in this issue.

### LINKED ISSUE #1684: feat(tts): implement DashScope TTS API
**Background**

This issue tracks the implementation of the DashScope (Alibaba Cloud) standard TTS (Sambert) model in AgentScope 2.0, as part of the overall TTS module migration (blocked by #1678).

**Task Breakdown**

- Implement `DashScopeTTSModel` as a subclass of `TTSModelBase` for the standard DashScope TTS (Sambert) API
- Enable `stream=True` if the API supports streaming output
- Audio chunks must be returned as `DataBlock` with the correct `media_type` (e.g. `"audio/mpeg"`, `"audio/pcm"`)
- Add unit tests

### LINKED ISSUE #1686: feat(tts): implement DashScope realtime TTS API
**Background**

This issue tracks the implementation of the DashScope (Alibaba Cloud) realtime (streaming-input) TTS model in AgentScope 2.0, as part of the overall TTS module migration (blocked by #1678).

**Task Breakdown**

- Implement `DashScopeRealtimeTTSModel` as a subclass of `TTSModelBase` with `supports_streaming_input = True`, including the `connect` / `close` / `push` / `synthesize` lifecycle
- Audio chunks must be returned as `DataBlock` with the correct `media_type` (e.g. `"audio/pcm"`)
- Add unit tests, including the async context manager lifecycle and incremental `push` → audio-chunk behavior

qbc2016:
## AgentScope Version

2.0.1

## Description

Add a text-to-speech abstraction (`TTSModelBase`) supporting both non-realtime and realtime (streaming-input) modes, with a DashScope implementation (`DashScopeTTSModel`) that wraps the MultiModalConversation API. Includes `TTSMiddleware` that intercepts agent text blocks and injects synthesized audio as `DATA_BLOCK_*` events into the event stream.

## Checklist


src/agentscope/credential/_base.py
```diff
@@ -36,6 +37,30 @@ def get_chat_model_class(cls) -> Type["ChatModelBase"]:
...
+    @classmethod
+    def get_tts_model_class(cls) -> Type["TTSModelBase"] | None:
+        """Return the :class:`TTSModelBase` subclass that consumes this
+        credential, or ``None`` if this provider does not support TTS.
+
+        Returns:
+            `Type[TTSModelBase] | None`:
+                The TTS model class, or ``None``.
+        """
+        return None
```
DavdGao: Keep consistent with `get_chat_model_class` to raise not implemented error

src/agentscope/tts/_dashscope/_model.py
```diff
@@ -0,0 +1,266 @@
...
+class DashScopeTTSModel(TTSModelBase):
+    """DashScope TTS model implementation using the MultiModalConversation
+    API. For more details please see the `official document
+    <https://bailian.console.aliyun.com/?tab=doc#/doc/?type=model&url=2879134>`_.
+    """
+
+    class Parameters(BaseModel):
+        """Frontend-exposed parameters for DashScope TTS models."""
+
+        voice: str = Field(
```
DavdGao, line 60: `Literal["xxx", "xxx"]` rather than str?

src/agentscope/tts/_tts_model_card.py
```diff
@@ -0,0 +1,128 @@
+# -*- coding: utf-8 -*-
```
DavdGao, line 1: I'm not sure if we need a new tts model card class or just reuse the ModelCard class

src/agentscope/tts/_tts_response.py
```diff
@@ -0,0 +1,56 @@
...
+from typing import Literal
+
+from .._utils._common import _get_timestamp
+from .._utils._mixin import DictMixin
+from ..message import DataBlock
+from ..types import JSONSerializableObject
+
+
+@dataclass
+class TTSUsage(DictMixin):
```
DavdGao, line 13: How should we track TTS usage within the `Agent` class? Currently, the `usage` field in `Msg` only contains `input_tokens` and `output_tokens`, which does not distinguish between LLM and TTS usage.

src/agentscope/tts/_tts_response.py
```diff
@@ -0,0 +1,56 @@
...
+
+    type: Literal["tts"] = field(default_factory=lambda: "tts")
+    """The type of the usage, must be `tts`."""
+
+
+@dataclass
+class TTSResponse(DictMixin):
+    """The response of TTS models."""
+
+    content: DataBlock | None
```
DavdGao, line 33: Should we support `None` here?

src/agentscope/middleware/_tts_middleware.py
```diff
@@ -0,0 +1,191 @@
...
+      :meth:`TTSModelBase.push`; any audio produced is emitted immediately.
+      On ``TextBlockEndEvent`` :meth:`TTSModelBase.synthesize` is called
+      to drain remaining audio, and the data block is closed.
+
+    Each ``DataBlockDeltaEvent.data`` carries an **incremental** base64 PCM
+    chunk; the full audio is the concatenation of every delta's decoded
+    bytes (the data block is keyed by ``block_id``).
+    """
+
+    def __init__(self, tts_model: TTSModelBase) -> None:
```
DavdGao, line 41: Where does the output audio sample rate (and other audio parameters) get specified for TTS models? It doesn't seem like the middleware exposes any related configuration options.

src/agentscope/credential/_base.py
qbc2016: This is intentional — unlike chat which every credential must support, TTS is optional. Returning `None` lets `list_tts_models()` degrade gracefully (empty list) and the service layer returns 400 for unsupported providers. If we raise here, every credential without TTS (e.g. Anthropic, Ollama) would need to override just to return `None`.

src/agentscope/tts/_dashscope/_model.py
qbc2016: The available voices vary by model and are defined in YAML model cards (e.g. `qwen3-tts-flash.yaml`). The `from_yaml()` method injects them as enum into the parameter schema for frontend validation. Using `str` here keeps the code decoupled from specific model offerings — adding a new voice only requires a YAML change, not a code change. This is also consistent with `DashScopeChatModel.Parameters.voice` which uses `str | None` for the same reason.

src/agentscope/tts/_tts_response.py
qbc2016: `content=None` means "no audio available yet" — this is used by the default `TTSModelBase.push()` return in the realtime streaming-input scenario, where the caller pushes a text chunk but the model hasn't produced any audio back yet.

src/agentscope/tts/_tts_response.py
qbc2016: How about making `Msg.usage` a list — `usage: list[ChatUsage | TTSUsage]`, appending one entry per model call or TTS call? Both `ChatUsage` and `TTSUsage` already carry a type discriminator (`"chat"` vs `"tts"`), so they're naturally suited for a union list. This would require introducing a new event type (e.g. `TTSCallEndEvent`) for the middleware to emit, and handling it in `Msg.apply_event`.

Or keep `Msg.usage` for LLM only, track TTS usage separately — e.g. via `Msg.metadata["tts_usage"]` or a dedicated `tts_usage` field.

src/agentscope/middleware/_tts_middleware.py
qbc2016: Audio format details are the responsibility of each concrete `TTSModelBase` subclass, not the middleware. If a future TTS provider supports configurable sample rates, the right place to add that would be in that provider's `Parameters` class (like `DashScopeTTSModel.Parameters`), not in the middleware layer.

src/agentscope/app/_router/_schema/_session.py
```diff
@@ -68,6 +74,10 @@ class CreateSessionRequest(BaseModel):
         description="Fallback model used when the primary model fails. "
         "Can be set later via PATCH.",
     )
+    tts_model_config: TTSModelConfig | None = Field(
+        default=None,
+        description="TTS model configuration. " "Can be set later via PATCH.",
```
DavdGao: concat the two strings

src/agentscope/credential/_base.py
DavdGao: If so, how about returning empty list?

src/agentscope/middleware/_tts_middleware.py
```diff
@@ -0,0 +1,192 @@
...
+                blocks produced during reasoning.
+        """
+        self.tts = tts_model
+
+    async def on_reasoning(
+        self,
+        agent: "Agent",
+        input_kwargs: dict,
+        next_handler: Callable[..., AsyncGenerator],
+    ) -> AsyncGenerator:
```
DavdGao, line 56: Docstrings are missing

src/agentscope/middleware/_tts_middleware.py
```diff
@@ -0,0 +1,192 @@
...
+    async def _emit_chunk(
+        agent: "Agent",
+        text_block_id: str,
+        tts_res: TTSResponse,
+        audio_block_ids: dict[str, str],
+        audio_media_types: dict[str, str],
+    ) -> AsyncGenerator:
+        """Emit one TTSResponse chunk as ``DATA_BLOCK_START`` (if needed)
+        followed by ``DATA_BLOCK_DELTA``."""
+        if tts_res is None or tts_res.content is None:
```
DavdGao, line 158: The tts_res typing doesn't include the `None` type

src/agentscope/middleware/_tts_middleware.py
```diff
@@ -0,0 +1,192 @@
...
+            return
+        media_type = tts_res.content.source.media_type
+        data = tts_res.content.source.data
+        if not data:
+            return
+
+        audio_block_id = audio_block_ids.get(text_block_id)
```
DavdGao: tbd

src/agentscope/model/_model_card.py
```diff
@@ -48,18 +48,27 @@ class ModelCard(BaseModel):
     """The model supported output types."""
 
     context_size: int = Field(
+        default=0,
         title="Context size",
         description="The context size.",
-        gt=0,
+        ge=0,
```
DavdGao: Why does this include ge=0 (i.e., greater than or equal to 0)? A context size of 0 doesn't seem meaningful for chat models

src/agentscope/model/_model_card.py
```diff
@@ -48,18 +48,27 @@ class ModelCard(BaseModel):
...
+        default=0,
         title="Max output tokens",
         description="The maximum number of tokens.",
-        gt=0,
+        ge=0,
     )
-    """The model max output tokens."""
+    """The model max output tokens. 0 for non-LLM models (e.g. TTS)."""
+
+    voices: list[str] = Field(
```
DavdGao: Shouldn't `voices` be included in the `parameter` argument？

src/agentscope/tts/_dashscope/_model.py
```diff
@@ -0,0 +1,266 @@
...
+
+    async def synthesize(
+        self,
+        msg: Msg | None = None,
+        **kwargs: Any,
+    ) -> TTSResponse | AsyncGenerator[TTSResponse, None]:
+        """Call the DashScope TTS API to synthesize speech from text.
+
+        Args:
+            msg (`Msg | None`, optional):
```
DavdGao: Should we consider Msg or pure text here?

src/agentscope/tts/_dashscope/_model.py
```diff
@@ -0,0 +1,266 @@
...
+                generator yielding ``TTSResponse`` chunks when ``stream=True``.
+        """
+        if msg is None:
+            return TTSResponse(content=None)
+
+        text = msg.get_text_content()
+
+        import dashscope
+
+        response = await asyncio.to_thread(
```
DavdGao: Why we use a thread here to run the tts model?

src/agentscope/tts/_dashscope/_model.py
```diff
@@ -0,0 +1,266 @@
...
+            text=text,
+            voice=self.parameters.voice,
+            stream=True,
+            **kwargs,
+        )
+
+        if self.stream:
+            return self._parse_into_async_generator(response)
+
+        return await asyncio.to_thread(self._aggregate_sync, response)
```
DavdGao: Similar as above

src/agentscope/tts/_tts_base.py
```diff
@@ -0,0 +1,212 @@
...
+    credential: CredentialBase
+    """The credential used to authenticate against the TTS provider."""
+
+    model: str
+    """The name of the TTS model."""
+
+    parameters: BaseModel
+    """The TTS model parameters."""
+
+    stream: bool
```
DavdGao, line 46: Does the argument `stream` work for tts models?

src/agentscope/tts/_tts_base.py
```diff
@@ -0,0 +1,212 @@
...
+                Additional keyword arguments to pass to the TTS API call.
+
+        Returns:
+            `TTSResponse`:
+                The TTSResponse containing the audio block.
+        """
+        return TTSResponse(content=None)
+
+    @classmethod
+    def list_models(
```
DavdGao, line 141: The `type` field isn't changed to `tts_model` here and defaults to `chat_model`

tests/tts_middleware_test.py
```diff
@@ -0,0 +1,309 @@
...
+        async def next_handler(**_kwargs: Any) -> AsyncGenerator:
+            for evt in upstream_events:
+                yield evt
+
+        emitted = []
+        async for evt in middleware.on_reasoning(agent, {}, next_handler):
+            emitted.append(evt)
+
+        # Upstream events are passed through
+        self.assertIs(emitted[0], upstream_events[0])
```
DavdGao: Assert the dict as a whole object rather than assert their attributes

examples/web_ui/frontend/src/api/types.ts
```diff
@@ -422,3 +440,21 @@ export interface ListModelResponse {
 	models: ModelCard[];
 	total: number;
 }
+
+export interface TTSModelCard {
```
DavdGao, line 444: [TBD] Reuse the previous card?

examples/web_ui/frontend/src/components/popover/ModelParametersPopover.tsx
```diff
@@ -400,6 +408,182 @@ export function ModelParametersPopover({
...
+
+											return (
+												<DropdownMenuSub key={`${credential.id}-${m.name}`}>
+													<DropdownMenuSubTrigger>
+														<span
+															className={
+																isSelected ? 'font-medium' : ''
+															}
+														>
+															{isSelected && '✓ '}
```
DavdGao: Use `DropdownMenuCheckboxItem` directly rather than rending checking icon here

examples/web_ui/frontend/src/hooks/useChat.ts
```diff
@@ -4,15 +4,15 @@ import { chatApi } from '@/api';
 import type { ChatRequest } from '@/api';
 
 /**
- * Fire-and-forget chat trigger.
+ * Chat trigger with request-lifecycle tracking.
```
DavdGao: Remove unnecessary changes

examples/web_ui/frontend/src/hooks/useChat.ts
```diff
@@ -4,15 +4,15 @@ import { chatApi } from '@/api';
...
  *
  * Sends a ``POST /chat/`` request that kicks off a chat run on the
  * backend. Events are **not** returned here — they arrive via the
  * session's SSE stream (``GET /sessions/{sid}/stream``), consumed by
- * :func:`useMessages`.
+ * ``useMessages``.
```
DavdGao: Similar as above

DavdGao (commented):
plz see inline comments

DavdGao (approved):
lgtm

PR #1839, opened 2026-06-09
feat(utils): add configurable ID factory via set_id_factory()

### LINKED ISSUE #1782: [Feature]: Allow customizable ID generator
All IDs are hardcoded to uuid.uuid4().hex, Please allow customization of the ID generation strategy.

Our production database uses UUID7 (time-ordered) for all entities, but AgentScope's internal UUID4 IDs cannot align with this, causing block_id mismatch between streaming events and persisted records.

Suggested approach: expose a configurable factory function, e.g.:
```
from agentscope import config
config.set_id_factory(lambda: uuid7().hex)
```
I'd be happy to contribute a PR for this feature if it's aligned with the project's direction.

lajxw:
## AgentScope Version

2.0.1

## Description

Background: AgentScope hardcodes uuid.uuid4().hex for all entity IDs. This prevents alignment with production databases that use other ID strategies such as UUIDv7 (time-ordered), causing block_id mismatch between streaming events and persisted records.

Changes:
     - Added _id_factory global and set_id_factory() in src/agentscope/_utils/_common.py, allowing users to override the default ID generation strategy at startup.
     - Exported set_id_factory from src/agentscope/__init__.py as public API.
     - Replaced 34 occurrences of uuid.uuid4().hex with _generate_id / _generate_id() across 18 entity files.
     - Security-sensitive tokens (gateway tokens, Redis lock tokens) intentionally remain as uuid.uuid4().hex.

Close #1782 

## Checklist

- [x]  Related documentation has been updated (e.g. links, examples, etc.) in [documentation repository](https://github.com/agentscope-ai/docs)


tests/id_factory_test.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
+"""Tests for the configurable ID factory."""
+import re
+import unittest
+
+from agentscope._utils._common import (
```
DavdGao: Since we are testing the factory method, please import it as `from agentscope import set_id_factory`, the same way end users would.

DavdGao (changes requested):
Thanks for the PR. Two suggestions on the test file:

1. Make the tests async to match our template. Please follow `tests/test_template.py`, which uses `IsolatedAsyncioTestCase` with `asyncSetUp / asyncTearDown` instead of the sync `unittest.TestCase`.
2. Trim the test cases. A few of the current tests add little value (e.g. asserting the public export is the internal function, the non-callable TypeError check). Please keep just the two that capture the intent:  
 - the default id_factory still returns uuid.uuid4().hex  
 - after from agentscope import set_id_factory, a custom factory actually takes effect on entities

DavdGao:
@lajxw Plz format the code by running `pre-commit run --all-files`

lajxw:
I have made the requested changes; please review again.

DavdGao (approved):
LGTM

src/agentscope/_utils/_common.py
```diff
@@ -18,6 +18,47 @@
...
 
 
+def _default_id_factory() -> str:
+    return uuid.uuid4().hex
+
+
+_id_factory: Callable[[], str] = _default_id_factory
+
+
+def set_id_factory(factory: Callable[[], str]) -> None:
```
DavdGao, line 25: Exporting this through `agentscope/__init__.py` is what's tripping the 18 ubuntu test failures in `workspace_docker_test.py`.

The gateway container built by the docker workspace installs `agentscope` via `uv pip install --no-deps` (see `Dockerfile.install_src.template`) and only pre-installs a minimal pin set — `mcp`, `uvicorn`, `fastapi` — defined by `_GATEWAY_BASE_REQUIREMENTS` in `src/agentscope/workspace/_utils.py`. The gateway script only needs `agentscope.mcp.MCPClient`, so this slim setup used to work.

This PR adds `from ._utils._common import set_id_factory` to `src/agentscope/__init__.py`, which now eagerly executes `_common.py` (and thus its top-level `import requests` on line 14) on every `import agentscope`, including inside the gateway container. Since `requests` is not in the gateway's requirements, the container fails to boot with `ModuleNotFoundError: No module named 'requests'`, the gateway never becomes healthy within 30s, and all 18 docker workspace tests fail. macOS/Windows CI passes because those runners skip the docker tests.

Suggested fix: make `requests` a lazy import — drop the top-level `import requests` in this file and move it into the function around line 167 that actually calls `requests.get(url)`. The gateway never reaches that code path, so the problem disappears without touching the docker requirements pin list.

PR #1841, opened 2026-06-09
feat(agent-team): add team worker report loop

### LINKED ISSUE #1797: feat(agent-team): add ralph loop for team members to ensure task completion
**Background**

When a team member completes a task, it is expected to call `TeamSay` to send the result back to the team leader, which wakes the leader up. 

However, the sub-agent sometimes forgets to call `TeamSay`, causing both the leader and the member to exit their ReAct loops prematurely, and the entire team task gets stuck.

**Changes**

- Introduce a ralph loop for team members: treat calling `TeamSay` to report results to the leader as the exit condition for a team member's run. If the member finishes its ReAct loop without calling `TeamSay`, inject a hint to remind it and re-run the `reply` function, repeating until the leader has been notified.
- The current plan is to implemented as an agent middleware under `agentscope.app.middleware`, which will be equipped to the team member when created within `AgentCreate` tool.

YingchaoX:
## AgentScope Version

2.0.1

## Description

Closes #1797.

This PR adds a bounded report loop for Agent Team workers so a team member does not silently finish a task without notifying the leader.

Background:
When a worker completes its assigned task, it is expected to call `TeamSay` to report the result back to the team leader. In practice, the model can sometimes produce a private final answer and exit without calling `TeamSay`, leaving the leader waiting and the team task stuck.

Changes made:
- Added `TeamReportMiddleware` under `agentscope.app.middleware`.
- Attached the middleware only to team worker agents created with `source == "team"`.
- Added structured `team_say` metadata to successful `TeamSay` results so middleware can tell whether the leader was actually notified.
- Treat direct worker-to-leader messages and worker broadcasts as satisfying the report contract.
- Keep reminders bounded to avoid infinite loops.
- Added unit coverage for silent worker finishes, successful reports, peer-only messages, state reset between replies, `TeamSay` metadata, and ChatService middleware assembly.

How to test:
- Ran local checks:
  - `uv run pre-commit run --all-files`
  - `uv run pytest tests/middleware_test.py tests/service_team_tools_test.py tests/service_chat_middleware_test.py`
- Ran local code e2e testing for the Agent Team report flow.
- Ran browser/page e2e testing through the local web UI.
- Verified with screenshots from the page e2e test.

截屏2026-06-09 20 54 44
截屏2026-06-09 20 55 03

## Checklist


YingchaoX:
Resolved the conflicts with the latest main and pushed merge commit 3cfd85f9.

The resolution also adapts the PR to the current codebase:
- updated the renamed agentscope.app._tool import path;
- aligned the tests with the current workspace/resource-access and runtime-hint interfaces;
- attached TeamReportMiddleware based on the actual session team role, so both created and invited workers are covered while leaders and standalone agents are not.

Validation completed successfully:
- uv run --extra dev pytest tests/middleware_test.py tests/service_team_tools_test.py tests/service_chat_middleware_test.py — 65 passed;
- pre-commit hooks for all seven PR files — all passed;
- PR diff whitespace check against the latest main — passed.

DavdGao:
- **Report state is lost on HITL resume.** `_reported_to_leader` is instance state, but `_run_impl` rebuilds the Agent and all middlewares on every run, including confirm/external-result continuations (`_chat.py:485`) — the flag is always `False` at resume and the `on_reply` guard never fires. A worker that already reported gets pushed to `TeamSay` again, duplicating the report and the leader wakeup. `BudgetControlMiddleware` keys this on `AgentState.middle_context` for exactly this reason. Note an offloaded tool keeps this middleware's `on_acting` generator alive inside the abandoned `drain_task`, so it can write state after the reply ends — guard that once the state is persisted.

- **Background-offloaded tools aren't exempted.** The offload placeholder tells the model "simply give a text reply without calling any tool, which ends the current reasoning loop" (`_tool_offload_middleware.py:379`) — exactly what this middleware suppresses. The worker then either reports a result that isn't ready (worse than the hang: the leader acts on a false result) or burns both reminders; the real result arrives later via inbox wakeup and is reported properly anyway. Suggest letting the reply end when the session has live background tasks — `registry_getall(MessageBusKeys.bg_tasks(session_id))` (cf. `_session.py:620`), needs the `MessageBus` injected. Also `TeamSay` is itself offload-eligible (`is_state_injected` / `is_external_tool` both default `False`); past 10s `on_acting` sees a synthetic `ToolResponse` with empty metadata and misreads it as "not reported".

- **Iteration exhaustion isn't handled.** `cur_iter < max_iters - 1` only stops the middleware from causing the fallback; a worker that exhausts `max_iters` on its own yields from `_reply_impl` directly, never through `_reasoning()`, so `on_reasoning` never runs — no `TeamSay`, leader waits forever. Same deadlock as #1797 by another route, made likelier by the reminder iterations. The budget is also off by one: compliance needs two extra iterations (suppressed answer → `TeamSay` → final answer) but the guard reserves one, so at `cur_iter == max_iters - 2` the worker reports and still ends on `ExceedMaxItersEvent`. Suggest intercepting on the *last allowed* iteration instead of after the loop:
  - set `input_kwargs["tool_choice"] = ToolChoice(mode="TeamSay")` alongside the hint (`BudgetControlMiddleware` pattern; a tool name doesn't alter the schema list, so prompt caching is unaffected) — deterministic, and `max_report_reminders` can drop to 1;
  - at `cur_iter == max_iters - 1` with nothing reported, force a `TeamSay` escalating to the leader: cap reached, work incomplete, reply to resume. Resume is free — the leader's `TeamSay` wakes the worker as a new reply and `_reply_impl:745` resets `cur_iter = 0`;
  - cap the escalations, stored per team task (must survive across replies, unlike the budget middleware's per-`reply_id` bucket).

- **`team_say` metadata should be one bool.** `delivered` is always `True`, `recipient_count` duplicates the result text, `broadcast` is `to is None`, and `reports_to_leader` is a verbatim copy of `includes_leader`. Only one is ever read: `metadata={"team_say": {"reports_to_leader": leader_target in recipients}}`.

Minor: `extend([hint])` → `append(hint)`; `max_report_reminders` is hardcoded at the `_chat.py` construction site; `TeamReportMiddleware` is exported in `__all__` but the docs checklist is unchecked.

DavdGao:
Supported in #2379 and #2386

PR #1858, opened 2026-06-11
fix(webui): show credential schema load errors

### LINKED ISSUE #1840: [Bug]: (web-ui): show an error when credential provider schemas fail to load
### Prerequisites

## Summary

When adding a model credential in the Web UI, Safari may show an empty provider
selector with only the placeholder visible. The dialog looks stuck because no
provider options are rendered and no error message is shown.

This appears to be a frontend error-handling gap rather than a backend schema
registration problem. The backend can return the credential schemas correctly,
but the create-credential dialog silently falls back to an empty provider list
when the schema request fails in the browser.

Image

Image

### Error Messages

```shell
## Diagnosis

The backend schema endpoint returns the expected provider schemas when queried
directly with a valid server address and user id header:


curl -H 'x-user-id: <user-id>' http://localhost:8000/credential/schemas


The likely frontend failure path is:

1. The create-credential dialog opens.
2. The dialog requests credential provider schemas.
3. The request fails in the browser, for example when the saved server URL is
   not a browser-reachable address.
4. Loading ends, but no error state is stored.
5. The schema list remains empty, so the provider selector renders only its
   placeholder.

One common local trigger is saving a server bind address such as
`http://0.0.0.0:8000` as the browser target. Browsers should use
`http://localhost:8000` or `http://127.0.0.1:8000` instead.
```

## Reproduction

1. Start the AgentScope app service.
2. Start the Web UI.
3. Open the Web UI in Safari.
4. Connect to the app service.
5. Open the credential page.
6. Click "Add credential".

## Actual Behavior

- The provider selector shows only the placeholder.
- No provider options are available.
- The create button remains disabled.
- No inline error explains why providers failed to load.

## Expected Behavior

The dialog should either:

- show the available credential providers, or
- show an actionable error message explaining that provider schemas failed to
  load.

### Environment

- AgentScope Version: 2.0.1
- Python Version: 3.13.12
- OS: macOS 26.3.1

Premsenareddy:
## Summary
- show inline errors when credential provider schema loading fails in create/edit credential dialogs
- add a credential page provider-load error state instead of silently showing no providers
- keep the sidebar in a loading state while provider schemas are still loading

Fixes #1840

## Tests
- corepack pnpm exec prettier --check frontend/src/components/dialog/CreateCredentialDialog.tsx frontend/src/components/dialog/EditCredentialDialog.tsx frontend/src/pages/credential/index.tsx frontend/src/i18n/locales/en.json frontend/src/i18n/locales/zh.json
- corepack pnpm --filter frontend build
- corepack pnpm --filter frontend lint


YingchaoX:
Thanks for working on this.

I tested the credential page locally and found a related failure mode that this PR makes more visible but does not fully prevent: if the setup page stores the server URL as `localhost:8000` instead of `http://localhost:8000`, the API client throws before the request is sent:

`TypeError: Failed to construct 'URL': Invalid URL`

This still results from the same credential schema loading path, but the root cause is that the setup flow accepts or preserves a browser-unusable server URL.

Would you be open to including a small normalization step in this PR, for example converting `localhost:8000` / `127.0.0.1:8000` to `http://localhost:8000` before saving `server_url`?

Premsenareddy:
Thanks for testing that path. I added the normalization in the follow-up commit:

- `localhost:8000` / `127.0.0.1:8000` are normalized to `http://...` before saving from setup.
- the setup input now allows schemeless local URLs to reach submit instead of relying on native `type=url` validation.
- API base URL reads also normalize existing stored values, so a previously saved schemeless URL can recover without throwing before the request is sent.

Local checks passed: prettier check, frontend build, and frontend lint.

Premsenareddy:
Refreshed this PR against latest `main` after upstream moved to `61cdeae4` and resolved the Web UI conflicts.\n\nThe refresh keeps the newer setup health-check flow while preserving this PR's URL normalization behavior: schemeless local URLs are normalized before the health check and before saving `server_url`. It also preserves the credential provider load error copy in both English and Chinese locales.\n\nValidation on refreshed head `a83616d0`:\n- Web UI production build passed\n- Web UI lint passed with existing warnings only\n- file-scoped pre-commit passed\n\nGitHub now reports the PR as mergeable and review-blocked instead of dirty.

PR #1866, opened 2026-06-12
feat: add DashScope CosyVoice TTS model

### LINKED ISSUE #1685: feat(tts): implement DashScope CosyVoice TTS API
**Background**

This issue tracks the implementation of the DashScope (Alibaba Cloud) CosyVoice TTS model (non-realtime) in AgentScope 2.0, as part of the overall TTS module migration (blocked by #1678).

**Task Breakdown**

- Implement `DashScopeCosyVoiceTTSModel` as a subclass of `TTSModelBase` for the CosyVoice API
- Enable `stream=True` if the API supports streaming output
- Audio chunks must be returned as `DataBlock` with the correct `media_type` (e.g. `"audio/mpeg"`, `"audio/pcm"`)
- Add unit tests

> 📢 This is an open issue and help is wanted. If you'd like to work on this, please leave a comment below to claim it first before starting.
Please also follow the existing [DashScopeTTSModel implementation](https://github.com/agentscope-ai/agentscope/tree/main/src/agentscope/tts/_dashscope) as a reference.

YingchaoX:
## AgentScope Version

```text
2.0.1
```

## Description

This PR addresses #1685 by adding DashScope CosyVoice TTS support.

It introduces `DashScopeCosyVoiceTTSModel`, a new DashScope TTS model class
for CosyVoice V3 models. The implementation uses DashScope's
`dashscope.audio.tts_v2.SpeechSynthesizer` WebSocket SDK path and supports
both streaming and non-streaming audio synthesis.

Main changes:

- Add `DashScopeCosyVoiceTTSModel`.
- Support `stream=True` incremental audio output.
- Support `stream=False` full audio aggregation.
- Default to `cosyvoice-v3-flash` and voice `longanhuan`.
- Add media type mapping for `wav`, `mp3`, `pcm`, and `opus`.
- Add dedicated CosyVoice model cards for:
  - `cosyvoice-v3-flash`
  - `cosyvoice-v3-plus`
- Register CosyVoice models under `DashScopeCredential`.
- Export the new model from `agentscope.tts`.
- Include CosyVoice model card YAML files in package data.
- Add mocked unit tests covering empty input, streaming, non-streaming,
  metadata, media type mapping, model card discovery, and credential wiring.

How to test:

```bash
uv run --extra dev pytest tests/tts_dashscope_test.py
```

Result:

```text
40 passed
```

A local Web UI E2E smoke test was also run with `CosyVoice-V3-Flash`. The
assistant response included an `audio/wav` data block and exposed a new
`Play audio` button.

Pre-commit status:

```bash
uv run pre-commit run --all-files
```

Most hooks passed, and the flake8 issues in this PR were fixed. The remaining
failures are tool/environment related under Python 3.14:

- `black` fails with `module 'ast' has no attribute 'Str'`.
- `pylint` / `astroid` crashes while checking an existing file
  `src/agentscope/credential/_factory.py`.
- `pylint` also reports existing import-order warnings in
  `src/agentscope/credential/_gemini.py`.

## Checklist


src/agentscope/tts/_dashscope/_cosyvoice_model.py
```diff
@@ -0,0 +1,415 @@
...
+
+
+def _make_callback_class() -> type["ResultCallback"]:
+    """Create a streaming callback class lazily after importing the SDK."""
+    from dashscope.audio.tts_v2 import ResultCallback
+
+    class _Callback(ResultCallback):
+        """Callback that exposes streaming audio through a thread queue."""
+
+        def __init__(self) -> None:
```
qbc2016: Please add docstring here

src/agentscope/tts/_dashscope/_cosyvoice_model.py
```diff
@@ -0,0 +1,415 @@
...
+_DEFAULT_AUDIO_FORMAT = "wav"
+_DEFAULT_SAMPLE_RATE = 24000
+_DEFAULT_TIMEOUT_MILLIS = 600000
+
+
+class _StreamComplete:
+    """Sentinel for callback-driven streaming completion."""
+
+
+class _StreamError:
```
qbc2016: These sentinel classes could be replaced by a single `_Sentinel(Enum)` or even just string tags, since only their identity is checked (`event_type is _StreamComplete`).

src/agentscope/tts/_dashscope/_cosyvoice_model.py
```diff
@@ -0,0 +1,415 @@
...
+
+        def __init__(self) -> None:
+            super().__init__()
+            self.queue: queue.Queue[tuple[Any, Any]] = queue.Queue()
+            self._done = False
+
+        def on_data(self, data: bytes) -> None:
+            """Receive one binary audio chunk."""
+            if data:
+                self.queue.put((bytes, bytes(data)))
```
qbc2016: Using the `bytes` type itself as a sentinel/tag in `self.queue.put((bytes, bytes(data)))` is clever but slightly unconventional. Consider a small enum or named constant (like `_EventType.DATA`) for clarity — the reader needs to trace back to understand what `event_type is bytes` means in the consumer loop.

qbc2016 (commented):
Thank you for this PR. 
Since `DashScopeCosyVoiceRealtimeTTSModel` has already been implemented in the `main` branch, I think we could extract some common code, similar to how we implemented [DashScopeCosyVoiceRealtimeTTSModel](https://github.com/agentscope-ai/agentscope/blob/v1/src/agentscope/tts/_dashscope_cosyvoice_realtime_tts_model.py) and [DashScopeCosyVoiceTTSModel](https://github.com/agentscope-ai/agentscope/blob/v1/src/agentscope/tts/_dashscope_cosyvoice_tts_model.py) in the `v1` branch.

src/agentscope/tts/_dashscope/_cosyvoice_model.py
```diff
@@ -0,0 +1,322 @@
...
+if TYPE_CHECKING:
+    from dashscope.audio.tts_v2 import AudioFormat, ResultCallback
+
+
+_DEFAULT_AUDIO_FORMAT = "wav"
+_DEFAULT_SAMPLE_RATE = 24000
+_DEFAULT_TIMEOUT_MILLIS = 600000
+
+
+def _make_callback_class() -> type["ResultCallback"]:
```
qbc2016: Maybe we should reuse the `_make_cosyvoice_callback_class()` in `_cosyvoice_realtime_model.py` and move it to `_utils.py`

YingchaoX: Thanks, updated. I moved the CosyVoice callback factories into _cosyvoice_utils.py and reused them from both the non-realtime and realtime CosyVoice models. I kept separate factories because the non-realtime callback emits queue events while the realtime callback owns PCM buffering and chunk response helpers.

YingchaoX:
This workflow seems to fail for fork PR review events because `secrets.STORAGE_REPO_TOKEN` is unavailable (`Secret source: None`), so `actions/checkout` receives an empty `token` while checking out `DavdGao/agentscope-monitor`.
@DavdGao

src/agentscope/tts/_dashscope/_cosyvoice_utils.py
```diff
@@ -0,0 +1,284 @@
...
+        usage=usage,
+        is_last=is_last,
+    )
+
+
+def _make_cosyvoice_streaming_callback_class() -> type["ResultCallback"]:
+    """Create the CosyVoice callback used by non-realtime streaming TTS."""
+    from dashscope.audio.tts_v2 import ResultCallback
+
+    class _CosyVoiceStreamingCallback(ResultCallback):
```
qbc2016: Nice refactor extracting the shared utilities into `_cosyvoice_utils.py` 

One suggestion: since `SpeechSynthesizer.call(text)` triggers the same `on_data` → `on_complete` callback sequence as `streaming_call` + `streaming_complete`, the Event+buffer-based `_make_cosyvoice_callback_class` should be reusable for non-realtime streaming too:

```python
callback = _make_cosyvoice_callback_class(..., prepend_header=False)()
synthesizer = SpeechSynthesizer(model, voice, format, callback=callback)
call_task = asyncio.create_task(asyncio.to_thread(synthesizer.call, text))
return callback.get_audio_chunks()
```
The only adjustment needed is a `prepend_header` flag (defaulting to `True` for realtime PCM), since non-realtime already gets container-formatted output from the SDK.

This could let us consolidate to a single callback class and drop `_make_cosyvoice_streaming_callback_class`,` _parse_into_async_generator`, `_call_streaming_synthesizer`, and `_EventType`, similar to the approach in the v1 branch. Happy to hear if there are edge cases where the queue-based approach is preferable!

src/agentscope/tts/_dashscope/_cosyvoice_models/cosyvoice-v3-flash-realtime.yaml
```diff
@@ -0,0 +1,28 @@
+name: cosyvoice-v3-flash
```
qbc2016: Small suggestion: instead of maintaining two nearly-identical directories (`_cosyvoice_models/` and `_cosyvoice_tts_models/`), consider putting all CosyVoice YAML cards in a single `_cosyvoice_models/` directory, with the filename distinguishing the variants — e.g. `cosyvoice-v3-flash.yaml` (non-realtime) and `cosyvoice-v3-flash-realtime.yaml` (realtime), mirroring the existing Qwen TTS pattern (`qwen3-tts-flash.yaml` vs `qwen3-tts-flash-realtime.yaml`). Since `TTSModelBase.list_models()` already filters by `card.realtime`, both classes can scan the same directory and pick the right cards. This would let us remove the `_cosyvoice_tts_models/` directory and its `pyproject.toml` entry.

qbc2016 (commented):

src/agentscope/tts/_dashscope/_cosyvoice_models/cosyvoice-v3-flash-realtime.yaml
YingchaoX: Thanks, updated. I merged the CosyVoice realtime and non-realtime cards into `_cosyvoice_models/`, using `*-realtime.yaml` for the realtime variants. Both classes now scan the same directory and rely on `TTSModelBase.list_models()` to filter by `realtime`.

Also removed the obsolete `_cosyvoice_tts_models/*.yaml` package-data entry.

Validated with `uv run pytest tests/tts_dashscope_test.py` and targeted pre-commit.

qbc2016 (commented):
Looking at the non-realtime paths, there are two `asyncio.to_thread` usages that may not be necessary:

**Non-streaming path** (`_synthesize_synthesizer.call()` is blocking, but other TTS implementations like `DashScopeTTSModel` call `MultiModalConversation.call()` directly in `synthesize()` without `to_thread` — they just block the event loop. For consistency, `_synthesize_sync` could be called directly too, without the `asyncio.to_thread` wrapper.

**Streaming path** (`_stream_audio_chunks`): `asyncio.create_task(asyncio.to_thread(_call_synthesizer))` exists because `SpeechSynthesizer.call()` blocks. But the realtime model avoids this by using `streaming_call()` + `streaming_complete()` (non-blocking), letting `callback.get_audio_chunks()` drive the waiting via its internal `to_thread(event.wait)`. If the non-realtime streaming path also used `streaming_call()` + `streaming_complete()`, the `create_task`/`to_thread`/`await call_task` block could be removed entirely.

cc @YingchaoX

YingchaoX:
> Looking at the non-realtime paths, there are two `asyncio.to_thread` usages that may not be necessary:
> 
> **Non-streaming path** (`_synthesize_synthesizer.call()` is blocking, but other TTS implementations like `DashScopeTTSModel` call `MultiModalConversation.call()` directly in `synthesize()` without `to_thread` — they just block the event loop. For consistency, `_synthesize_sync` could be called directly too, without the `asyncio.to_thread` wrapper.
> 
> **Streaming path** (`_stream_audio_chunks`): `asyncio.create_task(asyncio.to_thread(_call_synthesizer))` exists because `SpeechSynthesizer.call()` blocks. But the realtime model avoids this by using `streaming_call()` + `streaming_complete()` (non-blocking), letting `callback.get_audio_chunks()` drive the waiting via its internal `to_thread(event.wait)`. If the non-realtime streaming path also used `streaming_call()` + `streaming_complete()`, the `create_task`/`to_thread`/`await call_task` block could be removed entirely.
> 
> cc @YingchaoX

Updated.

Non-streaming now calls _synthesize_sync() directly. Streaming now uses streaming_call(text) + streaming_complete(complete_timeout_millis=timeout_millis) with the shared callback and prepend_header=False.

Since streaming_complete() blocks in the SDK, it runs in a background task so callback.get_audio_chunks() can still yield chunks incrementally. Added tests for this path and final metadata/usage.

qbc2016:
Thanks for the continued work on this. The direction of extracting shared code is correct, but the non-realtime model (`DashScopeCosyVoiceTTSModel`) has become significantly more complex than it needs to be. I'd like the implementation to be much closer in simplicity to the [v1 branch `DashScopeCosyVoiceTTSModel`](https://github.com/agentscope-ai/agentscope/blob/v1/src/agentscope/tts/_dashscope_cosyvoice_tts_model.py).

### Main concerns

**1. `synthesize()` is over-parameterized**

The current `synthesize()` pops 7 kwargs (`audio_format`, `format`, `sample_rate`, `voice`, `api_key`, `timeout_millis`, `bit_rate`) to allow per-call overrides. This is inconsistent with how `DashScopeTTSModel.synthesize()` works (which takes just `text` + pass-through kwargs to the SDK). These should be constructor-time configuration, not runtime variability:

Current (too complex):
```python
async def synthesize(self, text=None, **kwargs):
    audio_format = kwargs.pop("audio_format", None)
    format_alias = kwargs.pop("format", None)
    sample_rate = kwargs.pop("sample_rate", _DEFAULT_SAMPLE_RATE)
    voice = kwargs.pop("voice", self.parameters.voice)
    api_key = kwargs.pop("api_key", ...)
    timeout_millis = kwargs.pop("timeout_millis", ...)
    bit_rate = kwargs.pop("bit_rate", None)
    ...
```

Expected (simple, like v1):
```python
async def synthesize(self, text=None, **kwargs):
    if not text:
        return TTSResponse(content=None)
    synthesizer, callback = self._create_synthesizer()
    if self.stream:
        synthesizer.call(text=text)
        return callback.get_audio_chunks()
    else:
        audio = synthesizer.call(text=text)
        ...
```

**2. Streaming path is still complex**

The `_stream_audio_chunks()` uses `asyncio.create_task(asyncio.to_thread(_complete_synthesizer))` which contradicts the simplification goal. For a non-realtime model (one-shot text input, not streaming input), the simplest approach is what v1 does: call `synthesizer.call(text)` (which blocks but triggers callbacks), then yield from the callback's async generator. No `streaming_call` + `streaming_complete` + background task needed.

**3. `_cosyvoice_utils.py` has unnecessary abstractions**

- `_audio_format_enum()` (30+ lines) — maps string to SDK enum with complex fallback logic. For 99% of use cases, `AudioFormat.PCM_24000HZ_MONO_16BIT` or `AudioFormat.WAV_24000HZ_MONO_16BIT` suffices. If needed, expose `audio_format` as a constructor parameter (an enum, not a string).
- `_response_metadata()` — trivial dict construction that could be inline.
- `_parse_usage()` — always returns zeros; questionable value.

The shared utils should primarily contain just the callback factory (`_make_cosyvoice_callback_class`), similar to v1's `_utils.py`.

**4. `_make_cosyvoice_callback_class` has too many parameters**

The `prepend_header` parameter was added to support the non-realtime streaming path's desire to not prepend WAV headers. But if both models use the same approach (always prepend header on first chunk), this parameter is unnecessary.

### Suggested structure

```
_cosyvoice_utils.py          ~100 lines  (callback class only)
_cosyvoice_model.py          ~100 lines  (simple non-realtime model)
_cosyvoice_realtime_model.py  unchanged  (complex realtime with retry/reconnect)
```

The non-realtime model should mirror v1's simplicity:
1. `__init__()` — store config (credential, model, voice, stream, audio_format)
2. `_create_synthesizer()` — create a fresh `SpeechSynthesizer` + callback per request
3. `synthesize(text)` — if stream: `synthesizer.call(text)` → yield from callback; else: `synthesizer.call(text)` → encode → return

Please simplify `_cosyvoice_model.py` along these lines. The model card integration, credential wiring, and test coverage are all great — it's just the core synthesis logic that needs to be leaner.

YingchaoX:
> Thanks for the continued work on this. The direction of extracting shared code is correct, but the non-realtime model (`DashScopeCosyVoiceTTSModel`) has become significantly more complex than it needs to be. I'd like the implementation to be much closer in simplicity to the [v1 branch `DashScopeCosyVoiceTTSModel`](https://github.com/agentscope-ai/agentscope/blob/v1/src/agentscope/tts/_dashscope_cosyvoice_tts_model.py).
> 
> ### Main concerns
> **1. `synthesize()` is over-parameterized**
> 
> The current `synthesize()` pops 7 kwargs (`audio_format`, `format`, `sample_rate`, `voice`, `api_key`, `timeout_millis`, `bit_rate`) to allow per-call overrides. This is inconsistent with how `DashScopeTTSModel.synthesize()` works (which takes just `text` + pass-through kwargs to the SDK). These should be constructor-time configuration, not runtime variability:
> 
> Current (too complex):
> 
> ```python
> async def synthesize(self, text=None, **kwargs):
>     audio_format = kwargs.pop("audio_format", None)
>     format_alias = kwargs.pop("format", None)
>     sample_rate = kwargs.pop("sample_rate", _DEFAULT_SAMPLE_RATE)
>     voice = kwargs.pop("voice", self.parameters.voice)
>     api_key = kwargs.pop("api_key", ...)
>     timeout_millis = kwargs.pop("timeout_millis", ...)
>     bit_rate = kwargs.pop("bit_rate", None)
>     ...
> ```
> 
> Expected (simple, like v1):
> 
> ```python
> async def synthesize(self, text=None, **kwargs):
>     if not text:
>         return TTSResponse(content=None)
>     synthesizer, callback = self._create_synthesizer()
>     if self.stream:
>         synthesizer.call(text=text)
>         return callback.get_audio_chunks()
>     else:
>         audio = synthesizer.call(text=text)
>         ...
> ```
> 
> **2. Streaming path is still complex**
> 
> The `_stream_audio_chunks()` uses `asyncio.create_task(asyncio.to_thread(_complete_synthesizer))` which contradicts the simplification goal. For a non-realtime model (one-shot text input, not streaming input), the simplest approach is what v1 does: call `synthesizer.call(text)` (which blocks but triggers callbacks), then yield from the callback's async generator. No `streaming_call` + `streaming_complete` + background task needed.
> 
> **3. `_cosyvoice_utils.py` has unnecessary abstractions**
> 
> * `_audio_format_enum()` (30+ lines) — maps string to SDK enum with complex fallback logic. For 99% of use cases, `AudioFormat.PCM_24000HZ_MONO_16BIT` or `AudioFormat.WAV_24000HZ_MONO_16BIT` suffices. If needed, expose `audio_format` as a constructor parameter (an enum, not a string).
> * `_response_metadata()` — trivial dict construction that could be inline.
> * `_parse_usage()` — always returns zeros; questionable value.
> 
> The shared utils should primarily contain just the callback factory (`_make_cosyvoice_callback_class`), similar to v1's `_utils.py`.
> 
> **4. `_make_cosyvoice_callback_class` has too many parameters**
> 
> The `prepend_header` parameter was added to support the non-realtime streaming path's desire to not prepend WAV headers. But if both models use the same approach (always prepend header on first chunk), this parameter is unnecessary.
> 
> ### Suggested structure
> ```
> _cosyvoice_utils.py          ~100 lines  (callback class only)
> _cosyvoice_model.py          ~100 lines  (simple non-realtime model)
> _cosyvoice_realtime_model.py  unchanged  (complex realtime with retry/reconnect)
> ```
> 
> The non-realtime model should mirror v1's simplicity:
> 
> 1. `__init__()` — store config (credential, model, voice, stream, audio_format)
> 2. `_create_synthesizer()` — create a fresh `SpeechSynthesizer` + callback per request
> 3. `synthesize(text)` — if stream: `synthesizer.call(text)` → yield from callback; else: `synthesizer.call(text)` → encode → return
> 
> Please simplify `_cosyvoice_model.py` along these lines. The model card integration, credential wiring, and test coverage are all great — it's just the core synthesis logic that needs to be leaner.

Thanks! Updated: non-realtime CosyVoice now follows the simpler v1-style SpeechSynthesizer.call() flow without background tasks/to_thread wrapping. I also aligned the v3 voice lists with DashScope docs and verified them with live API calls.

src/agentscope/tts/_dashscope/_cosyvoice_model.py
```diff
@@ -0,0 +1,189 @@
...
+        audio_data = synthesizer.call(text=text)
+        metadata = self._response_metadata(synthesizer)
+        if not audio_data:
+            return TTSResponse(content=None, metadata=metadata)
+        return self._build_wav_response(
+            audio_data,
+            metadata=metadata,
+        )
+
+    def _create_synthesizer(
```
qbc2016, line 163: The `callback` parameter is a boolean that controls whether to create a callback, but reading the call site `self._create_synthesizer(callback=self.stream)` doesn't convey intent. The v1 approach is simpler — no parameter, just use `self.stream` internally:

```python
def _create_synthesizer(self):
    import dashscope
    from dashscope.audio.tts_v2 import AudioFormat, SpeechSynthesizer

    dashscope.api_key = self.credential.api_key.get_secret_value()

    callback_instance = (
        _make_cosyvoice_callback_class()() if self.stream else None
    )
    synthesizer = SpeechSynthesizer(
        model=self.model,
        voice=self.parameters.voice,
        format=AudioFormat.PCM_24000HZ_MONO_16BIT,
        callback=callback_instance,
    )
    return synthesizer, callback_instance
```

src/agentscope/tts/_dashscope/_cosyvoice_model.py
```diff
@@ -0,0 +1,189 @@
...
+        synthesizer = SpeechSynthesizer(
+            model=self.model,
+            voice=self.parameters.voice,
+            format=AudioFormat.PCM_24000HZ_MONO_16BIT,
+            callback=callback_instance,
+        )
+        return synthesizer, callback_instance
+
+    @staticmethod
+    def _response_metadata(
```
qbc2016: Suggest removing it entirely to keep consistency with other TTS models. If we need metadata in the future, we can add it uniformly across all TTS models. This also removes the need for `from ...types import JSONSerializableObject` (L19) and the `metadata` parameter in `_build_wav_response`.

`synthesize()` would simplify to:

```python
audio_data = synthesizer.call(text=text)
if not audio_data:
    return TTSResponse(content=None)
return self._build_wav_response(audio_data)
```

src/agentscope/tts/_dashscope/_cosyvoice_utils.py
```diff
@@ -0,0 +1,160 @@
...
+from ...types import JSONSerializableObject
+
+if TYPE_CHECKING:
+    from dashscope.audio.tts_v2 import ResultCallback
+
+
+_SAMPLE_RATE = 24000
+_CHANNELS = 1
+_BITS_PER_SAMPLE = 16
+_MEDIA_TYPE = "audio/wav"
```
qbc2016, line 20: `_SAMPLE_RATE`, `_CHANNELS`, `_BITS_PER_SAMPLE`, `_MEDIA_TYPE` are defined identically in both `_cosyvoice_model.py` and `_cosyvoice_utils.py`. Consider importing from `_cosyvoice_utils` to avoid duplication.

src/agentscope/tts/_dashscope/_cosyvoice_model.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        super().__init__(
+            credential=credential,
+            model=model,
+            parameters=parameters,
+            stream=stream,
+        )
+
+    @classmethod
+    def list_models(
```
qbc2016, line 107: This class uses `Path(__file__).parent / "_cosyvoice_models"`, while `DashScopeCosyVoiceRealtimeTTSModel` uses a `_MODELS_DIR` class variable with `os.path.join`. Suggest aligning — either both use `_MODELS_DIR` or both use `Path`.

examples/web_ui/frontend/src/components/popover/ModelParametersPopover.tsx
```diff
@@ -242,7 +242,7 @@ export function ModelParametersPopover({
 
 	useEffect(() => {
 		setValues(selectedModel?.parameters ?? {});
-	}, [selectedModel?.model]);
+	}, [selectedModel?.model, selectedModel?.parameters]);
```
qbc2016: This `useEffect` controls the **chat model** parameter form (`selectedModel`), not TTS. TTS model parameters are rendered directly from `selectedTTSModel.parameters` (fully controlled, no `useEffect` needed), and already handle same-name model disambiguation via the `realtime` flag. Adding `selectedModel?.parameters` (an object reference) to the chat model dependency array is unrelated to the CosyVoice feature and risks resetting the form on every parent re-render if the parent creates a new `parameters` object. Suggest reverting this change.

src/agentscope/tts/_dashscope/_cosyvoice_utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def _build_audio_response(
+    audio_data: bytes,
+    media_type: str,
+    *,
+    is_last: bool = True,
+    metadata: dict[str, JSONSerializableObject] | None = None,
```
qbc2016: The `metadata` parameter is never passed from any call site and can be removed too, along with the `from ...types import JSONSerializableObject` import.

tests/tts_dashscope_test.py
```diff
@@ -318,6 +323,191 @@ async def test_empty_stream_yields_terminal(self) -> None:
...
+            "payload": {"output": {}},
+        }
+        self.mock_synthesizer_class.side_effect = self._make_synthesizer
+
+    def tearDown(self) -> None:
+        """Tear down the test case."""
+        self.patcher.stop()
+
+    @staticmethod
+    def _wait_for_callback_consumed(callback: Any, length: int) -> None:
```
qbc2016: This static method is defined but never called anywhere in the test file. Remove it.

src/agentscope/app/_service/_tts_model.py
```diff
@@ -60,8 +64,16 @@ async def get_tts_model(
...
+    realtime: bool | None = None,
 ) -> Type[TTSModelBase]:
     """Pick the TTS class that lists the given model name."""
+    if realtime is not None:
+        for cls in classes:
+            if cls.realtime != realtime:
+                continue
+            if any(card.name == model for card in cls.list_models()):
+                return cls
+
```
qbc2016: When `realtime` is explicitly provided but no matching class is found in the first (filtered) loop, the function silently falls through to the second loop which ignores the `realtime` flag. This can return a realtime class when `realtime=False` was explicitly requested (or vice versa). Consider skipping the unfiltered fallback when `realtime` is explicitly provided.

qbc2016 (commented):
Good Job! Much cleaner now. 
A few nits remain before merging. Please see the inline comments.

src/agentscope/tts/_dashscope/_cosyvoice_model.py
```diff
@@ -0,0 +1,165 @@
...
+            assert callback is not None
+            synthesizer.call(text=text)
+            return callback.get_audio_chunks()
+
+        audio_data = synthesizer.call(text=text)
+        if not audio_data:
+            return TTSResponse(content=None)
+        return self._build_wav_response(audio_data)
+
+    def _create_synthesizer(
```
qbc2016, line 163: `DashScopeCosyVoiceTTSModel._create_synthesizer` and `DashScopeCosyVoiceRealtimeTTSModel._create_synthesizer` contain nearly identical logic — set API key, create callback, create `SpeechSynthesizer`. Consider extracting a shared helper into `_cosyvoice_utils.py`:

qbc2016 (commented):
Thanks for all the iterations — the PR has improved a lot through each round. Really appreciate the responsiveness.
However, since `cosyvoice-v3-flash` and `cosyvoice-v3-plus` each have both a realtime and a non-realtime variant (same model name, different mode), we discussed this and think it might be cleaner to have a single `DashScopeCosyVoiceTTSModel` with `realtime` as a constructor parameter rather than maintaining two separate classes.

The base class `TTSModelBase` already supports this pattern — `__aenter__`/`__aexit__` gate `connect()`/`close()` behind `if self.realtime:`, and `push()` has a default no-op for non-realtime mode. The two `_create_synthesizer` methods are nearly identical and would unify naturally. `synthesize()` would branch on `self.realtime` — the two paths are independent and don't share mutable state:

```python
async def synthesize(self, text=None, **kwargs):
    if not self.realtime:
        # Simple one-shot path (current _cosyvoice_model.py logic)
        ...
    else:
        # Streaming-input path with retry/reconnect (current _cosyvoice_realtime_model.py logic)
        ...
```

That said, this is a design-level suggestion — not a blocker. If you'd prefer to keep the two classes as-is, that's totally fine — we can take it from here and do the merge on our side. Let us know your preference.

src/agentscope/tts/_dashscope/_cosyvoice_model.py
```diff
@@ -0,0 +1,399 @@
...
+        realtime: bool = Field(
+            default=False,
+            title="Realtime",
+            description="Whether to enable streaming input mode.",
+        )
+
+    type: Literal["dashscope_cosyvoice_tts"] = "dashscope_cosyvoice_tts"
+    """The type of the TTS model."""
+
+    realtime: bool = False
```
qbc2016: Since `__init__` already sets `self.realtime = self.parameters.realtime`, the class-level declaration:

```python
realtime: bool = False
```

is always overridden and can be removed. Keeping it is misleading — a reader might assume the model is always non-realtime, when in fact the value is driven by `Parameters`.

examples/web_ui/frontend/src/pages/credential/index.tsx
```diff
@@ -290,7 +290,7 @@ function DetailPanel({ credential, schema, onEdit, onDelete }: DetailPanelProps)
 						</h3>
 						<div className="grid grid-cols-2 xl:grid-cols-3 gap-3">
 							{ttsModels.map((m) => (
-								<TTSModelCardItem key={m.name} model={m} />
+								<TTSModelCardItem key={`${m.name}-${m.realtime}`} model={m} />
```
qbc2016: Now that there are only two YAML model cards (`cosyvoice-v3-flash.yaml` and `cosyvoice-v3-plus.yaml`), each model name is unique in the list. The composite `(name, realtime)` key introduced in `credential/index.tsx` was added to disambiguate two entries sharing the same model name — that scenario no longer exists. This file can be reverted to its pre-PR state.

src/agentscope/app/_service/_tts_model.py
```diff
@@ -46,10 +46,19 @@ async def get_tts_model(
             detail=f"Provider {config.type!r} does not support TTS models.",
         )
 
-    tts_cls = _resolve_tts_class(tts_classes, config.model)
-    parameters = (
-        tts_cls.Parameters(**config.parameters) if config.parameters else None
+    tts_cls = _resolve_tts_class(
```
qbc2016: The `realtime`-based class-selection logic in `_resolve_tts_class` is no longer needed — a CosyVoice model name always resolves to `DashScopeCosyVoiceTTSModel`, and `realtime` is just a field inside `Parameters`.

src/agentscope/app/storage/_model/_session.py
```diff
@@ -44,6 +44,9 @@ class TTSModelConfig(BaseModel):
     model: str
     """The TTS model name."""
 
+    realtime: bool | None = None
```
qbc2016: The `realtime: bool | None = None` field added to `TTSModelConfig` is also redundant — `realtime` is already carried inside `parameters` (e.g. `{"voice": "longanyang", "realtime": true}`), so the top-level field can be removed.

qbc2016 (commented):
Please see the inline comments and revert the unnecessary changes accordingly. Everything else LGTM.

DavdGao (approved):
lgtm

PR #1870, opened 2026-06-13
fix(service): skip empty wake-up runs

### LINKED ISSUE #1868: [Bug]: Wake-up / scheduled runs execute N times across instances (1 real reply + N empty "your message was empty" replies)
### Prerequisites

## Summary

In a multi-instance deployment (multiple app processes sharing one Redis + DB), a single wake-up — a cron `Schedule` fire, a team message, or a background-tool completion — makes the agent run **N times** on the same session instead of once. The first run consumes the inbox content and replies correctly; the remaining N-1 runs find an empty inbox and reply with "It looks like your message was empty / didn't include a specific request…".

Two root causes, both present on `main` (HEAD `43ce70b2`):

### Defect A — wake-up queue consumed non-atomically across instances
`enqueue_wakeup` (`src/agentscope/app/message_bus/_base.py`) does `XADD` (one entry) + `PUBLISH` to the broadcast channel `agentscope:wakeup_signal`. **Every** instance's `WakeupDispatcher` subscribes and, on each signal, calls `dequeue_wakeups` → `queue_drain` (`_redis_message_bus.py`), which is `XRANGE` then `XDEL`. Its docstring states this is "destructive and idempotent **under the single-consumer-per-key invariant**." With N dispatchers on the single `agentscope:wakeups` key that invariant is violated: multiple dispatchers `XRANGE` the same entry before any `XDEL`, so each processes it. `WakeupDispatcher._drain_and_dispatch`'s `session_is_running` check + spawn is a TOCTOU race; the per-session `session_run` lock only *serializes* the duplicates, it does not prevent them.

### Defect B — empty wake-up still invokes the LLM
In `ChatService.run` (`src/agentscope/app/_service/_chat.py`), the only guard for `input_msg is None` is the "agent parked on an awaiting tool call" case; otherwise it unconditionally calls `agent.reply_stream(inputs=None)`. After the first run drains the inbox, the serialized duplicates run against an empty inbox and emit the spurious "your message was empty" replies.

## Impact
Scheduled tasks, team messages, and background-tool result deliveries can each execute multiple times in any multi-instance deployment: wasted LLM calls/tokens, polluted session history, and **incorrect behavior for non-idempotent agents** (a scheduled action runs N times). Severity scales with instance count.

## Proposed solution
**Primary — idempotent empty wake-up no-op (fixes B):** in `ChatService.run`, when `input_msg is None` and the agent is not parked on an awaiting tool call, return early when there is nothing to deliver (inbox empty). This embraces at-least-once delivery + idempotent consumers, so duplicate wake-ups become harmless regardless of delivery semantics. Needs a cheap `inbox_size()` / peek on `MessageBus` (only `inbox_push` / `inbox_drain` exist today).

**Hardening — exactly-once wake-up delivery (fixes A):** switch the wake-up queue consumption from `XRANGE`+`XDEL` to a Redis Streams **consumer group** (`XREADGROUP GROUP … STREAMS key >` + `XACK`/`XDEL`), with all dispatchers in one group and a periodic `XAUTOCLAIM` to recover entries from crashed consumers, so each entry is delivered to exactly one instance.

B is the robust foundation (consumer-group redelivery on crash still needs idempotency); A removes the wasted duplicate runs. Both together is ideal.

## Acceptance criteria
- With N instances on one Redis, a single wake-up yields **exactly one** run that produces output; no "your message was empty" replies.
- No wake-up is lost when an instance crashes mid-processing.

### Error Messages

```shell
One cron fire (non-stateful schedule, cron `* * * * *`) produced ONE session containing 4 assistant messages within ~6s:


[0] assistant content=[hint <scheduled-task>…</scheduled-task>, text "我是一个智能助手…"]   <- correct
[1] assistant content=[text "It looks like your message didn't include a specific request…"]
[2] assistant content=[text "It looks like your message didn't include a specific request…"]
[3] assistant content=[text "It looks like your message was empty…"]


Evidence it is wake-up fan-out, not the scheduler:
- Scheduler logs exactly ONE `Trigger fired` + ONE `Wakeup enqueued` per cron tick.
- `PUBSUB NUMSUB agentscope:wakeup_signal` = 5 (five live WakeupDispatchers).
- Redis `CLIENT LIST` shows 3 distinct host IPs sharing the same Redis.
- Reproduces with a single local process too (other instances on the shared Redis still race); unrelated to `--reload`.
```

### Steps to Reproduce

1. Run >=2 app instances built via `create_app(...)` with `RedisMessageBus` + `PostgresStorage`, all pointing at the **same** Redis + DB (e.g. multiple pods, or multiple machines sharing config).
2. Create a non-stateful cron `Schedule` that fires every minute with a trivial prompt:
   - `description = "introduce yourself in one sentence"`
   - `cron_expression = "* * * * *"`
   - `stateful = False`
3. Wait for one fire, then inspect the session it created (`GET /schedule/{id}/sessions`, then read that session's messages).
4. Observe N assistant messages where N ≈ number of live `WakeupDispatcher`s: one real reply, the rest "your message was empty".

### Environment

- AgentScope Version: `main` @ commit 43ce70b2 (== origin/main)
- Python Version: 3.12
- OS: macOS (dev) + Linux (prod); reproduced with >=2 app instances sharing one Redis + PostgreSQL

Premsenareddy:
## Summary
- pre-drain inbox content for wake-up driven ChatService runs
- skip agent execution when a wake-up finds no pending inbox entries
- add regression coverage for empty and non-empty wake-up runs

## Tests
- .venv/bin/python -m pytest -p no:cacheprovider tests/service_wakeup_dispatcher_test.py tests/service_inbox_middleware_test.py tests/service_chat_wakeup_test.py -q

Addresses #1868


Premsenareddy:
Refreshed this PR against the latest `main` and resolved the service/message-bus conflicts.

The wake-up guard now:
- drains the session inbox inside the current distributed session lock;
- skips the agent run when a duplicate wake-up has no deliverable inbox entries;
- publishes non-empty `HintBlockEvent`s through the current replay/event-projector path before reasoning;
- keeps normal `InboxMiddleware` behavior for user messages and continuation runs.

Validation:
- focused wake-up/inbox/dispatcher slice: 16 passed
- broader service/team/scheduler/offload slice: 98 passed
- file-scoped pre-commit: all hooks passed

GitHub now reports the PR as mergeable; refreshed CI is running.

DavdGao:
Thanks for the patch. We ended up fixing this at the root in #2476 — `queue_drain` now runs the XRANGE+XDEL as one Lua script, so the duplicate wake-ups that caused the empty runs don't happen in the first place and the consumer-side guard isn't needed. Closing this one.

PR #1873, opened 2026-06-14
fix: prevent streaming JSON dict regression on incomplete chunks (#1563)

### LINKED ISSUE #1563: [Bug]:RealtimeAgent tool call path allows remote users to trigger server-side Python execution when execute_python_code is exposed
**Describe the bug**  
When an AgentScope RealtimeAgent session exposes `execute_python_code` as an available tool, a remote WebSocket user can prompt the agent to call that tool and run Python code in the service environment. In the validated path, `RealtimeAgent._acting` forwards the model-produced tool call to `Toolkit.call_tool_function`, which invokes `execute_python_code` without an additional approval or isolation boundary on that path.

**To Reproduce**  

1. Use AgentScope `1.0.19.post1` and start the realtime voice agent service with `execute_python_code` included in the session tool list.

2. Connect a WebSocket client to the realtime endpoint, for example:

```text
ws://127.0.0.1:8010/ws/test-user/<session_id>
```

3. Create a realtime session and send a user message asking the agent to call `execute_python_code` once with a marker-writing Python snippet:

```python
from pathlib import Path
Path(r"<reviewer-controlled-temp-dir>\agentscope-realtime-marker.txt").write_text(
    "AGENTSCOPE_REALTIME_REPRO",
    encoding="utf-8",
)
print("done")
```

4. Observe that the service forwards the tool call through:

```text
src/agentscope/agent/_realtime_agent.py:RealtimeAgent._acting
src/agentscope/tool/_toolkit.py:Toolkit.call_tool_function
src/agentscope/tool/_coding/_python.py:execute_python_code
```

5. Check the marker file on disk.

**Expected behavior**  
A remote realtime user should not be able to cause server-side Python code to run unless an explicit approval, sandbox, or equivalent containment boundary is enforced before the tool invocation.

**Error messages**  
No exception is required to reproduce the issue. In the validated run, the tool result returned successfully with return code `0` and stdout containing:

```text
done
```

The marker file was created with:

```text
AGENTSCOPE_REALTIME_REPRO
```

- AgentScope Version: `1.0.19.post1`; also source-continuity checked in `1.0.18` and `1.0.19`
- Python Version: Python `3.12` in the local validation environment
- OS: Windows

**Additional context**  
This report does not claim that every RealtimeAgent deployment is affected by default. The confirmed affected condition is a RealtimeAgent WebSocket service/session where `execute_python_code` is registered or exposed to model tool calls without an additional approval, sandbox, or containment boundary.

nguyenthanhthe:
## Description

When the LLM generates tool calls with long string parameters (e.g., `write_file` with ~33KB of content), the streaming JSON parser sometimes receives incomplete chunks where a string value is truncated mid-token. The existing `_parse_streaming_json_dict` function falls back to `json_repair`, but when repair produces an empty dict `{}`, the accumulated state from previous successful chunks is lost.

This causes `write_file` to write only ~6KB instead of the full 33KB — the `content` parameter silently becomes an empty or incomplete dict.

**Root cause:** In `_parse_streaming_json_dict`, when `_json_loads_with_repair` fails and returns `{}`, and `last_input` is also `{}` (both empty), the regression guard `len(json.dumps(last_input)) > len(json.dumps(repaired_input))` evaluates to `2 > 2 = False`, so the function returns the empty repaired dict instead of falling back.

**Fix:** Added an explicit check that if `repaired_input` is empty/falsy but we had accumulated state from previous chunks, keep the previous state. Also added clearer comments documenting the regression-prevention logic.

**Related Issue:** Fixes #1563

**Security Considerations:** None — this is a data integrity fix for streaming JSON parsing.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [ ] Refactoring

## Component(s) Affected

- [x] Core (agentscope._utils._common)
- [ ] Agent
- [ ] Model
- [ ] Memory
- [ ] Tool
- [ ] Formatter
- [ ] MCP

## Checklist

- [ ] I ran `pre-commit run --all-files` locally and it passes
- [ ] If pre-commit auto-fixed files, I committed those changes and reran checks
- [ ] I ran tests locally and they pass
- [ ] Documentation updated (if needed)
- [x] Ready for review

## Testing

1. Create an agent equipped with `write_file` tool
2. Ask it to write a ~30KB+ file using `write_file`
3. Verify the full content is written instead of being truncated


nguyenthanhthe:
Closing this PR — it was created from a divergent fork and bundles unrelated changes alongside the streaming JSON fix. Replacing with a clean PR from upstream/main.

PR #1876, opened 2026-06-15
fix(dashscope): correct qwen max 3.7 model id

Rise1024:
## AgentScope Version

2.0.1

## Description

DashScope's OpenAI-compatible model list exposes Qwen Max 3.7 as `qwen3.7-max`. The previous bundled model card used `qwen-max-3.7`, which causes DashScope API calls selected from `DashScopeChatModel.list_models()` to fail with `model_not_found`.

This PR updates the DashScope Qwen Max 3.7 model card to use the correct model ID, renames the bundled YAML file to match the model ID, and adds a regression test to ensure the invalid ID is no longer exposed.

Changes made:
- Rename `src/agentscope/model/_dashscope/_models/qwen-max-3.7.yaml` to `src/agentscope/model/_dashscope/_models/qwen3.7-max.yaml`.
- Change the model card `name` from `qwen-max-3.7` to `qwen3.7-max`.
- Add a regression test for `DashScopeChatModel.list_models()`.

Tested with:

```bash
.venv/bin/python -m pytest tests/model_dashscope_test.py -q

### DIFF
--- a/tests/model_dashscope_test.py
+++ b/tests/model_dashscope_test.py
@@ -27,18 +27,6 @@ from agentscope.tool import ToolChoice
 A = AnyString()
 
 
-class TestDashScopeListModels(unittest.TestCase):
-    """Tests for DashScope chat model cards."""
-
-    def test_qwen_max_37_uses_dashscope_model_id(self) -> None:
-        """Qwen Max 3.7 should use the DashScope-compatible model ID."""
-        cards = DashScopeChatModel.list_models()
-        names = {card.name for card in cards}
-
-        self.assertIn("qwen3.7-max", names)
-        self.assertNotIn("qwen-max-3.7", names)
-
-
 # ---------------------------------------------------------------------------
 # Helpers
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/model_dashscope_test.py [hunk 1]
```diff
@@ -27,6 +27,18 @@
 A = AnyString()
 
 
+class TestDashScopeListModels(unittest.TestCase):
```
qbc2016: Please remove this.

qbc2016 (commented):
Thank you for the fix. 
Please remove the unittest `TestDashScopeListModels`, so we can merge it.

tests/model_dashscope_test.py
Rise1024: Removed TestDashScopeListModels and kept the model ID fix.

Rise1024:
Done

qbc2016 (approved):
LGTM.

PR #1883, opened 2026-06-16
fix(model): handle Gemini function calls without an id

he-yufeng:
## AgentScope Version

2.0.2 (current `main`)

## Description

Gemini does not always return an `id` on a function call. On the non-thinking
models there is no `thought_signature` to fall back on either, so the parser
ends up passing `id=None` into `ToolCallBlock`, whose `id` is a required `str`.
That raises:

```
pydantic_core.ValidationError: 1 validation error for ToolCallBlock
id
  Input should be a valid string [type=string_type, input_value=None]
```

which kills the whole tool-calling turn. It hits both code paths:

- `_parse_stream_response` (`src/agentscope/model/_gemini/_model.py:350`)
- `_parse_completion_response` (`src/agentscope/model/_gemini/_model.py:434`)

Fix is to mint a fallback id when neither the function call id nor a thought
signature is present:

```python
call_id = part.function_call.id or uuid.uuid4().hex
```

`uuid` is already imported, and the accumulated text/thinking blocks already
generate their ids the same way. The formatter round-trips `block.id` as both
`function_call.id` and `function_response.id`
(`src/agentscope/formatter/_gemini_formatter.py:200,224`), so a synthetic id
stays consistent across the tool-call / tool-result pair.

## How I tested

Added two tests to `tests/model_gemini_test.py`:

- non-streaming: an id-less function call now returns a `ToolCallBlock` with a
  generated id instead of crashing
- streaming: two parallel id-less calls in one chunk get distinct ids
  (previously they both crashed, and would have collided on the `None`
  accumulator key)

Both fail on current `main` with the ValidationError above and pass with the
fix. Full file stays green:

```
python -m pytest tests/model_gemini_test.py -q
14 passed
```

`black --line-length=79` and `flake8` are clean on both files.

## Checklist

Used Claude Code to assist; I reviewed the diff and ran the tests myself.

### DIFF
--- a/src/agentscope/model/_gemini/_model.py
+++ b/src/agentscope/model/_gemini/_model.py
@@ -347,7 +347,7 @@ class GeminiChatModel(ChatModelBase):
                                 part.thought_signature,
                             ).decode("utf-8")
                         else:
-                            call_id = part.function_call.id
+                            call_id = part.function_call.id or uuid.uuid4().hex
                         input_str = json.dumps(
                             keyword_args,
                             ensure_ascii=False,
@@ -431,7 +431,7 @@ class GeminiChatModel(ChatModelBase):
                             part.thought_signature,
                         ).decode("utf-8")
                     else:
-                        call_id = part.function_call.id
+                        call_id = part.function_call.id or uuid.uuid4().hex
                     content_blocks.append(
                         ToolCallBlock(
                             id=call_id,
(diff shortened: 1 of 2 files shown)


qbc2016 (approved):
LGTM.

DavdGao (approved):

PR #1894, opened 2026-06-18
fix(formatter): merge parallel tool_results into one user message in AnthropicChatFormatter

### LINKED ISSUE #1892: [Bug]: AnthropicChatFormatter splits parallel tool_results into multiple user messages -> 400 on DeepSeek
### Prerequisites

### Background / Description

Using agentscope with DeepSeek (Anthropic-compatible endpoint). When the model returns several tool_use blocks in one reply, the next call fails with 400.

AnthropicChatFormatter splits the tool_results into multiple user messages, but they should be in ONE user message right after the assistant message. DeepSeek checks this strictly and rejects it. Anthropic/GLM endpoints are lenient so it stays hidden.

Root cause: in agentscope/formatter/_anthropic_formatter.py, the ToolResultBlock branch flushes whenever content_blocks is non-empty, so each tool_result becomes its own user message. One-line fix works (see Steps to Reproduce).

### Error Messages

```shell
anthropic.BadRequestError: 400 - {
  'error': {
    'message': "messages.7:`tool_use` ids were found without `tool_result` blocks immediately after: call_00_..., call_01_..., call_02_..., call_04_.... Each `tool_use` block must have a corresponding `tool_result` block in the next message.",
    'type': 'invalid_request_error'
  }
}

messages.7 = assistant turn with 5 tool_use blocks. The 5 tool_results ran fine but were sent as separate user messages.
```

### Steps to Reproduce

Triggers when the model returns 2+ tool_use blocks in one reply. agentscope stores them + their tool_results in the same AssistantMsg.content:

[text, tool_use x3, tool_result x3]

Run the formatter -> you get 1 assistant + 3 user messages (one per tool_result), instead of 1 assistant + 1 user. DeepSeek rejects it.

Setting tools to sequential (is_concurrency_safe=False) does NOT help — _save_to_context still merges all tool_results into one assistant message.

Suggested fix in agentscope/formatter/_anthropic_formatter.py, ToolResultBlock branch:

  - if content_blocks:
  + if content_blocks and not has_tool_result:

First tool_result still flushes the assistant content (unchanged); following tool_results then merge into one user message. Single-tool output identical before/after.

### Environment

- AgentScope Version: 2.0.0
- Python Version: 3.11
- OS: macOS (darwin, arm64)
- Model: DeepSeek deepseek-v4-pro via Anthropic-compatible endpoint

qbc2016:
## AgentScope Version

2.0.2

## Description

When the model returns multiple tool_use blocks in one turn, AnthropicChatFormatter was flushing content_blocks on every 
ToolResultBlock, splitting N parallel results into N separate  user messages. Strict Anthropic-compatible endpoints (e.g. DeepSeek) reject this with HTTP 400.

Fix: only flush the preceding assistant content on the *first* ToolResultBlock (`if content_blocks and not has_tool_result`), 
so all results accumulate into a single user message as the spec requires.

- Fix flush condition in `_anthropic_formatter.py`
- Update `test_chat_formatter_complex_multi_step` ground truth
- Add regression test `test_chat_formatter_parallel_tool_results_merged`

Fixes #1892

## Checklist

### DIFF
--- a/tests/formatter_anthropic_test.py
+++ b/tests/formatter_anthropic_test.py
@@ -666,11 +666,6 @@ class TestAnthropicFormatter(IsolatedAsyncioTestCase):
                                 {"type": "text", "text": "result_1"},
                             ],
                         },
-                    ],
-                },
-                {
-                    "role": "user",
-                    "content": [
                         {
                             "type": "tool_result",
                             "tool_use_id": "call_2",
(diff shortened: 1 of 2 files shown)


tests/formatter_anthropic_test.py
```diff
@@ -830,3 +825,139 @@ async def test_chat_formatter_hint_block_multimodal(self) -> None:
...
+                        input='{"city": "Shanghai"}',
+                    ),
+                    ToolCallBlock(
+                        id="call_03",
+                        name="get_weather",
+                        input='{"city": "Guangzhou"}',
+                    ),
+                ],
+            ),
+            AssistantMsg(
```
DavdGao: The `ToolResultBlock` should be within the same `AssistantMsg` with the corresponding `ToolCallBlock`s, rather than splitted into two Msg objects

DavdGao (changes requested):
Plz see inline comments

DavdGao (approved):

PR #1899, opened 2026-06-19
fix(model): use flat estimate for multimodal data blocks

dongfeng3692:
## AgentScope Version

2.0.2

## Description

This PR fixes the fallback token estimation for multimodal `DataBlock` inputs in `ChatModelBase.count_tokens`.

Previously, base64 inputs were estimated from the raw base64 string length, while URL inputs were estimated from the URL text itself. This could produce very large overestimates for inline base64 image/audio inputs, and inconsistent estimates between equivalent `Base64Source` and `URLSource` payloads.

After checking how several agent projects (including Claude Code, Hermes, Pi) handle local multimodal token estimation, this PR uses a conservative flat estimate for each `DataBlock`. This follows the common fallback pattern of treating binary multimodal payloads as provider-side media inputs rather than prompt text. The selected value, `2000` tokens per `DataBlock`, is intentionally conservative, so it avoids underestimating context usage while preventing base64 payloads from dominating the estimate.

Changes made:

Updated multimodal `DataBlock` token estimation to use a conservative flat estimate, and added tests for base64 and file URL inputs.

### DIFF
--- a/src/agentscope/model/_base.py
+++ b/src/agentscope/model/_base.py
@@ -24,13 +24,12 @@ from ..message import (
     ThinkingBlock,
     ToolResultBlock,
     DataBlock,
-    URLSource,
-    Base64Source,
     HintBlock,
 )
 from ..tool import ToolChoice
 
 _TOOL_CHOICE_LITERAL_MODES = {"auto", "none", "required"}
+_MULTIMODAL_DATA_BLOCK_TOKEN_ESTIMATE = 2000
 
 
 class ChatModelBase:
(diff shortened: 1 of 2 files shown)


dongfeng3692:
@qbc2016 Could you take a look? The multimodal input flow is currently blocked.
This token size estimation error makes the program incorrectly believe the context is full, triggering unnecessary compression. In reality, file contents do not consume context tokens—they are calculated independently and typically remain under 2000. Other agents also use an estimated reference value of around 1000.

DavdGao (approved):
LGTM

Issue #956, opened 2025-11-24
[Bug]: ToolResponse.metadata not preserved when storing tool_result in memory


Adoubf:
**Describe the bug**  
When a tool function returns a `ToolResponse(..., metadata=some_dict)`, the `metadata` field is not preserved in the stored `Msg` within memory. In my use‐case I expected the `Msg.metadata` to equal the tool response `metadata`, but found it `None`.

**To Reproduce**  
1. Define a tool function that returns e.g.  
     ```python
     from agentscope.tool import ToolResponse
     from agentscope.message import TextBlock
  
     def my_tool(...):
         return ToolResponse(
             content=[TextBlock(text="Result text")],
             metadata={"key": "value", "other": 123}
         )
     ````
2. Register the tool via `Toolkit.register_tool_function(my_tool, ...)`
3. Use the tool in a ReActAgent call, e.g. tool call:

   ```json
   {"type": "tool_use", "name": "my_tool", "input": {...}}
   ```
4. After execution, inspect agent memory. Find the corresponding `Msg` with `"type": "tool_result"`; verify `msg.metadata` is `None`.

**Expected behavior**
The `msg.metadata` associated with the tool result should contain the same dictionary passed as `ToolResponse.metadata`.

**Environment**

* AgentScope version: 1.0.7
* Python version: 3.12.10
* OS: Ubuntu 22.04

**Additional context**
This is important for scenarios where structured data from tool invocation must be stored/referenced downstream (e.g., audit logs, database persistence) without additional parsing of `content`. Without `metadata`, one must parse JSON from `content.text`, increasing risk of parse errors, inconsistency, and extra processing.

---

**Possible workaround**
As a temporary workaround, I serialize structured result into `content.text` (e.g., via `json.dumps(metadata)`) and then parse it downstream. But this is not ideal and violates the design intent of `metadata`.

**Would like to propose**
Either preserve `ToolResponse.metadata` into `Msg.metadata`, or explicitly document that metadata will not be stored and provide a recommended approach for structured tool outputs (e.g., specialized `JSONBlock`).

PR #951, opened 2025-11-21
add memory with compression

Osier-Yi:
## AgentScope Version

[The version of AgentScope you are working on, e.g. `import agentscope; print(agentscope.__version__)`]

## Description

[Please describe the background, purpose, changes made, and how to test this PR]

## Checklist

Please check the following items before code is ready to be reviewed.

- [x]  Code has been formatted with `pre-commit run --all-files` command
- [x]  All tests are passing
- [x]  Docstrings are in Google style
- [x]  Related documentation has been updated (e.g. links, examples, etc.)
- [x]  Code is ready for review

PR #1905, opened 2026-06-19
fix(agent): prevent orphan tool_calls from bricking sessions

### LINKED ISSUE #1888: [Bug]: Mid-stream failures can leave orphan tool_calls and brick sessions
### Describe the bug

In AgentScope Python 2.0.2, a chat run can fail after an assistant tool call has been appended to `AgentState.context`, but before the matching `ToolResultBlock` is recorded and persisted.

The persisted session can then contain an assistant message with `tool_calls` but no matching tool result. On the next turn, OpenAI-compatible providers reject the conversation history with an error like:

```text
An assistant message with "tool_calls" must be followed by tool messages responding to each "tool_call_id"
```

After that, every retry of the same session can fail again, so the session is effectively bricked until the stored context is manually repaired or deleted.

### Expected behavior

If a run is interrupted or crashes while a tool call is pending, AgentScope should persist a valid recovery state before the exception escapes. For example, the pending tool call could be closed with an interrupted/error `ToolResultBlock`, so the next run does not send invalid tool-call history to the model provider.

### Actual behavior

A mid-stream failure can leave an orphan tool call in persisted session state. The next run reuses that invalid context and fails at provider validation before the agent can recover naturally.

### Related issues / prior art

This appears related to earlier Python reports of provider-side tool-call validation errors:

- https://github.com/agentscope-ai/agentscope/issues/772
- https://github.com/agentscope-ai/agentscope/issues/815
- https://github.com/agentscope-ai/agentscope/issues/1711

The Java implementation also had a very similar pending-tool-call recovery problem:

- https://github.com/agentscope-ai/agentscope-java/issues/951
- https://github.com/agentscope-ai/agentscope-java/pull/956

The Java fix added a `PendingToolRecoveryHook` that synthesizes error tool results for pending tool calls. Python may need an equivalent recovery point in the app/chat service or agent middleware layer.

### Suggested fix

Add a Python-side recovery mechanism for persisted agent state:

1. Before a run, scan `AgentState.context` for assistant tool calls that do not have matching tool results, and synthesize interrupted/error `ToolResultBlock`s.
2. During a run, wrap the streaming/chat execution path in `try` / `finally` or equivalent middleware so pending tool calls are repaired and persisted when an exception occurs.
3. Add regression tests that simulate a failure after a tool call is emitted but before its tool result is persisted, then verify the next run does not fail with provider validation errors.

This would avoid downstream applications needing to monkey-patch private `ChatService` internals to keep sessions recoverable.

### Environment

- AgentScope Python: 2.0.2
- Python: 3.12
- Storage: persisted app session state
- Providers: OpenAI-compatible tool-call validation

nexiouscaliver:
## Summary

Fixes a session-bricking bug where a mid-stream failure can leave an assistant tool call in `AgentState.context` without its matching tool result. On the next run the provider rejects the unpaired tool call, and every retry of that session fails until the stored context is repaired or deleted by hand.

Closes #1888.

## Root cause

When the model produces a tool call, the `ToolCallBlock` is appended to `self.state.context` at the end of `_reasoning_impl` (`_agent.py`) **before** the tool runs. The matching `ToolResultBlock` is only appended later in `_execute_tool_call`, after the tool executes. If the run is interrupted in that window, the context is left with a tool call that has no result.

On the next turn a new user message lands after the orphan, so `_check_next_action` (which only inspects the last message) sees nothing to act on and chooses "reasoning". `_prepare_model_input` then extends the model input with the whole context, sending the unpaired tool call to the provider:

```
An assistant message with "tool_calls" must be followed by tool messages responding to each "tool_call_id"
```

The session is bricked until the stored context is repaired or removed.

Note that `toolkit.call_tool` already wraps tool execution in a `try` that turns exceptions (including `CancelledError`) into error/interrupted results, so an exception raised *inside* a tool does not produce an orphan. Orphans come from failures **outside** `call_tool`, for example a process crash or interruption, or an `on_acting` middleware that raises between the context write and the result write. This is the same class of problem fixed in agentscope-java by `PendingToolRecoveryHook` (#951 / #956).

## The fix

Add `Agent._recover_interrupted_tool_calls()`, the Python counterpart of `PendingToolRecoveryHook`. Before a new run reaches the model, it scans the context and, for every tool call that still has no matching tool result, synthesizes an `interrupted` `ToolResultBlock` attached to the same message that holds the orphan tool call (mirroring the structure `_execute_tool_call` produces) and marks the tool call `finished`.

It is called once at the start of a fresh reply, in the new-user-message branch of `_reply_impl`, and **not** inside `_prepare_model_input`. This placement is what makes it safe:

- At the start of a fresh reply every tool call in the context belongs to a previous, already-finished run, so any that still lack a result are genuine orphans.
- Tool calls legitimately awaiting an outside event (`asking` / `submitted`) never reach this point: a fresh message while the agent is awaiting is rejected by `_check_incoming_event` before recovery runs, and an event-driven resume takes the other branch.
- It does not run during `compress_context` (which calls `_prepare_model_input` internally), so it cannot interfere with context compression or the reserved read cache.

This matches the recovery behavior requested in the issue: the pending tool call is closed with an interrupted result so the next run never sends invalid tool-call history to the model provider.

## Tests

Adds `tests/orphan_tool_call_recovery_test.py`:

- `test_orphan_recovered_after_in_run_acting_failure` reproduces the real orphan window with an `on_acting` middleware that raises (which, unlike an exception inside a tool, is not converted into a result), then asserts the next run recovers instead of bricking.
- `test_recovery_synthesizes_result_in_original_message` injects a persisted orphan directly and asserts the interrupted result is attached to the original message, the tool call is marked finished, and no tool call anywhere in the context lacks a matching result.

Both tests fail on `main` and pass with this change. Locally, `coverage run -m pytest tests` reports 956 passed / 0 failed / 1 skipped (the e2b credential skip) and `pre-commit run --all-files` is clean.

## Notes

- No new dependencies; no public API changes.
- The synthesized result is intentionally `interrupted` (not `error`) since the tool did not fail, it was simply never completed.


nexiouscaliver:
Hi @DavdGao @qbc2016, tagging you both since you maintain the agent module.

This fixes #1888, where a mid-stream failure can leave an orphan tool_call in `AgentState.context` (the call is appended before its result is produced) and brick the session on the next run when the provider rejects the unpaired tool_call. The change is small and focused: at the start of a fresh reply, any tool call that still has no result is closed with an `interrupted` result before the context reaches the model, mirroring the `PendingToolRecoveryHook` fix from agentscope-java. I kept it out of `_prepare_model_input` on purpose so it does not interfere with context compression or the reserved read cache, and I added regression tests that fail on `main` and pass with the change.

It is ready for review whenever you have time, and I am happy to adjust based on your feedback. Thanks for taking a look.

PR #1911, opened 2026-06-21
fix: recover orphan tool calls on mid-stream failure (#1888)

michaelxer:
## Problem

When tool execution fails after an assistant message with `tool_calls` is saved to context but before matching tool results are recorded, the next conversation turn sends invalid history to the model provider, causing validation errors.

## Fix

- Wraps the tool execution loop in a try/except block
- On exception, calls `_recover_orphan_tool_calls()` to synthesize error `ToolResultBlock` with `INTERRUPTED` state for any tool calls missing results
- Ensures context remains valid for the next conversation turn

## Changes

- `src/agentscope/agent/_agent.py` — Core fix: try/except around tool execution + new `_recover_orphan_tool_calls()` method
- `tests/test_orphan_tool_call_recovery.py` — 3 test cases: normal execution, single failure, multi-tool-call scenarios

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -624,46 +624,63 @@ class Agent:
             #  - If not, finish the loop by yielding RunFinishedEvent and exit
             #  - Otherwise, execute by batch and continue the loop
             # ===============================================================
-            for batch in await self._batch_tool_calls():
-                if batch.type == "sequential":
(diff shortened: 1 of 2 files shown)


michaelxer:
Closing — CI workflow approval hasn't come through. Happy to reopen if the project wants to enable external contributor CI.

PR #1917, opened 2026-06-22
fix(app): convert AG-UI SSE stream events

### LINKED ISSUE #1913: [Bug]: AGUIProtocolMiddleware does not work
### Prerequisites

### Background / Description

Looking forward to the transition from agentscope Events to ag-ui Events

### Error Messages

```shell
AgentScope Events have not been converted into ag-ui Events
```

### Steps to Reproduce

1. Configure AGUIProtocolMiddleware to start the agent_service under the examples module
```create_app(...,extra_middlewares=[
        Middleware(
            CORSMiddleware,
            allow_origins=["*"],
            allow_methods=["*"],
            allow_headers=["*"],
        ),
        Middleware(
            AGUIProtocolMiddleware
        ))
```
2. Request /sessions/{id}/stream
3. Observation The output remains in the native format of AgentScope, without being converted to the AG-UI protocol

### Environment

- AgentScope Version: 2.0.2
- Python Version: 3.12.13
- OS: Linux 6.6.142-amd64-desktop-hwe
- FastAPI: 0.137.0

YingchaoX:
## AgentScope Version

2.0.3

## Description

This PR fixes #1913.

`AGUIProtocolMiddleware` could convert raw AgentScope event JSON into AG-UI events, but it did not work for the session SSE endpoint. `/sessions/{session_id}/stream` sends events as Server-Sent Events frames, for example:

```text
data: {"type": "REPLY_START", ...}
```

The protocol middleware previously attempted to parse the whole chunk as JSON, so SSE-framed events failed deserialization and were passed through unchanged.

This PR updates the protocol middleware to:

- convert AgentScope event JSON inside `text/event-stream` `data:` frames
- preserve the outer SSE frame format
- pass heartbeat/comment frames through unchanged
- keep raw JSON stream conversion compatible
- avoid converting non-SSE JSON responses

Tests were added for raw JSON conversion, SSE data-frame conversion, heartbeat passthrough, FastAPI SSE integration, and non-SSE JSON response passthrough.

Validation run locally:

```bash
uv run --python 3.12 --extra dev pre-commit run --all-files
uv run --python 3.12 --extra dev pytest tests
```

Result:

```text
987 passed, 26 skipped, 1 warning
```

## Checklist

- [x] I have read the [[CONTRIBUTING.md](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md)](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md)
- [x] Related documentation has been updated (e.g. links, examples, etc.) in [[documentation repository](https://github.com/agentscope-ai/docs)](https://github.com/agentscope-ai/docs)

### DIFF
--- a/src/agentscope/app/middleware/_protocol/_base.py
+++ b/src/agentscope/app/middleware/_protocol/_base.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """Protocol middleware base class for converting AgentEvent stream to
 various protocols."""
+
 import json
 from abc import ABC, abstractmethod
 from typing import AsyncGenerator, Callable
@@ -61,11 +62,15 @@ class ProtocolMiddlewareBase(BaseHTTPMiddleware, ABC):
         # Call the next middleware or endpoint
         response = await call_next(request)
 
-        # Check if the response is a streaming response
-        if isinstance(response, StreamingResponse):
+        content_type = response.headers.get("content-type", "")
+        body_iterator = getattr(response, "body_iterator", None)
+
+        if (
+            content_type.startswith("text/event-stream")
+            and body_iterator is not None
+        ):
             # Wrap the original stream with our conversion logic
-            original_stream = response.body_iterator
-            converted_stream = self._convert_stream(original_stream)
+            converted_stream = self._convert_stream(body_iterator)
 
             # Create a new StreamingResponse with the converted stream
             return StreamingResponse(
@@ -91,36 +96,95 @@ class ProtocolMiddlewareBase(BaseHTTPMiddleware, ABC):
             Bytes in protocol format.
         """
         async for chunk in original_stream:
-            # Decode the chunk if it's bytes
             if isinstance(chunk, bytes):
                 chunk_str = chunk.decode("utf-8")
             else:
                 chunk_str = chunk
 
-            # Try to deserialize the chunk as AgentEvent
-            try:
-                # Parse the JSON string to dict
-                event_dict = json.loads(chunk_str)
+            converted = self._convert_sse_frame(chunk_str)
+            if converted is not None:
+                yield converted
+                continue
 
-                # Deserialize to AgentEvent based on the 'type' field
-                agent_event = self._deserialize_event(event_dict)
+            converted = self._convert_event_json(chunk_str)
+            if converted is not None:
+                yield converted
+                continue
 
-                # Convert AgentEvent to protocol format
-                protocol_data = self._convert_to_protocol(agent_event)
+            if isinstance(chunk, bytes):
+                yield chunk
+            else:
+                yield chunk.encode("utf-8")
 
-                # Serialize and yield the protocol data
-                yield json.dumps(protocol_data, ensure_ascii=False).encode(
+    def _convert_sse_frame(self, frame: str) -> bytes | None:
+        """Convert AgentEvent payloads inside an SSE frame.
+
+        Args:
+            frame: A server-sent event frame.
+
+        Returns:
+            Converted frame bytes if at least one ``data:`` payload was
+            converted, otherwise ``None``.
+        """
+        lines = frame.splitlines(keepends=True)
+        converted_lines: list[str] = []
+        converted_any = False
+
+        for line in lines:
+            if not line.startswith("data:"):
+                converted_lines.append(line)
+                continue
+
+            line_content, line_ending = self._split_line_ending(line)
+            payload = line_content.removeprefix("data:")
+            if payload.startswith(" "):
+                payload = payload[1:]
+
+            converted = self._convert_event_json(payload)
+            if converted is None:
+                converted_lines.append(line)
+                continue
+
+            converted_json = converted.decode("utf-8").rstrip("\n")
+            converted_lines.append(f"data: {converted_json}{line_ending}")
+            converted_any = True
+
+        if not converted_any:
+            return None
+
+        return "".join(converted_lines).encode("utf-8")
+
+    @staticmethod
+    def _split_line_ending(line: str) -> tuple[str, str]:
+        """Split a line into content and its original line ending."""
+        if line.endswith("\r\n"):
+            return line[:-2], "\r\n"
+        if line.endswith("\n"):
+            return line[:-1], "\n"
+        return line, ""
+
+    def _convert_event_json(self, chunk_str: str) -> bytes | None:
+        """Convert a serialized AgentEvent JSON string.
+
+        Args:
+            chunk_str: Serialized AgentEvent JSON.
+
+        Returns:
+            Converted protocol JSON bytes with trailing newline, or ``None``
+            when ``chunk_str`` is not a valid AgentEvent payload.
+        """
+        try:
+            event_dict = json.loads(chunk_str)
+            agent_event = self._deserialize_event(event_dict)
+            protocol_data = self._convert_to_protocol(agent_event)
+            return (
+                json.dumps(protocol_data, ensure_ascii=False).encode(
                     "utf-8",
-                ) + b"\n"
-
-            except (json.JSONDecodeError, KeyError, ValueError):
-                # If deserialization fails, pass through the original chunk
-                # or log the error
-                # For now, we'll pass through the original chunk
-                if isinstance(chunk, bytes):
-                    yield chunk
-                else:
-                    yield chunk.encode("utf-8")
+                )
+                + b"\n"
+            )
+        except (json.JSONDecodeError, KeyError, TypeError, ValueError):
+            return None
 
     def _deserialize_event(self, event_dict: dict) -> AgentEvent:
         """Deserialize event dictionary to AgentEvent object.
--- a/src/agentscope/app/middleware/_protocol/_base.py
+++ b/src/agentscope/app/middleware/_protocol/_base.py
@@ -17,9 +17,9 @@ from agentscope.event import AgentEvent
 class ProtocolMiddlewareBase(BaseHTTPMiddleware, ABC):
     """Base middleware for converting AgentEvent stream to protocol format.
 
-    This middleware intercepts streaming responses that yield AgentEvent
-    objects, deserializes them, and converts them to a specific protocol
-    format.
+    This middleware intercepts ``text/event-stream`` responses, deserializes
+    AgentEvent objects from SSE ``data:`` frames, and converts them to a
+    specific protocol format.
 
     Subclasses should implement the `_convert_to_protocol` method to define
     the conversion logic for their specific protocol (e.g., AGUI, A2A).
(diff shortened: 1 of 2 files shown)


src/agentscope/app/middleware/_protocol/_base.py
```diff
@@ -91,36 +96,95 @@ async def _convert_stream(
...
+        return "".join(converted_lines).encode("utf-8")
+
+    @staticmethod
+    def _split_line_ending(line: str) -> tuple[str, str]:
+        """Split a line into content and its original line ending."""
+        if line.endswith("\r\n"):
+            return line[:-2], "\r\n"
+        if line.endswith("\n"):
+            return line[:-1], "\n"
+        return line, ""
```
XiuShenAl, line 175: Nice helper! One small gap: `str.splitlines(keepends=True)` (used by `_convert_sse_frame`) recognizes `\r` as a valid line ending in addition to `\n` and `\r\n`. If a line happens to end with a bare `\r`, this method will fall through to `return line, ""`, leaving the `\r` inside the content — which could cause a subtle JSON parse failure downstream.

The SSE spec also treats `\r` as a valid line ending, so it might be worth covering:

```python
@staticmethod
def _split_line_ending(line: str) -> tuple[str, str]:
    """Split a line into content and its original line ending."""
    if line.endswith("\r\n"):
        return line[:-2], "\r\n"
    if line.endswith("\n"):
        return line[:-1], "\n"
    if line.endswith("\r"):
        return line[:-1], "\r"
    return line, ""
```

Admittedly rare in practice, but since `splitlines` can produce such lines, it would be nice to stay consistent.

src/agentscope/app/middleware/_protocol/_base.py
XiuShenAl: Now that `dispatch` only intercepts `text/event-stream` responses (great improvement, by the way!), the class docstring might need a small update. It currently says:

*This middleware intercepts streaming responses that yield AgentEvent objects*

This could mislead future subclass authors into thinking all `StreamingResponse` types are intercepted. Consider updating it to something like:

*This middleware intercepts `text/event-stream` responses, deserializes AgentEvent objects from SSE `data:` frames, and converts them to a specific protocol format.*

Similarly, the example in the `_convert_to_protocol` docstring at the bottom of the class still shows a generic `AGUIMiddleware` pattern that doesn't mention SSE — might be worth a quick refresh as well.

src/agentscope/app/middleware/_protocol/_base.py
```diff
@@ -91,36 +96,95 @@ async def _convert_stream(
...
-                event_dict = json.loads(chunk_str)
+            converted = self._convert_sse_frame(chunk_str)
+            if converted is not None:
+                yield converted
+                continue
+
+            converted = self._convert_event_json(chunk_str)
+            if converted is not None:
+                yield converted
+                continue
```
XiuShenAl, line 114: Minor observation: since `dispatch` now only enters this path for `text/event-stream` responses, the chunks will always be SSE-formatted (e.g. `data: {...}\n\n` or `:\n\n`). That means this raw-JSON fallback on the full chunk string will effectively never succeed at runtime — `json.loads("data: {...}\n\n")` will always raise `JSONDecodeError`.

I understand keeping it for defensive/forward-compatibility purposes (e.g. a subclass might override `dispatch` to also handle non-SSE streams), and that's totally fine. A short inline comment explaining the intent would help future readers understand why this path exists, something like:

```python
# Fallback: try parsing the entire chunk as raw JSON.
# Currently a no-op for SSE streams, but kept for subclasses
# that may override dispatch() to also handle non-SSE responses.
```

tests/agui_protocol_test.py
```diff
@@ -36,6 +44,128 @@
...
+
+        self.assertEqual(response.status_code, 200)
+        data = response.json()
+        self.assertEqual(data["type"], "REPLY_START")
+        self.assertEqual(data["session_id"], "sess_1")
+        self.assertNotIn("RUN_STARTED", response.text)
+
+    async def asyncTearDown(self) -> None:
+        """The async teardown method."""
+
```
XiuShenAl, line 187: The test coverage is solid — especially the FastAPI integration tests for SSE and JSON response passthrough. A few additional edge cases that might be worth covering:

1. **Multi-event SSE stream**: Verify that a stream with multiple consecutive `data: {...}\n\n` frames all get converted correctly (not just a single frame).

2. **Empty `data:` line**: What happens with `data:\n\n` (no JSON payload)? This should fall through gracefully.

3. **`\r\n` line endings**: Some SSE implementations use `\r\n`. A test like `data: {...}\r\n\r\n` would exercise `_split_line_ending` with a different ending.

4. *(Optional)* **Non-AgentEvent JSON in `data:` frame**: e.g. `data: {"arbitrary": "json"}\n\n` — should pass through unconverted.

These are all minor suggestions — the existing tests already cover the critical paths well.

src/agentscope/app/middleware/_protocol/_base.py
```diff
@@ -91,36 +96,95 @@ async def _convert_stream(
...
-                protocol_data = self._convert_to_protocol(agent_event)
+        Args:
+            frame: A server-sent event frame.
 
-                # Serialize and yield the protocol data
-                yield json.dumps(protocol_data, ensure_ascii=False).encode(
+        Returns:
+            Converted frame bytes if at least one ``data:`` payload was
+            converted, otherwise ``None``.
+        """
```
XiuShenAl, line 137: The implementation processes each `data:` line independently, which works well for the current SSE generator (which always yields complete frames with single-line JSON). Two implicit assumptions might be worth documenting in the docstring:

1. **Single-line JSON**: The SSE spec allows multi-line `data:` fields (e.g. `data: {"key":\ndata: "value"}\n\n`), where payloads are concatenated by the client. This method parses each `data:` line independently, so split-line JSON would fail `json.loads` and be passed through unconverted. This is fine for the current project since `_sse_generator` always emits `json.dumps()` on a single line.

2. **Complete frames per chunk**: The method assumes each chunk contains one or more complete SSE frames. If a frame were split across chunks (e.g. network-level fragmentation), the partial frame would not be recognized. Again, this is safe given that the upstream generator yields whole frames.

A small docstring addition would help future maintainers, e.g.:

```python
"""Convert AgentEvent payloads inside an SSE frame.

.. note::
    This method assumes each ``data:`` line contains a complete
    JSON payload and that each input ``frame`` contains one or
    more complete SSE frames (no cross-chunk splitting).

Args:
    frame (`str`):
        A server-sent event frame.

Returns:
    `bytes | None`:
        Converted frame bytes if at least one ``data:`` payload was
        converted, otherwise ``None``.
"""
```

src/agentscope/app/middleware/_protocol/_base.py
```diff
@@ -91,36 +96,95 @@ async def _convert_stream(
...
+
+            line_content, line_ending = self._split_line_ending(line)
+            payload = line_content.removeprefix("data:")
+            if payload.startswith(" "):
+                payload = payload[1:]
+
+            converted = self._convert_event_json(payload)
+            if converted is None:
+                converted_lines.append(line)
+                continue
```
XiuShenAl, line 155: When a `data:` line contains valid-looking content but `_convert_event_json` returns `None`, the line is silently passed through. In production, this could make it hard to notice if a new event type is missing from the conversion map.

Adding a debug-level log specifically here (inside `_convert_sse_frame`, not inside `_convert_event_json`) would provide observability without noise — this path is only hit for actual `data:` payloads that failed conversion, not for expected fallback failures in `_convert_stream`:

```python
converted = self._convert_event_json(payload)
if converted is None:
    logger.debug(
        "SSE data payload was not converted: %.120s", payload,
    )
    converted_lines.append(line)
    continue
```

Note: I'd recommend against adding logging inside `_convert_event_json` itself, because `_convert_stream` also calls it as a fallback on the full SSE frame string (which is expected to fail) — that would produce a noisy log line on every chunk.

tests/agui_protocol_test.py
```diff
@@ -36,6 +44,128 @@
...
+            ":\n\n",
+        )
+
+    async def test_fastapi_sse_response_is_converted(self) -> None:
+        """Test middleware converts a real FastAPI SSE response."""
+        app = FastAPI()
+        app.add_middleware(AGUIProtocolMiddleware)
+
+        @app.get("/sessions/sess_1/stream")
+        def stream() -> StreamingResponse:
```
XiuShenAl: Nit: the real `stream_session_events` endpoint is an `async def`. Starlette handles sync and async endpoints through slightly different internal paths (sync endpoints run in a threadpool executor), so using `async def` here would more closely mirror the production behavior:

```python
@app.get("/sessions/sess_1/stream")
async def stream() -> StreamingResponse:
```

Not a functional issue — just makes the test a bit more representative.

XiuShenAl (commented):
Nice fix overall — the change makes the middleware behavior much clearer and safer. I left a few minor inline suggestions around edge cases, documentation, and test coverage. These are mostly follow-up / polish items rather than blockers.

src/agentscope/app/middleware/_protocol/_base.py
YingchaoX: Addressed in the latest commit: `_split_line_ending` now preserves bare `\r` consistently with `splitlines(keepends=True)`.

src/agentscope/app/middleware/_protocol/_base.py
YingchaoX: Addressed. I updated the class docstring to describe the `text/event-stream` / SSE `data:` frame scope explicitly.

src/agentscope/app/middleware/_protocol/_base.py
YingchaoX: Addressed. I added a short comment explaining that the raw JSON fallback is retained for subclasses or future overrides that may reuse the converter outside the default SSE-only dispatch path.

tests/agui_protocol_test.py
YingchaoX: Thanks. I added the CRLF coverage because it directly exercises the line-ending helper. I kept the other optional edge cases out for now to keep this PR focused on the reported SSE frame shape.

src/agentscope/app/middleware/_protocol/_base.py
YingchaoX: Addressed. I documented the intentional assumptions: complete single-line JSON payloads in `data:` lines and complete SSE frames per input chunk; multiline data concatenation and cross-chunk reassembly remain out of scope for this fix.

src/agentscope/app/middleware/_protocol/_base.py
YingchaoX: Good point. I left logging out for now to keep the middleware behavior minimal and avoid adding log noise/semantics in this bugfix. The passthrough behavior remains documented as intentional.

tests/agui_protocol_test.py
YingchaoX: Addressed. The FastAPI SSE integration test endpoint now uses `async def` to better mirror the production route.

DavdGao (approved):

PR #1927, opened 2026-06-23
feat(middleware): add file-based long-term memory middleware

y1y5:
## AgentScope Version

2.0.3

## Description

This PR adds `FileLongTermMemoryMiddleware`, a lightweight and transparent
long-term memory solution that stores memories as Markdown files within an
AgentScope workspace. It requires no external memory service, vector database,
or embedding model.

The middleware separates user profiles, durable knowledge, and daily working
memory. It supports periodic extraction and agent-managed updates through
`static`, `auto`, and `both` modes, with `memory_read`, `memory_search`, and
`memory_manage` tools.

### Changes

- Added the file-based memory implementation under:
  `src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/`
  - `_middleware.py`: middleware lifecycle, context injection, and extraction.
  - `_store.py`: Markdown storage and memory update operations.
  - `_accessor.py`: workspace-constrained file access.
  - `_tools.py`: memory read, search, and management tools.
  - `__init__.py`: public exports.
- Added examples and usage documentation under:
  `examples/long_term_memory/fileLongTermMemory/`
- Added unit tests in:
  `tests/file_long_term_memory_middleware_test.py`

## Checklist


src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_accessor.py
```diff
@@ -0,0 +1,122 @@
...
+from pathlib import PurePosixPath
+from typing import TYPE_CHECKING
+
+from ....tool import BackendBase
+
+if TYPE_CHECKING:
+    from ....workspace import WorkspaceBase
+
+
+class WorkspaceFileAccessor:
```
DavdGao: Suggested to add an abstract method `get_backend` to the `WorkspaceBase` class, and return `self._backend` in its all children classes, rather than resolving it in a new class.

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_middleware.py
```diff
@@ -0,0 +1,592 @@
...
+
+    The middleware can use an explicit workspace, resolve one dynamically from
+    ``Agent.offloader``, or lazily create and own a local fallback workspace.
+    No AgentScope core changes or backend-specific file branches are required.
+    """
+
+    def __init__(
+        self,
+        *,
+        workspace: WorkspaceBase | None = None,
```
DavdGao: Use `backend: BackendBase | None = None` directly, rather than workspace

src/agentscope/middleware/_longterm_memory/_filesystem_memory/_middleware.py
```diff
@@ -0,0 +1,592 @@
...
+        Args:
+            workspace:
+                Optional explicit workspace shared by every agent using this
+                middleware. When omitted, each call resolves
+                ``agent.offloader`` or creates a local fallback.
+            mode:
+                ``"static"`` for periodic extraction, ``"auto"`` for
+                agent-managed writes, or ``"both"`` for both paths.
+            extraction_interval:
+                Number of completed user/assistant turns between static
```
DavdGao: Only valid in `static` and `both` mode?

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_middleware.py
```diff
@@ -0,0 +1,592 @@
...
+                Flush pending static memory before context compression removes
+                old messages.
+            memory_dir:
+                Workspace-relative root containing all LTM files.
+            user_max_chars, memory_max_chars, daily_max_chars:
+                Per-document character caps enforced after every mutation.
+            max_prompt_chars:
+                Character cap applied to the memory snapshot block before
+                optional manage instructions are appended.
+            fallback_workspace_root:
```
DavdGao: What does 'fallback' mean here?

src/agentscope/middleware/_longterm_memory/_filesystem_memory/_middleware.py
```diff
@@ -0,0 +1,592 @@
...
+            WorkspaceFileAccessor(workspace),
+            memory_dir=self._memory_dir,
+            user_max_chars=self._user_max_chars,
+            memory_max_chars=self._memory_max_chars,
+            daily_max_chars=self._daily_max_chars,
+        )
+        self._stores[key] = store
+        return store
+
+    async def _resolve_workspace(self, agent: "Agent") -> WorkspaceBase:
```
DavdGao: The middleware should not be aware of the workspace

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_middleware.py
```diff
@@ -0,0 +1,592 @@
...
+                "File LTM extraction failed for session_id=%s: %s",
+                agent.state.session_id,
+                error,
+            )
+
+    def _render_recent_dialogue(self, context: list[Msg]) -> str:
+        """Render recent non-system dialogue for structured extraction."""
+        rendered = []
+        for message in context[-self._extraction_window :]:
+            if message.name == "memory" or message.role == "system":
```
DavdGao: When will the message.name equal "memory"?

DavdGao (commented):
plz see inline comments

src/agentscope/middleware/_longterm_memory/_filesystem_memory/_middleware.py
y1y5: The middleware cannot be completely independent of the workspace. The backend and workdir need to be obtained from the workspace.

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_middleware.py
y1y5: To simplify usage, here removed the parameter for explicitly passing workspace.backend. Users no longer need to pass backend or workspace, making the process completely transparent. The middleware will now handle file read/write operations internally based on backend and workdir.

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_accessor.py
y1y5: Because the backend is restricted to absolute paths and lacks access to workdir, the current implementation retains the _accessor.py file. It resolves workdir into absolute paths and then delegates file I/O to the backend.

y1y5:
This update refines the file-based LTM architecture from 3 areas.

1. **Reduced workspace coupling**

   `WorkspaceBase.get_backend()` was added as a small, generic interface for accessing workspace file operations. `FileLongTermMemoryMiddleware` no longer accepts a `workspace` or `backend` directly. Instead, it resolves the workspace from `agent.offloader`, or creates a `LocalWorkspace` when `local_workspace_dir` is configured.

   `_accessor.py` is still necessary because a backend alone does not provide the workspace-relative path context required by LTM. It combines `BackendBase` with the workspace `workdir` and exposes a small, consistent file API to the memory store. As a result, the store remains independent of Local, Docker, and E2B workspace implementations.

2. **More reliable snapshot management**

   Snapshot caching is now based on the modification times of `USER.md` and `MEMORY.md`. Cached content is reused while the files remain unchanged and automatically reloaded after an external update. Writes performed through the middleware also invalidate the snapshot explicitly.

   This avoids repeatedly reading unchanged files while preventing stale memory when files are modified outside the current middleware instance.

3. **Hybrid static extraction**

   Static extraction now reuses the agent’s existing state and configured model. The full conversation context is preserved, including multimodal messages, while a `HintBlock` provides the current `USER.md`, `MEMORY.md`, and daily memory as additional reference.

   The extraction result is produced through structured output and then applied by the memory store. This is a deliberate compromise between manually rendering a limited text window and implementing a separate tool-calling loop. It keeps the extraction flow simple, preserves richer context, and avoids requiring an additional chat model or duplicating the agent runtime.

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_accessor.py
```diff
@@ -0,0 +1,98 @@
...
+backend operation runs. The adapter owns no backend lifecycle.
+"""
+from __future__ import annotations
+
+import re
+from pathlib import PurePosixPath
+from ....tool import BackendBase
+
+
+class BackendFileAccessor:
```
DavdGao: Remove this BackendFileAccessor class and directly use the backend instance instead to avoid unnecessary abstraction.

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_store.py
```diff
@@ -0,0 +1,572 @@
...
+            memory_dir:
+                Workspace-relative root for all LTM files.
+            user_max_chars:
+                Maximum size of ``USER.md`` after an update.
+            memory_max_chars:
+                Maximum size of ``MEMORY.md`` after an update.
+            daily_max_chars:
+                Maximum size of any one dated daily-memory file.
+        """
+        self._accessor = accessor
```
DavdGao: FileLTMStore is the only consumer of the BackendFileAccessor, so remove the BackendFileAccessor and directly use the backend instance here to avoid unnecessary abstraction.

examples/long_term_memory/fileLongTermMemory/app_demo.py
```diff
@@ -0,0 +1,114 @@
+# -*- coding: utf-8 -*-
```
DavdGao: Plz remove this app_demo.py file, and we should keep only one entry for the agent service to reduce maintenance overhead.

There is a fundamental issue here — the backend is obtained from the Workspace abstraction and is not available at `create_app` time. It only gets bound to a specific sandbox after `WorkspaceManager` calls `get_workspace` within the app. Therefore, the long-term memory middleware should not be initialized in `create_app`'s `extra_middlewares` — it needs to be set up at a later stage where the backend is already resolved.

examples/long_term_memory/fileLongTermMemory/oss_demo.py
```diff
@@ -0,0 +1,344 @@
+# -*- coding: utf-8 -*-
+"""Standalone file-based long-term memory middleware demo.
```
DavdGao: What does oss stand for here, and could we use a more descriptive name for this file?

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_middleware.py
```diff
@@ -0,0 +1,559 @@
...
+    The middleware resolves a workspace dynamically from ``Agent.offloader``.
+    When no workspace is supplied, it lazily creates and owns a LocalWorkspace
+    at ``local_workspace_dir``. Backend details are not part of the public
+    middleware configuration.
+    """
+
+    def __init__(
+        self,
+        *,
+        mode: Literal["static", "auto", "both"] = "both",
```
DavdGao: Keep consistent with the mem0 middleware, which uses `mode: Literal["static_control", "agent_control", "both"]`

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_middleware.py
```diff
@@ -0,0 +1,559 @@
...
+                Flush pending static memory before context compression removes
+                old messages.
+            memory_dir:
+                Workspace-relative root containing all LTM files.
+            user_max_chars, memory_max_chars, daily_max_chars:
+                Per-document character caps enforced after every mutation.
+            max_prompt_chars:
+                Character cap applied to the memory snapshot block before
+                optional manage instructions are appended.
+            local_workspace_dir:
```
DavdGao: Why we need the `local_workspace_dir` when we already have the `memory_dir` argument?

src/agentscope/middleware/_longterm_memory/_filesystem_memory/_middleware.py
```diff
@@ -0,0 +1,559 @@
...
+            user_max_chars, memory_max_chars, daily_max_chars:
+                Per-document character caps enforced after every mutation.
+            max_prompt_chars:
+                Character cap applied to the memory snapshot block before
+                optional manage instructions are appended.
+            local_workspace_dir:
+                Directory used for the middleware-owned LocalWorkspace when
+                the Agent has no workspace offloader. Set this when callers do
+                not want to create and manage a workspace themselves.
+            memory_instructions:
```
DavdGao: Could we combine `memory_instructions` and `manage_instructions` into a single argument to simplify the interface?

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_store.py
```diff
@@ -0,0 +1,572 @@
...
+            match.group(1).strip()
+            for match in re.finditer(
+                r"(?m)^##[ \t]+([^\r\n]+?)[ \t]*$",
+                text,
+            )
+        ]
+
+    @staticmethod
+    def _search_terms(query: str) -> set[str]:
+        """Build lightweight word and CJK-bigram search terms."""
```
DavdGao: How does the search function compute the similarty since no embedding model is used here.

DavdGao (changes requested):
Please check the inline comments, and consider renaming FileLongTermMemory to something more specific — it might be worth referencing how other libraries or papers name similar concepts.

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_store.py
y1y5: BackendFileAccessor has been removed. FileSystemMemoryStore now directly receives backend and workdir.

examples/long_term_memory/fileLongTermMemory/app_demo.py
y1y5: app_demo.py has been removed. The example directory now keeps a single demo.py entry point focused on FileSystemMemory behavior and the three control modes.

examples/long_term_memory/fileLongTermMemory/oss_demo.py
y1y5: The old oss_demo.py naming was unclear, so the example has been renamed to demo.py under examples/long_term_memory/fileSystemMemory/. The demo now uses a more descriptive FileSystemMemory-focused structure instead of OSS-specific wording.

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_middleware.py
y1y5: The demo, tests, README, and tool exposure logic have all been updated accordingly.

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_middleware.py
y1y5: local_workspace_dir has been removed from the public constructor to avoid confusion with memory_dir. memory_dir remains the workspace-relative directory for memory files, while the fallback workspace path is now an internal implementation detail.

src/agentscope/middleware/_longterm_memory/_filesystem_memory/_middleware.py
y1y5: manage_instructions has been removed from the public API. memory_instructions is now an optional extra prompt appended after the default memory instructions, so users can customize the prompt without replacing the built-in memory protocol.

src/agentscope/middleware/_longterm_memory/_fileLongTermMemory/_store.py
y1y5: Clarified in code comments and README.

y1y5:
**Update Summary**

- Renamed the feature to `FileSystemMemoryMiddleware`.
- Moved the implementation package to `_filesystem_memory`.
- Removed `BackendFileAccessor`; `FileSystemMemoryStore` now uses `backend` and `workdir` directly.
- Removed `app_demo.py` and replaced the old demo naming with a single `demo.py`.
- Simplified the public constructor API by removing `local_workspace_dir`, `manage_instructions`, and `max_prompt_chars`.
- Aligned mode names with `Mem0Middleware`: `static_control`, `agent_control`, and `both`.
- Reworked the demo to focus on memory behavior across the three modes.
- Expanded documentation for lightweight lexical search, including CJK bigram matching and examples.

DavdGao (approved):

PR #1929, opened 2026-06-23
fix(agent): emit thinking end before answer text

### LINKED ISSUE #1885: [Bug]: Streaming events out of order — TextBlockDelta emitted before ThinkingBlockEnd at the reasoning→answer boundary

WilliamK112:
## AgentScope Version

`2.0.3` from a local editable install.

## Description

Fixes #1885.

Streaming chat chunks were converted to events by handling text before thinking. At the reasoning-to-answer boundary, that could emit the first `TEXT_BLOCK_DELTA` before `THINKING_BLOCK_END`, which makes downstream renderers place the first answer token inside the thinking section.

This PR keeps data-only chunk behavior intact, emits thinking transitions before text transitions, and closes thinking before the answer text starts when thinking and text arrive in the same streamed chunk.

Validation run locally:

- `.venv/bin/python -m pytest tests/agent_basic_test.py`
- `.venv/bin/python -m pytest tests`
- `.venv/bin/pre-commit run --all-files`

## Checklist

- [x] An issue has been created for this PR: #1885
- [x] Docstrings are in Google style; this PR does not add public API docstrings
- [x] Related documentation has been updated or is not required for this internal bug fix

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -2421,14 +2421,54 @@ class Agent:
             elif isinstance(block, DataBlock):
                 data_blocks.append(block)
 
-        # Handle the text blocks. We only auto-close the open text block
-        # when the current chunk has neither text NOR data — a chunk that
-        # carries only data (e.g. an omni-style audio PCM delta arriving
-        # between two text deltas) must keep the text stream alive so the
-        # frontend doesn't fragment one logical text stream into many
-        # separate bubbles. A chunk with tool calls (and no text/data)
-        # still closes text, which preserves text → tool → text render
-        # order via distinct text blocks.
+        # Pure data chunks (e.g. omni-style audio PCM deltas) keep the
+        # currently open text/thinking stream alive so the frontend doesn't
+        # fragment one logical stream into many separate bubbles.
+        data_only = bool(data_blocks) and not (
+            text_blocks or thinking_blocks or tool_call_blocks
+        )
+
+        # Close an open text stream before switching to another non-data
+        # modality. Tool-only chunks still close text, preserving text → tool
+        # → text render order via distinct text blocks.
+        if block_ids.get("text") and not text_blocks and not data_only:
+            yield TextBlockEndEvent(
+                reply_id=self.state.reply_id,
+                block_id=block_ids["text"],
+            )
+            block_ids["text"] = None
+
+        # Handle thinking before text so reasoning ends before answer text
+        # begins at the reasoning → answer boundary.
+        if thinking_blocks:
+            # Generate a new thinking block id and start event
+            if not block_ids.get("thinking"):
+                block_ids["thinking"] = _generate_id()
+                yield ThinkingBlockStartEvent(
+                    reply_id=self.state.reply_id,
+                    block_id=block_ids["thinking"],
+                )
+            # Generate the thinking delta event with the existing id
+            yield ThinkingBlockDeltaEvent(
+                reply_id=self.state.reply_id,
+                block_id=block_ids["thinking"],
+                delta="".join([_.thinking for _ in thinking_blocks]),
+            )
+            if text_blocks:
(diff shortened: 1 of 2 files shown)


DavdGao:
Fixed in #1887

PR #1943, opened 2026-06-24
feat(workspace): add daytona workspace backend

BreezyB1n:
## AgentScope Version

`2.0.3`

### Related Issue

Closes #1916

### Background

AgentScope currently supports local filesystem, Docker, and E2B workspace backends. This PR adds Daytona as a self-hostable, open-source sandbox workspace backend while preserving the existing workspace / backend / manager layering used by Docker and E2B.

### What Changed

- Added `DaytonaWorkspace`, `DaytonaBackend`, and `DaytonaWorkspaceManager`.
- Ported `DaytonaWorkspace` onto the shared `SandboxedWorkspaceBase` abstraction introduced on `main`, so shared gateway lifecycle, MCP management, skill management, offload, reset, and builtin tool listing are inherited instead of duplicated in the provider implementation.
- Kept Daytona-specific logic scoped to provider lifecycle hooks: create/attach by labels, state normalization, SDK-derived workdir/home paths, first-time bootstrap, and graceful `stop(force=False)` teardown.
- Implemented Daytona backend primitives for command execution and file I/O, including SDK result mapping, default cwd exposure, raw bytes read/write, missing-file mapping, and inherited filesystem helpers through `BackendBase`.
- Kept Daytona's public config surface intentionally minimal and aligned with E2B-style usage: `api_key`, `api_url`, `target`, `timeout_seconds`, `gateway_port`, `env`, `sandbox_metadata`, `extra_pip`, `default_mcps`, `skill_paths`, and optional `os_user`.
- Wired Daytona exports and the optional `workspace` dependency entry.
- Updated the README workspace backend list to include Daytona.
- Added mock and credential-gated live test coverage for backend, workspace, manager, bootstrap, inherited workspace behavior, persistence, and app toolkit Write/Read usage.

### Validation

| Layer | Coverage | Result |
| --- | --- | --- |
| Daytona local/mock regression | Backend, workspace, manager, bootstrap, inherited shared-base behavior, and builtin file-cache regression | 61 passed |
| Daytona live E2E | Real Daytona sandbox with `DAYTONA_API_KEY`: backend exec/file I/O/helpers, workspace lifecycle, skills, offload/reset, persistent reattach, MCP listing, Toolkit Write/Read | 20 passed |
| Daytona total | Local/mock regression + credential-gated live coverage | 81 passed |
| Static checks | `compileall`, `git diff --check`, `uv lock --check` | Passed |
| Pre-commit | Targeted Daytona refactor files | Passed |

Commands run:

[...]

PR #1933, opened 2026-06-24
feat(workspace): add K8sWorkspace with Pod/PVC lifecycle, tar-stream backend and MCP gateway support

XiuShenAl:
## AgentScope Version

2.0.3

## Description

Closes #1842

### Background

AgentScope currently supports three workspace backends: `LocalWorkspace` (dev), `DockerWorkspace` (single-node), and `E2BWorkspace` (cloud sandbox). There is no production-grade, multi-tenant workspace implementation that leverages Kubernetes-native orchestration, resource quotas, namespace isolation, and automatic pod recovery.

### Purpose

Add a `K8sWorkspace` implementation that provides production-ready containerized execution environments backed by Kubernetes Pods and PVCs, fully reusing the existing MCP Gateway + GatewayClient architecture.

### Changes Made

**New files:**
- `src/agentscope/workspace/_k8s/__init__.py` — Package exports for `K8sWorkspace` and `K8sBackend`
- `src/agentscope/workspace/_k8s/_k8s_bootstrap.py` — Constants, RFC-1123 name sanitization (`_k8s_safe_name`), bootstrap command sequence generator, released/dev install mode renderers
- `src/agentscope/workspace/_k8s/_k8s_backend.py` — `K8sBackend(BackendBase)` implementing `exec_shell` (WebSocket exec with channel-demuxed stdout/stderr/exit-code), `read_file` (tar stream extraction), and `write_file` (tar stream injection via stdin channel 0 with proper EOF signaling)
- `src/agentscope/workspace/_k8s/_k8s_workspace.py` — `K8sWorkspace(WorkspaceBase)` with full lifecycle: namespace/PVC/Pod creation & reuse, bootstrap provisioning, MCP Gateway startup, dual network modes (`pod-ip` for in-cluster, `port-forward` for local dev), all abstract method implementations
- `src/agentscope/app/workspace_manager/_k8s_workspace_manager.py` — `K8sWorkspaceManager(WorkspaceManagerBase)` with TTL-based caching, background sweeper, double-checked locking, parallel `close_all`
- `tests/workspace_k8s_test.py` — Unit tests covering name sanitization, bootstrap commands, backend exec/read/write, manager cache/TTL, PVC/Pod lifecycle, and Pending-state early detection

**Modified files:**
- `src/agentscope/workspace/__init__.py` — Added `K8sBackend`, `K8sWorkspace` exports
- `src/agentscope/app/workspace_manager/__init__.py` — Added `K8sWorkspaceManager` export
[...]

PR #1951, opened 2026-06-27
feat(workspace): share workspace across different agents

### LINKED ISSUE #1796: feat(workspace): support multiple agents for the same workspace
**Background**

1. Currently, workspaces and workspace managers are isolated at the agent level. However, in agent team scenarios, team members also need to operate within the same workspace. Therefore, a workspace should be able to support multiple agents sharing it.

2. Even for the same agent, stateful MCPs (e.g., `browser-use`) should not be shared across different sessions, as doing so would cause state conflicts. On the other hand, creating a new MCP instance per session may overload the workspace, causing too many connections to accumulate before TTL expiration.

**TBD**

- Need to discuss how to properly manage and control resource usage while supporting multi-agent workspaces.
- A potential approach: add `agent_id` and `session_id` as parameters to `list_mcps`. For stateful (`is_stateful=True`) MCP clients, maintain separate connections for each unique `(agent_id, session_id)` combination.

### LINKED ISSUE #1919: feat(workspace): share workspace across different agents
**Background**

In agent-team, the subagents and leader need to operate the same workspace or directory. However, their tools should be isolated from each other, espeicially for the stateful MCP clients, e.g. brower-user MCP client, which requires us to improve the current implement workspace implementation to support multiple agents

**Changes**
- support to maintain different MCP client instances for different agent ids.

iluv7:
## PR Title

feat(workspace): isolate MCP client sessions per agent (#1919)

## AgentScope Version

2.0.3

## Background

In agent-team setups, the leader and sub-agents share the same `workspace_id`,
causing them to share the same `MCPClient` instances. For stateful MCP clients
(e.g., browser-user), this results in session state interference between
agents — the leader's browser tab operations affect the worker's browsing
session.

Closes #1919.

## Changes

Core idea: isolate MCP client sessions by `agent_id` while keeping the
workspace directory, skills, and offload storage shared.

When an agent calls `list_mcps(agent_id)` for the first time, independent
instances are deep-copied from `default_mcps` and independently connected.
Workspace-level MCP storage changes from `list[MCPClient]` to
`dict[agent_id, list[MCPClient]]`.

### Layer-by-layer changes

| Layer | File | Change |
|-------|------|--------|
| Interface | `workspace/_base.py` | Add `agent_id` parameter to `list_mcps`/`add_mcp`/`remove_mcp` |
| LocalWorkspace | `workspace/_local_workspace.py` | `_mcps` → per-agent dict; lazy clone + connect on first access; `.mcp` file format → `{agent_id: [configs]}` with old-format auto-migration |
| MCP Gateway | `workspace/_mcp_gateway/_mcp_gateway_app.py` | `state.clients` → `{name: {agent_id: client}}`; all 6 endpoints accept `agent_id` query param; `_connect_initial` supports both old and new config formats |
| Gateway Client | `workspace/_gateway_client.py` | `GatewayMCPClient`/`GatewayMCPTool`/`GatewayClient` pass `agent_id` through to gateway HTTP calls |
| Docker/E2B | `workspace/_docker/`, `workspace/_e2b/` | `_gateway_clients` → per-agent two-level dict; `_write_gateway_config` servers format → `{agent_id: [configs]}`; lazy clone via `gateway.make_client` + `connect` |
| App layer | `app/_service/_toolkit.py`, `app/_router/_workspace.py` | Thread existing `agent_id` into workspace MCP methods; `ValueError` → 409 in add_mcp endpoint |

### Test Summary

**274 tests, 273 passed** (1 pre-existing failure in MCP SSE `$defs` formatting, unrelated to this change; 19 E2B tests require `E2B_API_KEY`).

### Existing Test Regression (265 tests)

| Scope | Count | Coverage |
|-------|-------|----------|
| LocalWorkspace (`workspace_local_test.py`) | 17 | Offload context/tool results, skills init/dedup/list, MCP init connect failure removal, `.mcp` old-format compatibility |
| DockerWorkspace (`workspace_docker_test.py`) | 18 | Offload context/tool results/DataBlock dedup/URL source, skills add/list/init, lifecycle (close/reset), `.mcp` persistence across restart |
| Toolkit / Service (`service_toolkit_test.py`, `toolkit_test.py`) | 19 | Tool assembly (user agent full set, worker agent TeamSay only), schedule guard, extra factory, function registration, tool groups, meta tools |
| MCP client (`mcp_*_test.py`) | 6 | SSE/streamable HTTP client |
| Other unit tests | 205 | Agent, Model, Permission, Middleware, State, etc. |

#### Full HTTP API E2E Tests (4 tests)

`tests/test_e2e_api.py` — starts a complete FastAPI service via pytest + uvicorn,
verifying the full stack through HTTP requests:

- **Lazy clone**: leader and worker each create agent + session, verify via `GET /workspace/mcp` that each gets an independent `default-search` MCP cloned from `default_mcps`
- **Per-agent add isolation**: leader dynamically adds `leader-extra` via `POST /workspace/mcp`, confirms leader sees it and worker does not
- **Duplicate add within same agent → 409**: same agent re-adds an MCP with the same name, verifies `409 Conflict` with `"already exists"` in the response body
- **Per-agent remove isolation**: leader deletes `leader-extra` via `DELETE /workspace/mcp/leader-extra`, confirms leader's list recovers and worker is never affected

#### E2BWorkspace E2E Tests (4 tests, requires `E2B_API_KEY`)

`tests/test_e2e_e2b_mcp.py` — verifies `agent_id` routing through the in-sandbox
Gateway via pytest + E2B cloud sandbox:

- **Lazy clone from default_mcps**: two agents each get an independent `default-fs` via first `list_mcps`; same agent's second access hits cache
- **Per-agent add/remove isolation**: agent-A `add_mcp("extra")` does not affect B; `remove_mcp` only removes A's entry
- **Duplicate add within same agent → ValueError**: re-adding the same name for the same agent raises `ValueError`
- **`.mcp` persistence**: cross-agent add operations correctly persist to the sandbox `.mcp` file

#### DockerWorkspace E2E Tests (4 tests, requires Docker)

`tests/test_e2e_docker_mcp.py` — verifies `agent_id` HTTP query param forwarding
through the Gateway layer via pytest + Docker container:

- **Lazy clone from default_mcps**: two agents each get an independent `default-fs` via Gateway `GET /mcps?agent_id=xxx`; same agent's second access hits cache
- **Per-agent add/remove isolation**: agent-A `add_mcp("extra")` does not affect B; `remove_mcp` only removes A's entry
- **Duplicate add within same agent → ValueError**: re-adding the same name for the same agent raises `ValueError`
- **`.mcp` persistence as `{agent_id: [configs]}` format**: file written with correct per-agent dict structure containing both agent-A and agent-B keys

### Backward Compatibility

- Old `.mcp` format (flat list `[{...}]`) is auto-detected in `_restore_or_seed_mcps` and migrated to `{"_default": [{...}]}`, verified in LocalWorkspace tests
- `WorkspaceBase` abstract method signature changes are synchronized across all subclasses (Local/Docker/E2B) and callers (router/toolkit)
- Router `add_mcp` endpoint now converts `ValueError` → `409 HTTPException`, returning the correct HTTP status for duplicate add scenarios

### Code Quality

- `pre-commit run --all-files`: all 17 hooks pass

## Checklist

- [x] Issue created for this PR
- [x] CONTRIBUTING.md read
- [x] Docstrings follow Google style
- [ ] Related documentation updated
- [x] Code ready for review


iluv7:
Design Decisions

1. Runtime add_mcp visibility — "complete isolation" over "broadcast"

The issue discussed a "broadcast + independent instance" approach (leader's add_mcp visible to workers, each with their own client). We chose complete isolation instead: MCPs added at runtime by a leader are only visible to that leader; workers must add their own. Reasoning: (1) simpler implementation without cross-agent broadcast; (2) avoids side effects from broadcasting (if A deletes an MCP, should B's be deleted? what if B doesn't want it?); (3) a broadcast layer can be added on top later without breaking the existing API.

2. Per-agent isolation rather than per-session isolation (#1796)

Issue #1796 proposed isolating stateful MCPs by (agent_id, session_id). We chose per-agent isolation: all sessions of the same agent share the same MCP client instances. Per-session isolation would spawn a new Chrome/browser process for every new chat, causing connections to accumulate rapidly. Per-agent isolation matches expected behavior (same agent identity = same browser login state), and per-session isolation can be added later without breaking the current API.

DavdGao (commented):
@iluv7 First off — thanks a lot for the contribution, and apologies in advance for the friction I'm about to introduce.

We've been reviewing this PR alongside #1953, #1943, and #1933. Each of them adds a new sandbox-backed workspace on top of the previous abstraction, but none of them include the "shared across different agents" capability.

To avoid having to bolt on this shared functionality separately to each individual workspace subclass, we've introduced a new base class `SandboxWorkspaceBase` in #1971. With this in place on the current `main` branch, we can build the cross-agent workspace sharing feature just once at the base class level, rather than reimplementing it everywhere.

Could you please rebase / merge `main` into this branch, and then implement the cross-agent workspace sharing functionality on top of the new base class? That way we get it right in one shot.

Thanks again for your patience and continued effort on this!

iluv7:
@DavdGao 
Thanks for the heads-up and for introducing `SandboxedWorkspaceBase` in #1971 — it's exactly the right abstraction.

I've merged main into this branch. The original per-agent MCP isolation approach is unchanged, except now it will be implemented at the `SandboxedWorkspaceBase` level instead of duplicating across each workspace subclass. Main changes:

- **`WorkspaceBase`** — `list_mcps`, `add_mcp`, `remove_mcp` gain `agent_id: str` parameter
- **`SandboxedWorkspaceBase`** — `_gateway_clients` and `_mcps` become per-agent; gateway config updated accordingly — covers Docker / E2B / future sandboxes in one place
- **`LocalWorkspace`** — same per-agent storage for its own `_mcps`
- **Gateway app / GatewayClient** — endpoints gain `agent_id` query param

Skills remain shared, everything else is already per-session. Will push once ready.

iluv7:
@DavdGao 
Quick update on this PR — I've merged main and re-implemented the per-agent MCP isolation on top of the new `SandboxedWorkspaceBase` (#1971).

**Summary of changes:**

Per-agent MCP logic is implemented at the `SandboxedWorkspaceBase` level (#1971) so Docker/E2B workspaces don't need individual changes:

- `SandboxedWorkspaceBase` handles all gateway-routed per-agent MCP (lazy clone, add/remove, gateway config)
- `DockerWorkspace` and `E2BWorkspace` — **zero changes**, they inherit everything from the base class
- `LocalWorkspace` — per-agent `_mcps` dict, duplicate name check, uses parent `_restore_or_seed_mcps()`
- `WorkspaceBase` — `_mcps` → `dict[agent_id, list[MCPClient]]`, lazy clone in base `list_mcps`, `_save_mcp_file`/`_restore_or_seed_mcps` support per-agent format + old flat list auto-migration
- `GatewayClient` / `GatewayMCPClient` / `GatewayMCPTool` — `_agent_id` attribute, all HTTP requests carry `?agent_id=` param
- `_mcp_gateway_app.py` — `state.clients` becomes `{name: {agent_id: client}}`, all 6 endpoints require `agent_id` query param

**Tests: 48/48 passing across all three workspace backends:**

| Layer | Tests | Status |
|-------|-------|--------|
| LocalWorkspace | 17 (regression) | ✅ |
| DockerWorkspace | 18 (regression) | ✅ |
| E2BWorkspace | 1 (regression) | ✅ |
| Docker per-agent e2e | 4 (new) | ✅ |
| E2B per-agent e2e | 4 (new) | ✅ |
| HTTP API per-agent e2e | 4 (new) | ✅ |

In addition to the automated tests, manually verified per-agent MCP isolation via the Swagger API:

- Added an MCP to leader agent → visible to leader
1e1ad49cf6c99a77de0ccf4642d4c1fb
- Queried from worker agent → returned empty, leader's MCP not visible to worker
5a98794bfc8c590c4de947b788a9e90c

DavdGao (commented):
@iluv7 Thanks for your contribution! We're currently in the process of merging several workspace-related PRs (K8s, OpenSandbox, and Daytona). The main branch has decoupled all cloud-based sandbox operations into `SandboxWorkspaceBase`, which should be largely stable and independent of this PR — though some conflicts may still be unavoidable. Feel free to sync with main now or wait until we've landed those PRs — either way works. We'll push to get them merged as soon as possible. Thanks for your patience!

iluv7:
Hi @DavdGao, merge conflicts resolved. Pre-commit + Local + Docker + E2E API tests all pass locally.

In Addition, the E2B sandbox tests (`workspace_e2b_test.py` and `test_e2e_e2b_mcp.py`) are failing with gateway startup timeouts. Confirmed pre-existing — tested against a clean main checkout (same failure). Had a quick look at `_setup_mcp_gateway()`: `pkill -f _mcp_gateway_app.py` matches its own `sh -c` wrapper on E2B, killing the shell before `nohup` runs. Docker isn't affected. Happy to send a fix today if you'd like.

src/agentscope/workspace/_base.py
```diff
@@ -332,30 +337,79 @@ async def list_tools(self) -> list[ToolBase]:
             Write(backend=backend),
         ]
 
-    async def list_mcps(self) -> list[MCPClient]:
-        """Return the currently registered MCP clients."""
-        return list(self._mcps)
+    async def list_mcps(self, agent_id: str) -> list[MCPClient]:
```
DavdGao: - For backward compatibility, we should set a default value for the new `agent_id` argument, e.g., `"default_agent"`.

- I'm also wondering whether we need an additional `session_id` parameter to handle cases where MCP state should **not** be shared across sessions, even for the same agent. For example, with browser-use MCP, each session should maintain its own browser state independently — even if the workspace is already isolated per agent. That said, this would require adding `session_id` to `list_mcps`, `add_mcp`, and `remove_mcp` as well, which increases the interface complexity. @iluv7 What do you think?

DavdGao: Alternatively, for simplicity, we could enforce session-level isolation for all MCPs within a workspace, regardless of whether they are stateful or not.

iluv7: Hi, @DavdGao 
Thanks for the review and the great suggestion! I also feel that per-agent isolation has its limitations. While it solves the leader-vs-worker MCP sharing problem, there's still a gap: the same agent can run multiple sessions concurrently. The distributed lock in `ChatService._run_impl` is keyed on `session_id`, not `agent_id`, so two sessions of the same agent can run in parallel and still share the same `_mcps["agent-leader"]` instances — interfering with each other.

I'm in favor of the latter approach — session-level isolation for all MCPs within a workspace. `session_id` is already a finer-grained dimension than `agent_id`, and the interface would be cleaner with just one parameter. What do you think?

That said, session-level isolation raises one concern: stateful MCPs obviously need isolation, but stateless ones (no session state to pollute) would also get cloned per session. If many sessions use the same stateless MCP, that's a lot of pointless cloning. Fortunately, there's already a safeguard in the code:

```py
if self.mcp_config.type == "stdio_mcp" and not self.is_stateful:
    raise ValueError("STDIO MCP must be stateful (is_stateful=True).")
```

This means stateless MCPs can only use HTTP — no local child processes — which avoids the memory explosion of massive and pointless `stateless + stdio` clones. But I think we can take it a step further and use `is_stateful` for an explicit distinction:

- `is_stateful=True` → clone per session, each with its own connection/process
- `is_stateful=False` → share the same instance across all sessions (no session state to pollute)

This way stateless MCPs don't even need to be cloned — just share the reference directly. What do you think?

Also, the backward compatibility concern is already addressed — I've added a default value of `"default_agent"` to the `agent_id` parameter.

iluv7:
Hi, @DavdGao , just a friendly ping on this PR. Let me know if there's anything I can adjust or if you'd like me to rebase. Thanks!

src/agentscope/workspace/_base.py
iluv7: Beyond how to establish connections in the per-session model, another problem occurred to me — cleanup. In the per-agent approach, the number of connections is small and reclaiming them alongside the workspace is sufficient. But with per-session, the connection count scales to agents × sessions — each session may clone its own independent stateful MCP connection. If you only create and never destroy, closing a browser tab leaves those connections orphaned, consuming resources indefinitely.

## Current approach (per-agent)

Connection lifecycle is bound to the workspace. The WorkspaceManager has a global TTL (default 3600 seconds); when the workspace is evicted, `close()` cascades to shut down all MCP connections across all agents. There is no agent-level TTL — once a connection is lazy-initialized, it stays alive until the workspace dies or the user manually calls `remove_mcp`.

An agent is a long-lived entity. Users typically reuse the same agent across many sessions; tearing down and rebuilding the connection each time doesn't make sense — a stdio cold start of Chrome takes 1-2 seconds. Keeping connections alive and reusing them across sessions fits the per-agent semantics of "same agent identity, shared MCP." The current cleanup relies on workspace TTL as a safety net, which is acceptable for now.

## The new problem with per-session

At the per-session level, the connection count becomes agents × sessions, making lifecycle management a hard requirement:

**When does a session actually "end"?** A normal close can trigger active cleanup, but what happens when the user simply closes a browser tab and the session lingers on the server?

## My preferred approach

Record the last-access time for each `(agent_id, session_id)` connection. A periodic task (e.g., every 5 minutes) scans the registry and automatically closes + cleans up any connection idle beyond the TTL threshold (e.g., 30 minutes).

- Simple to implement — no cross-process coordination needed.
- Pairs with `is_stateful`: stateless MCPs have nothing to close; TTL only governs stateful ones.
- The default TTL maps to a session's typical idle duration (30 minutes), covering most real-world usage.
- Active release on normal session close handles the common path; TTL is the safety net for abnormal scenarios.
- Workspace reset/close provides a global fallback — the last safety net.

DavdGao (approved):
Thanks for your contribution, and lgtm

PR #1954, opened 2026-06-29
fix(tool): reject negative grep pagination values

### LINKED ISSUE #1936: [Bug]: Grep accepts negative pagination values
The builtin `Grep` tool accepts negative pagination values for `head_limit` and `offset`.

Because the implementation applies these values to Python list slicing, negative values can trigger Python negative-index behavior instead of normal pagination semantics. For example, `offset=-1` with unlimited results can return the last result instead of behaving like a valid non-negative pagination offset.

Expected behavior: `head_limit` and `offset` should reject negative values because they represent pagination limits and offsets.

I would like to work on this issue. Could you please assign it to me?

VectorPeak:
Re-submitted after fork recovery. Original PR #1932 was closed due to fork deletion.


DavdGao (approved):
lgtm

Issue #1947, opened 2026-06-26
SubAgentTemplate workers cannot declare or receive custom business tools — only the framework default toolset


75ACOL:
## Context

The `SubAgentTemplate` feature (#1825) works well for configuring a worker's *system prompt*, *context*, *react*, and *permission* config, and routing to templates via `AgentCreate(subagent_type=...)` is clean. ✅

However, there's a gap regarding **business tools**: a worker spawned from a `SubAgentTemplate` has no way to declare "I need tool X" and actually receive it. This forces a workaround, and I'd like to discuss whether the framework can support this elegantly.

## The problem

A `SubAgentTemplate` lets you configure a worker's *system prompt*, but the prompt is the **only** thing that survives into the worker's identity. `AgentCreate` builds the worker `AgentRecord` as follows (`_tools/_agent_create.py:423`):

```python
worker_agent = AgentRecord(
    user_id=...,
    source="team",
    data=AgentData(
        name=name,                 # ← leader-assigned, free-form, NOT the template type
        system_prompt=system_prompt,
        context_config=...,
        react_config=...,
        # no field carries subagent_type / template identity / tool requirements
    ),
)
```

Two consequences that together block clean tool injection:

1. **The worker's `name` is not the template `type`.** `name` is whatever the leader passes to `AgentCreate` (e.g. `"analyst-1"`, different every run), while `subagent_type` (e.g. `"analyst"`) is discarded after routing. So `custom_agent_cls.__new__(name)` cannot dispatch the worker to a type-specific subclass — the routing key is gone.

2. **`get_toolkit` assembles the worker's tools from generic sources only** (`_service/_toolkit.py`): workspace builtins, planning tools, and `TeamSay` for workers. The `extra_factory` hook only receives `(user_id, agent_id, session_id)` — it cannot see `system_prompt` or template type, so it can't decide *which* tools *this* worker needs without hacky heuristics. Applying it globally also pollutes every worker (e.g. one worker type receiving tools another type never asked for).

**Result:** a worker can be instructed (via prompt) to use project-specific tools, but has no mechanism to actually receive them. It silently falls back to framework builtins, which work but lose the value of the tools the template intended.

## What we tried (and why each is a workaround)

- **\`@register(name)\` a per-type subclass with \`_assemble_tools\`** → doesn't work. Worker `name` ≠ template `type` and isn't stable (consequence #1).
[...]

PR #1981, opened 2026-07-02
feat(agent-team): support template-scoped worker tools

RerankerGuo:
## Summary

Lets `SubAgentTemplate`-backed workers declare and receive their own
custom business tools, instead of being limited to the framework default
toolset. Issue #1947 reports that workers spawned from a template cannot
receive tools scoped to that template.

Changes:

- Persist `subagent_type` on `AgentData` so the service can recover which
  template, and therefore which tool factory, a worker belongs to.
- Add an optional, non-serializable `extra_tools_factory` to
  `SubAgentTemplate`.
- Recreate template-scoped tools when assembling a matching worker's
  toolkit, while preserving the existing defaults for legacy and invited
  workers.

Refs #1947.

## Tests

- A created worker persists its originating template type.
- A matching worker receives tools from its template factory.
- A worker without a matching template receives no unrelated tools.
- Existing team creation, invitation, routing, and toolkit tests remain
  green.

```bash
uv pip install -e ".[dev]"
pytest tests/service_team_tools_test.py tests/service_toolkit_test.py
```

### DIFF
--- a/src/agentscope/app/_service/_toolkit.py
+++ b/src/agentscope/app/_service/_toolkit.py
@@ -73,7 +73,8 @@ async def get_toolkit(
        that is its team's leader gets the full leader-side toolset
        (``TeamCreate / AgentCreate / TeamSay / TeamDelete``, plus
        ``AgentInvite`` when the user has at least one invitable agent).
-    6. Caller-supplied extras (``extra_factory``)
+    6. Template-scoped extras for matching created team workers
+    7. Caller-supplied extras (``extra_factory``)
 
     Plus the workspace's skills and MCPs, which become the toolkit's
     ``skills_or_loaders`` and ``mcps`` parameters.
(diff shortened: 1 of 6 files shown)


RerankerGuo:
Rebased onto current main (`1aeb03d0`) after #1998 and resolved the toolkit test fixture against the new resource-access API. The refreshed branch is mergeable at `5e506a0`; it now contains only the six #1947-related files (+179/-4). Targeted tests, Black, flake8, and mypy pass locally.

RerankerGuo:
Closing because this implementation no longer matches the current sub-agent template contract. `SubAgentTemplate` is now deliberately pure data and serializable, while this PR introduces a non-serializable callable tool factory and predates the recent agent/workspace architecture changes. A future solution should be discussed against the current design before implementation.

PR #1989, opened 2026-07-03
docs(agent): fix copy-paste and mismatched docstrings in agent, state and tool modules

orangeCatDeveloper:
## Summary

Documentation-only fixes for several incorrect or copy-pasted docstrings in the `agent`, `state`, and `tool` modules. No behavior change.

**`src/agentscope/agent/_config.py`**
- `SummarySchema.context_to_preserve` carried a leftover attribute docstring about "executing multiple tool calls in parallel", which is unrelated to the field; it now describes the field itself.
- The `ContextConfig` summary-template docstring referenced a non-existent `compression_summary_model`; the template fields actually come from `summary_schema`.

**`src/agentscope/agent/_agent.py`**
- `Agent.__init__` documented `context_config` as a non-existent `CompressionConfig` (actual type: `ContextConfig`), and several params (`middlewares`, `state`, `model_config`, `context_config`, `react_config`) were missing the `| None` / `optional` markers that match the actual signature defaults.
- `_into_queue` documented `tool_call` as `ToolBlockCall` instead of `ToolCallBlock`.
- `_handle_error_tool_call` had a truncated sentence for the `state` param.

**`src/agentscope/state/_state.py`**
- Grammar: "when feed into the LLM" → "when fed into the LLM" (×2) in `AgentState` field docstrings.

**`src/agentscope/tool/_response.py`**
- `ToolChunk.id` docstring said "The identity of the tool response" (copied from `ToolResponse`); it now says "tool chunk".

## Notes

- Docstring/comment changes only — no runtime behavior, public API, or breaking changes.
- `pre-commit` and `pytest tests` pass locally.
- No companion docs PR needed (inline docstrings only).

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -121,23 +121,23 @@ class Agent:
             toolkit (`Toolkit | None`, optional):
                 The toolkit used for registering tools, MCPs and skills as the
                 sole source.
-            middlewares (`list[MiddlewareBase] | None`):
+            middlewares (`list[MiddlewareBase] | None`, optional):
                 Middlewares applied to the agent to modify its behavior
                 without altering its source code. Supported hook points
                 include: reply, reasoning, acting, model call, and system
                 prompt retrieval.
-            state (`AgentState`):
+            state (`AgentState | None`, optional):
                 The agent state. A new state will be created if not provided.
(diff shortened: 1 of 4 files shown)


DavdGao (approved):
LGTM

PR #1992, opened 2026-07-03
fix(model): preserve OpenAI chat choice metadata

RerankerGuo:
## Summary

Preserves OpenAI Chat Completions choice metadata that
`OpenAIChatModel` currently drops on the floor:

- `finish_reason` — exposed in `ChatResponse.metadata` (streaming
  and non-streaming paths).
- `function_call` — exposed in `ChatResponse.metadata` for the
  non-streaming path (legacy option).
- `tool_calls` accumulated from the streaming delta — exposed in
  `ChatResponse.metadata` for the non-streaming path; the
  streaming path forwards each delta into the existing
  `delta_res.append_tool_call` so tool calls are already available
  on the content blocks (this PR does not duplicate them into
  metadata).
- `id` (response id) — preserved on both paths.

Two small helpers were added locally to `_openai_chat/_model.py`
to serialise SDK objects safely (`_safe_getattr` for `unittest.mock`
safety, `_dump_openai_obj` for the legacy `function_call` shape).

Refs agentscope-ai/agentscope#1386.

## Streaming-path refactor note

The original implementation attached metadata via a post-loop
`_final_kwargs` block. Between the original commit and the rebase
on 2026-07-08, upstream `Agent._execute_tool_call` refactored the
streaming response path from "build chunks → yield final
`_final_kwargs` at the end" to "yield a `delta_res` per chunk".
That made the post-loop block invalid.

To preserve the original PR intent on the new structure, the
fix attaches `finish_reason` to `delta_res.metadata` on the chunk
that first sees it (i.e. the final chunk). Non-streaming path is
unchanged because it already builds a single `ChatResponse`.

## Tests

`tests/model_openai_chat_test.py`:

- Streaming: assert `finish_reason` lands on the `ChatResponse`
  whose `is_last=True`.
- Non-streaming: assert `finish_reason`, `function_call`,
  `tool_calls`, and `id` all land on `ChatResponse.metadata`.
- `unittest.mock`-style attributes do not crash the helpers.

```
uv pip install -e .[dev]
pytest tests/model_openai_chat_test.py
```

### DIFF
--- a/src/agentscope/model/_openai_chat/_model.py
+++ b/src/agentscope/model/_openai_chat/_model.py
@@ -33,6 +33,39 @@ else:
     AsyncStream = Any
 
 
+def _safe_getattr(obj: Any, name: str) -> Any:
+    """Return an attribute value unless it is an unresolved mock-like attr."""
+    value = getattr(obj, name, None)
+    if value.__class__.__module__.startswith("unittest.mock"):
+        return None
+    return value
+
+
+def _dump_openai_obj(obj: Any) -> Any:
+    """Convert OpenAI SDK objects to JSON-safe metadata values."""
+    if obj is None:
+        return None
+    if isinstance(obj, (str, int, float, bool)):
+        return obj
+    if isinstance(obj, list):
+        return [_dump_openai_obj(_) for _ in obj]
+    if isinstance(obj, tuple):
+        return [_dump_openai_obj(_) for _ in obj]
+    if isinstance(obj, dict):
+        return {str(k): _dump_openai_obj(v) for k, v in obj.items()}
+    if hasattr(obj, "model_dump"):
+        return obj.model_dump(mode="json", exclude_none=True)
+    if obj.__class__.__module__.startswith("unittest.mock"):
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Friendly ping — this preserves chat choice metadata (currently dropped by `OpenAIChatModel`), addressing #1386. Rebased onto `main` (latest commit `ded7ccd7`) today. Non-streaming path keeps `finish_reason`/`tool_calls`/`function_call` in `ChatResponse.metadata`; streaming path attaches `finish_reason` to `delta_res.metadata` (the upstream per-chunk-yield refactor required restructuring the post-loop block). Fork PR workflows need a one-time maintainer approval to run `Python Unittest Coverage`. Would appreciate review when convenient.

RerankerGuo:
Closing because the linked #1386 was marked completed by the maintainer on July 8 while this implementation remained unreviewed. The current model API has since been substantially refactored; reviving this old metadata contract without a new accepted requirement would create avoidable review and compatibility risk.

Issue #1722, opened 2026-05-29
feat(service): support multi-process and distributed deployment


DavdGao:
**Background**

The current agent service is a single-process implementation. To support **multi-process, distributed deployment**, we need to eliminate the architectural barriers that prevent horizontal scaling. The core principle is to establish a **single source of truth** by moving shared state out of local memory and into a centralized store (e.g., Redis, database).

Key barriers that need to be addressed include:

- Background task state and session replay data must be persisted in a centralized database rather than in-process memory
- Local filesystem and in-process managers are not compatible with a distributed environment

**Proposed Changes**

The following components require refactoring:

- **`SessionManager`**: Migrate to a database-backed implementation to ensure session state is accessible across all nodes
- **`BackgroundManager`**: Migrate to a database-backed implementation (e.g., Redis) to ensure high-speed, cross-process access to task state
- **`ScheduleManager`**: Requires a fundamental redesign. Key questions to resolve include:
  - On which node/process should a scheduled task run and be triggered?
  - How should cron tasks be distributed and assigned across different nodes and processes upon creation?

**Known Constraints**

- In distributed mode, **local filesystem workspaces and Docker workspaces cannot be used** and will need alternative solutions

---

> ⚠️ This issue is still under discussion. Feedback and suggestions are welcome.

YingchaoX:
Just sharing a thought here. I wonder if we can start with a lighter deployment model before introducing a full distributed manager design.

From my understanding, the current architecture already has some multi-process-friendly pieces: shared `StorageBase`, `MessageBus`, session locks, inbox, wakeup queue, event replay, etc. So maybe we can distinguish two deployment modes:

```mermaid
flowchart TB
  subgraph Single["Single-instance deployment"]
    A["Agent Service<br/>API Worker + Scheduler"]
  end

  subgraph Multi["Multi-instance deployment"]
    LB["Load Balancer"] --> W1["Service Worker"]
    LB --> W2["Service Worker"]
    LB --> WN["Service Worker"]

    Scheduler["Single Scheduler / External Scheduler"]
  end

  A --> Redis["RedisStorage + RedisMessageBus"]
  W1 --> Redis
  W2 --> Redis
  WN --> Redis
  Scheduler --> Redis

  Redis --> Queue["Wakeup Queue"]
  Queue --> W1
  Queue --> W2
  Queue --> WN
```

For single-instance deployment, keeping the scheduler embedded in the service process seems perfectly fine and matches the current code.

[...]

PR #1993, opened 2026-07-03
feat(service): allow disabling embedded scheduler

RerankerGuo:
## Summary

Adds `create_app(enable_scheduler=False)` so API-only service workers
don't double-spawn scheduled jobs in multi-process deployments:

- New lifespan option `enable_scheduler: bool = True` on `create_app`.
- When `False`, `SchedulerManager` skips local APScheduler
  start/restore/job registration.
- Schedule records remain usable for storage/listing so a dedicated
  scheduler process can still own firing.

Refs agentscope-ai/agentscope#1722.

## Maintainer design context

This implements the lighter-weight direction that maintainers
already endorsed on #1722:

> @YingchaoX (2026-06-10): "Just sharing a thought here. I wonder
> if we can start with a lighter deployment model before
> introducing a full distributed manager design. From my
> understanding, the current architecture already has some
> multi-process-friendly pieces: shared `StorageBase`, `MessageBus`,
> session locks, inbox, wakeup queue, ev..."

> @DavdGao (2026-06-11): "Agree with this. For the current
> scheduler refactoring, we're planning to add a configuration
> option to control whether the embedded scheduler thread is
> enabled..."

(`enable_scheduler=False` is exactly that configuration option.)

## Tests

`tests/service_scheduler_test.py` covers:

- `enable_scheduler=False` ⇒ `SchedulerManager.start()` does not
  start the APScheduler and skips restore/registration.
- `enable_scheduler=True` (default) preserves existing behavior.
- Lifespan path with `enable_scheduler=False` exits cleanly.

```
uv pip install -e .[dev]
pytest tests/service_scheduler_test.py tests/app_lifespan_dedicated_test.py
```

### DIFF
--- a/src/agentscope/app/_app.py
+++ b/src/agentscope/app/_app.py
@@ -49,6 +49,7 @@ def create_app(
     knowledge_chunker: ChunkerBase | None = None,
     blob_store: BlobStoreBase | None = None,
     enable_index_worker: bool = True,
+    enable_scheduler: bool = True,
     *,
     extra_credentials: list[Type[CredentialBase]] | None = None,
     extra_middlewares: list[FastAPIMiddleware] | None = None,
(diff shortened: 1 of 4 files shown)


RerankerGuo:
Rebased onto current main (`1aeb03d0`) after #1998 and resolved the lifespan composition with the new resource-access service. The refreshed branch is mergeable at `26d7d31`; all seven targeted scheduler/lifespan tests plus Black, flake8, and mypy pass locally.

RerankerGuo:
Closing because this old branch no longer forms a complete change against the current service architecture. It only disables the embedded scheduler and does not provide the dedicated scheduler startup path needed by #1722; it is also conflicted after the lifespan/service refactors. A renewed implementation should first define that current end-to-end deployment contract rather than expanding this stale PR.

PR #1997, opened 2026-07-03
feat(channel): implement channel

### LINKED ISSUE #2027: feat(channel): support app channels in agent service
**Background**

As discussed in #1834, while we don't want to get bogged down in the implementation details of individual channels, a multi-tenant service like agent service still requires well-defined channel abstractions, enabling developers to easily integrate with different apps and platforms.

**Changes**
- Add core abstractions for app channels
- Add several reference implementations, e.g. DingTalk, Feishu, and Discord

qbc2016:
## AgentScope Version

2.0.4dev

## Description

1. Implement channel
2. Add feishu channel

## Checklist


src/agentscope/app/storage/_model/_channel.py
```diff
@@ -0,0 +1,66 @@
+# -*- coding: utf-8 -*-
+"""The channel storage model."""
+from typing import Any, Literal
+
+from pydantic import BaseModel, Field
+
+
+DmScope = Literal["MAIN", "PER_PEER", "PER_CHAT", "PER_CHANNEL_PEER"]
+PermissionModeType = Literal["dont_ask", "bypass", "default"]
```
DavdGao: Why do we need to add new permission mode here? since we already have `agentscope.permission.PermissionMode`

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+    metadata: dict[str, Any] = Field(default_factory=dict)
+    """Platform-specific metadata: chat_id, chat_type, tenant_key, etc."""
+
+    received_at: str = Field(
+        default_factory=lambda: datetime.now().isoformat(),
+    )
+
+
+class ChannelCapability(BaseModel):
+    """Platform capability declaration for Gateway degradation decisions."""
```
DavdGao: Please add more detailed docstrings for this class. For example: what does `markdown=True` mean here? Is the `bool` bidirectional? What is the unit for `max_message_length`? Are there any maximum count limits for `image` and `file`?

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+    channel_id: str
+    """Source channel instance identifier."""
+
+    channel_user_id: str
+    """Platform-side unique user identifier."""
+
+    channel_message_id: str | None = None
+    """Platform-side message id for reply referencing."""
+
+    message: str
```
DavdGao: Should we consider using a unified content field here, similar to our other key abstractions: `list[TextBlock | DataBlock]`? We'd need to confirm that `TextBlock` and `DataBlock` cover all the data types required for attachments though.

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+    """Plain-text message content after normalisation."""
+
+    attachments: list[dict] = Field(default_factory=list)
```
DavdGao: How exactly is the data in `attachments` structured and consumed? As a unified `ChannelEvent` abstraction, this should be more explicitly defined rather than leaving it as an loosely-typed `dict`.

DavdGao: IMO, if `attachments` are meant to be consumed by the agent, they should be convertible into `msg.content`. If so, the conversion should be handled here rather than leaking channel-specific inconsistencies up to the agent layer.

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+class ChannelBase(ABC):
+    """Abstract base class for platform adapters.
+
+    A Channel handles three concerns:
+    1. Establish and maintain a long-lived connection (WebSocket/stream).
+    2. Normalise platform payloads into ``ChannelEvent``.
+    3. Send agent replies back to the platform.
+    """
+
+    _gateway: ChannelGateway | None = None
```
DavdGao: - What is the _gateway attribute used for?
- Please add detailed docstrings to improve readability and understandability.

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+    @abstractmethod
+    def channel_id(self) -> str:
+        """The unique channel instance identifier."""
+
+    @abstractmethod
+    async def start_listening(self) -> None:
+        """Establish long-lived connection and loop receiving events.
+
+        Must implement automatic reconnection with exponential back-off
+        (max 30s).
```
DavdGao: How is this 30s timeout enforced/constrained?

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+
+    @abstractmethod
+    async def normalize(self, raw_payload: dict) -> ChannelEvent | None:
+        """Convert a raw platform payload into a ``ChannelEvent``.
+
+        Returns ``None`` to indicate the event should be ignored.
+        Must extract ``chat_id`` into ``metadata["chat_id"]``.
```
DavdGao: If it must have a chat_id field, why not adding a `chat_id` field to the `ChannelEvent`?

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+        separate internal hook (e.g. ``_normalize_sdk_event``). In that case,
+        this method may return None unconditionally. It is retained in the
+        interface to support HTTP-webhook-based channels where raw JSON
+        payloads are POSTed to the server.
+        """
+
+    @abstractmethod
+    async def send_response(
+        self,
+        event: ChannelEvent,
```
DavdGao, line 290: At first glance, if `send_response` is meant to send the agent's response back to the platform (e.g. DingTalk), why is the `event` parameter typed as `Msg` rather than `ChannelEvent`?

That is, agentscope abstraction ==> different channel platforms

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+        response: str,
+    ) -> None:
+        """Send agent reply back to the platform."""
+
+    async def add_reaction(  # pylint: disable=unused-argument
+        self,
+        event: ChannelEvent,
+        emoji_type: str,
+    ) -> str | None:
+        """Add emoji reaction to message. Returns reaction id or None."""
```
DavdGao: - What's the direction of this method — agent service → channel platform, or the other way around?
- Why does `add_reaction` take a `ChannelEvent` as input?

Please add detailed docstrings, otherwise this code is quite hard to follow.

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+        return None
+
+    async def remove_reaction(
```
DavdGao, line 370: Similar as above

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+        self,
+        event: ChannelEvent,
+        reaction_id: str,
+    ) -> None:
+        """Remove a previously added reaction."""
+
+    def build_approval_card(  # pylint: disable=unused-argument
```
DavdGao: I'm curious how does the current design handle the hitl events here, and how does a platform that doesn't support card rendering handle the hitl events in their subclasses?

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,213 @@
...
+        self,
+        *,
+        request_id: str,
+        tool_name: str,
+        tool_input_summary: str = "",
```
DavdGao: What does `tool_input_summary` represent here, and what is its relationship to the JSON dictionary input?

src/agentscope/app/channel/_config.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+"""Channel module configuration."""
+from __future__ import annotations
+
+from typing import Any
+
+from pydantic import BaseModel, Field
+
+
+class DefaultSessionConfig(BaseModel):
```
DavdGao: Since we already have SessionConfig under `agentscope/app/storage`, do we need a duplicated session config class here?

src/agentscope/app/channel/_gateway.py
```diff
@@ -0,0 +1,950 @@
...
+from ._repository import ChannelRecord, ChannelStorageBase
+from ..storage import StorageBase
+from .._service import ChatService
+from .._manager import ChatRunRegistry
+
+_LOCK_PREFIX = "agentscope:channel:user_lock:"
+
+_TOOL_EVENT_TYPES: frozenset[str] = frozenset(
+    {
+        "TOOL_CALL_START",
```
DavdGao: Reuse `EventTypes` under `agentscope.event` rather than adding magic strings here

src/agentscope/app/channel/_gateway.py
```diff
@@ -0,0 +1,950 @@
...
+        "TOOL_RESULT_START",
+        "TOOL_RESULT_TEXT_DELTA",
+        "TOOL_RESULT_DATA_DELTA",
+        "TOOL_RESULT_END",
+    },
+)
+
+_THINKING_EVENT_TYPES: frozenset[str] = frozenset(
+    {
+        "THINKING_BLOCK_START",
```
DavdGao: Similar as above

src/agentscope/app/channel/_gateway.py
```diff
@@ -0,0 +1,950 @@
...
+        storage: StorageBase,
+        message_bus: MessageBus,
+        chat_service: ChatService,
+        chat_run_registry: ChatRunRegistry,
+        mapper: SessionMapperBase,
+        channel_storage: ChannelStorageBase,
+        session_config: DefaultSessionConfig,
+        response_timeout: float = 60.0,
+        concurrent_users_limit: int = 1000,
+    ) -> None:
```
DavdGao, line 54: Missing docstrings

src/agentscope/app/_router/_channel.py
```diff
@@ -0,0 +1,504 @@
+# -*- coding: utf-8 -*-
+"""Channel HTTP API routes.
+
+Endpoints:
```
DavdGao: Move routers into `agentscope/app/_routers`

src/agentscope/app/channel/_manager.py
```diff
@@ -0,0 +1,349 @@
...
+
+def _gen_node_id() -> str:
+    """Generate a unique node identifier for this process."""
+    import uuid
+
+    return uuid.uuid4().hex[:12]
+
+
+class ChannelManager:
+    """Manages the full lifecycle of channel instances."""
```
DavdGao: What's the difference between this manager and the channel gateway? They seem to serve a similar purpose.

src/agentscope/app/channel/_manager.py
```diff
@@ -0,0 +1,349 @@
...
+        self,
+        storage: StorageBase,
+        message_bus: MessageBus,
+        chat_service: ChatService,
+        chat_run_registry: ChatRunRegistry,
+        session_mapper: SessionMapperBase,
+        channel_storage: ChannelStorageBase,
+        config: ChannelConfig,
+        type_registry: ChannelTypeRegistry | None = None,
+    ) -> None:
```
DavdGao: missing docstrings here

src/agentscope/app/channel/_registry.py
```diff
@@ -0,0 +1,308 @@
...
+
+from pydantic import BaseModel, Field
+
+if TYPE_CHECKING:
+    from ._base import ChannelBase
+
+ChannelFactory = Callable[..., "ChannelBase"]
+
+
+class ChannelTypeSchema(BaseModel):
```
DavdGao, line 22: Should we use a plain `dict` here or define platform-specific classes instead, similar to how we handle model credentials? 

With the current approach, it's unclear how a developer would know how to correctly configure platform-specific fields — for example, how would they know what credentials are required for Feishu without digging into the implementation?

```Python
ChannelTypeSchema(
                channel_type="feishu",
                display_name="飞书 (Feishu/Lark)",
                description="飞书机器人，通过长连接模式接收消息",
                credentials_schema={
                    "type": "object",
                    "required": ["app_id", "app_secret"],
                    "properties": {
                        "app_id": {
                            "type": "string",
                            "title": "App ID",
                            "description": "飞书应用的 App ID",
                        },
                        "app_secret": {
                            "type": "string",
                            "title": "App Secret",
                            "description": "飞书应用的 App Secret",
                            "format": "password",
                        },
                    },
                },
                config_schema={
                    "type": "object",
                    "properties": {
                        "only_at_reply": {
                            "type": "boolean",
                            "title": "仅@时回复",
                            "description": "群聊中仅在被@时回复",
                            "default": True,
                        },
                    },
                },
                platform_bot_id_field="app_id",
            ),
```

src/agentscope/app/channel/_repository.py
```diff
@@ -0,0 +1,190 @@
...
+__all__ = [
+    "ChannelRecord",
+    "ChannelStorageBase",
+    "InMemoryChannelStorage",
+    "RoutingRule",
+    "StorageBackedChannelStorage",
+]
+
+
+class ChannelStorageBase(ABC):
```
DavdGao: Why should we add new storage class here rather than reuse the existing storage classes under `agentscope/app/storage`

DavdGao (changes requested):
@qbc2016 Please see my inline comments. I'll start by reviewing the core abstractions in this PR — once we've aligned on those, we can move on to discussing the rest of the implementation.

src/agentscope/app/storage/_model/_channel.py
qbc2016: fixed.

src/agentscope/app/channel/_base.py
qbc2016: Agreed. Added detailed docstrings for all `ChannelCapability` fields. Clarified that all boolean flags describe the **send** direction (agent → platform). `max_message_length` unit is specified as Unicode characters. Also added `interactive_card: bool` to indicate HITL card support (see comment 12).

src/agentscope/app/channel/_base.py
qbc2016: Agreed — migrated `ChannelEvent` to use `content: list[TextBlock | DataBlock]` as the **single source of truth**. Removed `message: str` and `attachments: list[dict]` as stored fields. A read-only `message` property is provided for convenience (concatenates all `TextBlock.text`). Platform adapters now construct content blocks directly during normalisation, and the gateway passes `event.content` straight to `UserMsg` — no intermediate conversion step needed.

src/agentscope/app/channel/_base.py
qbc2016: Agreed. Replaced the loosely-typed `list[dict]` with the unified `list[TextBlock | DataBlock]` content model (see comment 3). `ChannelAttachment` has been removed entirely — platform adapters now construct `DataBlock(source=Base64Source(...))` or `DataBlock(source=URLSource(...))` directly during normalisation.

src/agentscope/app/channel/_base.py
qbc2016: Added detailed docstring to `_gateway`, `set_gateway()`, and the `ChannelBase` class docstring. Clarified that: 
1. it's injected by `ChannelManager` at startup; 
2. channel adapters should only use it to dispatch events via `handle_event()`; 
3. they must not access gateway internals.

src/agentscope/app/channel/_base.py
qbc2016: Updated the `start_listening()` docstring to require exponential back-off for automatic reconnection without prescribing specific parameters — each platform adapter should choose values that suit its connection characteristics.

src/agentscope/app/channel/_base.py
qbc2016: Good point. Promoted `chat_id` to an explicit optional field in `ChannelEvent` (`chat_id: str | None = None`). Updated `normalize()` docstring to require populating this field. Gateway now uses `event.chat_id` directly instead of `event.metadata.get("chat_id")`.

src/agentscope/app/channel/_base.py
qbc2016: The `event` parameter is intentionally `ChannelEvent`, not `Msg`. Its purpose is to provide routing context (e.g. `chat_id`, `channel_message_id`) for delivering the reply to the correct chat/thread — not to carry the content being sent (that's the `response: str` parameter). Added detailed docstring clarifying the direction (agent service → channel platform) and the role of each parameter.

src/agentscope/app/channel/_base.py
qbc2016: Added detailed docstrings for `add_reaction` and `remove_reaction`. Clarified direction: **agent service → channel platform**. The gateway calls `add_reaction(event, "OnIt")` to signal "processing" on the original message, then `remove_reaction` after completion. The `event` parameter provides the `channel_message_id` for targeting the correct message.

src/agentscope/app/channel/_base.py
qbc2016: See comment above. Both methods now have full docstrings.

src/agentscope/app/channel/_base.py
qbc2016: Added `interactive_card: bool = False` to `ChannelCapability` and documented the full HITL degradation chain:
- Platforms with card support (e.g. Feishu) set `interactive_card=True` and override `build_approval_card`, `send_interactive_card`, `register_approval`, `resolve_approval`
- Platforms without card support use the base class defaults: `send_interactive_card` returns `None` → `_handle_approval` returns `None` → approval cannot be completed, treated as timeout/abort, tool execution will not proceed
- This behavior is documented in a comment block above the HITL methods

As a follow-up improvement, we plan to add **text-based fallback approval** for platforms without card support. The gateway would send a tool-call summary and wait for the user to reply "yes" / "no" (case-insensitive). The main implementation challenge is the per-user distributed lock: during `_handle_approval()`, the lock blocks incoming messages, so the "yes"/"no" reply cannot be received directly. Our preferred approach is to intercept approval replies at the message entry point (before or at the earliest stage of lock acquisition) and route them to `resolve_approval()`, bypassing the normal message flow. Input validation would handle yes/y/no/n responses and re-prompt on unrecognized input.

src/agentscope/app/channel/_base.py
qbc2016: Added detailed Args docstring to `build_approval_card`. `tool_input_summary` is a text summary of the tool call's input parameters — the gateway passes up to 500 characters of the original tool input JSON. Each parameter's purpose and relationship is now documented.

src/agentscope/app/channel/_config.py
qbc2016: Not a true duplicate — `DefaultSessionConfig` provides **module-level fallback values** (just `workspace_id` and `chat_model_config`), while `SessionConfig` in `storage` is the **full session configuration**. However, the naming was confusing. Renamed to `ChannelSessionDefaults`. Added docstring explaining the relationship with `SessionConfig`.

src/agentscope/app/channel/_gateway.py
qbc2016: Agreed. Replaced event-type magic strings in `_gateway.py` with `EventType` enum values from `agentscope.event`. Both `_TOOL_EVENT_TYPES` / `_THINKING_EVENT_TYPES` frozensets and the `_collect_from_bus` comparisons now use `EventType.REPLY_START`, `EventType.TEXT_BLOCK_DELTA`, etc.

src/agentscope/app/channel/_gateway.py
qbc2016: Both frozensets and all string comparisons are updated.

src/agentscope/app/channel/_gateway.py
qbc2016: Added comprehensive `Args` docstring to `ChannelGateway.__init__` covering all 9 parameters. Also updated the class-level docstring to clarify its role as "event-processing/data plane" with a cross-reference to `ChannelManager`.

src/agentscope/app/_router/_channel.py
qbc2016: Done.

src/agentscope/app/channel/_manager.py
qbc2016: They do have clearly distinct responsibilities:
- `ChannelManager` = **lifecycle / control plane**: CRUD, start/stop, enable/disable, multi-node lifecycle sync
- `ChannelGateway` = **event-processing / data plane**: per-event routing, session management, chat run, response collection, HITL approval

Updated both class-level docstrings with cross-references to make this distinction explicit.

src/agentscope/app/channel/_manager.py
qbc2016: Added.

src/agentscope/app/channel/_repository.py
qbc2016: This is not a new persistence backend. The production implementation (`StorageBackedChannelRepository`, renamed from `StorageBackedChannelStorage`) delegates to the existing `StorageBase` (Redis). The repository layer exists because:
1. `StorageBase` methods are user-scoped (`get_channel(user_id, channel_id)`), but channel runtime events arrive with only `channel_id`
2. The repository handles `channel_id → tenant_user_id` resolution and caching
3. `ChannelManager` needs a global `list_channels(enabled_only=True)` at startup

Renamed `ChannelStorageBase` → `ChannelRepositoryBase` and added a module-level docstring explaining this design rationale.

src/agentscope/app/channel/_registry.py
qbc2016: Done. Defined Pydantic models for all built-in platform credentials and configs: `FeishuCredentials`, `FeishuChannelConfig`, `DiscordCredentials`, `DingTalkCredentials`, `WeComCredentials`. The `_register_builtin_types()` method now uses `.model_json_schema()` to auto-generate the JSON Schema from these models. Benefits: developers can see required fields directly from the model class, IDE autocomplete works, and parameter validation is automatic. Third-party plugins can still supply raw JSON Schema dicts via `ChannelTypeSchema`.

src/agentscope/app/channel/_base.py
```diff
@@ -0,0 +1,416 @@
...
+    channel_id: str
+    """Source channel instance identifier."""
+
+    channel_user_id: str
+    """Platform-side unique user identifier."""
+
+    channel_message_id: str | None = None
+    """Platform-side message id for reply referencing."""
+
+    chat_id: str | None = None
```
DavdGao: Maybe str rather than str | None

src/agentscope/app/channel/_manager.py
```diff
@@ -0,0 +1,375 @@
...
+            exc = task.exception() if not task.cancelled() else None
+            return {
+                "status": "error" if exc else "stopped",
+                "error": str(exc) if exc else None,
+            }
+        return {"status": "running"}
+
+    # ── Internal lifecycle ──
+
+    async def _load_and_start_channels(self) -> None:
```
DavdGao: We should consider the distributed deployment with multiple nodes

src/agentscope/app/channel/_repository.py
```diff
@@ -0,0 +1,211 @@
...
+        self,
+        platform_bot_id: str,
+    ) -> ChannelRecord | None:
+        for record in self._store.values():
+            if record.platform_bot_id == platform_bot_id:
+                return record
+        return None
+
+
+class StorageBackedChannelRepository(ChannelRepositoryBase):
```
DavdGao: TODO: remove the repository

src/agentscope/app/channel/_gateway.py
```diff
@@ -0,0 +1,945 @@
...
+            )
+
+            if not done:
+                collect_task.cancel()
+                if run_task is not None:
+                    run_task.cancel()
+                raise TimeoutError("No response within timeout")
+
+            if collect_task in done:
+                text, confirm_data = collect_task.result()
```
DavdGao: If we need a time out setting here

DavdGao (approved):
lgtm

PR #2005, opened 2026-07-05
fix(tool): reject negative grep context flags

nolanchic:
## Summary

The `Grep` tool's `-A`, `-B`, `-C`, and `context` parameters were declared as plain integers with **no `minimum`** in the input schema and **no in-code guard**. A negative value was stringified and forwarded to ripgrep, which failed with an opaque message:

```
ripgrep error (code 2): rg: error parsing flag -C: value is not a valid number: invalid digit found in string
```

…instead of a clear validation error.

This is the **same class of bug as #1936** (negative `head_limit`/`offset`), which was fixed with a schema `minimum` plus an in-code guard. This PR applies the same treatment to the four context-line parameters.

## Changes

- `src/agentscope/tool/_builtin/_grep.py`
  - Added `"minimum": 0` to the input schema for `-A`, `-B`, `-C`, and `context`.
  - Added an in-code guard in `call()` (right after the flags are parsed) that returns a clear `ToolChunk(state=ERROR)` reading `Error: <flag> must be non-negative.` for any negative value — mirroring the existing `head_limit`/`offset` guards.
- `tests/builtin_grep_test.py`
  - `test_negative_context_returns_clear_error` — `context=-5` → ERROR mentioning `context` + `non-negative`.
  - `test_negative_context_flags_return_clear_error` — `-A`/`-B`/`-C` negative → ERROR naming the flag.
  - `test_positive_context_works` — `context=1` still returns SUCCESS (no regression).
  - `test_context_schema_rejects_negative` — asserts all four schema entries have `minimum: 0`.

## Verification

- [x] `pytest tests/builtin_grep_test.py -v` → 17 passed (13 existing + 4 new)
- [x] `pytest tests/builtin_grep_test.py tests/builtin_glob_test.py tests/builtin_read_test.py` → 37 passed, no regression
- [x] Manual repro: `context=-5` now returns `Error: context must be non-negative.` (state=error) instead of the opaque ripgrep error
- [x] `pre-commit run --files src/agentscope/tool/_builtin/_grep.py tests/builtin_grep_test.py` → all hooks pass (mypy / black / flake8 / pylint / pyroma)
- [ ] CI

### DIFF
--- a/src/agentscope/tool/_builtin/_grep.py
+++ b/src/agentscope/tool/_builtin/_grep.py
@@ -93,21 +93,25 @@ class Grep(ToolBase):
                 "type": "integer",
                 "description": "Number of lines to show after each match. "
                 "Requires output_mode: 'content'.",
+                "minimum": 0,
             },
             "-B": {
                 "type": "integer",
                 "description": "Number of lines to show before each match. "
                 "Requires output_mode: 'content'.",
+                "minimum": 0,
             },
             "-C": {
                 "type": "integer",
                 "description": "Alias for context.",
+                "minimum": 0,
             },
             "context": {
                 "type": "integer",
                 "description": "Number of context lines to show before and "
                 "after matches. Requires output_mode: "
                 "'content'.",
+                "minimum": 0,
             },
             "n": {
                 "type": "boolean",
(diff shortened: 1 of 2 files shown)


nolanchic:
Closing this — the underlying issue #1936 has already been closed as completed (by @DavdGao), so the negative-pagination gap it targets was resolved upstream through another change. Closing to keep the queue clean; no action needed from maintainers.

PR #2006, opened 2026-07-05
fix(formatter): include tool_name in ollama tool result messages

nolanchic:
## Summary

The Ollama chat formatter emitted tool-result messages as `{"role": "tool", "content": ...}`, **omitting the `tool_name` field** that Ollama's tool-calling API requires to correlate a result with its function.

Per the [Ollama tool-calling spec](https://docs.ollama.com/capabilities/tool-calling), the correct shape is:

```json
{"role": "tool", "tool_name": "<function>", "content": "..."}
```

Note this is **NOT** the OpenAI-style `tool_call_id` — Ollama uses `tool_name`. (My initial audit assumed OpenAI-compat `tool_call_id`; checking the official docs corrected this to `tool_name`.)

Without `tool_name`, **multi-turn and parallel tool use on Ollama cannot match results back to calls**, producing 400s or mis-pairing.

## Changes

- `src/agentscope/formatter/_ollama_formatter.py`
  - `OllamaChatFormatter.format` now emits `"tool_name": block.name` on every `role: tool` message. (`ToolResultBlock.name` is a required `str`, so it is always present.)
  - The `OllamaMultiAgentFormatter` delegates tool-sequence formatting to `OllamaChatFormatter.format` (`_format_tool_sequence`), so it is covered by the same change.
- `tests/formatter_ollama_test.py`
  - Updated the four existing ground-truth expectations that contained `role: tool` messages.
  - Added a dedicated regression test `test_tool_result_message_carries_tool_name` that asserts `tool_name` is present and that the OpenAI-style `tool_call_id` is **not** emitted.

## Verification

- [x] `pytest tests/formatter_ollama_test.py -v` → 9 passed (8 existing + 1 new regression)
- [x] `pre-commit run --files ...` → all hooks pass (mypy / black / flake8 / pylint / pyroma)
- [ ] CI

### DIFF
--- a/tests/formatter_ollama_test.py
+++ b/tests/formatter_ollama_test.py
@@ -112,7 +112,11 @@ class TestOllamaFormatter(IsolatedAsyncioTestCase):
                     },
                 ],
             },
-            {"role": "tool", "content": "The capital of Japan is Tokyo."},
+            {
+                "role": "tool",
+                "tool_name": "get_capital",
+                "content": "The capital of Japan is Tokyo.",
+            },
             {"role": "assistant", "content": "The capital of Japan is Tokyo."},
         ]
 
@@ -146,6 +150,7 @@ class TestOllamaFormatter(IsolatedAsyncioTestCase):
         }
         self._gt_tool_result = {
             "role": "tool",
+            "tool_name": "get_capital",
             "content": "The capital of Japan is Tokyo.",
         }
 
@@ -312,6 +317,7 @@ class TestOllamaFormatter(IsolatedAsyncioTestCase):
                 },
                 {
                     "role": "tool",
+                    "tool_name": "get_map",
                     "content": expected_tool_content,
                 },
                 {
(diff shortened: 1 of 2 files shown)


tests/formatter_ollama_test.py
```diff
@@ -574,3 +580,39 @@ async def test_chat_formatter_hint_block_multimodal(self) -> None:
             ],
             res,
         )
+
+    async def test_tool_result_message_carries_tool_name(self) -> None:
```
DavdGao, line 613: Please remove the unnecessary unit test, as tool_name is already covered by the tests above.

DavdGao (commented):
Plz solve the conflict and others lgtm

nolanchic:
Rebased onto the latest `main` and resolved the conflict in `tests/formatter_ollama_test.py` (kept both the `tool_name` regression test from this PR and the incomplete-input test from #2012). CI/pre-commit green, mergeable.

DavdGao (approved):

PR #2007, opened 2026-07-05
fix(formatter): drop empty text blocks for anthropic

nolanchic:
## Summary

The Anthropic formatter forwarded empty `TextBlock`s verbatim as `{"type": "text", "text": ""}`, which Anthropic rejects with a **400** ("text blocks must be non-empty"). Empty text blocks occur in practice after a **tool-call-only assistant turn** whose streamed text is empty.

## Changes

Three changes in `OllamaChatFormatter`'s sibling — `AnthropicChatFormatter` (`src/agentscope/formatter/_anthropic_formatter.py`):

1. Skip a top-level `TextBlock` whose `text` is empty (assistant/user turns).
2. Skip an empty `TextBlock` inside a `ToolResultBlock`'s output list, and skip an empty-string tool `output`.
3. If (2) empties the `tool_result` content list entirely, fall back to a non-empty placeholder `"(empty tool output)"` — Anthropic also rejects an empty `tool_result` content list, so this prevents trading one 400 for another.

## Tests (`tests/formatter_anthropic_test.py`)

Three new regression tests:
- `test_empty_text_block_is_dropped` — empty `TextBlock` before a `tool_use` → no empty text emitted
- `test_empty_tool_result_text_is_dropped` — empty `TextBlock` as a tool result → placeholder used
- `test_empty_string_tool_output_uses_placeholder` — empty-string tool `output` → placeholder used

## Verification

- [x] `pytest tests/formatter_anthropic_test.py -v` → 14 passed (11 existing + 3 new)
- [x] `pre-commit run --files ...` → all hooks pass (mypy / black / flake8 / pylint / pyroma)
- [ ] CI

### DIFF
--- a/tests/formatter_anthropic_test.py
+++ b/tests/formatter_anthropic_test.py
@@ -988,3 +988,180 @@ class TestAnthropicFormatter(IsolatedAsyncioTestCase):
         ][0]
         # Repaired to a dict rather than crashing.
         self.assertIsInstance(tool_use["input"], dict)
+
+    async def test_empty_text_block_is_dropped(self) -> None:
+        """Empty ``TextBlock`` must be skipped, not forwarded.
+
+        Anthropic rejects ``{"type": "text", "text": ""}`` with a 400
+        ("text blocks must be non-empty"). Empty text blocks arise in
+        practice after a tool-call-only assistant turn whose streamed text
+        block is empty. The formatter must drop them rather than emit them.
+        """
+        fmt = AnthropicChatFormatter()
+        msgs = [
+            AssistantMsg(
+                name="assistant",
+                content=[
+                    TextBlock(text=""),
+                    ToolCallBlock(
+                        id="call_1",
+                        name="get_weather",
+                        input='{"city": "Tokyo"}',
+                    ),
+                ],
+            ),
+        ]
+
+        res = await fmt.format(msgs)
+
+        # The empty TextBlock is dropped — the assistant message contains
+        # only the tool_use block, nothing else.
+        self.assertListEqual(
+            [
+                {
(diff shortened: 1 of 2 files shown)


tests/formatter_anthropic_test.py
```diff
@@ -956,3 +956,133 @@ async def test_chat_formatter_parallel_tool_results_merged(
             ],
             res,
         )
+
+    async def test_empty_text_block_is_dropped(self) -> None:
```
qbc2016, line 992: Could we assert the whole formatted result directly here, e.g. `self.assertEqual(expected, res) / self.assertListEqual(expected, res)`, instead of checking individual fields?
This would be more consistent with the existing formatter tests and would make the regression coverage stronger by catching any unexpected extra/missing blocks or structural changes in the full response.

qbc2016 (commented):
Thank you for this pr. Please see the inline comments.

nolanchic:
Good point — thanks. Updated the three empty-block regression tests to assert the whole formatted result with `self.assertListEqual(expected, res)` instead of checking individual fields, matching the existing formatter-test style. All 14 anthropic formatter tests pass.

qbc2016:
Please merge the latest main branch and fix the conflicts.

nolanchic:
Rebased onto the latest `main` and resolved the conflict in `tests/formatter_anthropic_test.py` (kept both the incomplete-input test from #2012 and the empty-block tests here). Also applied the earlier `assertListEqual` feedback so all three empty-block tests assert the whole formatted result. 15 anthropic formatter tests pass, mergeable.

qbc2016 (approved):
LGTM.

PR #2011, opened 2026-07-06
fix(permission): require review for mutating find actions

### LINKED ISSUE #2003: find mutating actions can be classified as read-only

Premsenareddy:
## Summary
- detect mutating find actions before Bash read-only prefix auto-allow
- keep normal read-only find searches auto-allowable, including action-like values passed to predicates such as -name
- add parser and permission-engine regressions for find -delete and related actions

## Root cause
find was treated as a read-only command by prefix, so actions such as find . -delete and find . -exec rm {} \; could be auto-allowed before the normal permission path.

## Validation
- .venv/bin/python -m pytest -p no:cacheprovider tests/permission_bash_parser_test.py tests/permission_engine_test.py -q
- .venv/bin/pre-commit run --files src/agentscope/tool/_builtin/_bash_parser.py tests/permission_bash_parser_test.py tests/permission_engine_test.py

Closes #2003

### DIFF
--- a/src/agentscope/tool/_builtin/_bash_parser.py
+++ b/src/agentscope/tool/_builtin/_bash_parser.py
@@ -132,6 +132,46 @@ READ_ONLY_COMMANDS = {
     "pip show",
 }
 
+FIND_MUTATING_ACTIONS = {
+    "-delete",
+    "-exec",
+    "-execdir",
+    "-ok",
+    "-okdir",
+    "-fls",
+    "-fprint",
+    "-fprintf",
+}
+
+FIND_PREDICATES_WITH_VALUE = {
+    "-anewer",
+    "-cnewer",
+    "-context",
+    "-exec",
+    "-execdir",
+    "-fprint",
+    "-fprintf",
+    "-fstype",
+    "-gid",
+    "-group",
+    "-ilname",
+    "-iname",
+    "-inum",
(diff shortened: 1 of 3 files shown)


DavdGao:
@Premsenareddy Please note the following guidelines to avoid duplicate efforts on the same Issue:

1. **Claim the Issue first**: Before starting development, please leave a comment on the Issue page to claim it, so other contributors know it's already being worked on.

2. **PR ownership rules**: For unassigned Issues with multiple PRs from the open-source community, we will merge based on **the order in which the PRs were created**.

3. **Avoid duplicate PRs**: If a PR has already been submitted for this Issue, please refrain from opening another one to reduce unnecessary maintenance overhead.

Closing this for now.

PR #1988, opened 2026-07-02
fix(agent): recover failed acting tool calls

RerankerGuo:
## Summary
- catch runtime exceptions raised during the agent acting/tool execution phase
- persist an error ToolResultBlock and mark the tool call finished instead of crashing the ReAct loop
- keep DeveloperOrientedException behavior unchanged
- add a regression test that simulates an acting-layer MCP failure and verifies the agent continues

## Tests
- /tmp/agentscope-ripgrep-test/bin/python -m pytest tests/agent_basic_test.py

Fixes #1985

PR #1995, opened 2026-07-03
feat(agent): support agent interruption with graceful context and tool state handling

iluv7:
## AgentScope Version
2.0.4dev

### Related Issue

Closes #1806
Related design discussion: https://github.com/agentscope-ai/agentscope/issues/1806#issuecomment-4840709638

### Background

Currently, there is no way to interrupt a running agent mid-execution. The existing cancel pattern (`CancelDispatcher` → `asyncio.Task.cancel()`) is used only for session deletion and tears down the task abruptly, leaving context potentially inconsistent.

After discussion, we aligned on using `asyncio.CancelledError` for real-time interruption: catch and normalize it at the Model and Tool layers into an `is_interrupted` flag, let the Agent layer check this flag and exit gracefully, add an outermost catch in `reply_stream` / `reply` for middleware-originated cancellations, protect the ChatService persistence path with `asyncio.shield`, and auto-trigger interruption when the SSE stream disconnects to prevent orphan tasks.

### Interrupt signal path

- `MessageBusKeys` adds `session_interrupt_channel`, separate from the existing `session_cancel_channel` (hard cancel for session deletion)
- `CancelDispatcher` adds `_interrupt_loop` subscription loop; on receiving a broadcast, cancels the local chat-run task via `ChatRunRegistry`
- Session router adds `POST /sessions/{session_id}/interrupt` endpoint

### SSE disconnect auto-interrupt

When a client disconnects (tab close, refresh, network drop), the `_sse_generator` finally block checks if the agent is still running (`is_locked`) and publishes an interrupt signal, preventing orphan tasks from wasting resources.

### Model layer CancelledError catch (9 providers)

All 9 providers catch `(CancelledError, GeneratorExit)` inside their `_parse_stream_response`:

- `openai_chat`, `deepseek`, `moonshot`, `xai`, `dashscope`, `gemini`, `ollama`, `openai_response`, `anthropic`

Each preserves accumulated partial content from `acc_text` / `acc_thinking` / `acc_tool_calls` and returns a final `ChatResponse` with `is_interrupted=True`.

### Tool layer CancelledError catch

`Toolkit.call_tool` catches `CancelledError` during streaming tool execution, preserves accumulated partial content, appends `<system-reminder>The tool call is interrupted by the user.</system-reminder>`, and returns `ToolResponse` with `is_interrupted=True`.

### Agent layer graceful exit

- Adds `_interrupted_in_cycle: bool` flag to Agent
[...]

Issue #1985, opened 2026-07-02
[Bug]: Agent crashes when tool call fails due to missing exception handling in _acting()


Marlin-Phone:
## 问题描述

当 Agent 调用 MCP 工具失败时（如 MCP Server 返回 HTTP 401），`ReActAgent._acting()` 方法中的工具调用异常未被捕获，直接向上传播导致整个 Agent 循环崩溃。表现为：

- 微信渠道：typing 指示器永久旋转，用户收不到任何回复
- 消费者线程死锁，后续消息被丢弃

## 根因分析

`_acting()` 方法（`_react_agent.py:657`）的 `try` 块只包含 `finally` 子句来记录工具结果消息，但缺少 `except` 子句。当 `toolkit.call_tool_function(tool_call)` 抛出异常时，由于没有异常处理：

1. 异常直接向上传播到调用方
2. Agent 的推理循环被中断
3. 消费者永远等待不到响应

```python
try:
    tool_res = await self.toolkit.call_tool_function(tool_call)
    # ... process chunks ...
finally:
    await self.memory.add(tool_res_msg)  # 只记录，不处理异常
```

## 修复方案

在 `try` 和 `finally` 之间添加 `except Exception` 块，将异常信息作为工具返回结果写入 `ToolResultBlock`，使 Agent 能够看到错误并继续推理循环，而非直接崩溃：

```python
except Exception as e:
    error_msg = f"Tool call failed: {e}"
    tool_res_msg.content[0]["output"] = error_msg
    await self.print(tool_res_msg, True)
    return None
```

## 修改文件

- `src/agentscope/agent/_react_agent.py`：在 `_acting()` 方法的 `try` 块中添加 `except Exception` 子句

DavdGao:
@Marlin-Phone I'm a bit confused about this issue. Within the [toolkit](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/tool/_toolkit.py#L339), we already have a `try/except` block that catches MCP errors, converts them into error tool responses, and handles tool execution failures gracefully. Could you provide more details about the specific error class being raised in your case?

Marlin-Phone:
@DavdGao Hi, thanks for the reply! Let me clarify — this is not about the MCP toolkit layer.

In **agentscope v1.0.20** (the version QwenPaw currently bundles), `ReActAgent._acting()` in `_react_agent.py` has a `try...finally` structure but **no `except`** clause:

```python
[...]

PR #2023, opened 2026-07-08
fix(agent): recover runtime exceptions raised by tool transports (#1985)

RerankerGuo:
## Summary

Wraps the `_acting` iteration inside `Agent._execute_tool_call`
with a focused `try/except` so that runtime exceptions raised by
tool transports (e.g. an MCP HTTP transport that bypasses the
`MCPError` path) are converted into a `ToolResultState.ERROR`
result instead of tearing down the agent reply loop.

- `DeveloperOrientedException` is re-raised so programming and
  misconfiguration errors still surface during development.
- Any other `Exception` is logged, delegated to the existing
  `_handle_error_tool_call` helper, and the agent loop continues
  with the tool error visible to the next LLM call.

Refs agentscope-ai/agentscope#1985.

## Maintainer context

This supersedes #1988, which closed after @DavdGao asked for
root-cause investigation on #1985. After @DavdGao's initial
reading ("the toolkit already catches MCP errors and converts them
to error tool responses"), reporter @Marlin-Phone pushed back on
2026-07-06 with the actual root cause:

> "In agentscope v1.0.20 (the version QwenPaw currently bundles),
> `ReActAgent._acting()` in `_react_agent.py` has a `try...finally`
> structure but no `except` clause... I also checked v2.0.3: the
> architecture changed, but a similar gap exists at
> `_execute_tool_call()` L1478 where `async for chunk in
> self._acting(tool_call)` runs without a try/except — only input
> validation errors (`AgentOrientedException`) are caught at
> L1360."

The protected code path is now `src/agentscope/agent/_agent.py`
`_execute_tool_call` at line ~1720 (was 1478 in v2.0.3; the line
number shifted as the function grew during the `_agent.py`
refactors in #1995 / 0c807329).

## Tests

`tests/agent_acting_recovery_test.py` covers:

- `test_runtime_exception_in_acting_does_not_crash_loop`: a tool
  that raises `RuntimeError` triggers a `ToolResultEndEvent(state=
  error)` and the agent loop reaches `ReplyEndEvent`.
- `test_developer_oriented_exception_still_propagates`:
  `DeveloperOrientedException` raised from `_acting` propagates
  out of `reply_stream` (not silently swallowed).

```
uv pip install -e .[dev]
pytest tests/agent_acting_recovery_test.py
```

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -51,7 +51,10 @@ from ..event import (
     ReplyEndReason,
     UserInterruptEvent,
 )
-from ..exception import AgentOrientedException
+from ..exception import (
+    AgentOrientedException,
+    DeveloperOrientedException,
+)
 from ..model import (
     ChatResponse,
     ChatUsage,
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Friendly ping — focused fix for #1985. After my earlier #1988 was closed (@DavdGao asked for root-cause investigation), @Marlin-Phone pushed back on the toolkit-handles-it reading and pointed at the unprotected `_execute_tool_call` L1720 path on current 2.x main. This PR wraps the `_acting` iteration with a `try/except` that lets `DeveloperOrientedException` propagate (preserving the developer-facing semantics) and converts any other runtime failure into a `ToolResultState.ERROR` via the existing `_handle_error_tool_call` helper. Adds `tests/agent_acting_recovery_test.py` (recovery + propagation). Rebased onto `ded7ccd7`. Would appreciate review — supersedes the closed #1988 in scope.

lt91888:
+1 — this is the same issue I was trying to fix in #2031, but this PR does it properly with DeveloperOrientedException handling and tests. Looks good to me.

DavdGao:
@RerankerGuo Plz see comments in #1985, the cause of the error requires further investigation

RerankerGuo:
Thanks for tracing the current call path. I agree the proposed test bypasses Toolkit.call_tool by mocking _acting directly, so it does not demonstrate an exception that can escape the toolkit's existing generator-creation and generator-consumption handlers. Without a minimal reproduction on current main, the extra catch in Agent._execute_tool_call would be speculative and could hide framework-level failures. Closing this PR; it can be revisited if #1985 gains a concrete escaping code path.

Issue #1698, opened 2026-05-26
perf(event): add status event into the agent event system


DavdGao:
**Background**

We have established an event system to support frontend interactions and human-in-the-loop (HITL) workflows. Currently, when the agent performs time-consuming internal operations — such as compacting the context — there is no way to notify the frontend, leaving users uninformed about what is happening in the background.

**Changes**

- Introduced a general-purpose status event mechanism that allows the backend to emit status notifications to the frontend during time-consuming operations.
- These events are designed to be operation-agnostic, enabling users to customize the event content rather than being tied to any specific operation, ensuring flexibility and extensibility for future use cases.

ZxlDragonDoctor:
I'd like to work on this issue.

**Proposed scope (in / out)**
- In: introduce a general-purpose status event in the agent event system that backends can emit during long-running internal operations (e.g. context compaction), delivered to the frontend/HITL consumers through the existing event channel.
- Out: coupling the event to a single operation name, redesigning the whole event bus, or WebUI-only presentation changes beyond what the event payload already enables.
- Tests: unit coverage for status event emission/payload shape and that existing event consumers still receive their current events unchanged.

If this direction looks right, I'll proceed with a minimal implementation aligned with the current event APIs.

ZxlDragonDoctor:
Opened PR # for this: introduces an operation-agnostic `StatusEvent` in the agent event system (`EventType.STATUS`), exports it on `AgentEvent`, maps it in the AGUI protocol middleware, and emits a completed status when context compaction actually shortens the conversation.

Testing: `PYTHONPATH=src python -m pytest tests/event_test.py -q` ? `4 passed, 2 subtests passed`.

PR #2030, opened 2026-07-09
feat(event): add status event

RerankerGuo:
## Summary
- add a general-purpose StatusEvent to the agent event union
- export StatusEvent from agentscope.event and make message replay ignore it as a non-content event
- map StatusEvent to an AGUI CUSTOM status event and cover direct/SSE conversion paths

## Tests
- PYTHONPATH=/tmp/agentscope-issue1698/src:/tmp/agentscope-issue1698/tests /tmp/agentscope-issue1999/.venv/bin/python -m pytest tests/agui_protocol_test.py tests/event_to_message_test.py

Refs #1698


RerankerGuo:
Closing because this core `StatusEvent` addition has no maintainer design approval and the linked #1698 is stale. Current service-layer extensibility uses `CustomEvent` specifically to avoid adding application-specific signals to the core agent event union, so this older approach conflicts with the direction now present on `main`.

Issue #1388, opened 2026-03-31
[Bug]: Agent compress memory: summary_template can not handle summary without 'next_steps'


ceshihao:
**<u>AgentScope is an open-source project. To involve a broader community, we recommend asking your questions in English.</u>**

**Describe the bug**
```
  File "/home/admin/.local/lib/python3.11/site-packages/agentscope/pipeline/_functional.py", line 192, in stream_printing_messages
    raise exception from None
  File "/home/admin/.local/lib/python3.11/site-packages/agentscope/agent/_agent_base.py", line 455, in __call__
    reply_msg = await self.reply(*args, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/admin/.local/lib/python3.11/site-packages/agentscope/agent/_agent_meta.py", line 120, in async_wrapper
    current_output = await original_func(
                     ^^^^^^^^^^^^^^^^^^^^
  File "/home/admin/.local/lib/python3.11/site-packages/agentscope/tracing/_trace.py", line 431, in wrapper
    raise e from None
  File "/home/admin/.local/lib/python3.11/site-packages/agentscope/tracing/_trace.py", line 422, in wrapper
    res = await func(self, *args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/admin/.local/lib/python3.11/site-packages/agentscope/agent/_react_agent.py", line 434, in reply
    await self._compress_memory_if_needed()
  File "/home/admin/.local/lib/python3.11/site-packages/agentscope/agent/_react_agent.py", line 1111, in _compress_memory_if_needed
    self.compression_config.summary_template.format(
KeyError: 'next_steps'
```

**To Reproduce**

The issue only happens sometimes when compressed model does not contain 'next_steps'.

```
agentscope==1.0.16
```
my code simple react agent
```
        agent = ReActAgent(
            name="xxx",
            model=DashScopeChatModel(
                model_name="qwen3.5-plus",
                enable_thinking=False,
                multimodality=True,
                stream=True
            ),
            sys_prompt=sys_prompt,
            toolkit=toolkit,
            memory=RedisMemory(),
            formatter=DashScopeChatFormatter(),
            plan_notebook=plan_notebook,
            max_iters=20,
            compression_config=ReActAgent.CompressionConfig(
                enable=True,
                agent_token_counter=CharTokenCounter(),
                trigger_threshold=50000,
[...]

PR #2032, opened 2026-07-09
fix(agent): tolerate missing compression summary fields

RerankerGuo:
## Summary
- avoid crashing context compression when a structured summary omits a field referenced by the summary template
- render missing summary fields as empty strings and log a warning
- add a regression test for a missing next_steps field

## Tests
- PYTHONPATH=/tmp/agentscope-issue1388/src:/tmp/agentscope-issue1388/tests /tmp/agentscope-issue1999/.venv/bin/python -m pytest tests/compress_context_test.py
- PYTHONPATH=/tmp/agentscope-issue1388/src /tmp/agentscope-issue1999/.venv/bin/python -m py_compile src/agentscope/agent/_agent.py

Refs #1388

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -94,6 +94,18 @@ else:
     MiddlewareBase = Any
 
 
+class _SummaryFormatDict(dict):
+    """Format mapping that leaves missing summary fields empty."""
+
+    def __missing__(self, key: str) -> str:
+        """Return an empty value for optional/missing summary fields."""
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Closing because #1388 is a stale AgentScope 1.x report and this PR does not establish the same missing-field behavior against the current 2.x structured summary schema. The current schema requires the default template fields; a fallback that silently empties arbitrary missing values needs a current reproduction and maintainer-approved contract first.

PR #2046, opened 2026-07-10
fix(workspace): fix OpenSandbox bootstrap and state filtering

jwx0925:
## Summary

- use the OpenSandbox SDK's `SandboxState` constants when filtering running and paused sandboxes
- stop installing `procps` in the stock OpenSandbox bootstrap path

## Why

The lifecycle API expects title-cased state wire values such as `Running` and `Paused`; the previous uppercase strings are rejected by compatible servers.

Installing `procps` also made the shared `pkill -f _mcp_gateway_app.py || true; nohup ...` launch command active in the stock OpenSandbox image. The process match includes the launcher command itself, so the gateway never starts and no gateway log is created. Other workspace images do not install `procps`, and the guarded cleanup remains a no-op there.

## Impact

Fresh OpenSandbox workspaces can complete bootstrap and start the MCP gateway. Existing callers keep the same default image and public configuration surface.

## Validation

- Python syntax checks for the changed modules
- `git diff --check`
- OpenSandbox unit modules load successfully; 12 live tests skip as designed when no endpoint is configured
- local Docker OpenSandbox E2E passed against `http://127.0.0.1:32889`, covering sandbox creation, gateway health, DeepSeek tool calls, Bash, text and binary file I/O, Glob, Grep, list/stat/delete, pause, resume, and kill

### DIFF
--- a/src/agentscope/workspace/_opensandbox/_constants.py
+++ b/src/agentscope/workspace/_opensandbox/_constants.py
@@ -12,7 +12,7 @@ base class, not here.
 
 #: Default OpenSandbox image. The slim Python image is small but still
 #: has enough package-manager support for installing curl, certificates,
-#: ripgrep, and procps during bootstrap.
+#: and ripgrep during bootstrap.
 DEFAULT_IMAGE = "python:3.11-slim"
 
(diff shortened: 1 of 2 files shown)


DavdGao (approved):
lgtm

PR #2001, opened 2026-07-04
feat(middleware): add on_check_permission hook

Miracle778:
> [!NOTE]
> **Scope update**
>
> This PR originally implemented structured decision-path observability and
> separate read-only hooks for permission decisions and confirmations.
>
> Following the maintainer feedback in #2000, the PR has been narrowed to one
> standard onion-style `on_check_permission` hook. The structured evaluation
> types and confirmation hook have been removed. Decision-path observability
> will be discussed separately.

## AgentScope Version

2.0.4.post1

## Description

Closes #2000.

Adds a standard onion-style `on_check_permission` middleware hook around
permission checking.

The hook provides a public extension point for both permission auditing and
application-owned authorization policies, while keeping
`PermissionEngine.check_permission` as the sole built-in engine interface.

## Background

AgentScope performs permission checking after tool resolution and input
validation, but before `on_acting`.

This means existing middleware cannot observe the permission decision itself:

- ASK and DENY return before `on_acting`;
- ALLOW reaches `on_acting`, but only as an already-permitted execution;
- application policies based on user, role, or tenant cannot be composed around
  the built-in permission check through a public middleware interface.

### `on_check_permission` middleware hook

Adds the following hook to `MiddlewareBase`:

```python
async def on_check_permission(
    self,
    agent: Agent,
    input_kwargs: dict,
    next_handler: Callable[..., Awaitable[PermissionDecision]],
) -> PermissionDecision:
    ...
```

The hook runs after tool resolution and input validation, and before Agent
consumes the returned decision.

Its `input_kwargs` contains:

- `tool_call`;
- the resolved `tool`;
- parsed and validated `tool_input`.

As a standard middleware onion, a middleware may:

- delegate with `next_handler(**input_kwargs)`;
- observe or replace the downstream decision;
- return a decision without delegating, short-circuiting the remaining chain.

The built-in `PermissionEngine.check_permission` is the innermost handler.
With no permission middleware registered, Agent follows the existing direct
engine path.

### Confirmed tool calls

A tool call resumed after user confirmation still traverses
`on_check_permission` immediately before execution.

The innermost handler returns the already-confirmed ALLOW without re-evaluating
the built-in engine. This keeps application policies and auditing active while
preserving the existing confirmation behavior.

### Input isolation

The middleware chain receives copies of `tool_call` and `tool_input`.

Middleware may forward modified copies to downstream permission middleware or
the built-in permission check, but those changes do not alter the tool call or
input eventually executed by Agent.

### Runnable example

[...]

PR #2054, opened 2026-07-12
feat(permission): permission decision and confirmation observability

Premsenareddy:
## Summary

Add opt-in, read-only middleware hooks that make every permission decision and user-confirmation outcome observable through `extra_agent_middlewares`, closing the audit blind spot where ASK/DENY and BYPASS-suppressed decisions were invisible to middleware.

### Permission decision observability (`on_permission_decision`)

- `PermissionEvaluation` dataclass (effective + candidate decision + `PermissionResolution` enum) returned by new `evaluate_permission`; `check_permission` delegates to it for backward compatibility.
- All five `_check_<mode>` methods preserve the candidate ASK when a mode transforms it: `BYPASS_ASK_SUPPRESSED`, `ASK_CONVERTED_TO_DENY`, `ASK_OVERRIDDEN_BY_ALLOW_RULE`, `USER_CONFIRMED`, `DIRECT`.
- `on_permission_decision` read-only hook fires before the agent consumes the decision; Agent covers the user-confirmed reuse path.

### Permission confirmation observability (`on_permission_confirmation`)

- Separate hook for user approval/rejection results. Confirmation is user input, not an engine decision, so it is kept out of `PermissionEvaluation`.
- Fires before state transition, `add_rule`, and execution.

### Read-only contract enforced by isolation

Both hooks deep-copy evaluation/tool_input/tool_call (decision) and tool_call/rules (confirmation) per observer, so observers cannot mutate the consumed result, executed input, or applied rules. `agent` and `tool` are passed as-is (live runtime context and toolkit-shared instance, matching every other middleware hook's convention).

### Runnable permission audit example

- Adds `examples/permission_audit_service` and focused importlib-based tests demonstrating permission decision and confirmation audit records.

## Tests

- Engine: per-mode evaluation semantics (effective/candidate/resolution) including BYPASS suppression, DONT_ASK conversion, and allow-rule override.
- Agent integration: ALLOW/DENY/ASK/BYPASS-suppress/USER_CONFIRMED/external/multi-call/no-observer/exception/mutation-isolation.
- `check_permission` backward-compat regression.
- No real destructive commands run — side-effect-free demo tools emit safety ASKs instead.

No changes to event/message/storage/UI schemas or permission behavior.

Closes #2001


Miracle778:
This PR contains the same implementation, tests, documentation, and example assets as my earlier PR #2001. The corresponding commits have identical stable patch IDs.

Since #2001 is the original PR and is still under review, please close this duplicate PR.

Premsenareddy:
Thanks for pointing this out. Closing this PR as a duplicate of #2001 so review can stay focused on the original implementation.

PR #2056, opened 2026-07-12
fix(workspace): restore default glob helper path

### LINKED ISSUE #2055: [Bug]: LocalWorkspace.list_tools raises AttributeError on current main
## Description

Image

`LocalWorkspace.list_tools()` raises the following exception on the current `main` branch:

```text
AttributeError: 'LocalWorkspace' object has no attribute '_glob_helper_path'
```

This prevents the agent service from constructing the toolkit for a local workspace when a user starts a chat.

## Reproduction

Environment:

- AgentScope: editable installation from current `main`
- Python: 3.11.12
- Workspace: `LocalWorkspace`

```python
import asyncio
import tempfile

async def main() -> None:
    with tempfile.TemporaryDirectory() as path:
        workspace = LocalWorkspace(workdir=path)
        await workspace.initialize()
        try:
            await workspace.list_tools()
        finally:
            await workspace.close()


asyncio.run(main())
```

## Root cause

This appears to be a regression introduced by commit `81538d35`.

The v2.0.4 release and the parent of that commit both define the following default on `WorkspaceBase`:

```python
_glob_helper_path: str | None = None
```

The commit removed that attribute while retaining the access in `WorkspaceBase.list_tools()`. `SandboxedWorkspaceBase` overrides `_glob_helper_path`, but `LocalWorkspace` inherits directly from `WorkspaceBase`, so the attribute is missing.

## Suggested fix

Restore the default `_glob_helper_path: str | None = None` class attribute on `WorkspaceBase` and add a regression test covering `LocalWorkspace.list_tools()`.

@DavdGao, I would be happy to submit the small fix and regression test if this approach looks good.

dailingtao:
## AgentScope Version

`2.0.4.dev0` (current `main`)

## Description

Fixes #2055.

### Background

`WorkspaceBase.list_tools()` uses `_glob_helper_path` when constructing the builtin `Glob` tool. Commit `81538d35` removed the default class attribute from `WorkspaceBase`, while `LocalWorkspace` still inherits directly from that base class. As a result, starting a chat with a local workspace fails while building the toolkit:

```text
AttributeError: 'LocalWorkspace' object has no attribute '_glob_helper_path'
```

### Changes

- Restored the default `_glob_helper_path: str | None = None` class attribute on `WorkspaceBase`.
- Added a regression test verifying that `LocalWorkspace.list_tools()` returns all six builtin tools backed by `LocalBackend`.

Sandboxed workspaces continue to override `_glob_helper_path` with their backend-side helper script path.

### Validation

```text
python -m pytest tests/workspace_local_test.py -q
18 passed

pre-commit run --files src/agentscope/workspace/_base.py tests/workspace_local_test.py
All hooks passed

python -m compileall -q src/agentscope/workspace/_base.py tests/workspace_local_test.py
git diff --check
```

### Breaking Changes

None.

## Checklist

- [x] Related documentation has been considered; no documentation change is needed for this regression fix

### DIFF
--- a/src/agentscope/workspace/_base.py
+++ b/src/agentscope/workspace/_base.py
@@ -149,6 +149,15 @@ class WorkspaceBase:
     _skill_lock: asyncio.Lock
     """Guards mutation of the ``skills/`` directory."""
 
+    _glob_helper_path: str | None = None
+    """Optional path (backend-side) to the ``Glob`` helper script.
+
+    ``None`` means the :class:`Glob` builtin tool falls back to its
+    default behaviour (suitable for :class:`LocalBackend`). Remote
+    backends override this with a sandbox-/container-side script path
+    so :class:`Glob` can run efficiently inside the workspace.
+    """
+
     def __init__(
         self,
         *,
--- a/src/agentscope/workspace/_base.py
+++ b/src/agentscope/workspace/_base.py
@@ -149,14 +149,16 @@ class WorkspaceBase:
     _skill_lock: asyncio.Lock
     """Guards mutation of the ``skills/`` directory."""
 
-    _glob_helper_path: str | None = None
-    """Optional path (backend-side) to the ``Glob`` helper script.
+    @property
+    def _glob_helper_path(self) -> str | None:
+        """Optional path (backend-side) to the ``Glob`` helper script.
 
-    ``None`` means the :class:`Glob` builtin tool falls back to its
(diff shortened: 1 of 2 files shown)


dailingtao:
Effect after repair
image

DavdGao:
@dailingtao Thanks for pointing out the issue and plz run `pre-commit run --all-files` to format the code

DavdGao (approved):

PR #2057, opened 2026-07-12
feat(app): support forking sessions

### LINKED ISSUE #2041: feat(service): support forking a session for the same agent
**Background**

In multi-session agent services, users may want to branch off an existing conversation — for example, to explore an alternative direction without losing the original session context. Supporting session forking enables this workflow in a clean and controlled manner.

**Changes**
- Add a `fork session` endpoint on the backend, taking `user_id`, `agent_id`, and `session_id` as input and returning a new forked `session_id`; forking is restricted to sessions owned by the requesting user — cross-user and cross-agent forks are not permitted
- Add a **Fork** option in the dropdown menu of each session entry in the chat page sidebar on the frontend

Trefoi1:
## AgentScope Version

`2.0.4.dev0`

## Related Issue

Fixes #2041

## Background

AgentScope did not previously provide an end-to-end way to fork an existing session.

This PR adds session forking across the storage, service, API, and Web UI layers. A fork preserves the persisted session configuration, state, and message history while creating independent session records with new identities.

### Storage

* Added regular Session Fork support.
* Added Schedule Session Fork support.

  * Forked Schedule Sessions become normal user-created Sessions.
  * `source_schedule_id` is cleared.
* Added Team Session Fork support.

  * Supports Teams with created members.
  * Supports Teams with invited members.
  * Supports mixed created and invited members.
* Creates a new Team when forking a Team Leader Session.
* Clones created member Agents and creates new member Sessions.
* Reuses invited Agent records and Agent data while creating new owner-scoped Team Sessions.
* Preserves message history and existing TTL behavior.
* Uses Redis `WATCH` and `MULTI/EXEC` for optimistic concurrency control.
* Validates source graph identity, ownership, provenance, indexes, and key types.
* Detects target key collisions and rebuilds the complete Fork plan when required.
* Adds atomic and validated migration for legacy Team `member_ids`.

### Team Invitation

* Fixed invited Agent owner namespace handling.
* Keeps the invited Agent's standalone Session unchanged.
* Creates a separate Team-scoped Session under the invited Agent owner namespace.
* Does not modify the invited Agent index.

### Service

* Added `SessionService.fork_session()`.
* Validates Session ownership before reading runtime status.
* Rejects cross-user and cross-Agent requests as not found.
* Requires the source Session to be idle before invoking the Storage Fork operation.
* Checks normalized Team member statuses before forking a Team Leader.
* Leaves Team Worker and corrupted graph decisions to the Storage layer.
* Reads MessageBus run locks only for status checks.
* Does not copy or modify MessageBus runtime state.

### API

Added the following endpoint:

`POST /sessions/{session_id}/fork?agent_id={agent_id}`

Successful requests return `201 Created` with:

`{"session_id": "new-session-id"}`

Error mapping:

* Inaccessible or missing Session: `404 Not Found`
* Running, waiting, or unsupported Fork: `409 Conflict`
* Corrupted persisted graph: `500 Internal Server Error`

Internal Redis and graph error details are logged on the server and are not exposed to clients.

### Frontend

* Added “Fork” / “创建副本” to the Session menu.
* Calls the new Fork endpoint using the currently selected Agent.
* Refreshes the Session list after a successful Fork.
* Navigates directly to the newly created Session.
* Prevents duplicate Fork requests while a Fork is in progress.
* Keeps Team Leader, Worker, Schedule, and ownership decisions on the backend.

## Fork Semantics

* Each forked Session receives a new Session ID.
* Forked records are independent from the source records.
* Persisted messages are copied.
* Existing `workspace_id` references are preserved, so Workspace files remain shared.
* Schedule provenance is not inherited.
* Team Worker Sessions cannot be forked directly.
* Invited Agents are reused rather than cloned.
* MessageBus events, inboxes, locks, cancellation state, and transient runtime state are not copied.

## Safety and Consistency

* Cross-user Fork requests are rejected.
* Cross-Agent Fork requests are rejected.
* Team Worker Fork requests are rejected.
* Team Leader Fork requires the root Session and all normalized members to be idle.
* Source records are protected by Redis optimistic locking.
* Target records, messages, and indexes are validated before commit.
* Redis key physical types and business index categories are validated separately.
* Corrupted legacy Team graphs are rejected without modifying the source Team.
* Serialization and complete Fork planning occur before `MULTI`.

## Tests

* Service and Router tests: 20 passed
* Storage Fork regression tests: 47 passed
* RedisStorage regression tests: 56 passed
* Frontend format check: passed
* Frontend production build: passed
* ESLint: 0 errors, 16 pre-existing warnings
* pre-commit: passed
* `git diff --check`: passed

## Manual Validation

Verified:

* Regular Session menu displays the Fork action.
* Successful Fork returns `201` and a new Session ID.
* The UI navigates to the newly created Session.
* Forked messages match the source Session.
* Updating the Forked Session does not modify the source Session.
* Schedule Session Fork succeeds and is detached from its Schedule.
* Idle Team Leader Fork succeeds.
* Team Worker Fork returns `409`.
* Permission-waiting and external-result-waiting Sessions return `409`.
* Cross-user and cross-Agent requests return `404`.
* Rejected repeated requests do not create duplicate Sessions.
* Fork creates no copied MessageBus events, inboxes, locks, cancellation channels, or other transient runtime state.

A real running Agent was not manually triggered during validation. The `RUNNING` rejection path was not exercised manually, but is covered by automated Service tests.

### DIFF
--- a/src/agentscope/app/storage/_base.py
+++ b/src/agentscope/app/storage/_base.py
@@ -206,6 +206,22 @@ class StorageBase(ABC):
             `SessionRecord`: The created or updated record.
         """
 
+    async def fork_session(
+        self,
+        user_id: str,
+        agent_id: str,
+        session_id: str,
+    ) -> SessionRecord:
+        """Fork a session using the storage backend's native semantics.
+
+        Backends that do not implement session forking fail explicitly while
+        remaining instantiable during the phased rollout.
+        """
+        del user_id, agent_id, session_id
+        raise NotImplementedError(
+            f"{type(self).__name__} does not support session forking.",
+        )
+
     @abstractmethod
     async def set_session_team_id(
         self,
--- a/examples/web_ui/frontend/src/api/session.ts
+++ b/examples/web_ui/frontend/src/api/session.ts
@@ -20,6 +20,11 @@ export const sessionApi = {
 
 	create: (body: CreateSessionRequest) => client.post<CreateSessionResponse>('/sessions/', body),
 
+	fork: (sessionId: string, agentId: string) =>
+		client.post<CreateSessionResponse>(`/sessions/${sessionId}/fork`, undefined, {
+			agent_id: agentId,
+		}),
+
 	update: (sessionId: string, agentId: string, body: UpdateSessionRequest) =>
 		client.patch<SessionRecord>(`/sessions/${sessionId}`, body, { agent_id: agentId }),
 
--- a/examples/web_ui/frontend/src/hooks/useSessions.ts
+++ b/examples/web_ui/frontend/src/hooks/useSessions.ts
@@ -62,6 +62,22 @@ export function useSessions(agentId: string | null) {
 		[agentId, refetch],
 	);
 
+	/** Forks a session and returns the new Session ID. */
+	const fork = useCallback(
+		async (sessionId: string): Promise<string> => {
+			if (!agentId) throw new Error('No agent selected');
+			const response = await sessionApi.fork(sessionId, agentId);
+			const newSessionId = response.session_id;
+			try {
+				await refetch();
+			} catch {
+				// The API already created the Fork; navigation still uses its ID.
+			}
+			return newSessionId;
+		},
+		[agentId, refetch],
+	);
+
 	/** Deletes a session and refreshes the list. */
 	const remove = useCallback(
 		async (sessionId: string) => {
@@ -72,5 +88,5 @@ export function useSessions(agentId: string | null) {
 		[agentId, refetch],
 	);
 
-	return { sessions, loading, error, refetch, create, update, remove };
+	return { sessions, loading, error, refetch, create, update, remove, fork };
 }
--- a/examples/web_ui/frontend/src/i18n/locales/en.json
+++ b/examples/web_ui/frontend/src/i18n/locales/en.json
@@ -176,6 +176,7 @@
 		"placeholder": "Session name"
 	},
 	"session-menu": {
+		"fork": "Fork",
 		"rename": "Rename",
 		"delete": "Delete"
 	},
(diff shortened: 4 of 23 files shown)


src/agentscope/app/storage/_base.py
```diff
@@ -206,6 +206,22 @@ async def upsert_session(
             `SessionRecord`: The created or updated record.
         """
 
+    async def fork_session(
```
DavdGao, line 209: Considering that forking a session involves deep copying:
- The agent state
- Team members
  - Created agents: copy the agent records & team sessions
  - Invited agents: copy their team sessions
- Frontend display messages

Given this complexity, this would be better to be implemented as a service-layer operation rather than a storage-layer concern.

Trefoi1: Thanks for the suggestion. We agree that the service layer should own the
user-facing orchestration and runtime-state checks.

In this implementation, SessionService.fork_session() is intentionally kept
as the entry point for ownership validation, idle-state checks, and delegation.
The actual graph cloning remains behind StorageBase.fork_session() because a
Team Fork is a persistence-level transaction: it may update multiple
SessionRecords, AgentRecords, TeamRecords, message lists, and indexes, and must
use Redis WATCH/MULTI/EXEC with optimistic retries.

Keeping this operation in Storage also ensures that all backend-specific
details,
including Redis keys, TTLs, physical key types, and atomic writes, do not leak
into SessionService. The service does not perform the graph copy itself; it
only validates the request and delegates to Storage.

The frontend only consumes the newly created session ID and does not participate
in the graph-copy operation.

Trefoi1: The PR code checks are passing, but the GitHub Monitor workflow fails before running the notification script because STORAGE_REPO_TOKEN is not available:

Input required and not supplied: token

The failing step is the checkout of DavdGao/agentscope-monitor in
.github/workflows/github-monitor.yml. Could you configure or refresh this repository secret and rerun the workflow?

All code-related checks, including pre-commit and Python tests, are passing.

PR #2063, opened 2026-07-13
fix(model): reuse openai.AsyncClient across calls instead of new per call

dongfeng3692:
## PR Title Format
_call_api(及 tts 的 synthesize)在每次 API 调用时都新建一个 openai.AsyncClient。每个 client 各自持有一个独立的 httpx 连接池,因此这种写法等于:

每个请求都重新做 DNS + TCP + TLS 握手,连接无法复用;
高并发下连接建立本身成为瓶颈,延迟急剧恶化。
OpenAI 官方明确建议 client 实例化一次、全程复用。

下面时复用和非复用的性能报告
| 并发 | 策略 | p50 (ms) | p90 (ms) | p99 (ms) | 吞吐 (req/s) | 成功率 |
| --- | --- | --- | --- | --- | --- | --- |
| 5 | 旧-每次新建 | 19,325 | 31,784 | 34,813 | 0.1 | 100% |
| 5 | 新-复用 | 2,675 | 2,864 | 2,919 | 1.7 | 100% |
| 20 | 旧-每次新建 | 72,159 | 131,269 | 143,846 | 0.1 | 100% |
| 20 | 新-复用 | 2,340 | 2,538 | 2,578 | 7.7 | 100% |
| 50 | 旧-每次新建 | 171,398 | 308,195 | 337,874 | 0.1 | 100% |
| 50 | 新-复用 | 2,438 | 2,586 | 2,803 | 16.6 | 100% |


DavdGao:
@Osier-Yi Please review this PR. If I remember correctly, mem0 requires rebuilding the async HTTP client for each call; otherwise, the mem0 middleware throws an error.

Osier-Yi:
> @Osier-Yi Please review this PR. If I remember correctly, mem0 requires rebuilding the async HTTP client for each call; otherwise, the mem0 middleware throws an error.

Only ollama needs this~ I tested this PR with mem0 example, and pass!

Osier-Yi (commented):
LGTM

qbc2016:
[dongfeng3692](https://github.com/dongfeng3692) Thanks for the great work — the performance gains are really compelling!

A couple of things to note:

**Anthropic & Gemini**

`AnthropicChatModel._call_api` and `GeminiChatModel._call_api` also create a new client on every call. It would be nice to apply the same optimization there as well, either in this PR or as a quick follow-up.

**Test failures**

Since the client is now created eagerly in `__init__()`, the existing tests that use `@patch("openai.AsyncClient")` at the method level will fail — the model's `self.client` is already set before the mock takes effect. The affected tests will need to be updated accordingly (e.g., by creating the model inside the patch scope, or by directly assigning `model.client = mock_client`).

dongfeng3692:
Thank you for reviewing! I will solve these two problems later

qbc2016 (commented):
Please run `pip install pre-commit` and `pre-commit run --all-files` to format the code. @Osier-Yi Please check whether the behavior change in Ollama is feasible.

DavdGao:
Thanks for the PR — the benchmark numbers are compelling and the underlying problem is real 👍

A few things I'd like to settle before merging.

## 1. Reusing the client across event loops

The connection pool is lazy and loop-agnostic at construction time, but the **TCP connections** get bound to whichever loop is running when the first request goes out. Reusing a pooled connection from a different loop fails.

Worth being precise about the scope, because this is narrower than it first looks: **it's about loop identity, not concurrency.** Measured against a local server with retries disabled:

| Scenario | Concurrent | Crosses loops | Result |
|---|---|---|---|
| `gather` of 50 calls on one model | yes | no | ✅ 0 failures |
| `create_task` background work (e.g. `AgenticMemoryMiddleware:614`) | yes | no | ✅ |
| One model across two `asyncio.run()` calls, serial | no | yes | ❌ |
| One model used from a dedicated thread loop | no | yes | ❌ |

The bottom two rows aren't even concurrent, which is the clearest evidence that concurrency isn't the variable. Note also that this doesn't depend on where the model is constructed — module scope, inside `main()`, or inside the first loop all behave identically.

**So the scenario this PR optimizes for — one loop, high concurrency — is exactly the safe one.** The regression is confined to reusing a single model instance across loop boundaries.

The failing request evicts the dead connection, so a retry succeeds. That makes the impact depend entirely on whether a retry safety net exists:

| Backend | Behavior | Why |
|---|---|---|
| openai / anthropic / dashscope / deepseek / moonshot | ⚠️ Silent cost | SDK default `max_retries=2` swallows it; one wasted request + backoff per loop crossing |
| **ollama** | ❌ **Hard failure** | Raw httpx, no SDK retry, and `RuntimeError` isn't in the retryable set |

One client across 6 consecutive `asyncio.run()` calls:

```
max_retries=2 (openai default): ok=6 fail=0   <- masked by retries
max_retries=0 (retries off)   : ok=3 fail=3
```

Ollama fails hard because `_ollama/_model.py:119` lists only `httpx.ConnectError / ReadTimeout / RemoteProtocolError` as retryable, while the actual exception is `RuntimeError: Event loop is closed` — `_base.py:202` classifies it as non-retryable and re-raises immediately. This explains the earlier "only ollama needs per-call clients" observation: the other backends aren't fine, their retry layer is just absorbing the cost.

**Where this bites in-tree:** `_mem0/_agentscope_adapter.py:63-86` (`_AsyncBridge`) runs a dedicated event loop on its own daemon thread and submits AgentScope model coroutines to it via `run_coroutine_threadsafe`. Sharing one model instance between the agent (main loop) and mem0 (bridge loop) hits this — I measured 3/3 failures with retries off, 0/3 with two separate model instances.

`examples/long_term_memory/mem0/oss_demo.py:181-186` already documents this hazard and works around it with two separate model instances, which is also why the mem0 example passes under this PR. For the record, `AgenticMemoryMiddleware` and the ReMe middleware use `asyncio.create_task`, stay on the same loop, and are **not** affected.

The broader point: today, sharing one model across loops happens to work because clients are built per call. This PR makes "one model instance belongs to exactly one event loop" a hard constraint on `ChatModel`. That may well be an acceptable trade — but it should be a deliberate, documented contract change rather than an implicit side effect of a performance fix. At minimum I'd want it called out in the docstrings and changelog, and ideally ollama given a clearer failure mode than a bare `RuntimeError`.

## 2. `ChatModel` would need a `close()` method

The client now lives as long as the model object, but nothing in the repo ever calls `await client.aclose()`, so connections are never released. Under this design `ChatModel` needs a corresponding close method — which also means callers have to start managing model lifecycle. That cost should be weighed alongside the performance win.

## 3. `_xai` isn't covered

`model/_xai/_model.py:190` still constructs a client per call. It's grpc.aio-based with an explicit `await client.close()`, so the semantics differ — I'd suggest deliberately leaving it out of this PR, but noting that in the description so it doesn't later read as an oversight.

---

Interested in your thoughts on #1, particularly the ollama case.

dongfeng3692:
@DavdGao 
1 单独处理Ollama  只是掩盖了问题，所以我觉得还是做好约束让一个loop有单独的client
考虑到目前只有mem0 会单独产生一个loop，有没有可能 修改mem0 也能复用主loop呢

如果实在不行，可以让Mem0Middleware只接受 factory（构造函数）这个折中方案我觉得合理一点，这样可以避免误使用
```
Mem0Middleware(
    chat_model_factory=lambda: ChatModel(...),
    embedding_model_factory=lambda: EmbeddingModel(...),
)
```


2 我建议不引入close，不然会增加复杂度，openai的sdk会自动gc，但是引入不是坏处，目前可以先留空
image

3 _xai 我没改是因为 逻辑不同，我也没有对应的Xai的账号可以测试，所以保险起见 没有动

DavdGao (approved):
lgtm

PR #2068, opened 2026-07-13
feat(workspace): support apple container as workspace backend

### LINKED ISSUE #2035: feat(workspace): support apple container as workspace backend
## Background

Apple recently introduced **Apple Container**, a lightweight, native sandbox solution for macOS. As our agent service expands to local desktop deployments, we should evaluate and add support for this containerization approach.

## Note

Before implementation, we need to investigate and verify that Apple Container meets our core requirements, specifically:

- **File I/O** – supports file reading and writing within the sandbox
- **Shell execution** – allows running shell commands
- **Dependency installation** – supports installing third-party tools (e.g., `ripgrep`, `uv`)
- **MCP gateway** – capable of setting up and running an MCP gateway server

## Changes

Introduces three new classes to support Apple Container as a workspace backend:

- `AppleContainerBackend` — added in `src/agentscope/workspace/_apple_container`
- `AppleContainerWorkspace` — added in `src/agentscope/workspace/_apple_container`
- `AppleContainerWorkspaceManager` — added in `src/agentscope/app/workspace_manager`

iluv7:
## AgentScope Version

2.0.4.dev0

## Description

Closes https://github.com/agentscope-ai/agentscope/issues/2035

Adds an Apple Container workspace backend for AgentScope, leveraging macOS 26's native `container` CLI to run Linux containers as lightweight VMs on Apple silicon.

### What Changed

**Added AppleContainerBackend:**

Apple Container exposes only XPC — no Python SDK, no HTTP management API. Unlike Docker (`aiodocker` HTTP API) and E2B (official Python SDK), Apple Container can only be driven through its CLI binary. The backend uses `asyncio.create_subprocess_exec` to implement the three BackendBase primitives:

- `exec_shell`: `container exec --workdir <cwd> <cid> -- <command>`. Handles timeouts (`process.kill()` → `exit_code=-1`) and missing CLI (`FileNotFoundError` → `exit_code=127`).
- `read_file`: `container exec cat <path>`, captures stdout. `exit_code ≠ 0` raises `FileNotFoundError`.
- `write_file`: writes data to a host-side temp file, then `container cp` into the container, with cleanup in `finally`. Embedding raw bytes directly into a shell command (`echo '...' > path`) is unsafe — data may contain single quotes, null bytes, `$()`, backticks, etc., causing parsing errors, content truncation, or command injection.

Inherits all derived filesystem helpers from `BackendBase` (`file_exists`, `is_dir`, `list_dir`, `stat_mtime`, `delete_path`), implemented via `exec_shell` + POSIX commands with no additional code needed.

**Added AppleContainerWorkspace:**

- Extends `SandboxedWorkspaceBase`, reusing the MCP Gateway lifecycle.
- Container lifecycle: `container image pull` → `container run -d` → `container stop` → `container rm -f`.
- Pulls OCI images on demand. Reattaches to existing containers across process restarts via deterministic naming.
- UV-based bootstrap: `apt-get` → `curl uv` → `uv venv` → `uv pip install` gateway dependencies.
- Configurable CPUs, memory, environment variables, and extra pip packages.

**Added AppleContainerWorkspaceManager:**

- TTL-based workspace caching (default 3600s expiry, 300s sweep interval).
- `get_workspace` / `close` / `close_all` interface consistent with other managers.
- `async with` protocol: starts the background sweeper on enter, shuts down all containers in parallel on exit.

Exported new classes from `src/agentscope/workspace/__init__.py` and `src/agentscope/app/workspace_manager/__init__.py`.

### Environment Notes

- Apple Container (`github.com/apple/container`) must be installed separately from GitHub Releases, and the system service started via `container system start`.
- macOS 26+ with Apple silicon only. Uses the Virtualization framework to run each container as an independent lightweight VM.
- Call chain: Python subprocess → `container` CLI → XPC → `container-apiserver` → `container-runtime-linux` → Linux VM.

### Testing

- Apple Container Backend tests: 11 passed
- Apple Container Workspace tests: 8 passed, 1 skipped
- Pre-commit hooks: all passed
- Backend tests mock subprocess calls to verify CLI command construction. Workspace tests mock subprocess + Gateway startup to verify initialization, shutdown, idempotency, and bootstrap lifecycle.
- Live integration tests require macOS 26 + Apple silicon hardware with `APPLE_CONTAINER_LIVE=1` set (not yet tested — My macOS is not upgraded to 26).

## Checklist


src/agentscope/app/workspace_manager/_applecontainer_workspace_manager.py
```diff
@@ -0,0 +1,296 @@
...
+
+from ..._logging import logger
+from ...mcp import MCPClient
+from ...workspace._applecontainer._constants import (
+    DEFAULT_BASE_IMAGE,
+    DEFAULT_CPUS,
+    DEFAULT_GATEWAY_PORT,
+    DEFAULT_MEMORY,
+)
+from ...workspace._applecontainer._applecontainer_workspace import (
```
DavdGao: Should be `from ...workspace import AppleContainerWorkspace`

DavdGao (changes requested):
Please add unit tests for the Apple container backend, workspace, and manager. 

Since GitHub Actions allows specifying macOS versions, we should update the macOS version in `.github/workflows/unittest.yml` to use a newer one, and run the tests in macOS (skip windows and linux)

iluv7:
Hi @DavdGao, thanks for the review! The import issue has been fixed and unit tests have been added:

**Changes**

- Changed `AppleContainerWorkspace` import to use the public API (`from ...workspace import AppleContainerWorkspace`)
- Updated CI matrix from `macos-15` to `macos-26`
- Added 3 test files: `tests/backend_applecontainer_test.py`, `tests/workspace_applecontainer_test.py`, `tests/workspace_manager_applecontainer_test.py`

**Tests**

All Apple Container tests are mock-based and run on macOS only:

| | Cases | Coverage |
|--|:-----:|----------|
| `tests/backend_applecontainer_test.py` | 11 passed | Mock subprocess, verify CLI command construction |
| `tests/workspace_applecontainer_test.py` | 8 passed, 1 skipped | Lifecycle (init/close/idempotency/bootstrap) |
| `tests/workspace_manager_applecontainer_test.py` | 8 passed | Cache, TTL sweep, workspace_id assignment |

All 28 tests run on macOS. The 1 skipped is the e2e live test. All mock-based unit tests are gated with `@skipUnless(sys.platform == "darwin")` so they skip on Windows/Linux in CI.

I also ran the e2e test through a dedicated CI job on `macos-26` (since removed). The first 4 pipeline steps all succeeded, confirming our code path is correct:

1. ✅ `gh release download` — CLI 1.1.0 installed
2. ✅ `container system start` — daemon started
3. ✅ `container image pull python:3.11-slim` — image pulled
4. ✅ `container run -d` — CLI command constructed correctly

Only the final step (VM boot via Virtualization.framework) failed:

```
Error Domain=VZErrorDomain Code=2 "Virtualization is not available on this hardware."
```

Could this be because GitHub Actions runners are VMs without nested virtualization support? The e2e test would likely require bare metal macOS 26 + Apple silicon.

**Total: 27 passed, 1 skipped**

iluv7:
I initially misunderstood your suggestion — instead of updating the main test matrix, I had spun up a dedicated sub-job on `macos-26` just to run the e2e test. That's now corrected: the main test matrix uses `macos-26` directly, and Apple Container tests are skipped on Windows/Linux via `@skipUnless(sys.platform == "darwin")`. The e2e results reported above were from that sub-job, which I've since removed. Happy to add it back if you'd like to take a look.

tests/backend_applecontainer_test.py
```diff
@@ -0,0 +1,285 @@
...
+required.
+"""
+
+import asyncio
+import sys
+from unittest.async_case import IsolatedAsyncioTestCase
+from unittest.mock import AsyncMock, MagicMock, patch
+from unittest import skipUnless
+
+from agentscope.workspace._applecontainer._applecontainer_backend import (
```
DavdGao: ==> `from agentscope.workspace import AppleContainerBackend`

tests/workspace_applecontainer_test.py
```diff
@@ -0,0 +1,380 @@
...
+from unittest import skipUnless
+
+from agentscope.workspace._applecontainer._applecontainer_backend import (
+    AppleContainerBackend,
+)
+from agentscope.workspace._applecontainer._constants import (
+    CONTAINER_WORKDIR,
+    DEFAULT_BASE_IMAGE,
+)
+from agentscope.workspace._applecontainer._applecontainer_workspace import (
```
DavdGao: Similar as above

tests/workspace_applecontainer_test.py
```diff
@@ -0,0 +1,380 @@
...
+
+import asyncio
+import json
+import os
+import sys
+from unittest.async_case import IsolatedAsyncioTestCase
+from unittest.mock import AsyncMock, MagicMock, patch
+from unittest import skipUnless
+
+from agentscope.workspace._applecontainer._applecontainer_backend import (
```
DavdGao: Similar as above

tests/backend_applecontainer_test.py
```diff
@@ -0,0 +1,285 @@
...
+    AppleContainerBackend,
+)
+
+# ── mock helpers ────────────────────────────────────────────────────
+
+
+def _make_mock_process(
```
DavdGao: Since we're going to run the unit tests on macOS, we don't need to mock the execution anymore and should remove the mock logic from the tests.

tests/workspace_applecontainer_test.py
```diff
@@ -0,0 +1,380 @@
...
+
+# ── availability check ──────────────────────────────────────────────
+
+_APPLE_CONTAINER_LIVE = os.getenv("APPLE_CONTAINER_LIVE", "")
+
+
+# ── mock helpers ────────────────────────────────────────────────────
+
+
+def _mock_process(
```
DavdGao: Similar as above

tests/workspace_manager_applecontainer_test.py
```diff
@@ -0,0 +1,226 @@
...
+from unittest.mock import patch
+from unittest import skipUnless
+
+from agentscope.app.workspace_manager import (
+    AppleContainerWorkspaceManager,
+    IsolationPolicy,
+)
+
+
+class _FakeWorkspace:
```
DavdGao, line 16: Similar as above

DavdGao (commented):
Plz see inline comments

iluv7:
Hi, @DavdGao 
Updated `tests/backend_applecontainer_test.py` and `tests/workspace_applecontainer_test.py`:
- Removed all subprocess mocks — tests now run against a real `container` CLI, skipped when CLI not installed
- Changed imports to use the public API (`from agentscope.workspace import ...`)

YingchaoX:
# Local verification feedback for the Apple Container workspace backend

Hi, thanks for working on Apple Container support. 

I tested the current PR head, `c33af7a26f3b2fda5e9886f8c32144d644e74642`, on a bare-metal Apple Silicon Mac with:

- macOS 26.5.1
- Apple Container 1.1.0 (`5973b9c`)
- Python 3.14.5
- AgentScope installed from this PR worktree

I also confirmed that the local worktree SHA, GitHub PR head SHA, and `refs/pull/2068/head` all match `c33af7a`.

## GitHub Actions coverage

The current green macOS 26 job does not appear to exercise the real Apple Container implementation. Its log contains:

```text
tests/backend_applecontainer_test.py ssssssssssss
tests/workspace_applecontainer_test.py ...ssssss
tests/workspace_manager_applecontainer_test.py ........
```

The workflow changes the runner to `macos-26`, but it does not install Apple Container, run `container system start`, or set `APPLE_CONTAINER_LIVE=1`. Consequently:

- all 12 backend tests are skipped;
- the three constructor-only workspace tests run, while all six container-dependent workspace tests are skipped;
- the eight workspace-manager tests pass against `_FakeWorkspace`, without creating a container.

The overall job is therefore green because pytest treats skips as successful completion, not because an Apple Container was created. The job I inspected is:

https://github.com/agentscope-ai/agentscope/actions/runs/29318213442/job/87037021433

## Local pytest results

I ran the three PR-specific test files:

```bash
pytest \
  tests/backend_applecontainer_test.py \
  tests/workspace_applecontainer_test.py \
  tests/workspace_manager_applecontainer_test.py -vv
```

The completed results were:

```text
backend:                3 passed, 9 failed
workspace constructor
and manager tests:     11 passed
```

The combined run did not complete: the first workspace lifecycle test remained in `container image pull python:3.11-slim`, so I interrupted it after reproducing the backend failures. I have not run or claimed a passing repository-wide `pytest tests` result.

## 1. `container exec` command construction

The backend currently constructs commands in this form:

```text
container exec --workdir <cwd> <container-id> -- <command>
```

On both Apple Container 1.0.0 and the current stable 1.1.0, `--` is treated as the target executable:

```text
Error: failed to find target executable --
```

The same command succeeds when the separator is removed:

```text
container exec --workdir <cwd> <container-id> <command>
```

This caused the failures in shell execution, timeout behavior, derived filesystem helpers, and file read/write verification. Since the behavior is unchanged in Apple Container 1.1.0, this does not appear to be only an old-CLI compatibility issue.

## 2. Local image detection

The workspace compares the configured short image name directly:

```python
img.get("name") == "python:3.11-slim"
```

However, Apple Container reports the local image using its canonical reference:

```text
docker.io/library/python:3.11-slim
```

The workspace therefore fails to recognize an already-cached image and runs `container image pull` again. Normalizing both references before comparison, or comparing the resolved image identity, would avoid the repeated pull.

## 3. Real dependency and MCP gateway verification

Issue #2035 explicitly asks to verify dependency installation (`ripgrep` and `uv`) and the MCP gateway. The opt-in live test currently verifies `echo` and a file round-trip, but it does not:

- execute `rg --version` or `uv --version`;
- register an MCP server;
- list gateway-provided MCP tools; or
- invoke a tool through the gateway.

I attempted the unmodified workspace bootstrap in a real container. It reached container creation, but `apt-get update` failed because the container VM on my machine could resolve DNS while TCP/HTTPS connections timed out. The host network and DashScope API were working normally. I therefore cannot use that result to conclude that the gateway implementation itself is broken; it is an environment-specific blocker that prevented full gateway verification.

For CI, a true live test will likely require a bare-metal/self-hosted Apple Silicon runner because GitHub-hosted macOS runners do not provide the nested virtualization needed to boot Apple Container VMs.

## Positive end-to-end result

To isolate the core AgentScope integration from the two confirmed command/image issues and the local gateway-network blocker, I applied runtime-only adapters without editing the PR source:

1. removed the unsupported `--` separator;
2. skipped the redundant pull for the already-cached image;
3. skipped gateway bootstrap because container TCP egress was unavailable.

I then ran a real end-to-end flow using DashScope `qwen3.7-max`:

```text
qwen3.7-max
  -> AgentScope Agent
  -> Write tool
  -> real Apple Container
  -> Read tool
  -> independent host-side file verification
```

The result was:

```text
WORKSPACE_INIT_SECONDS=1.006
CONTAINER_IDENTITY=Linux aarch64 | as_ws_codex-llm-container-e2e
TOOLS=Read,Write
LLM_FINAL=E2E_OK
FILE_BYTES=b'QWEN37_APPLE_CONTAINER_E2E_OK\n'
E2E_RESULT=PASS
```

This is encouraging evidence that the core LLM-to-container built-in-tool path works once the CLI construction is corrected. It is not a substitute for the missing MCP gateway E2E test.

## Additional notes

- The constructor-only workspace tests are currently skipped on non-macOS platforms even though their class documentation says that no container is needed. Those tests could run cross-platform.
- `write_file()` calls `os.write(tmp_fd, data)` once and ignores the returned byte count. A full-write loop or file-object write would avoid possible partial writes for large payloads.
- The PR checklist still has the related documentation item unchecked. The contribution guide asks new workspace backends to include companion user documentation.

## Suggested next steps

Would you be open to the following changes before merging?

1. Remove the `--` separator from `container exec` and add deterministic tests for the exact argv construction.
2. Normalize OCI image references before checking whether an image is present.
3. Keep subprocess/JSON/error-path unit tests deterministic and runnable without Apple Container.
4. Keep real-container tests opt-in, but make the dedicated live job fail rather than silently skip when Apple Container is expected but unavailable.
5. Run the live job on a bare-metal/self-hosted Apple Silicon runner and cover dependency installation plus an actual MCP gateway tool call.
6. Document the supported Apple Container versions, network/proxy prerequisites, and user-facing workspace configuration.

I would be happy to rerun the local tests after an update to the PR.

examples/workspace/apple-container-workspace.md
YingchaoX, line 1: This tech documentation file doesn't need to be included in this PR.

iluv7: Done.

iluv7:
@YingchaoX 
Thanks for the detailed review — all three bugs confirmed and fixed:

**1. `--` separator in `container exec`**

Apple Container's CLI treats `--` as the target executable rather than a delimiter (unlike Docker/kubectl). Removed it from the argv construction and added 4 deterministic tests verifying exact command vectors with no `--`.

**2. Image reference normalization**

`container image list` returns canonical references like `docker.io/library/python:3.11-slim` while we configure the short form. Added `_normalize_image_ref` to strip `docker.io` and `library` from both sides before comparison, so a cached image is recognized and the pull is skipped. 7 unit tests cover short refs, canonical library refs, non-library images, other registries, and tags/digests.

**3. Partial write in `write_file()`**

Switched from `os.write(fd, data)` to `os.fdopen(fd, "wb").write(data)` — the Python file object handles partial writes internally.

**Tests**

Added 21 deterministic tests covering argv construction, subprocess error paths, and image normalization — all run cross-platform without a container.

The live test class went from one generic test to five, all behind `APPLE_CONTAINER_LIVE=1`:
- `test_bootstrap_dependencies` — `rg --version`, `uv --version`, `python3 --version`
- `test_builtin_tools_bound_and_callable` — `list_tools()` + `Bash.call()` through tool binding
- `test_gateway_healthy_and_mcp_list` — `list_mcps()` confirms gateway is alive
- `test_mcp_server_register_and_call` — writes a `FastMCP` echo server into the container, registers it via `add_mcp()`, then `list_mcps` → `list_raw_tools` → `get_tool` → `call` the full chain
- `test_file_read_write_roundtrip` — raw backend + Write/Read tool wrapper both paths

Constructor and manager tests no longer require macOS.

**CI**

- The matrix job now uses `--ignore` on the two Apple Container test files so they don't silently skip on non-macOS runners.
- A dedicated `apple-container-live` job runs on `macos-26` (switchable to a self-hosted Apple Silicon runner) and fails explicitly when the `container` CLI is missing or the system service is not running — no more silent passes. It sets `APPLE_CONTAINER_LIVE=1` so the full set of live integration tests (bootstrap deps + MCP echo server register-and-call) are exercised.

Thank you once again for taking the time to help test and review! Happy to iterate if you find anything else on a re-run.

DavdGao (approved):
lgtm

PR #2077, opened 2026-07-14
fix(tracing): avoid otel context detach error on cross-task stream close

### LINKED ISSUE #2076: TracingMiddleware logs Failed to detach context when reply_stream is closed from another asyncio task
## Summary

```text
Failed to detach context
```

## Environment

- AgentScope: 2.0.4
- Python: 3.12

```python
        yield item
```

## Minimal Reproduction

```python
import asyncio

async def gen():
        try:
            yield 1
        finally:

async def main():
    g = gen()
    print(await anext(g))

        await g.aclose()

asyncio.run(main())
```

Actual output includes:

```text
Failed to detach context
    yield span
    yield span
  File "...", line ..., in gen
GeneratorExit

```

## Expected Behavior

## Possible Fix Direction

alex-mm:
## AgentScope Version

`main` (reproduced on `2.0.4`)

### Background

`TracingMiddleware.on_reply` and `on_acting` wrapped an async generator with:

```python
        yield item
```

`start_as_current_span()` attaches the OpenTelemetry current-span context on
`__enter__` and detaches it on `__exit__`. Because the `with` body contains
`yield`, enter and exit happen at different suspension points. In streaming
scenarios (SSE / cancellation / client disconnect), the async generator is
often closed from a **different asyncio task** than the one that started it.
`asyncio.create_task` runs the coroutine in a copied `contextvars.Context`, so
the detach happens in a different context than the attach, and OpenTelemetry
logs:

```
Failed to detach context
```

The request itself usually succeeds, but this emits ERROR-level log noise that
pollutes production error monitoring. See #2076 for the minimal reproduction.

### Changes

- Create the reply / tool spans with `tracer.start_span(...)` instead of
  `start_as_current_span(..., end_on_exit=False)`, and manage `span.end()`
  manually (as before, via `_set_span_success_status` / `_set_span_error_status`).
- Attach/detach `set_span_in_context(span)` **within each `__anext__` step**
  and never hold the context across the `yield` to the external consumer. Attach
  and detach therefore always happen in the same context, even when the outer
  async generator is closed from another task.
- Child spans (chat / execute_tool) are still parented correctly because the
  span context is active while `next_handler` runs. Synthetic external-execution
  `execute_tool` spans are now parented explicitly via `context=span_context`.
- `on_model_call` already used a manual span wrapper that never attaches context
  across yields, so it was left unchanged.

### How to test

```bash
pytest tests/tracing_test.py
```

New regression tests `test_on_reply_close_from_other_task_no_detach_error` and
`test_on_acting_close_from_other_task_no_detach_error` drive the middleware one
step, then close the stream from another asyncio task and assert that no
`opentelemetry.context` ERROR is logged. Both fail on `main` (with the exact
`ValueError: ... was created in a different Context`) and pass with this fix.
All 15 existing tracing tests still pass, confirming span parenting/attributes
are preserved.

## Checklist

- [x] An issue has been created for this PR (#2076)

Fixes #2076

### DIFF
--- a/src/agentscope/middleware/_tracing/_trace.py
+++ b/src/agentscope/middleware/_tracing/_trace.py
@@ -13,6 +13,7 @@ from typing import (
 
 import aioitertools
 
+from opentelemetry import context as otel_context
 from opentelemetry import trace as otel_trace
 from opentelemetry.trace import StatusCode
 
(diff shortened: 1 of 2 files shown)


DavdGao (commented):
@alex-mm Plz run `pre-commit run --all-files` and format the code in this PR, others lgtm

alex-mm:
Done, thanks! I ran `pre-commit run --all-files` locally and fixed the pylint issue in this PR (`__anext__()` -> `anext()`).

The latest PR head is `43bf61a2`. The GitHub Pre-commit workflow is currently in `action_required` state for the fork PR, so it needs maintainer approval before it can run on GitHub.

src/agentscope/middleware/_tracing/_trace.py
```diff
@@ -154,91 +155,103 @@ async def on_reply(
...
+                    pass
+
+        has_error = False
+        error_exc: BaseException | None = None
+        last_msg: Msg | None = None
+        hitl_pending: list[str] = []
+        external_pending: list[str] = []
+        observed_reply_id: str | None = None
+
+        gen = next_handler(**input_kwargs)
```
qbc2016, line 206: Please move `gen = next_handler(**input_kwargs)` inside the `try` block.

Although AgentScope’s internal `next_handler` is normally an async generator function, a callable returning an async iterator may still raise synchronously during construction. With the current placement, such an exception bypasses the error handling, leaving the span neither marked as `ERROR` nor ended.

This also changes the previous behavior, where `next_handler(...)` was evaluated inside the guarded `async for`. The same adjustment should be made in both `on_reply` and `on_acting`.

DavdGao (approved):
lgtm

PR #2053, opened 2026-07-12
fix(skill): expand user-home paths

helloJamest:
﻿## Summary

- Normalize local skill paths with `expanduser` before `abspath`.
- Apply the normalization to `LocalSkillLoader`, `WorkspaceBase`, and `LocalWorkspace` skill seed paths.
- Add regression tests for `~` skill paths in both skill loading and workspace initialization.

## Root cause

`os.path.abspath()` treats `~` as a literal path segment, so paths such as `~/.agentscope/skills/foo` point under the current working directory instead of the user's home directory.

## Tests

- `python -m pytest tests\skill_loader_test.py::SkillLoaderTest::test_expanduser_directory -q` (fails before the fix)
- `python -m pytest tests\workspace_local_test.py::TestLocalWorkspaceSkills::test_initialize_with_tilde_skill_path -q` (fails before the fix)
- `python -m pytest tests\skill_loader_test.py tests\workspace_local_test.py::TestLocalWorkspaceSkills -q`
- `python -m py_compile src\agentscope\_utils\_common.py src\agentscope\skill\_local_loader.py src\agentscope\workspace\_base.py src\agentscope\workspace\_local_workspace.py`
- `pre-commit run --files src\agentscope\_utils\_common.py src\agentscope\skill\_local_loader.py src\agentscope\workspace\_base.py src\agentscope\workspace\_local_workspace.py tests\skill_loader_test.py tests\workspace_local_test.py`

Closes #2052

PR #2086, opened 2026-07-15
fix(skill): expand user home in skill paths

### LINKED ISSUE #2052: [Bug]: skill paths using ~ are not expanded before loading or seeding skills
### Prerequisites

### Background / Description

AgentScope 2.0 supports loading skills from local directories through `LocalSkillLoader` and seeding workspace skills through `skill_paths`.

However, user-home paths using `~` are not expanded before validation/loading. For example:

```python
loader = LocalSkillLoader("~/.agentscope/skills/my_skill")
skills = await loader.list_skills()
```

`LocalSkillLoader.__init__` currently normalizes with `os.path.abspath(directory)` only, so `~` is treated as a literal path segment instead of the current user's home directory. This makes the loader look in the wrong directory and return no skills.

The same ergonomics issue can affect workspace `skill_paths`, which are commonly provided through config files, environment variables, or CLI arguments where `~` is a natural shorthand.

This is a narrow compatibility fix related to the broader AgentScope 2.0 skill update tracked in #2044.

### Expected Behavior

Skill paths should expand user-home shorthand before becoming absolute paths, e.g.:

```python
os.path.abspath(os.path.expanduser(path))
```

This should apply consistently to local skill loading and workspace skill seeding paths.

### Proposed Fix

- Normalize `LocalSkillLoader` directories with `expanduser` before `abspath`.
- Normalize workspace `skill_paths` before they are used to validate/copy seed skills.
- Add regression tests for `LocalSkillLoader("~/...")` and workspace `skill_paths` using `~`.

### Environment

- AgentScope Version: `main`
- Python Version: 3.11+
- OS: all platforms supporting `os.path.expanduser`

nolanchic:
## Summary

Skill paths supplied via config files, env vars, or CLI args commonly use the `~` shorthand (e.g. `~/.agentscope/skills/my_skill`). `~` was treated as a literal path segment instead of the user home directory, so `LocalSkillLoader` looked in the wrong place and workspace `skill_paths` silently failed to seed.

Closes #2052.

## Root cause

Path normalisation used `os.path.abspath(path)` only, never `os.path.expanduser`. So `~/.agentscope/skills/foo` resolved to `<cwd>/~/.agentscope/skills/foo`.

## Fix

Expand `~` before resolving to an absolute path at every boundary where skill paths enter the system:

- `LocalSkillLoader.__init__`
- `WorkspaceBase.__init__` — covers Docker/E2B/K8S seed paths (they delegate here)
- `LocalWorkspace.__init__` — sets `skill_paths` itself, bypassing the base
- `WorkspaceBase.add_skill` — covers ad-hoc runtime adds and the base seed path

Each normalisation is `os.path.abspath(os.path.expanduser(path))`.

## Test plan

- [x] `tests/skill_loader_test.py::test_tilde_directory_is_expanded` — loads a skill under the real home dir via a `~/...` path; asserts both the normalised `directory` attribute and that the skill is actually loadable.
- [x] `tests/workspace_local_test.py::TestLocalWorkspaceSkills::test_skill_paths_expand_tilde_before_seeding` — seeds via a `~/...` `skill_paths` entry; asserts stored paths are expanded and the skill is copied into `${workdir}/skills`.
- [x] Existing skill/workspace tests still pass (`skill_loader_test`, `workspace_local_test`, `toolkit_skill_test`).
- [x] `pre-commit run --all-files` passes (black, flake8, mypy, pylint).

```bash
pytest tests/skill_loader_test.py tests/workspace_local_test.py tests/toolkit_skill_test.py
# 35 passed
```

### DIFF
--- a/src/agentscope/skill/_local_loader.py
+++ b/src/agentscope/skill/_local_loader.py
@@ -25,7 +25,7 @@ class LocalSkillLoader(SkillLoaderBase):
                 Whether to scan subdirectories. Defaults to False (only
                 scan current directory).
         """
-        self.directory = os.path.abspath(directory)
+        self.directory = os.path.abspath(os.path.expanduser(directory))
         self.scan_subdir = scan_subdir
         self._cache: dict[str, Skill] = {}
 
--- a/src/agentscope/workspace/_base.py
+++ b/src/agentscope/workspace/_base.py
@@ -184,14 +184,18 @@ class WorkspaceBase:
                 without a persisted ``.mcp`` file.
             skill_paths (`list[str] | None`, optional):
                 Local skill directories to copy into ``skills/`` on
-                first start.
+                first start. Each path is normalised to an absolute
+                path with ``~`` expanded to the user home directory.
         """
         self.workspace_id = workspace_id or _generate_id()
         self.is_alive = False
         self._backend = None
 
         self.default_mcps = list(default_mcps or [])
-        self.skill_paths = list(skill_paths or [])
+        self.skill_paths = [
+            os.path.abspath(os.path.expanduser(path))
+            for path in (skill_paths or [])
+        ]
 
         self._mcps = []
         self._mcp_lock = asyncio.Lock()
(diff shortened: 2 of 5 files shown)


DavdGao:
@nolanchic, thanks for your interest in contributing to agentscope.

This issue has already been assigned to another contributor who submitted PR #2053 earlier. We'll proceed with that PR for now. Plz feel free to look for other open issues — contributions are always welcome!

PR #2087, opened 2026-07-15
fix(tracing): avoid cross-context detach on stream close

### LINKED ISSUE #2076: TracingMiddleware logs Failed to detach context when reply_stream is closed from another asyncio task

nolanchic:
## Summary

`TracingMiddleware` kept OpenTelemetry spans current across async-generator `yield` points via `start_as_current_span`. When a stream is closed (`aclose`) from a different asyncio task — normal for SSE-style replies on client disconnect, cancellation, or cross-task cleanup — the span's `detach` runs in the closing task's `contextvars` Context, so resetting the token raises `ValueError`. OpenTelemetry swallows that and logs `\"Failed to detach context\"` at ERROR level, polluting production monitoring even though the request itself does not fail.

Closes #2076.

## Root cause

`on_reply` and `on_acting` are async generators. `with tracer.start_as_current_span(..., end_on_exit=False)` wraps an `async for ... yield item` loop, so the OTel current-span `ContextVar` stays set across `yield`. `use_span.__exit__` → `context.detach(token)` runs in the closing task's Context (a different `contextvars.Context` than where the token was created). `ContextVar.reset` raises `ValueError`, which OTel's `context.detach` catches and logs via `logger.exception(\"Failed to detach context\")`.

`on_model_call` is unaffected — its `with` block has no `yield` (it returns the generator wrapper), so detach always happens in the originating context.

## Fix

- Replace `start_as_current_span` in `on_reply` and `on_acting` (the two hooks whose `with` block spans a `yield`) with manual `start_span` + new `_attach_span`/`_detach_span` helpers.
- A probe `ContextVar` detects whether cleanup runs in the originating context: `_detach_span` resets the probe token first; only if that succeeds (same context) does it call `context.detach`. In the foreign-context case (cross-task close) it is a no-op — the span was never attached there, so there is genuinely nothing to detach.
- Span parent/child relationships are preserved (the span is still current during `next_handler` iteration, so nested model/tool spans parent correctly) and `span.end()` is still called manually.
- Clean stream shutdowns (`GeneratorExit` / `asyncio.CancelledError`) are now treated as normal completion across `on_reply`, `on_acting`, and the streaming model-call wrapper (`_trace_async_generator_wrapper`), so they no longer mark spans as ERROR.

## Test plan

- [x] New `test_span_detach_survives_cross_task_close` — a minimal async generator mirroring `on_reply`/`on_acting` (attach span, yield, detach in `finally`) is closed from a different asyncio task; asserts no `\"Failed to detach context\"` ERROR log and the span is still ended/exported. Verified it FAILS on the naive unconditional-detach version with the exact `ValueError` from the issue, and PASSES with the probe guard.
- [x] All existing tracing tests pass (`on_reply` / `on_acting` / `on_model_call` span attributes, parenting via shared `conversation.id`, HITL, external execution).
- [x] Broad middleware/agent suite green (63 tests across `tracing`, `middleware`, `middleware_budget`, `agent_basic`, `agent_interrupt`, `hitl_external_execution`, `hitl_user_confirmation`).
- [x] `pre-commit run` passes (black, flake8, mypy, pylint).

```bash
pytest tests/tracing_test.py tests/middleware_test.py tests/middleware_budget_test.py tests/agent_basic_test.py tests/agent_interrupt_test.py tests/hitl_external_execution_test.py tests/hitl_user_confirmation_test.py
# 63 passed
```

### DIFF
--- a/src/agentscope/middleware/_tracing/_trace.py
+++ b/src/agentscope/middleware/_tracing/_trace.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 """TracingMiddleware and supporting utilities for OpenTelemetry tracing."""
+import asyncio
+import contextvars
 import json
 from typing import (
     Any,
@@ -13,6 +15,7 @@ from typing import (
 
 import aioitertools
 
+from opentelemetry import context as otel_context
 from opentelemetry import trace as otel_trace
 from opentelemetry.trace import StatusCode
 
@@ -50,6 +53,15 @@ if TYPE_CHECKING:
 T = TypeVar("T")
 
 
+# Probe ContextVar used to detect whether span cleanup is running in the
+# same ``contextvars`` Context that attached the span. See
+# :func:`_detach_span`.
+_probe_ctxvar = contextvars.ContextVar(
+    "_agentscope_tracing_ctx_probe",
+    default=False,
+)
+
+
 # ---------------------------------------------------------------------------
 # Utility helpers
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


nolanchic:
The only failing check is `test (macos-15, 3.11)`, and the single failing test is
`rag_vdb_milvus_lite_test.py::MilvusLiteStoreTest::test_insert_and_search`
(gRPC "Method not implemented!" → similarity 0.0). This is a macOS runner
environment issue unrelated to this PR, which only touches the tracing
middleware — all tracing tests pass and pre-commit is green.

Closing and reopening to re-trigger CI for a clean run.

qbc2016:
Thanks @nolanchic for the detailed investigation and thoughtful solution. Your analysis of the cross-task OpenTelemetry context detach issue is very helpful.

After comparing the overlapping proposals for #2076, we plan to consolidate the fix in #2077. That implementation avoids keeping an OTel context attached across `yield` boundaries, scopes the attachment to each downstream `anext()` call, and includes regression coverage for both `on_reply` and `on_acting`.

The handling of `GeneratorExit` and `CancelledError` in this PR is also worth considering, but it changes span-status semantics beyond the context-detach issue. If needed, that part would be better discussed in a focused follow-up after #2077 lands.

To avoid maintaining multiple fixes for the same root cause, we’ll close this PR in favor of #2077. This is purely to consolidate the work—the investigation and test coverage here were valuable.

Thanks again for the contribution!

PR #2088, opened 2026-07-15
fix(service): return newest messages and paginate by cursor

### LINKED ISSUE #2080: [Bug]: `list_messages` returns oldest messages instead of newest and lacks cursor-based pagination
### Prerequisites

### Summary
`RedisStorage.list_messages()` has two issues:

1. **Returns oldest messages instead of newest**: Messages are appended via `rpush`, so the tail of the list is the newest. However, `lrange(key, offset, offset + limit - 1)` reads from the head. With default `offset=0, limit=50`, only the first 50 messages ever stored are returned. Once a session exceeds 50 messages, newer messages become invisible after page refresh or SSE reconnection.
2. **No pagination support**: The response lacks a `has_more` field and there is no cursor-based pagination mechanism. Numeric offset pagination is fundamentally flawed for real-time message streams — new messages pushed via SSE while the user is paginating cause the offset window to drift, resulting in duplicates or gaps. Industry standard for chat/message APIs is cursor-based pagination using message IDs as anchors.

### Expected Behavior

**Request parameters:**
- Deprecate `offset` (int); introduce `before` (str, message ID)
- Omit `before` → return the latest `limit` messages
- Pass `before=msg_id` → return `limit` messages before that message
- Messages within each page remain in chronological order

**Response body:**
- Add `has_more: bool` indicating whether older messages exist before the current page
- No `next_cursor` needed; the frontend uses `messages[0].id` as the `before` value for the next request

### Suggested Fix

**Storage layer** (`_redis_storage.py`):
```python
async def list_messages(
    self, user_id, session_id,
    limit: int = 50,
    before: str | None = None,
) -> tuple[list[Msg], bool]:
    key = self._message_key(user_id, session_id)
    total = await self._client.llen(key)

    if before is None:
        end = total - 1
    else:
        idx = await self._find_message_index(key, before)
        if idx is None:
            return [], False
        end = idx - 1

    start = max(end - limit + 1, 0)
    if end < 0:
        return [], False

    raw_list = await self._client.lrange(key, start, end)
    has_more = start > 0
    return [Msg.model_validate_json(r) for r in raw_list], has_more
```

**Response schema** (`_session.py`):
```python
class ListMessagesResponse(BaseModel):
    messages: list = Field(description="Messages in chronological order.")
    is_running: bool = Field(description="Whether the session is currently running.")
    has_more: bool = Field(
        description="Whether there are older messages before this page.",
    )
```

**Router layer** (`_session.py`):
- Replace `offset: int = Query(0)` with `before: str | None = Query(None)`
- Call storage to obtain `(messages, has_more)` and pass both into the response

### Affected Files
- `src/agentscope/app/storage/_redis_storage.py` L991-1001
- `src/agentscope/app/_router/_schema/_session.py` L189-195
- `src/agentscope/app/_router/_session.py` L508-554

nolanchic:
## Summary

`RedisStorage.list_messages` read messages from the **head** of the list (`lrange(key, offset, ...)`), but messages are appended with `rpush`, so the **tail is the newest**. With the default `offset=0, limit=50`, only the first 50 messages ever stored were returned — once a session exceeded 50 messages the newest ones became invisible after a page refresh or SSE reconnect. Numeric offset pagination is also unstable for real-time streams: new messages pushed mid-pagination shift the window and cause duplicates/gaps.

Closes #2080.

## Changes

- **`RedisStorage.list_messages`** → `list_messages(user_id, session_id, limit, before) -> tuple[list[Msg], bool]`:
  - `before=None` returns the newest `limit` messages; `before=msg_id` returns the `limit` messages preceding that id (exclusive upper bound).
  - The window is always returned oldest→newest, so the frontend can anchor the next page on `messages[0].id`.
  - Returns `has_more` indicating older messages exist before the window.
  - New `_find_message_index` helper locates the cursor message (Redis lists aren't indexed by id; scans newest→oldest since the cursor is usually recent).
- **`StorageBase`** abstract `list_messages` updated to the new contract (only `RedisStorage` implements it).
- **`ListMessagesResponse`** schema gains `has_more`.
- **Session router** `GET /sessions/{id}/messages` exposes `before` instead of `offset`.
- **Web UI** API client (`examples/web_ui/.../api/session.ts`) drops `offset` for an optional `before` and reads `has_more`. The default load (no `before`) now fetches the newest page — which also fixes the UI silently showing the oldest messages today.

This is a breaking change to the `before`/`has_more` shape of `GET /sessions/{id}/messages` (drops `offset`). The only in-tree caller is the web UI, updated here.

## Test plan

- [x] Updated/added `TestMessage` tests (against `fakeredis`): newest-window default, `has_more` true/false, full backwards cursor paging across three pages (every older message returned exactly once), unknown `before` id → empty, empty session, chronological order, session isolation.
- [x] Full `storage_redis_test.py` green (59) and `service_team_tools_test.py` + `app_lifespan_dedicated_test.py` green (38) — no regressions.
- [x] `pre-commit run` passes (black, flake8, mypy, pylint).

```bash
pytest tests/storage_redis_test.py tests/service_team_tools_test.py tests/app_lifespan_dedicated_test.py
# 97 passed
```

### DIFF
--- a/examples/web_ui/frontend/src/api/session.ts
+++ b/examples/web_ui/frontend/src/api/session.ts
@@ -13,6 +13,7 @@ import type {
 export interface MessagesResponse {
 	messages: Msg[];
 	is_running: boolean;
+	has_more: boolean;
 }
 
 export const sessionApi = {
@@ -41,10 +42,18 @@ export const sessionApi = {
 			agent_id: agentId,
 		}),
 
-	messages: (sessionId: string, agentId: string, offset = 0, limit = 50) =>
+	/**
+	 * Fetch a backwards page of persisted messages.
+	 *
+	 * Omit `before` to get the newest `limit` messages; pass the oldest
+	 * id of the previous page as `before` to page further back. The
+	 * response includes `has_more` so the caller knows whether another
+	 * older page exists.
+	 */
+	messages: (sessionId: string, agentId: string, before?: string, limit = 50) =>
 		client.get<MessagesResponse>(`/sessions/${sessionId}/messages`, {
 			agent_id: agentId,
-			offset: String(offset),
+			...(before ? { before } : {}),
 			limit: String(limit),
 		}),
 
--- a/src/agentscope/app/_router/_schema/_session.py
+++ b/src/agentscope/app/_router/_schema/_session.py
@@ -193,6 +193,13 @@ class ListMessagesResponse(BaseModel):
     is_running: bool = Field(
         description="Whether the session is currently running.",
     )
+    has_more: bool = Field(
+        description=(
+            "Whether older messages exist before this page. Use "
+            "``messages[0].id`` as the ``before`` cursor for the next "
+            "request."
+        ),
+    )
 
 
 class SessionStatusResponse(BaseModel):
--- a/src/agentscope/app/_router/_session.py
+++ b/src/agentscope/app/_router/_session.py
@@ -513,25 +513,37 @@ async def update_session(
 async def list_messages(
     session_id: str,
     agent_id: str = Query(description="Agent the session belongs to."),
-    offset: int = Query(0, ge=0, description="Pagination offset."),
     limit: int = Query(50, ge=1, le=200, description="Max messages."),
+    before: str
+    | None = Query(
+        None,
+        description=(
+            "Cursor: return messages older than this message id. "
+            "Omit to fetch the newest page."
+        ),
+    ),
     user_id: str = Depends(get_current_user_id),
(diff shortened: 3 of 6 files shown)


nolanchic:
Closing this — I noticed after submitting that #2080 was filed by @DavdGao, who already opened #2081 with the same cursor-based approach a few hours earlier and is actively iterating on it. My implementation is algorithmically equivalent and offers nothing #2081 doesn't already cover (and #2081 additionally preserves `offset` with a deprecation warning). Happy to step aside and let #2081 land. Sorry for the duplicate noise.

Issue #1648, opened 2026-05-23
[Bug]: Read→Edit chain broken: Read cache not populated due to `is_state_injected=False`


YingchaoX:
**Describe the bug**

In the real Agent execution chain (Write → Read → Edit), the `Read` tool successfully reads the file, but the subsequent `Edit` still fails with:

> Error: To edit a file, you must first read it using the Read tool.

The model then enters a Read/Edit retry loop and eventually hits `EXCEED_MAX_ITERS`.

**Root cause:** In PR #1549, `is_state_injected` was introduced with inconsistent assignments across builtin tools:

- `Read.is_state_injected = False`
- `Write.is_state_injected = False`
- `Edit.is_state_injected = True`

In `Toolkit.call_tool` (`src/agentscope/tool/_toolkit.py:299-305`), `_agent_state` is only injected when `is_state_injected=True`. Therefore:

1. `Read` never receives `_agent_state` → `Read.__call__` skips `cache_file()` → the read cache is **never populated**.
2. `Edit` receives `_agent_state` → checks the cache → cache is empty → throws the "must first read" error.
3. `Write` never receives `_agent_state` → skips the "file must be read first" safety check when overwriting existing files.

This means the read-cache safety mechanism is completely broken in real Agent execution.

**To Reproduce**

Run the `read_edit_chain` scenario from `demo/real_agent_e2e.py` (or any real Agent with `Read` + `Edit` tools), or use the following minimal reproduction that simulates the real `Toolkit.call_tool` path:

```python
import asyncio
from agentscope.message import ToolCallBlock
from agentscope.state import AgentState
from agentscope.tool import Edit, Read, Toolkit

async def main() -> None:
    state = AgentState()
    toolkit = Toolkit(tools=[Read(), Edit()])

    # Simulate real Agent execution: Read via Toolkit.call_tool
    read_call = ToolCallBlock(
        id="1",
        name="Read",
        input='{"file_path": "/tmp/test.txt"}',
    )
    async for chunk in toolkit.call_tool(read_call, state):
        if hasattr(chunk, "content"):
            print("Read:", chunk)

    # Simulate real Agent execution: Edit via Toolkit.call_tool
    edit_call = ToolCallBlock(
        id="2",
        name="Edit",
        input='{"file_path": "/tmp/test.txt", "old_string": "alpha", "new_string": "beta"}',
    )
    async for chunk in toolkit.call_tool(edit_call, state):
        if hasattr(chunk, "content"):
[...]

PR #2092, opened 2026-07-15
fix(tool): make read cache backend-aware so Read->Edit works in workspaces

### LINKED ISSUE #2084: [Bug]:  Read→Edit loop in DockerWorkspace because ToolContext uses host getmtime
### Prerequisites

### Background / Description

When using AgentScope 2.0.4 with `DockerWorkspace`, the builtin `Read` tool successfully reads a file inside the workspace container, but a subsequent `Edit` call for the same path fails with:

> Error: To edit a file, you must first read it using the Read tool.

The agent then retries `Read` and `Edit` repeatedly until it reaches the maximum number of iterations.

This looks similar to #1648, but the root cause appears to be different. In AgentScope 2.0.4, `Read.is_state_injected` is already `True`, so the state-injection issue described in #1648 has been fixed.

The current issue appears to be caused by `ToolContext.cache_file()` and `ToolContext.get_cache()` using the backend application's local filesystem:

```python
updated_at = await aiofiles.os.path.getmtime(file_path)
For a Docker workspace, a path such as /workspace/test.txt exists inside the workspace container but does not exist in the backend/application container.
The actual sequence is:
Read successfully reads the file through DockerBackend.read_file().
Read calls state.tool_context.cache_file(file_path, lines).
ToolContext.cache_file() calls host-local aiofiles.os.path.getmtime(file_path).
The call fails because the workspace path does not exist on the host filesystem.
The exception is silently caught, so no cache entry is created.
Edit cannot find the cache entry and reports that the file must be read first.
Expected behavior:
Files read through a Docker or remote backend should be added to the read cache.
Cache freshness checks should use the same backend as Read and Edit.
A successful Read followed by Edit for the same path should not enter a retry loop.
BackendBase already provides a backend-aware stat_mtime() method. The cache could use that method instead of host-local getmtime(), or use a backend-derived content fingerprint.
Related issues and changes:
#1648 — same visible Read/Edit loop, but caused by missing state injection.
#1903 — added builtin tool support for Docker and E2B workspaces.

### Error Messages

```shell
Error: To edit a file, you must first read it using the Read tool.
```

### Steps to Reproduce

1. Create a `DockerWorkspace` and write a test file inside it:

```python
import asyncio
import tempfile

from agentscope.workspace import DockerWorkspace


async def run_tool(toolkit, state, tool_call):
    async for chunk in toolkit.call_tool(tool_call, state):
        print(chunk)


async def main():
    with tempfile.TemporaryDirectory() as host_workdir:
        workspace = DockerWorkspace(
            workspace_id="read-edit-cache-repro",
            host_workdir=host_workdir,
        )
        await workspace.initialize()

        try:
            backend = workspace.get_backend()
            await backend.write_file(
                "/workspace/test.txt",
                b"alpha\n",
            )

            toolkit = Toolkit(tools=await workspace.list_tools())
            state = AgentState()

            await run_tool(
                toolkit,
                state,
                ToolCallBlock(
                    id="read-1",
                    name="Read",
                    input='{"file_path": "/workspace/test.txt"}',
                ),
            )

            print(
                "Cached files after Read:",
                state.tool_context.read_file_cache,
            )

            await run_tool(
                toolkit,
                state,
                ToolCallBlock(
                    id="edit-1",
                    name="Edit",
                    input=(
                        '{"file_path": "/workspace/test.txt", '
                        '"old_string": "alpha", '
                        '"new_string": "beta"}'
                    ),
                ),
            )
        finally:
            await workspace.close()


if __name__ == "__main__":
    asyncio.run(main())
```

2. Run the script with Docker available.

3. Observe that `Read` successfully returns the file content, but `state.tool_context.read_file_cache` remains empty.

4. Observe that the subsequent `Edit` call fails with:

```text
Error: To edit a file, you must first read it using the Read tool.
```

### Environment

- AgentScope Version: 2.0.4
- Python Version: 3.11.15
- OS: Linux
- Workspace Backend: DockerWorkspace

nolanchic:
## Problem

`ToolContext.cache_file()` and `get_cache()` used `aiofiles.os.path.getmtime(file_path)` on the **host** filesystem. For a sandbox backend (`DockerWorkspace`/`E2B`), a path like `/workspace/test.txt` exists only inside the container, so the host `getmtime` raised `FileNotFoundError`:

- `cache_file()` swallowed the exception and returned without caching → no cache entry was created.
- `get_cache()` then found no entry → `Edit`/`Write` failed with `Error: To edit a file, you must first read it using the Read tool.`

The agent then retried `Read`+`Edit` until it hit the max iteration count. Reproducer and analysis are in #2084.

`Read.is_state_injected` is already `True` in 2.0.4, so this is a different root cause from the state-injection issue in #1648.

## Fix

`Read`/`Edit`/`Write` already hold a `self._backend: BackendBase`. Before calling the cache they now obtain the mtime from the **backend** (`BackendBase.stat_mtime`, which `stat -c %Y` inside the sandbox for Docker/E2B and `os.stat().st_mtime` for `LocalBackend`) and pass it into `cache_file`/`get_cache`:

```python
mtime = await self._backend.stat_mtime(file_path)
cache = await tool_context.get_cache(file_path, mtime=mtime)
...
await tool_context.cache_file(file_path=file_path, lines=lines, mtime=mtime)
```

`ToolContext` changes:
- `get_cache(file_path, mtime=None)`: uses the supplied mtime for the staleness check. A `None` mtime (backend could not stat) skips the check and returns the cached entry as-is, so a freshly read sandbox file stays cacheable.
- `cache_file(file_path, lines, mtime=None)`: uses the supplied mtime; falls back to host `getmtime` only when no mtime is supplied, preserving the previous behavior for any caller that does not pass one.

Because `LocalBackend.stat_mtime` is `os.stat().st_mtime`, the local-backend path is behaviorally unchanged — the new code just routes the same value through the backend.

## Tests

New `tests/builtin_file_cache_backend_test.py` reproduces the #2084 shape without Docker: an in-memory backend whose paths (`/workspace/test.txt`) do **not** exist on the host filesystem. It asserts:

- `Read` populates `tool_context.read_file_cache` for a host-invisible path
- `Edit` after `Read` succeeds (the loop from #2084)
- `Write` after `Read` succeeds
- `Edit` without `Read` still fails with "must first read"

All existing `builtin_file_cache_test` / `builtin_read_test` / `builtin_edit_test` / `builtin_write_test` / `backend_local_test` continue to pass (71 tests in those files). `black`/`flake8` clean.

Fixes #2084


DavdGao (approved):
lgtm

PR #2093, opened 2026-07-16
fix(agent): preserve multi-tool pairs during context compression

### LINKED ISSUE #2090: [Bug]: Context compression can split a multi-tool call/result batch and send an invalid provider request
## Prerequisites

## Background / Description

Context compression can split a valid multi-tool assistant message between a tool call and its matching result.

I reproduced this on the current `main` branch at commit `1f433fd7ff2b27f9d98f9385779b5c436c6487ee`.

A normal Agent run with two tool calls naturally generated one `AssistantMsg` with the following `Msg.content` shape. The block order and IDs below are from the real run; only the repeated tool output text is abbreviated:

```json
[
  {
    "type": "thinking",
    "id": "fb50b86ab2174054888b74b52850d400"
  },
  {
    "type": "tool_call",
    "id": "call_00_uER2QtaMLuqheXpzyCT47644",
    "name": "first_long_tool",
    "input": "{\"value\": \"hello\"}"
  },
  {
    "type": "tool_call",
    "id": "call_01_BII3hlgWzRcrVWaM7HTk7265",
    "name": "second_long_tool",
    "input": "{\"value\": \"world\"}"
  },
  {
    "type": "tool_result",
    "id": "call_00_uER2QtaMLuqheXpzyCT47644",
    "name": "first_long_tool",
    "output": ["FIRST:hello: ... repeated ..."]
  },
  {
    "type": "tool_result",
    "id": "call_01_BII3hlgWzRcrVWaM7HTk7265",
    "name": "second_long_tool",
    "output": ["SECOND:world: ... repeated ..."]
  },
  {
    "type": "thinking",
    "id": "a6b20208e3b34407b8692f14245c2fc8"
  },
  {
    "type": "text",
    "id": "70b6fc8e0cc445cd8992ad89d2c4a345",
    "text": "Both tools have been called..."
  }
]
```

With a reserve budget of 548 tokens, `Agent._split_context_for_compression()` produced:

```text
to_compress:
thinking, call1, call2, result1

to_reserve:
result2, thinking, text
```

The compression request therefore contains two tool calls but only the result for the first call. A strict OpenAI-compatible provider rejects the request before AgentScope can update `state.context`.

### Expected behavior

- Context compression should never split matching tool calls and results.
- If a complete multi-tool batch cannot fit in the reserve budget, the complete batch should be moved to the compressed side.
- Both the compression request and the reserved context should remain valid.

### Root cause

The current boundary adjustment is performed only once.

Starting from a boundary after `call1`, it detects that `result1` is unmatched in the reserved suffix and moves the boundary through `result1`. This also moves `call2` to the compressed side while leaving `result2` reserved, creating a new unmatched pair that is not checked again:

```text
Original:
call1, call2, result1, result2

Initial boundary:
call1 | call2, result1, result2

After the one-pass adjustment:
call1, call2, result1 | result2
```

This is an independently reproducible context-compression bug. I am not claiming that it is the root cause of the separately reported #2067.

## Error Messages

```shell
BadRequestError: Error code: 400 - {
  'error': {
    'message': "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. (insufficient tool messages following tool_calls message)",
    'type': 'invalid_request_error',
    'param': None,
    'code': 'invalid_request_error'
  }
}
```

## Steps to Reproduce

Run the following from an AgentScope source checkout so that `tests.utils.MockModel` is available:

```python
import asyncio
import json

from tests.utils import MockModel

    AssistantMsg,
    TextBlock,
    ToolCallBlock,
    ToolResultBlock,
    ToolResultState,
    UserMsg,
)


async def main():
    agent = Agent(
        name="Friday",
        system_prompt="",
        model=MockModel(context_size=1000),
        toolkit=Toolkit(),
        state=AgentState(
            session_id="multi-tool-compression",
            context=[
                UserMsg("User", "old" * 80),
                AssistantMsg(
                    "Friday",
                    [
                        ToolCallBlock(
                            id="tc1",
                            name="first_tool",
                            input=json.dumps({"value": "a" * 80}),
                        ),
                        ToolCallBlock(
                            id="tc2",
                            name="second_tool",
                            input=json.dumps({"value": "b" * 80}),
                        ),
                        ToolResultBlock(
                            id="tc1",
                            name="first_tool",
                            output=[TextBlock(text="first result " * 8)],
                        ),
                        ToolResultBlock(
                            id="tc2",
                            name="second_tool",
                            output=[TextBlock(text="second result " * 8)],
                        ),
                        TextBlock(text="Both tools completed."),
                    ],
                ),
                UserMsg("User", "latest question"),
            ],
        ),
    )

    to_compress, to_reserve = (
        await agent._split_context_for_compression(
            to_reserved_tokens=86,
            tools=[],
        )
    )

    print([
        [(block.type, block.id) for block in msg.get_content_blocks()]
        for msg in to_compress
    ])
    print([
        [(block.type, block.id) for block in msg.get_content_blocks()]
        for msg in to_reserve
    ])

    formatted = await OpenAIChatFormatter().format(to_reserve)
    print(json.dumps(formatted, indent=2))

    assert formatted[0]["role"] == "tool"
    assert formatted[0]["tool_call_id"] == "tc2"


asyncio.run(main())
```

Run:

```shell
python reproduce.py
```

The reserved context starts with an orphan tool result for `tc2`.

## Environment

- AgentScope Version: 2.0.4 / main commit `1f433fd7ff2b27f9d98f9385779b5c436c6487ee`
- Python Version: 3.11.9
- OS: Windows 10.0.22621
- Live verification provider: DeepSeek `deepseek-v4-pro` through the OpenAI-compatible API

## Proposed fix

A focused fix could repeatedly validate the adjusted boundary until it becomes stable. This preserves the existing block-level compression behavior while ensuring that moving one result cannot create another split pair.

No public API change or new dependency should be needed. If this scope looks appropriate, I would be happy to submit a focused PR with a deterministic multi-tool regression test.

qiuC123:
## AgentScope Version

2.0.4

## Problem

Context compression adjusts its block boundary once when the reserved suffix
contains a tool result without its matching tool call.

For an interleaved multi-tool batch such as:

```text
call1, call2, result1, result2
```

moving `result1` into the compressed side also moves `call2`, while `result2`
remains reserved. This creates a new split pair that is not checked again,
causing invalid provider requests.

## Solution

Recheck the reserved suffix after every boundary adjustment until no unmatched
tool results remain. This reaches a stable boundary while preserving the
existing block-level compression strategy.

## Changes

- Replace the one-pass tool-pair boundary adjustment with a fixed-point loop.
- Add a deterministic regression test with two tool calls followed by their
  two results.
- Verify that the complete multi-tool batch stays on the compressed side while
  later text remains reserved.

## Testing

- `pytest tests/compress_context_test.py::ContextCompressionTest::test_split_multi_tool_pairs_reaches_stable_boundary -vv`
- `pytest tests/compress_context_test.py tests/compress_tool_result_test.py tests/formatter_openai_chat_test.py tests/agent_basic_test.py -q`
  - 30 passed
- `pre-commit run --files src/agentscope/agent/_agent.py tests/compress_context_test.py`
  - all hooks passed
- Manual smoke test confirmed both call/result pairs remain together.

## Notes for Reviewer

Fixes #2090.

This is an internal correctness fix with no public API, dependency, or
documentation changes.

## Checklist

- [x] Related documentation is not required for this internal bug fix

### DIFF
--- a/tests/compress_context_test.py
+++ b/tests/compress_context_test.py
@@ -700,31 +700,117 @@ class ContextCompressionTest(IsolatedAsyncioTestCase):
         )
 
         self.assertListEqual(
-            [msg.id for msg in to_compress],
-            ["old-user", "multi-tool-msg"],
+            [msg.model_dump() for msg in to_compress],
+            [
+                {
+                    "id": "old-user",
+                    "created_at": AnyString(),
+                    "finished_at": AnyString(),
+                    "name": "User",
+                    "role": "user",
+                    "content": [
+                        {
+                            "id": AnyString(),
+                            "type": "text",
+                            "text": "old" * 80,
+                        },
+                    ],
+                    "metadata": {},
+                    "usage": None,
+                },
+                {
+                    "id": "multi-tool-msg",
+                    "created_at": AnyString(),
+                    "finished_at": None,
+                    "name": "Friday",
+                    "role": "assistant",
+                    "content": [
+                        {
+                            "id": "tc1",
+                            "type": "tool_call",
+                            "name": "first_tool",
+                            "input": json.dumps({"value": "a" * 80}),
+                            "state": "pending",
+                            "suggested_rules": [],
+                        },
+                        {
+                            "id": "tc2",
+                            "type": "tool_call",
+                            "name": "second_tool",
+                            "input": json.dumps({"value": "b" * 80}),
+                            "state": "pending",
+                            "suggested_rules": [],
+                        },
+                        {
+                            "id": "tc1",
+                            "type": "tool_result",
+                            "name": "first_tool",
+                            "output": [
+                                {
+                                    "id": AnyString(),
+                                    "type": "text",
+                                    "text": "first result " * 8,
+                                },
+                            ],
+                            "state": "success",
+                            "metadata": {},
+                        },
+                        {
+                            "id": "tc2",
+                            "type": "tool_result",
+                            "name": "second_tool",
+                            "output": [
+                                {
+                                    "id": AnyString(),
+                                    "type": "text",
+                                    "text": "second result " * 8,
+                                },
+                            ],
+                            "state": "success",
+                            "metadata": {},
+                        },
+                    ],
+                    "metadata": {},
+                    "usage": None,
+                },
+            ],
         )
         self.assertListEqual(
+            [msg.model_dump() for msg in to_reserve],
             [
-                (block.type, block.id)
-                for block in to_compress[-1].get_content_blocks()
+                {
+                    "id": "multi-tool-msg",
+                    "created_at": AnyString(),
+                    "finished_at": None,
+                    "name": "Friday",
+                    "role": "assistant",
+                    "content": [
+                        {
+                            "id": "final-text",
+                            "type": "text",
+                            "text": "Both tools completed.",
+                        },
+                    ],
+                    "metadata": {},
+                    "usage": None,
+                },
+                {
+                    "id": "latest-user",
+                    "created_at": AnyString(),
+                    "finished_at": AnyString(),
+                    "name": "User",
+                    "role": "user",
+                    "content": [
+                        {
+                            "id": AnyString(),
+                            "type": "text",
+                            "text": "latest question",
+                        },
+                    ],
+                    "metadata": {},
+                    "usage": None,
+                },
             ],
-            [
-                ("tool_call", "tc1"),
-                ("tool_call", "tc2"),
-                ("tool_result", "tc1"),
-                ("tool_result", "tc2"),
-            ],
-        )
-        self.assertListEqual(
-            [msg.id for msg in to_reserve],
-            ["multi-tool-msg", "latest-user"],
-        )
-        self.assertListEqual(
-            [
-                (block.type, block.id)
-                for block in to_reserve[0].get_content_blocks()
-            ],
-            [("text", "final-text")],
         )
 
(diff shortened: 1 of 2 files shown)


tests/compress_context_test.py
```diff
@@ -634,6 +636,97 @@ async def test_split_function(self) -> None:
...
+                    ),
+                ],
+            ),
+            toolkit=Toolkit(),
+        )
+
+        to_compress, to_reserve = await agent._split_context_for_compression(
+            to_reserved_tokens=86,
+            tools=[],
+        )
```
qbc2016, line 700: The existing block-level split tests in this file all use [`_.model_dump() for _ in to_compress`] / `to_reserve` for full structural assertions (see the tests just above this one). Please switch to `model_dump()` to stay consistent with the rest of the file.

qbc2016 (commented):
Thank you for the pr. Please see the inline comments.

tests/compress_context_test.py
qiuC123: Thanks for the suggestion. I updated the regression test to use full `model_dump()` assertions for both `to_compress` and `to_reserve`, consistent with the existing block-level split tests. The related test suite passes with 30 tests, and all pre-commit hooks pass. The change is included in commit c528a2b.

qbc2016 (approved):
LGTM.

PR #2100, opened 2026-07-16
fix(webui): scroll newly loaded sessions to bottom

### LINKED ISSUE #2097: fix(webui): auto-scroll to bottom when navigating to or opening a session
**Background**

When a user navigates to a session (either by clicking a session in the sidebar, switching agents, or directly via URL), the message list loads the persisted history but does not automatically scroll to the bottom. Since the most recent messages are at the end of the list, users land at the top of the conversation and must manually scroll down to see the latest content. This is a poor first-impression experience, especially for sessions with long histories.

The root cause is in `ChatContent.tsx`: the auto-scroll `useEffect` only triggers when `msgs.length` increases relative to `prevMsgCountRef.current`, and `prevMsgCountRef` starts at 0. On initial load, the condition `currentCount > prevCount && prevCount > 0` evaluates to false because `prevCount` is 0, so the scroll never fires.

**Changes**

- In `ChatContent.tsx`, adjust the auto-scroll `useEffect` to also scroll to the bottom on the initial message load (when `prevMsgCountRef.current === 0` and `msgs.length > 0`). This ensures that whenever a session's messages are first populated, the view jumps to the latest content.
- Alternatively, expose an imperative scroll-to-bottom method from `ChatContent` (or use a ref callback) that `ChatViewport` can call after messages are loaded, providing more explicit control over the timing.
- Ensure the scroll happens after the DOM has rendered the full message list (e.g., in a `useLayoutEffect` or after a `requestAnimationFrame`) to avoid scrolling to an incomplete height.

**Affected files**

- `examples/web_ui/frontend/src/components/chat/ChatContent.tsx` — auto-scroll `useEffect` logic and `prevMsgCountRef` initialization handling

Part of #2094

JanCong:
## Summary

- scroll to the latest message when a session first loads a non-empty history
- preserve the existing near-bottom guard for later messages and active streaming updates
- keep the fix scoped to `ChatContent.tsx` without adding dependencies or changing component APIs

## Root cause

The auto-scroll effect only considered message growth when `prevMsgCountRef.current` was already greater than zero. On the first persisted-history load, the previous count is zero, so the effect skipped scrolling entirely.

## Validation

- `pnpm exec prettier --check frontend/src/components/chat/ChatContent.tsx`
- `pnpm --filter frontend lint` (0 errors; 16 existing warnings also present on `main`)
- `pnpm --filter frontend build`
- local decision-matrix assertions for empty, initial-load, later-growth, and active-update cases (6/6 passed)

Fixes #2097

### DIFF
--- a/examples/web_ui/frontend/src/components/chat/ChatContent.tsx
+++ b/examples/web_ui/frontend/src/components/chat/ChatContent.tsx
@@ -65,21 +65,18 @@ const ChatContentComponent: React.FC<ChatContentProps> = ({
 		const prevCount = prevMsgCountRef.current;
 
 		const isActive = phase !== 'idle';
-		const shouldCheck =
+		const isInitialLoad = prevCount === 0 && currentCount > 0;
+		const hasRelevantUpdate =
 			(currentCount > prevCount && prevCount > 0) || (isActive && prevCount > 0);
+		const shouldScroll = isInitialLoad || (hasRelevantUpdate && wasNearBottomRef.current);
 
-		if (shouldCheck && scrollAreaRef.current) {
+		if (shouldScroll && scrollAreaRef.current) {
 			const { scrollHeight } = scrollAreaRef.current;
 
-			// Check if user was near bottom before content changed
-			const isNearBottom = wasNearBottomRef.current;
-
-			if (isNearBottom) {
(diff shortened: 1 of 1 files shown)


DavdGao (approved):
lgtm

PR #2101, opened 2026-07-16
fix(webui): auto-open task panel on updates

### LINKED ISSUE #2095: fix(webui): auto-open task panel when task updates arrive via CustomEvent
**Background**

When the agent creates or updates tasks during a reply, the backend pushes a `CustomEvent` with `name="state_updated"` containing the latest `tasks_context`. The `ChatViewport` component receives this via the `onStateUpdated` callback and updates the local `tasksContext` state. However, if the Task Panel (right-side dockable panel) is not already open, the user has no visual indication that tasks have been created or updated — they must manually open it from the panel dropdown menu. This defeats the purpose of real-time task tracking and makes it easy to miss plan progress during long-running agent replies.

**Changes**

- In `ChatViewport.tsx`, modify the `handleStateUpdated` callback to automatically open the `'plan'` panel (via `openPanelInLayout`) when a `tasks_context` update arrives and the panel is not already visible.
- Ensure the auto-open only triggers on genuine task data changes (not on every `state_updated` event), e.g., by comparing the incoming `tasks_context` against the previous one to avoid redundant panel toggling.
- Respect user intent: if the user has explicitly closed the Task Panel during the current session, consider whether to suppress auto-reopen until the next session switch. A simple approach is to always auto-open on first task update per session, which aligns with the expectation that task visibility is important.

**Affected files**

- `examples/web_ui/frontend/src/pages/chat/ChatViewport.tsx` — `handleStateUpdated` callback and `panelLayout` state management

Part of #2094

RerankerGuo:
## Summary

- compare incoming `tasks_context` snapshots before updating task UI state
- automatically open the plan panel on the first non-empty task change in each session
- remember an explicit plan-panel close so later updates in the same session do not reopen it
- reset the auto-open guard and task snapshot when the agent/session changes

Permission-only `state_updated` events carry the current task context too. Snapshot comparison prevents those events from opening the task panel when task data did not change. `openPanelInLayout` keeps the operation idempotent when the panel is already visible.

Fixes #2095

## Tests

- `pnpm --filter frontend build`
- `pnpm --filter frontend lint` (no errors; existing warnings only)
- `pnpm exec prettier --write frontend/src/pages/chat/ChatViewport.tsx`
- `git diff --check`

### DIFF
--- a/examples/web_ui/frontend/src/pages/chat/ChatViewport.tsx
+++ b/examples/web_ui/frontend/src/pages/chat/ChatViewport.tsx
@@ -1,7 +1,7 @@
 import type { PermissionContext } from '@agentscope-ai/agentscope/permission';
 import type { TaskContext } from '@agentscope-ai/agentscope/state';
 import { BookText, ChevronDown, Database, ListTodo, PanelRight, ShieldCheck } from 'lucide-react';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import type { ChatModelConfig, SessionKnowledgeConfig, TTSModelConfig } from '@/api';
 import { sessionApi } from '@/api';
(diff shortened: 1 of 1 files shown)


DavdGao:
Thanks for the PR!

While reviewing I noticed the change duplicates logic that already exists on the backend: `StateChangeMiddleware` computes a hash to detect whether `tasks_context` changed, and this PR adds a second `JSON.stringify` comparison in `ChatViewport` to detect the same thing. The reason it's needed is that `_state_hash` hashes both fields together and `_publish_state` always sends both, so a permission-only change still carries an unchanged `tasks_context`.

If the backend only sends the part that actually changed, receiving `tasks_context` becomes sufficient proof that the tasks changed, and the frontend just opens the panel. That reduces the fix to:

1. **`src/agentscope/app/middleware/_state_change_middleware.py`** — hash `tasks_context` and `permission_context` separately, and include only the changed field(s) in the `CustomEvent` payload. Backward compatible: both frontend handlers are already guarded by `if (value.x)`. Worth noting in `CustomEvent`'s docstring that `state_updated`'s `value` is partial — a missing key means "unchanged".

2. **`ChatViewport.tsx`** — drop `tasksContextSnapshotRef` and the snapshot write in the `view` effect. In `handleStateUpdated`, on `tasks_context`: `setTasksContext` and, if `tc.tasks?.length` and the user hasn't closed the panel in this session, `openPanelInLayout(layout, 'plan')` (idempotent, so no auto-open-once guard needed). Keep one ref for "user dismissed", reset on `agentId`/`sessionId` change, and set it inside the `setPanelLayout` updater so `togglePanel` keeps its empty dep array.

One bug in the current diff regardless of which route you take: `nextTasksContext.tasks.length` is unguarded while `value.tasks_context` is only checked for truthiness — a payload without `tasks` would throw inside the SSE handler.

RerankerGuo:
Superseded by #2183, which implements the backend/frontend design requested in the maintainer review here. Closing this earlier frontend-only version to keep one active implementation for #2095.

PR #2105, opened 2026-07-16
fix(tool): use backend mtime for file cache validation

### LINKED ISSUE #2084: [Bug]:  Read→Edit loop in DockerWorkspace because ToolContext uses host getmtime

qiboyu301-crypto:
## PR Title Format

`fix(tool): use backend mtime for file cache validation`

## AgentScope Version

`2.0.4.post1`

## Description

Fixes #2084.

The builtin Read tool could read files through a Docker or remote workspace backend, but the subsequent Edit or Write call still failed with a "must first read" error. The file cache was validating workspace paths with the host filesystem's `getmtime()`, where those paths do not exist.

This change:

- validates the cache with the filesystem backend's `stat_mtime()`;
- passes the mtime result at call time without storing the backend in `ToolContext`;
- keeps the existing host-filesystem fallback for callers that omit `mtime`;
- treats an explicit `mtime=None` as unverifiable and does not reuse the cache;
- adds regression tests using a non-local backend whose files are not visible to the host filesystem.

This overlaps with #2092, but explicitly distinguishes an omitted mtime from a backend that returns `None`, so an unverifiable cache entry is not trusted.

Testing performed:

- `pre-commit run --all-files` passed.
- Windows/Python 3.12: 72 passed, 2 skipped.
- Linux/Python 3.11 against the final commit: 82 relevant tests passed.
- Broader Linux suite excluding the two Docker image-building test files: 1295 passed, 55 skipped, and 300 subtests passed.
- A real DockerBackend verification passed for Read → Edit, Read → Write, unread-file rejection, external mtime invalidation, AgentState serialization, and workspace teardown.

The repository's Docker workspace test suites require rebuilding an image from external resources unavailable on the test server, so the affected behavior was verified separately against a real DockerBackend.

## Checklist

- [x] Related documentation is not required because this is an internal cache fix with no public API or user-facing documentation changes

### DIFF
--- a/src/agentscope/state/_state.py
+++ b/src/agentscope/state/_state.py
@@ -20,6 +20,13 @@ from ..message import (
 from ..permission import PermissionContext
 
 
+class _UseHostMtime:
+    """Sentinel indicating that the host filesystem should be inspected."""
+
+
+_USE_HOST_MTIME = _UseHostMtime()
+
+
 class ReadCacheEntry(BaseModel):
     """The read file cache."""
 
@@ -43,45 +50,74 @@ class ToolContext(BaseModel):
     """The names of the activated tool groups, each group contains a set of
     tools."""
 
-    async def get_cache(self, file_path: str) -> ReadCacheEntry | None:
+    async def get_cache(
+        self,
+        file_path: str,
+        mtime: float | None | _UseHostMtime = _USE_HOST_MTIME,
+    ) -> ReadCacheEntry | None:
         """Get cached file content if still valid.
 
         Args:
-            file_path: The absolute path of the file.
+            file_path (`str`):
+                The absolute path of the file.
+            mtime (`float | None`, optional):
+                Modification time obtained from the filesystem that owns the
+                file. When omitted, the host filesystem is inspected for
+                backward compatibility. An explicit ``None`` means the
+                modification time could not be determined, so the cache is
+                treated as invalid.
 
         Returns:
-            The cached entry if valid, otherwise None.
+            `ReadCacheEntry | None`:
+                The cached entry if valid, otherwise None.
         """
 
         # Find the cache entry
         for entry in self.read_file_cache:
             if entry.file_path == file_path:
                 # Check if cache is still valid
-                try:
-                    updated_at = await aiofiles.os.path.getmtime(file_path)
-                    if updated_at == entry.updated_at:
-                        return entry
-                    else:
-                        # Cache is outdated, remove it
-                        self.read_file_cache.remove(entry)
-                        return None
-                except Exception:
-                    # File might not exist anymore
+                if isinstance(mtime, _UseHostMtime):
+                    try:
+                        mtime = await aiofiles.os.path.getmtime(file_path)
+                    except Exception:
+                        mtime = None
+
+                if mtime is None or mtime != entry.updated_at:
+                    # File might no longer exist, be unverifiable, or have
+                    # changed since it was read.
                     self.read_file_cache.remove(entry)
                     return None
+                return entry
         return None
 
-    async def cache_file(self, file_path: str, lines: list[str]) -> None:
+    async def cache_file(
+        self,
+        file_path: str,
+        lines: list[str],
+        mtime: float | None | _UseHostMtime = _USE_HOST_MTIME,
+    ) -> None:
         """Cache file content with LRU eviction.
 
         Args:
-            file_path: The absolute path of the file.
-            lines: The lines of the file content.
+            file_path (`str`):
+                The absolute path of the file.
+            lines (`list[str]`):
+                The lines of the file content.
+            mtime (`float | None`, optional):
+                Modification time obtained from the filesystem that owns the
+                file. When omitted, the host filesystem is inspected for
+                backward compatibility. An explicit ``None`` means the
+                modification time could not be determined and caching is
+                skipped.
         """
-        try:
-            updated_at = await aiofiles.os.path.getmtime(file_path)
-        except Exception:
-            # Cannot get mtime, skip caching
+        if isinstance(mtime, _UseHostMtime):
+            try:
+                mtime = await aiofiles.os.path.getmtime(file_path)
+            except Exception:
+                mtime = None
+
+        if mtime is None:
+            # Cannot validate the entry later, so do not cache it.
             return
 
         # Calculate size in KB
@@ -113,7 +149,7 @@ class ToolContext(BaseModel):
         self.read_file_cache.append(
             ReadCacheEntry(
                 lines=lines,
-                updated_at=updated_at,
+                updated_at=mtime,
                 bytes=new_entry_bytes,
                 file_path=file_path,
             ),
--- a/src/agentscope/tool/_builtin/_edit.py
+++ b/src/agentscope/tool/_builtin/_edit.py
@@ -295,7 +295,11 @@ Usage:
 
         content = None
         if _agent_state is not None:
-            cache = await _agent_state.tool_context.get_cache(file_path)
+            mtime = await self._backend.stat_mtime(file_path)
+            cache = await _agent_state.tool_context.get_cache(
+                file_path,
+                mtime=mtime,
+            )
             if cache is None:
                 # Haven't read this file before
                 return ToolChunk(
--- a/src/agentscope/tool/_builtin/_read.py
+++ b/src/agentscope/tool/_builtin/_read.py
@@ -225,8 +225,13 @@ Usage:
         try:
             # Read file content via backend
             lines = None
+            mtime = None
             if _agent_state is not None:
-                cache = await _agent_state.tool_context.get_cache(file_path)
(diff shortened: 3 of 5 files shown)


HermanShi:
We hit this same root cause in a different deployment shape and can confirm the patch addresses it. Adding an independent data point since the PR is still awaiting review.

**Setup**: a custom `BackendBase` implementation backed by our own sandbox containers, not `DockerWorkspace`. The agent worker runs in a *different* container from the sandbox, so a path like `/workspace/sessions/<id>/notes.txt` exists only in the sandbox's mount namespace.

**Symptom**: identical to #2084 — `Read` succeeds, then `Write`/`Edit` reject every call with "must first read it". In our case the failure is fully deterministic rather than intermittent: `read_file_cache` stays empty even immediately after a successful `Read`, so no number of retries can ever satisfy the precondition.

**Measured against a live sandbox** (not mocks), same path, same instant:

| call | result |
|---|---|
| `BackendBase.stat_mtime(path)` | `1785670357.0` |
| `aiofiles.os.path.getmtime(path)` | `FileNotFoundError` |

That is the bug in two lines: the file the tool just read is invisible to the process validating the cache.

**Why this patch shape is good for third-party backends**: we never implemented `stat_mtime` ourselves. `BackendBase`'s default (`stat -c %Y || stat -f %m` over `exec_shell`) already returns the correct value through our sandbox, as measured above. So the fix requires **zero changes on our side** — a nice property for anyone maintaining a custom backend.

**Our current workaround**, for anyone else who lands here: we withhold `AgentState` from `Write`/`Edit` at mount time, which skips the read-cache precondition entirely. It restores functionality but permanently forfeits the read-before-write guard, so we are tracking this PR rather than treating that as a fix.

Happy to test the branch against our sandbox backend if that would help move this along.

qiboyu301-crypto:
One more finding while testing this fix with `DockerBackend`:

Making the cache backend-aware fixes the original Read → Edit failure, but it also exposes an existing cache consistency issue. After a successful `Edit` or `Write`, the corresponding entry in `read_file_cache` is not refreshed or invalidated.

This matters because `DockerBackend` writes may leave the file mtime unchanged (`0.0` in my test). In that case, `get_cache()` still considers the old entry valid, and a later `Edit` rebuilds the whole file from stale `cache.lines`.

For example:

1. `Read` caches `A + B`
2. `Edit 1` writes `A1 + B`, but the cache still contains `A + B`
3. The mtime remains unchanged
4. `Edit 2` uses the old cache and writes `A + B2`, silently losing the first change

I don't think #2105 creates the underlying cache lifecycle issue, but it makes it reachable for remote backends because the cache can now be populated successfully.

Should we refresh or invalidate the cache after a successful `Edit`/`Write`, and add a regression test for `Read → Edit → Edit` with an unchanged mtime? I'm happy to update #2105 accordingly.

qiboyu301-crypto:
I also posted this on #2105 since it directly affects the proposed fix.

syyy44:
Following up on your Aug 11 note about the cache lifecycle — I hit the same thing and have a working fix with tests, so flagging it here before anyone duplicates work.

**Reproduced the shape you described.** Modelling a backend whose write leaves the mtime unchanged (you measured a constant on DockerBackend; I used an in-memory backend returning a constant `stat_mtime`, with the path also resolvable on the host so today's host-`getmtime` check likewise reports "fresh"):

```
Read       caches  alpha = 1  /  beta = 2
Edit1  ->  writes  alpha = 111 / beta = 2      (cache still pre-edit1)
Edit2  ->  writes  alpha = 1   / beta = 222    <- edit1 silently gone
```

Both calls report `Successfully replaced 1 occurrence`. The mechanism is the one you identified: `_edit.py` rebuilds the entire file from `cache.lines` (L317 → L368 → L378), so a stale cache is written back wholesale.

**One extra hole that matters for your patch.** In `ToolContext.cache_file`, the "remove existing cache for this file" step (L94) sits *after* the mtime lookup (L84), so the `except: return` at L86 exits with the previous entry still in the list. Since #2105 treats an explicit `mtime=None` as unverifiable, re-caching a path on a backend that can't stat it would keep serving the older lines rather than dropping them. Moving the removal ahead of the lookup closes that.

**What I implemented**, if it's useful:

- `Edit` refreshes the cache with the content it just wrote — keeps `Edit → Edit` correct without forcing an extra `Read` round-trip.
- `Write` replaces the file wholesale and doesn't own a line view of the result, so it invalidates the entry explicitly (new `ToolContext.invalidate_cache`). This preserves the existing "must first read" contract — and notably keeps `test_write_invalidates_cache` passing, which currently only passes as a side effect of the mtime changing.
- `cache_file` drops the previous entry up front, per above.

Verified that the post-`Edit` cached lines are byte-identical to what a fresh `Read` would cache, across no-trailing-newline, CRLF, and edit-to-empty inputs — that was my main worry, given the earlier `"\n".join` doubling regression. Plus 4 regression tests (`Edit→Edit`, `replace_all` then `Edit`, `Write` invalidation, and the `cache_file` early-return hole); full suite **1966 passed** with no regressions. My verification is against a fake backend and the test suite only — I don't have a real Docker environment here, so I can't independently confirm the DockerBackend measurement you made.

Since this is your finding and you offered to fold it into #2105, I'd rather not cut across you. Happy either way:

1. you take it into #2105 — I can paste the diff or the test file here, or
2. I open a small PR stacked on #2105 (or #2092, whichever lands) that only does the post-write cache maintenance.

Which do you prefer?

DavdGao:
Fixed in #2092

PR #2115, opened 2026-07-17
fix(webui): recover failed agent replies

### LINKED ISSUE #2110: [Bug]: WebUI message stalls indefinitely (spinner never stops) when LLM backend crashes/freezes
### Prerequisites

### Background / Description

When sending a message in the WebUI, if the underlying LLM model/service crashes, hangs, or becomes unresponsive (e.g., OOM, process killed, network partition), the frontend UI enters an unrecoverable state. The loading spinner continues to animate indefinitely—sometimes for hours—and the "Stop" button becomes non-functional. The user is forced to manually refresh the browser to recover the session.
Expected behavior: When the LLM backend fails, the WebUI should detect the connection drop or timeout, stop the spinner, and display an error message (e.g., "Model response failed" or "Connection lost"). The "Stop" button should either work to cancel the pending request or be replaced by a "Retry/Refresh" option.

### Error Messages

```shell
Backend logs typically show the LLM crash (e.g., CUDA out of memory, grpc timeout, connection reset by peer). However, the browser console shows no explicit error; the WebSocket/SSE connection often remains technically "open" on the client side even though the backend stopped streaming. No ReplyEndEventor error event is received by the frontend to trigger the UI reset.
```

### Steps to Reproduce

1. Start the AgentScope service and open the WebUI.

2. Configure a chat session with an LLM model (e.g., a local vLLM instance or a remote endpoint).
3. Simulate a model crash or freeze

4. Observe the WebUI:
The message bubble shows a loading spinner.
The spinner continues spinning indefinitely (tested up to 2+ hours).

### Environment

- AgentScope Version: 2.0.4
- Python Version: 3.12
- OS: windows 11

Jupiter363:
## Summary

- emit a terminal `ReplyEndEvent(finished_reason=ReplyFinishedReason.ERROR)` when an active agent reply fails, while preserving the original exception behavior
- map errored reply termination to AG-UI `RUN_ERROR`
- finalize unfinished WebUI messages locally when the event stream fails or closes, the server reports an idle unfinished tail, or Stop fallback handling is triggered
- use the SDK 0.0.14 first-class `Msg.finished_reason` field for client-local recovery

## Root cause

When model or runtime execution raised inside `Agent._reply_impl()`, direct consumers of `agent.reply_stream()` received the exception without a matching terminal event. Those consumers could remain in a streaming state indefinitely. The WebUI also needed a client-side fallback for cases where the server process or event stream disappears before any terminal event can arrive.

This branch is rebased onto current `main` and keeps the error model introduced by #2133: `ReplyFinishedReason`, `Msg.finished_reason`, and `Msg.error`. It does not add metadata-based reason plumbing or duplicate `MessageBubble` error rendering.

## Validation

- 47 targeted Python tests passed, covering agent reply failures, message aggregation, AG-UI mapping, error classification, and interruption
- changed-file pre-commit checks passed, including AST, mypy, Black, Flake8, Pylint, and Pyroma
- WebUI production build passed against `@agentscope-ai/agentscope` 0.0.14
- ESLint completed with 0 errors (16 existing warnings outside the changed code)
- `git diff --check` passed

Fixes #2110

### DIFF
--- a/examples/web_ui/frontend/src/hooks/useMessages.ts
+++ b/examples/web_ui/frontend/src/hooks/useMessages.ts
@@ -1,4 +1,4 @@
-import { EventType } from '@agentscope-ai/agentscope/event';
+import { EventType, ReplyFinishedReason } from '@agentscope-ai/agentscope/event';
 import type {
 	AgentEvent,
 	CustomEvent,
@@ -56,6 +56,15 @@ const hitlKey = (e: { worker_session_id: string; reply_id: string }) =>
 	`${e.worker_session_id}:${e.reply_id}`;
 
 /**
+ * Finalize an in-memory reply when the server cannot emit REPLY_END.
+ * This client-local stamp is not persisted and will not survive a refresh.
+ */
+const finishMessage = (msg: Msg, reason: ReplyFinishedReason, finishedAt: string) => {
+	if (!msg.finished_at) msg.finished_at = finishedAt;
+	msg.finished_reason = reason;
+};
+
+/**
  * Lifecycle phase of the reply currently owned by this session.
  *
  * - ``idle`` — no in-flight reply; the send button is enabled.
@@ -72,7 +81,7 @@ const hitlKey = (e: { worker_session_id: string; reply_id: string }) =>
  */
 export type ReplyPhase = 'idle' | 'streaming' | 'interrupting';
 
-/** Safety fallback: force phase back to idle if REPLY_END is not seen. */
+/** Safety fallback: terminate locally if REPLY_END is not seen. */
 const INTERRUPT_TIMEOUT_MS = 10_000;
 
 /**
@@ -160,6 +169,19 @@ export function useMessages(
 		});
 	}, []);
 
+	const finalizeCurrentReply = useCallback(
+		(reason: ReplyFinishedReason) => {
+			if (currentReplyRef.current) {
+				finishMessage(currentReplyRef.current, reason, new Date().toISOString());
+			}
+			currentReplyRef.current = null;
+			clearInterruptTimer();
+			setPhase('idle');
+			scheduleUpdate();
+		},
+		[clearInterruptTimer, scheduleUpdate],
+	);
+
 	/** Apply a single AgentEvent to the in-progress reply. */
 	const processEvent = useCallback(
 		(event: AgentEvent) => {
@@ -199,9 +221,9 @@ export function useMessages(
 				if (currentReplyRef.current) {
 					appendEvent(currentReplyRef.current, event);
 				}
+				currentReplyRef.current = null;
 				clearInterruptTimer();
 				setPhase('idle');
-				currentReplyRef.current = null;
 			} else if (currentReplyRef.current) {
 				appendEvent(currentReplyRef.current, event);
 			}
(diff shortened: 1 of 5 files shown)


DavdGao:
Thanks for this, and apologies for the slow turnaround — you were first on #2110 and the diagnosis here is exactly right.

Some context: #2133 has since merged and overlaps with part of this PR. It closes the reply from the service layer with a terminating `ReplyEndEvent(finished_reason=ReplyFinishedReason.ERROR, error=ErrorInfo)`, and adds first-class `Msg.finished_reason` / `Msg.error` fields plus an `ErrorType` taxonomy for localization.

Two things here are still uncovered and worth landing:

1. **Agent-layer emission.** `Agent._reply_impl` still has no `except Exception` path, so anyone consuming `agent.reply_stream()` directly — outside the app service — gets an exception with no terminal event and hangs. Your `reply_active` guard closes that at the source, which is a better location than the service-layer fix.
2. **AG-UI mapping.** `_to_agui_event` still returns `AGUIRunFinishedEvent` unconditionally; mapping an errored reply to `AGUIRunErrorEvent` is still missing.

The frontend `finalizeCurrentReply` on stream close is also still useful as a client-side backstop for the case where the server dies before it can emit anything.

Could you rebase onto main and adapt to the merged vocabulary?

- `ReplyEndReason` is now deprecated — use `ReplyFinishedReason.ERROR` from `agentscope.types`
- drop the `metadata["finished_reason"]` plumbing in `message/_base.py`; `append_event` already stamps the first-class `finished_reason` / `error` fields
- on the frontend, `finishMessage` should stamp `msg.finished_reason` rather than `msg.metadata.finished_reason` — the merged `MessageBubble` reads the first-class field, so a locally recovered reply currently would not render as failed. Worth a comment noting the stamp is client-local and will not survive a refresh (the server never persisted a terminal event in that path)
- `ReplyEndEventWithReason` can go — `finished_reason` is properly typed on `ReplyEndEvent` as of SDK 0.0.14
- optionally populate `ErrorInfo` on the event so the UI can localize per error type

The `MessageBubble` failed-state rendering will likely conflict with what merged, so that part can probably be dropped or reconciled against the existing error `Alert`.

Jupiter363:
Rebased onto current `main` and addressed the requested vocabulary/model changes:

- switched the agent and AG-UI paths to `agentscope.types.ReplyFinishedReason.ERROR`
- removed the deprecated `ReplyEndReason` addition and all `metadata[finished_reason]` plumbing
- dropped the conflicting `MessageBubble`/locale changes in favor of the existing first-class error Alert
- updated the WebUI recovery path to stamp `Msg.finished_reason` locally (with a note that the fallback is not persisted)
- kept regression coverage for direct `reply_stream()` consumers and AG-UI `RUN_ERROR`

All refreshed GitHub checks are passing on commit `b50456dc`.

Jupiter363:
@DavdGao Thanks for the detailed guidance. I have rebased the branch onto the latest `main` and updated the implementation according to your suggestions, including adopting the first-class `ReplyFinishedReason`/message fields and removing the deprecated metadata plumbing and conflicting UI changes. All refreshed CI checks are now passing. When you have a chance, I would appreciate another look.

PR #2116, opened 2026-07-17
fix(rag): normalize Milvus Lite cosine scores

alex-mm:
## Summary
- normalize Milvus Lite COSINE scores by computing cosine similarity from returned vectors
- keep scores stable across Python/Milvus Lite wheel combinations that expose raw COSINE values differently
- add a top_k=1 score assertion for Milvus Lite search

## Tests
- uvx pre-commit run --all-files
- /tmp/agentscope-ci311.Wz3hRC/bin/python -m pytest tests/rag_vdb_milvus_lite_test.py -q
- .venv/bin/python -m pytest tests/rag_vdb_milvus_lite_test.py -q

### DIFF
--- a/src/agentscope/rag/_vdb/_milvus_lite.py
+++ b/src/agentscope/rag/_vdb/_milvus_lite.py
@@ -7,6 +7,7 @@ is convenient for local development, tests, and small RAG workloads.
 """
 import asyncio
 import json
+import math
 import os
 import uuid
 from typing import TYPE_CHECKING, Any, Literal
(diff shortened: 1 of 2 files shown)


qbc2016:
Thanks for looking into this! I noticed that #2089 already pinned `milvus-lite>=3.1.0` and removed the score conversion, since 3.1.0+ returns cosine similarity directly. If certain platform wheels still behave differently, would it be simpler to tighten the version pin rather than adding a runtime workaround? What do you think?

PR #2120, opened 2026-07-17
fix(agent): continue after thinking-only responses

### LINKED ISSUE #1596: [Bug]: ReActAgent exits loop prematurely when LLM returns only thinking blocks (no text, no tool_use)
**<u>AgentScope is an open-source project. To involve a broader community, we recommend asking your questions in English.</u>**

**Describe the bug**
In _react_agent.py, the main ReAct loop's exit condition (around line 514) only checks for the absence of tool_use blocks:

```python
elif not msg_reasoning.has_content_blocks("tool_use"):
    # Exit the loop when no structured output is required (or
    # already satisfied) and only text response is generated
    msg_reasoning.metadata = structured_output
    reply_msg = msg_reasoning
    break
```
This assumes that if there's no tool_use, the response must contain text. However, when using models with extended thinking (e.g., Claude with thinking enabled, Qwen with thinking enabled), the LLM may return a response containing only thinking blocks — no text and no tool_use.

In this case, not has_content_blocks("tool_use") evaluates to True, causing the loop to break and return msg_reasoning as the final reply — which contains no user-visible text content, effectively producing an empty response.

**To Reproduce**

1、Use a model with extended thinking enabled (e.g., Claude 3.5/4 with thinking capability, or Qwen with thinking mode)
2、Send a complex query that triggers extensive reasoning
3、In some iterations, the model returns only a thinking block without text or tool_use
4、The agent incorrectly breaks out of the loop, returning an empty/thinking-only reply

**Expected behavior**

When the LLM response contains only thinking blocks (no text, no tool_use), the agent should continue the loop instead of breaking out, allowing the LLM another iteration to produce actual text or tool calls.
```python
elif not msg_reasoning.has_content_blocks("tool_use"):
    # Only exit if there's actual text content for the user
    if msg_reasoning.has_content_blocks("text"):
        msg_reasoning.metadata = structured_output
        reply_msg = msg_reasoning
        break
    # else: only thinking blocks present, continue the loop
```

**Error messages**
none

agentscope version: 1.0.19
Model: Any model with thinking/extended_thinking support (qwen3.6-flash, etc.)
Python: 3.12

**Additional context**
none

March-77:
## Summary

- keep the ReAct loop running when a model response contains only thinking blocks
- preserve the existing final-response behavior for text, data, and empty responses
- add a regression test covering a thinking-only turn followed by a visible answer

## Root cause

`Agent._reasoning_impl()` treated every completed response without a tool call as a final answer. For extended-thinking models, a response can contain only `ThinkingBlock` content, so the agent returned an empty user-visible reply instead of continuing to reason.

## Impact

Extended-thinking models can now proceed to the next ReAct iteration after an intermediate thinking-only response and produce a visible answer or tool call.

## Validation

- `python -m pytest tests/agent_basic_test.py tests/agent_interrupt_test.py tests/event_test.py tests/model_response_test.py tests/compress_context_test.py -q` (41 passed)
- `python -m pre_commit run --files src/agentscope/agent/_agent.py tests/agent_basic_test.py`

Fixes #1596

### DIFF
--- a/tests/agent_basic_test.py
+++ b/tests/agent_basic_test.py
@@ -17,7 +17,12 @@ from agentscope.permission import (
     PermissionBehavior,
     PermissionContext,
 )
-from agentscope.message import TextBlock, ToolCallBlock, UserMsg
+from agentscope.message import (
+    TextBlock,
+    ThinkingBlock,
+    ToolCallBlock,
+    UserMsg,
+)
 
 
 class MockSequentialTool(ToolBase):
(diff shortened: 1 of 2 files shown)


tests/agent_basic_test.py
```diff
@@ -563,6 +568,30 @@ async def test_non_streaming_reasoning(self) -> None:
...
+                ChatResponse(
+                    content=[TextBlock(text="Final answer")],
+                    is_last=True,
+                ),
+            ],
+        )
+
+        msg = await self.agent.reply(UserMsg(name="user", content="Think"))
+
+        self.assertEqual(self.model.cnt, 2)
```
DavdGao, line 589: Plz assert outputs with complete structures (e.g. assertDictEqual, assertListEqual). Use utils.AnyString for random/dynamic strings. Otherwise LGTM.

DavdGao (approved):

PR #2124, opened 2026-07-18
fix(app): prevent session SSE event loss during replay

### LINKED ISSUE #2123: [Bug]: Session SSE silently loses team HITL events between replay and live subscribe
### Prerequisites

- [x] I searched existing issues, discussions, and open/closed PRs. #1890 / #1918 introduced leader-session HITL projection, while #2110 concerns missing terminal events after an LLM failure; neither covers this replay/live handoff race.
- [x] This is a correctness bug, not a usage question.

### Background / Description

`GET /sessions/{session_id}/stream` reads the replay log and performs team HITL projection **before** subscribing to the live channel. An event published in that window is silently absent from the already-open SSE connection.

This affects the leader-session HITL projection introduced by #1890 / #1918; it does not dispute that feature's intended behavior.

This reproduces through the production `stream_session_events()` route function using `RedisMessageBus` over a real Redis TCP connection (`redis-py -> TCP -> Redis 6.0.16`). Redis remains healthy and accepts the event.

Expected behavior: once the endpoint starts serving a healthy session stream, it must not silently omit events published while replay or team projection is awaiting storage; replay/live overlap should not produce duplicates.

Actual behavior: the event exists in the Redis Stream and is published successfully, but the SSE client never receives it.

### Root Cause

At base commit [`30ca3ef`](https://github.com/agentscope-ai/agentscope/commit/30ca3ef33189ce11cbd04a3475e009d2b879f021), the route performs:

1. [`log_read()` replay](https://github.com/agentscope-ai/agentscope/blob/30ca3ef33189ce11cbd04a3475e009d2b879f021/src/agentscope/app/_router/_session.py#L760-L766)
2. [`projection.list()` and per-worker storage awaits](https://github.com/agentscope-ai/agentscope/blob/30ca3ef33189ce11cbd04a3475e009d2b879f021/src/agentscope/app/_router/_session.py#L776-L806)
3. [live `subscribe()`](https://github.com/agentscope-ai/agentscope/blob/30ca3ef33189ce11cbd04a3475e009d2b879f021/src/agentscope/app/_router/_session.py#L824-L827)

The HITL projector independently performs [`upsert()` followed by `publish()`](https://github.com/agentscope-ai/agentscope/blob/30ca3ef33189ce11cbd04a3475e009d2b879f021/src/agentscope/app/_service/_projectors/_subagent_hitl.py#L170-L176). Therefore normal multi-worker concurrency is sufficient:

1. SSE captures the projection list and suspends on worker A's storage query.
2. Worker B upserts a new HITL projection and publishes its event.
3. Worker B is absent from the captured projection list, and Redis reports zero subscribers at publish time.
4. The route subscribes later and waits indefinitely; Pub/Sub does not replay the missed event.

This is a replay-then-subscribe TOCTOU. Its window grows with projection/storage latency and concurrent worker activity.

### Impact

For `subagent_require_user_confirm`, the leader UI is never shown the approval card, while the worker remains parked waiting for input. The SSE connection stays open and continues heartbeats, so reconnect-based recovery is not triggered.

The base cleanup also [deletes the entire replay Stream after persistence](https://github.com/agentscope-ai/agentscope/blob/30ca3ef33189ce11cbd04a3475e009d2b879f021/src/agentscope/app/_service/_chat.py#L653-L658), making a later reconnect unable to recover the event. This is a correctness failure rather than a rendering delay.

### Error Messages

There is normally no application or Redis error; the failure is silent. A deterministic probe adds a 1.5-second observation timeout and reports:

```text
replay= data: {... "name": "warmup" ...}
pubsub_numsub= [('agentscope:session:events:leader', 0)]
entry_id= 1784353855842-0 stream_len= 2
RESULT=LOST TimeoutError after 1.5s
```

The assigned entry ID and `stream_len=2` show that Redis accepted and retained worker B's event. The timeout occurs because the still-open SSE connection never delivers it.

Invoking the exact base cleanup operation against the same real Redis backend produces:

```text
BASE_LOG_TRIM entry_id=1784353757737-0 stream_len_before_trim=1
BASE_LOG_TRIM stream_len_after_trim=0 key_exists=0
```

### Steps to Reproduce

1. Start a real Redis server and verify TCP connectivity:

```bash
sudo apt-get install -y redis-server redis-tools
sudo systemctl start redis-server
redis-cli -h 127.0.0.1 -p 6379 ping
# PONG
```

2. Create a clean worktree at the current base commit and use the repository's development environment:

```bash
git fetch upstream main
git worktree add ../agentscope-sse-repro 30ca3ef33189ce11cbd04a3475e009d2b879f021
cd ../agentscope-sse-repro
python -m pip install -e '.[service,storage]'
```

3. Save the following as `/tmp/repro_sse_handoff.py`:

```python
import asyncio

from agentscope.app._router._session import stream_session_events
from agentscope.app._service import SessionProjection, SubagentHitlProjector
from agentscope.app.message_bus import MessageBusKeys, RedisMessageBus


class Storage:
    def __init__(self):
        self.worker_lookup_started = asyncio.Event()
        self.release_worker_lookup = asyncio.Event()

    async def get_session(self, _user_id, _agent_id, session_id):
        if session_id == "leader":
            return object()
        self.worker_lookup_started.set()
        await self.release_worker_lookup.wait()
        return None


def hitl(worker):
    return {
        "worker_session_id": worker,
        "worker_agent_id": f"{worker}-agent",
        "worker_agent_name": worker,
        "reply_id": f"{worker}-reply",
        "event_type": "require_user_confirm",
        "event": {},
        "created_at": "2026-07-18T00:00:00",
    }


async def main():
    bus = RedisMessageBus(host="127.0.0.1", port=6379, db=15)
    await bus.__aenter__()
    redis = bus.get_client()
    await redis.flushdb()
    projection = SessionProjection(bus)
    storage = Storage()
    key = MessageBusKeys.session_events("leader")

    worker_a = hitl("worker-a")
    await projection.upsert(
        "leader",
        SubagentHitlProjector.KIND,
        SubagentHitlProjector.entry_id("worker-a", "worker-a-reply"),
        worker_a,
    )
    await projection.publish("leader", "warmup", {"worker": "warmup"})

    response = await stream_session_events(
        "leader",
        agent_id="leader-agent",
        user_id="user",
        storage=storage,
        message_bus=bus,
    )
    stream = response.body_iterator
    try:
        print("replay=", (await asyncio.wait_for(anext(stream), 2)).strip())
        next_event = asyncio.create_task(anext(stream))
        await asyncio.wait_for(storage.worker_lookup_started.wait(), 2)
        print("pubsub_numsub=", await redis.pubsub_numsub(key))

        worker_b = hitl("worker-b")
        await projection.upsert(
            "leader",
            SubagentHitlProjector.KIND,
            SubagentHitlProjector.entry_id("worker-b", "worker-b-reply"),
            worker_b,
        )
        await projection.publish(
            "leader",
            SubagentHitlProjector.EVT_REQUIRE,
            worker_b,
        )
        entries = await bus.log_read(key, max_count=1000)
        print("entry_id=", entries[-1][0], "stream_len=", await redis.xlen(key))
        await asyncio.sleep(0.25)
        storage.release_worker_lookup.set()

        try:
            print("live=", (await asyncio.wait_for(next_event, 1.5)).strip())
        except asyncio.TimeoutError:
            print("RESULT=LOST TimeoutError after 1.5s")
    finally:
        storage.release_worker_lookup.set()
        await stream.aclose()
        await redis.flushdb()
        await bus.aclose()


asyncio.run(main())
```

4. Run it against the base source tree:

```bash
PYTHONPATH=src python /tmp/repro_sse_handoff.py
```

The output matches the failure shown above. The 250 ms controlled await is only a deterministic scheduling mechanism; any non-zero storage await creates the same ordering window.

For comparison, a local subscribe-first proof-of-fix produces:

```text
pubsub_numsub= [('agentscope:session:events:leader', 1)]
entry_id= 1784353864916-0 stream_len= 2
live= data: {... "name": "subagent_require_user_confirm", "value": {"worker_session_id": "worker-b", ...}}
```

The same race also reproduces with fakeredis, confirming that it is ordering logic rather than a Redis server failure.

### Proposed Scoped Fix

- Start the live subscription before replay and wait for Redis's subscribe acknowledgement through the existing `on_ready` barrier.
- Buffer live events during replay/projection and deduplicate the replay/live overlap by Redis Stream entry ID.
- Keep the scoped fix focused on the replay/live handoff; handle run-end trim and retention together in the separate cursor proposal.
- Add deterministic regression coverage for the projection window and cancellation before subscription readiness.

A longer-term follow-up can replace the Stream/Pub/Sub dual path with a single cursor-based Stream tail using SSE `id:` / `Last-Event-ID`, explicit retention-gap recovery, and bounded slow-consumer semantics. That larger protocol change is not required to close this correctness bug.

### Environment

- AgentScope Version: `2.0.4.post1` (`main` at `30ca3ef33189ce11cbd04a3475e009d2b879f021`)
- Python Version: `3.11.15`
- Redis: `6.0.16`, real TCP connection via `redis-py`; identical ordering result with fakeredis
- OS: Windows development environment with Redis on Ubuntu 22.04 / WSL2

yuminshen:
## AgentScope Version

`2.0.4.post1` (base: `agentscope-ai/agentscope@30ca3ef`)

Fixes #2123.

### Problem

The session SSE route replayed the buffered log and ran team HITL projection (one storage query per worker) before subscribing to live Pub/Sub. An event published in that window, such as worker B requesting confirmation while worker A's query is in flight, is stored in Redis but never reaches the open connection: it misses the completed replay and precedes the subscription. The existing run-end `log_trim()` then removes the replay log, so a later reconnect cannot recover that missed event. Full race analysis and real Redis reproduction are in #2123.

### Changes

- **Subscribe before replay.** The SSE generator starts the feeder and waits for the existing `on_ready` barrier before reading the log, so events published during replay or projection are already queued. On Redis, `on_ready` now fires only after the `SUBSCRIBE` acknowledgement is received.
- **Deduplicate the overlap.** Live payloads carry the Redis Stream entry ID; the route drops queued events whose ID was replayed and strips the internal `_entry_id` before sending. The SSE wire format is unchanged.
- **Clean up on early cancel.** Cancelling before readiness cancels and awaits the feeder, so no `sse-feeder:<session_id>` task leaks.

Scope: 3 files, 243 changed lines (132 production, 111 test). No new dependency or public API; `on_ready` already exists on the base commit for both Redis and in-memory backends.

### Behavior

Run-end replay-log cleanup remains unchanged. After `ReplyEnd`, a refresh loads the persisted message history and does not replay the completed run's event log. This patch changes only the replay/live handoff ordering, Redis readiness semantics, and internal overlap deduplication.

### Validation

```bash
python -m pytest tests/service_session_event_stream_test.py -q
# 2 passed - projection-window loss, overlap dedup, cancel-before-ready cleanup

python -m pytest tests/service_*_test.py -q
# 120 passed (13 service modules)
```

Changed-file pre-commit passes all configured hooks: AST, encoding, trailing commas, mypy, Black, Flake8, Pylint, and Pyroma.

GitHub Actions on `1773311` passes [pre-commit on Ubuntu](https://github.com/yuminshen/agentscope/actions/runs/29637974869) and the full [`pytest tests` matrix](https://github.com/yuminshen/agentscope/actions/runs/29637974852) on Ubuntu, macOS 15, and Windows with Python 3.11.

Real Redis A/B (`redis-py -> TCP -> Redis 6.0.16`; full output in #2123):

```text
base:      pubsub_numsub=0 at publish -> RESULT=LOST (TimeoutError, stream_len=2)
fix:       pubsub_numsub=1 before replay -> RESULT=DELIVERED
ack check: on_ready -> pubsub_numsub=1; immediate publish -> DELIVERED
run-end cleanup unchanged: stream_len 1->0, key_exists=0
```

WebUI contract smoke (`ReplyEnd` -> persist -> trim -> refresh):

```text
before refresh: reply_cards=1
after refresh:  reply_cards=1
console_errors=0
```

`git diff --check upstream/main` passes.

### Risks

- Duplicates on the replay/live seam are suppressed by Redis Stream entry ID.
- Reconnect after run-end cleanup still relies on persisted message history; cursor-based event resume remains a separate proposal.
- Slow consumers still use the per-connection queue; replacing the dual Stream/Pub/Sub path with a cursor-based tail is left to a separate proposal.

### Before merge

- [x] `pre-commit run --all-files` on Ubuntu for `1773311`
- [x] Full `pytest tests` on Ubuntu, macOS 15, and Windows for `1773311`
- [x] WebUI smoke: refresh after `ReplyEnd` retains exactly one persisted reply card without console errors

## Checklist

- [x] An issue has been created for this PR: #2123
- [x] Documentation not required (no public API or configuration change)

### DIFF
--- a/src/agentscope/app/_router/_session.py
+++ b/src/agentscope/app/_router/_session.py
@@ -756,88 +756,70 @@ async def stream_session_events(
         )
 
     async def _sse_generator() -> AsyncGenerator[str, None]:
-        # 1. Replay buffered events from the current run (if any).
-        for _entry_id, event in await message_bus.log_read(
-            MessageBusKeys.session_events(session_id),
-            max_count=MessageBusKeys.SESSION_REPLAY_MAX_LEN,
-        ):
-            yield f"data: {json.dumps(event, ensure_ascii=False)}\n\n"
-
-        # 1b. Inject pending subagent HITL cards projected onto this
-        #     session as a team leader (design §3.5). These live in a
-        #     durable Redis hash — NOT in the replay log (trimmed per
-        #     run) nor in the leader's own Msg history — so a fresh
-        #     reconnect after the worker parked still surfaces them.
-        #
-        #     Reconcile-on-read: the worker session's own context is the
-        #     SSOT. Inject only when the worker is still ASKING; drop and
-        #     delete ghosts (worker resolved/cancelled without clearing).
-        projection = SessionProjection(message_bus)
-        for payload in await projection.list(
-            session_id,
-            SubagentHitlProjector.KIND,
-        ):
-            if not await _worker_still_asking(
-                storage,
-                user_id,
-                payload["worker_agent_id"],
-                payload["worker_session_id"],
-                payload["reply_id"],
-            ):
-                await projection.delete(
-                    session_id,
-                    SubagentHitlProjector.KIND,
-                    SubagentHitlProjector.entry_id(
-                        payload["worker_session_id"],
-                        payload["reply_id"],
-                    ),
-                )
-                continue
-            custom = CustomEvent(
-                name=SubagentHitlProjector.EVT_REQUIRE,
-                value=payload,
-            )
-            data = json.dumps(
-                custom.model_dump(mode="json"),
-                ensure_ascii=False,
-            )
-
-            yield f"data: {data}\n\n"
-
-        # 2. Live subscribe via a background feeder task that pushes
-        #    events into a queue. The main loop reads from the queue
-        #    with a timeout so we can interleave heartbeat frames.
-        #
-        #    We avoid calling ``wait_for(__anext__())`` on the async
-        #    generator directly because cancelling a suspended
-        #    ``__anext__`` leaves the generator in a "running" state
-        #    that prevents ``aclose()`` from working.
+        key = MessageBusKeys.session_events(session_id)
         queue: asyncio.Queue[dict | None] = asyncio.Queue()
+        ready = asyncio.Event()
 
         async def _feeder() -> None:
-            """Read from the bus subscription and forward to the queue.
-
-            Pushes ``None`` as a sentinel when the subscription ends
-            (which in practice only happens if the bus shuts down).
-            """
             try:
-                async for evt in message_bus.subscribe(
-                    MessageBusKeys.session_events(session_id),
+                async for event in message_bus.subscribe(
+                    key,
+                    on_ready=ready.set,
                 ):
-                    await queue.put(
-                        {k: v for k, v in evt.items() if k != "_entry_id"},
-                    )
+                    await queue.put(event)
             except asyncio.CancelledError:
                 pass
             finally:
+                ready.set()
                 await queue.put(None)
 
         feeder_task = asyncio.create_task(
             _feeder(),
             name=f"sse-feeder:{session_id}",
         )
-
+        # Subscribe first; replayed ids suppress the overlap while events
+        # published during replay/projection stay queued for live delivery.
+        replayed_ids: set[str] = set()
         try:
+            await ready.wait()
+            for entry_id, event in await message_bus.log_read(
+                key,
+                max_count=MessageBusKeys.SESSION_REPLAY_MAX_LEN,
+            ):
+                replayed_ids.add(entry_id)
+                yield f"data: {json.dumps(event, ensure_ascii=False)}\n\n"
+
+            projection = SessionProjection(message_bus)
+            for payload in await projection.list(
+                session_id,
+                SubagentHitlProjector.KIND,
+            ):
+                if not await _worker_still_asking(
+                    storage,
+                    user_id,
+                    payload["worker_agent_id"],
+                    payload["worker_session_id"],
+                    payload["reply_id"],
+                ):
+                    await projection.delete(
+                        session_id,
+                        SubagentHitlProjector.KIND,
+                        SubagentHitlProjector.entry_id(
+                            payload["worker_session_id"],
+                            payload["reply_id"],
+                        ),
+                    )
+                    continue
+                custom = CustomEvent(
+                    name=SubagentHitlProjector.EVT_REQUIRE,
+                    value=payload,
+                )
+                data = json.dumps(
+                    custom.model_dump(mode="json"),
+                    ensure_ascii=False,
+                )
+                yield f"data: {data}\n\n"
+
             while True:
                 try:
                     item = await asyncio.wait_for(
@@ -846,7 +828,11 @@ async def stream_session_events(
                     )
                     if item is None:
                         break
-                    yield f"data: {json.dumps(item, ensure_ascii=False)}\n\n"
+                    entry_id = item.get("_entry_id")
+                    if entry_id in replayed_ids:
+                        continue
+                    event = {k: v for k, v in item.items() if k != "_entry_id"}
+                    yield f"data: {json.dumps(event, ensure_ascii=False)}\n\n"
                 except asyncio.TimeoutError:
                     yield ":\n\n"
         finally:
--- a/src/agentscope/app/message_bus/_redis_message_bus.py
+++ b/src/agentscope/app/message_bus/_redis_message_bus.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """The Redis-backed message bus implementation."""
+
 import asyncio
 import json
 import uuid
@@ -516,12 +517,10 @@ class RedisMessageBus(MessageBus):
         pubsub = self._client.pubsub()
         try:
             await pubsub.subscribe(key)
-            if on_ready is not None:
-                on_ready()
             while True:
                 try:
                     message = await pubsub.get_message(
-                        ignore_subscribe_messages=True,
+                        ignore_subscribe_messages=False,
                         timeout=self._SUBSCRIBE_POLL_TIMEOUT_SECS,
                     )
                 except redis_exceptions.TimeoutError:
@@ -532,6 +531,11 @@ class RedisMessageBus(MessageBus):
                     continue
                 if message is None:
                     # No payload within the poll window — keep waiting.
+                    continue
+                if message.get("type") == "subscribe":
+                    if on_ready is not None:
+                        on_ready()
+                        on_ready = None
                     continue
                 if message.get("type") != "message":
                     # Skip ``subscribe`` ack and similar control frames.
--- a/tests/service_session_event_stream_test.py
+++ b/tests/service_session_event_stream_test.py
@@ -1,0 +1,111 @@
+# -*- coding: utf-8 -*-
+"""Regression tests for the session SSE replay/live handoff."""
+# pylint: disable=missing-class-docstring,missing-function-docstring
+import asyncio
+import json
+from unittest import IsolatedAsyncioTestCase
+
+from agentscope.app._bus_ops import publish_session_event
+from agentscope.app._router._session import stream_session_events
+from agentscope.app._service import SessionProjection, SubagentHitlProjector
+from agentscope.app.message_bus import InMemoryMessageBus
+
+
+def _data(frame: str) -> dict:
+    line = next(x for x in frame.splitlines() if x.startswith("data: "))
+    return json.loads(line[6:])
+
+
+class _Storage:
+    def __init__(self) -> None:
+        self.lookup_started = asyncio.Event()
+        self.release_lookup = asyncio.Event()
+
+    async def get_session(self, *_args: str) -> object | None:
+        if _args[-1] == "leader":
+            return object()
+        self.lookup_started.set()
+        await self.release_lookup.wait()
+        return None
+
+
+class _OverlapBus(InMemoryMessageBus):
+    injected = False
+
+    async def log_read(self, key, since=None, max_count=100):  # type: ignore
+        if not self.injected:
+            self.injected = True
+            await publish_session_event(self, "leader", {"n": "overlap"})
+        return await super().log_read(key, since, max_count)
+
+
+class _NeverReadyBus(InMemoryMessageBus):
+    async def subscribe(self, key, *, on_ready=None):  # type: ignore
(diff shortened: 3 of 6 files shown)


src/agentscope/app/_router/_session.py
```diff
@@ -756,88 +756,70 @@ async def stream_session_events(
...
         feeder_task = asyncio.create_task(
             _feeder(),
             name=f"sse-feeder:{session_id}",
         )
-
+        # Subscribe first; replayed ids suppress the overlap while events
+        # published during replay/projection stay queued for live delivery.
+        replayed_ids: set[str] = set()
         try:
+            await ready.wait()
```
DavdGao, line 784: Blocking: unbounded, and nothing is written meanwhile — no replay, no heartbeat. Replay used to flush immediately; now a Redis server that accepts TCP but stalls before the ack leaves the client on a silent connection that proxies may cut. `_NeverReadyBus` builds exactly this state.

Please wrap in `asyncio.wait_for` and fall through to replay on timeout, so the worst case degrades to today's behaviour.

src/agentscope/app/_router/_session.py
```diff
@@ -756,88 +756,70 @@ async def stream_session_events(
...
+            for entry_id, event in await message_bus.log_read(
+                key,
+                max_count=MessageBusKeys.SESSION_REPLAY_MAX_LEN,
+            ):
+                replayed_ids.add(entry_id)
+                yield f"data: {json.dumps(event, ensure_ascii=False)}\n\n"
+
+            projection = SessionProjection(message_bus)
```
DavdGao, line 792: Blocking: please restore the `1. / 1b. / 2.` structure, the design §3.5 reference, the reconcile-on-read note, the `_feeder` docstring, and the `wait_for(__anext__())` rationale. None of it is recoverable from the code, and it's much of why the diff reads larger than the change.

src/agentscope/app/_router/_session.py
```diff
@@ -756,88 +756,70 @@ async def stream_session_events(
...
+            for payload in await projection.list(
+                session_id,
+                SubagentHitlProjector.KIND,
+            ):
+                if not await _worker_still_asking(
```
DavdGao, line 797: Not this PR: one storage round-trip per projected worker on the connect path is why the window was wide enough to hit #2123. Subscribing earlier avoids the race without narrowing it. Batching looks like a separate issue.

src/agentscope/app/_router/_session.py
```diff
@@ -846,7 +828,11 @@ async def _feeder() -> None:
                     )
                     if item is None:
                         break
-                    yield f"data: {json.dumps(item, ensure_ascii=False)}\n\n"
+                    entry_id = item.get("_entry_id")
```
DavdGao, line 831: Nit: `entry_id` and `event` are rebound after the replay loop used both in the same scope. `live_id` / `payload` would read clearer.

src/agentscope/app/message_bus/_redis_message_bus.py
```diff
@@ -533,6 +532,11 @@ async def subscribe(
                 if message is None:
                     # No payload within the poll window — keep waiting.
                     continue
+                if message.get("type") == "subscribe":
+                    if on_ready is not None:
+                        on_ready()
```
DavdGao, line 537: The docstring above still says `on_ready` fires "after ``SUBSCRIBE`` has been issued" — it now fires on the ack. Worth updating: `_cancel_dispatcher`, `_wakeup_dispatcher` and `_index_task_consumer` all block on this hook at startup.

src/agentscope/app/message_bus/_redis_message_bus.py
```diff
@@ -533,6 +532,11 @@ async def subscribe(
...
+                        on_ready = None
+                    continue
                 if message.get("type") != "message":
                     # Skip ``subscribe`` ack and similar control frames.
```
DavdGao, line 541: Stale — the ack is handled in the branch above now.

src/agentscope/app/message_bus/_redis_message_bus.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """The Redis-backed message bus implementation."""
+
```
DavdGao, line 3: Unrelated whitespace.

tests/service_session_event_stream_test.py
```diff
@@ -0,0 +1,111 @@
...
+            message_bus=bus,
+        )
+        iterator = response.body_iterator
+        pending = asyncio.create_task(anext(iterator))
+        await asyncio.sleep(0)
+        pending.cancel()
+        with self.assertRaises(asyncio.CancelledError):
+            await pending
+        await iterator.aclose()
+        tasks = asyncio.all_tasks()
```
DavdGao, line 107: Likely flaky: `finally` runs `await feeder_task` from an already-cancelling task, so it raises `CancelledError` immediately and gets swallowed — the feeder may not be reaped when `all_tasks()` is sampled. Assert on the task object instead.

tests/service_session_event_stream_test.py
```diff
@@ -0,0 +1,111 @@
...
+        return await super().log_read(key, since, max_count)
+
+
+class _NeverReadyBus(InMemoryMessageBus):
+    async def subscribe(self, key, *, on_ready=None):  # type: ignore
+        await asyncio.Event().wait()
+        yield {}  # pragma: no cover
+
+
+class TestSessionEventHandoff(IsolatedAsyncioTestCase):
```
DavdGao, line 48: These cover the route side, but the Redis ack change has no test. `service_message_bus_test.py` already uses `fakeredis`, so it is CI-testable: assert `PUBSUB NUMSUB == 1` when `on_ready` fires. Right now the subtlest line has the least coverage.

DavdGao (changes requested):
Thanks — the race is real and the fix is the right shape. I checked the dedup invariant: `log_append` precedes `publish`, so post-ack events are either in the log snapshot or strictly after it. Holds.

Two blockers inline: the unbounded `ready.wait()`, and the deleted rationale comments. Rest are nits.

Agree that cursor-based resume belongs in a separate proposal — happy to open an issue with the details.

PR #2132, opened 2026-07-20
feat(tool): add PowerShell support for Windows workspaces

### LINKED ISSUE #2130: feat(powershell): add powershell-based tool for windows os
### Discussed in https://github.com/agentscope-ai/agentscope/discussions/2125

<div type='discussions-op-text'>

<sup>Originally posted by **Liripo** July 19, 2026</sup>
有什么好的方式能自动识别 windowns环境, 工具执行时自动使用powershell的命令</div>

**Background**

AgentScope currently lacks native support for PowerShell as a tool-building environment. Now that AgentScope 2.0 has entered a stable development phase, it is a good time to start considering and implementing PowerShell-based tooling support, enabling broader accessibility for Windows-native developers and workflows.

**Open Questions**

Since all tools in AgentScope rely on the `BackendBase` class for backend switching, and currently only `LocalBackend` is implemented, we need to determine the best approach for supporting Windows/PowerShell. Specifically:

- Should we introduce a new `LocalWindowsBackend` as a dedicated Windows backend?
- Should we handle platform differences within the existing `LocalBackend` by detecting the current OS at runtime?
- Or should `LocalBackend` be refactored to support cross-platform execution through a unified interface?

**Changes**

- [TBD] Modify or introduce a new backend to support Windows, depending on the decision made in the Open Questions above
- Implement a PowerShell tool by subclassing `ToolBase`
- Update the `list_tools` interface in `LocalWorkspace` to return either PowerShell or Bash tools based on the current operating system

> **Note:** The above changes should only be planned and implemented **after** the open questions have been resolved.

qiuC123:
## Problem

`LocalWorkspace` always exposes the Bash tool, so Windows agents do not get a native PowerShell command interface. Adding a separate Windows backend would duplicate process and filesystem behavior that `LocalBackend` already implements portably.

## Solution

Add a standalone `PowerShell` tool backed by `LocalBackend`, and route only Windows local workspaces to it through a protected shell-tool factory. POSIX local workspaces and sandbox workspaces continue to use Bash.

## Changes

- add and publicly export `PowerShell(ToolBase)`
- invoke `powershell.exe` with `-NoLogo`, `-NoProfile`, and `-NonInteractive`
- configure UTF-8 output and handle cwd, timeout, errors, Unicode, and output truncation
- keep permissions conservative: commands are not auto-classified or suggested for allow rules
- add OS-based shell selection for `LocalWorkspace` while retaining Bash as the base workspace default
- add unit and Windows integration coverage for execution and workspace selection

## Testing

- `python -m pytest tests/builtin_powershell_test.py tests/workspace_local_test.py tests/backend_local_test.py tests/builtin_bash_test.py tests/workspace_docker_test.py -q`
  - 57 passed, 44 skipped
- pre-commit on all changed Python files
  - all hooks passed
- `python -m pytest tests -q`
  - 1158 passed, 240 skipped, 1 unrelated failure
  - the remaining failure is `tests/workspace_daytona_test.py::TestDaytonaWorkspaceBuiltinToolsMock::test_offload_context_tool_result_and_reset`; no Daytona source or tests are changed by this PR

## Notes for Reviewer

- Reuses `LocalBackend`; no `LocalWindowsBackend` is introduced.
- PowerShell command validation is intentionally out of scope and can be designed separately. Until then, the tool returns `PASSTHROUGH` and provides no automatic allow-rule suggestions.
- Sandbox workspaces keep their existing Bash behavior.

Closes #2130

### DIFF
--- a/src/agentscope/tool/_builtin/_powershell.py
+++ b/src/agentscope/tool/_builtin/_powershell.py
@@ -1,0 +1,212 @@
+# -*- coding: utf-8 -*-
+"""The PowerShell tool in agentscope."""
+
+import os
+from typing import AsyncGenerator, Any, List
+
+from ._backend import BackendBase, LocalBackend, _normalize_newlines
+from .._base import ToolBase, ToolMiddlewareBase
+from .._response import ToolChunk
+from ...message import TextBlock, ToolResultState
+from ...permission import (
+    PermissionBehavior,
+    PermissionContext,
+    PermissionDecision,
+    PermissionRule,
+)
+
+
+class PowerShell(ToolBase):
+    """Execute PowerShell commands through a workspace backend."""
+
+    name: str = "PowerShell"
+    """The tool name presented to the agent."""
+
+    description: str = (
+        "Executes a PowerShell command and returns its output. "
+        "Commands run without loading the user's PowerShell profile."
+    )
+    """The description presented to the agent."""
+
+    input_schema: dict[str, Any] = {
+        "type": "object",
+        "properties": {
+            "command": {
+                "type": "string",
+                "description": "The PowerShell command to execute.",
+            },
+            "description": {
+                "type": "string",
+                "description": (
+                    "Clear, concise description of what this command does."
+                ),
+            },
+            "timeout": {
+                "type": "integer",
+                "description": (
+                    "Optional timeout in milliseconds "
+                    "(default: 120000, max: 600000)"
+                ),
+                "default": 120000,
+                "maximum": 600000,
+                "minimum": 0,
+            },
+        },
+        "required": ["command"],
+    }
+
+    is_mcp: bool = False
+    is_read_only: bool = False
+    is_concurrency_safe: bool = False
+    is_external_tool: bool = False
+    is_state_injected: bool = False
+
+    def __init__(
+        self,
+        cwd: str | os.PathLike[str] | None = None,
+        middlewares: List[ToolMiddlewareBase] | None = None,
+        backend: BackendBase | None = None,
+    ) -> None:
+        """Initialize the PowerShell tool.
+
+        Args:
+            cwd (`str | os.PathLike[str] | None`, optional):
+                Working directory used when executing commands.
+            middlewares (`List[ToolMiddlewareBase] | None`, optional):
+                Tool middlewares wrapping command execution.
+            backend (`BackendBase | None`, optional):
+                Backend used for subprocess execution. Defaults to the
+                host-local backend.
+        """
+        super().__init__(middlewares=middlewares)
+        self._cwd = os.fspath(cwd) if cwd is not None else None
+        self._backend = backend or LocalBackend()
+
+    async def check_permissions(
+        self,
+        tool_input: dict[str, Any],
+        context: PermissionContext,
+    ) -> PermissionDecision:
+        """Defer every command to the configured permission mode.
+
+        PowerShell-specific command validation is intentionally outside this
+        implementation. No command is automatically classified as safe.
+        """
+        return PermissionDecision(
+            behavior=PermissionBehavior.PASSTHROUGH,
+            message="Execute PowerShell command",
+            decision_reason="PowerShell command validation is not enabled",
+        )
+
+    async def generate_suggestions(
+        self,
+        tool_input: dict[str, Any],
+    ) -> List[PermissionRule]:
+        """Return no automatic allow-rule suggestions.
+
+        A broad rule would weaken the conservative permission boundary before
+        PowerShell-specific command validation is available.
+        """
+        return []
+
+    async def call(  # type: ignore[override] # pylint: disable=unused-argument
+        self,
+        command: str,
+        description: str = "",
+        timeout: int = 120000,
+    ) -> AsyncGenerator[ToolChunk, None]:
+        """Execute a PowerShell command through the configured backend.
+
+        Args:
+            command (`str`):
+                PowerShell source text to execute.
+            description (`str`, optional):
+                Human-readable description of the command.
+            timeout (`int`, optional):
+                Timeout in milliseconds, capped at 600000.
+
+        Yields:
+            `ToolChunk`:
+                A final chunk containing the command output.
+        """
+        timeout_ms = min(timeout, 600000)
+        utf8_command = (
+            "$OutputEncoding = [Console]::OutputEncoding = "
+            "[System.Text.UTF8Encoding]::new($false); "
+            f"{command}"
+        )
+        try:
+            result = await self._backend.exec_shell(
+                [
+                    "powershell.exe",
+                    "-NoLogo",
+                    "-NoProfile",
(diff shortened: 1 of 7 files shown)


src/agentscope/tool/_builtin/_powershell.py
```diff
@@ -0,0 +1,212 @@
...
+        timeout_ms = min(timeout, 600000)
+        utf8_command = (
+            "$OutputEncoding = [Console]::OutputEncoding = "
+            "[System.Text.UTF8Encoding]::new($false); "
+            f"{command}"
+        )
+        try:
+            result = await self._backend.exec_shell(
+                [
+                    "powershell.exe",
```
DavdGao: Currently the implementation hardcodes `powershell.exe`. Since PowerShell 6+ ships as `pwsh` and is cross-platform, we should prefer it when available — falling back to `powershell.exe` if not found.

One approach is a lazy probe that runs once on first use, trying `pwsh` → `powershell.exe` in order and caching the result. Exit code `127` means the command was not found; anything else means it's usable. If neither is found, we fall back to `powershell.exe` and let the real error surface at execution time.

```python
_SHELL_CANDIDATES = ["pwsh", "powershell.exe"]

def __init__(self, ...):
    ...
    self._executable: str | None = None

async def _resolve_executable(self) -> str:
    """Probe pwsh (PowerShell 6+) before falling back to Windows PowerShell 5.1."""
    if self._executable is None:
        for candidate in _SHELL_CANDIDATES:
            probe = await self._backend.exec_shell(
                [candidate, "-NoLogo", "-NoProfile", "-NonInteractive",
                 "-Command", "exit 0"],
                timeout=10.0,
            )
            if probe.exit_code != 127:
                self._executable = candidate
                break
        else:
            self._executable = "powershell.exe"  # fallback; real errors surface later
    return self._executable
```

src/agentscope/tool/_builtin/_powershell.py
```diff
@@ -0,0 +1,212 @@
...
+            timeout (`int`, optional):
+                Timeout in milliseconds, capped at 600000.
+
+        Yields:
+            `ToolChunk`:
+                A final chunk containing the command output.
+        """
+        timeout_ms = min(timeout, 600000)
+        utf8_command = (
+            "$OutputEncoding = [Console]::OutputEncoding = "
```
DavdGao, line 196: - Quote mangling: exec_shell uses create_subprocess_exec, so on Windows the argv is re-joined into a command line under CRT rules — which do not match how powershell.exe parses its own arguments. Commands containing " (very common: Write-Output "a b") can be split incorrectly.
- Prefix pollution: a script starting with param(...) fails because a statement now precedes it; a single-line command starting with # swallows the rest; and reported error line numbers shift.

Recommend -EncodedCommand instead (Base64/UTF-16LE encode $OutputEncoding = ...; <command>). That single change eliminates both the escaping and the prefix-semantics classes of bug, and it's the standard approach on the PowerShell side.

src/agentscope/tool/_builtin/_powershell.py
```diff
@@ -0,0 +1,212 @@
...
+)
+
+
+class PowerShell(ToolBase):
+    """Execute PowerShell commands through a workspace backend."""
+
+    name: str = "PowerShell"
+    """The tool name presented to the agent."""
+
+    description: str = (
```
DavdGao: The PowerShell description is too minimal. Given its similarities with Bash, consider referencing the Bash description as a baseline and expanding the PowerShell description to a comparable level of detail.

src/agentscope/workspace/_base.py
```diff
@@ -315,29 +315,39 @@ async def get_instructions(self) -> str:
 
     # ── for Agent: tool & MCP discovery ────────────────────────────
 
+    def _create_shell_tool(self, backend: BackendBase) -> ToolBase:
```
DavdGao: Since this helper function is only used within the local workspace class, consider keeping the changes self-contained within `_local_workspace.py` rather than modifying it across multiple files. This would improve cohesion and reduce unnecessary cross-file dependencies.

tests/builtin_powershell_test.py
```diff
@@ -0,0 +1,288 @@
...
+            [],
+        )
+
+
+class PowerShellExecutionTest(IsolatedAsyncioTestCase):
+    """Test PowerShell command execution through the backend."""
+
+    async def test_call_uses_noninteractive_powershell_and_cwd(self) -> None:
+        """Wrap commands for PowerShell and pass cwd to the backend."""
+        backend = AsyncMock()
```
DavdGao, line 57: Since GitHub Actions supports PowerShell in its Windows CI environment, we should consider actually executing the PowerShell tool on the Windows runner instead of mocking it. (Skip in mac & linux)

DavdGao (commented):
@qiuC123 Plz see inline comments.

src/agentscope/tool/_builtin/_powershell.py
qiuC123: Implemented in `841518a`. `PowerShell` now lazily probes `pwsh` and then `powershell.exe`, treats exit code 127 as unavailable, caches the selected executable, and retains `powershell.exe` as the final fallback. I also added unit coverage for preference, fallback, and cache reuse.

src/agentscope/tool/_builtin/_powershell.py
qiuC123: Implemented in `841518a`. Commands now use `-EncodedCommand` with UTF-16LE/Base64. The user command is decoded into a separate `ScriptBlock`, keeping the UTF-8 setup outside so `param(...)`, leading comments, quotes, and user-relative error line numbers retain their semantics. Real Windows tests cover the preamble/quote and line-number cases.

src/agentscope/tool/_builtin/_powershell.py
qiuC123: Expanded in `841518a` using the Bash description as the baseline. The PowerShell description now covers dedicated filesystem tools, path and working-directory guidance, timeout limits, command descriptions, independent/dependent command execution, and git safety.

src/agentscope/workspace/_base.py
qiuC123: Adjusted in `841518a`. `WorkspaceBase.list_tools()` is restored to its original Bash behavior, and the Windows-only replacement now lives entirely in `LocalWorkspace.list_tools()`. The cumulative PR diff no longer changes `_base.py`.

tests/builtin_powershell_test.py
qiuC123: Extended in `841518a`. `PowerShellWindowsIntegrationTest` runs the real tool through `LocalBackend` on Windows and is skipped elsewhere. It now covers Unicode output, `param(...)`/leading comments/quoted arguments, and user-relative line numbers; the mock tests remain for deterministic argv, timeout, and error-path assertions.

DavdGao (approved):
LGTM

Issue #1334, opened 2026-03-17
[Bug]: `qwen3.5-35b-a3b` with a `structured_model` raise `openai.BadRequestError`


opposj:
**Describe the bug**

By calling an agent in the following common pattern with a `structured_model`:

```py
agent = ReActAgent(
    name="Vanilla",
    sys_prompt=prompt.metadata.get(
        "sys_prompt", "You're a helpful assistant named Vanilla."
    ),
    model=OpenAIChatModel(
        model_name=model,
        api_key=api_key,
        stream=stream,
        client_kwargs={
            "base_url": base_url,
            "timeout": timeout,
            "max_retries": max_retries,
        },
        generate_kwargs={
            "max_tokens": max_tokens,
            "temperature": temperature,
        },
    ),
    formatter=OpenAIChatFormatter(),
)

return await agent(prompt, structured_model=prompt.metadata.get("structured_model"))
```

When the `model` is set to `qwen3.5-35b-a3b`, a conflict about thinking mode and tool choice emerges:

```bash
openai.BadRequestError: Error code: 400 - {'error': {'message': '<400> ***.***.InvalidParameter: The tool_choice parameter does not support being set to required or object in thinking mode', 'type': 'invalid_request_error', 'param': '', 'code': 'invalid_parameter_error'}}
```

**To Reproduce**

Design any `structured_model` and pass it to a vanilla agent, whose model should be set to `qwen3.5-35b-a3b`.

**Expected behavior**

The `structured_model` parameter can be used with `qwen3.5-35b-a3b` without throwing exceptions.

**Error messages**

Refer to the above description.

**Environment (please complete the following information):**

- AgentScope Version: 1.0.16
- Python Version: 3.10
- OS: Ubuntu 22.04

**Additional context**

None

qbc2016:
This appears to be a compatibility issue between DashScope and OpenAI. For reasoning models, OpenAI doesn't support setting `tool_choice` to `required`. You can try using `DashScopeChatModel` and `DashScopeChatFormatter` instead.

opposj:
My provider does not support DashScope-style API calls. Is it possible to add support for this case?

qbc2016:
You can change the [code here](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/model/_openai_model.py#L260) to `kwargs["tool_choice"] = "auto"`.

opposj:
[...]

PR #2140, opened 2026-07-21
fix: disable thinking for structured output calls and add compression fallback

### LINKED ISSUE #2137: compress_context: structured summary relies on forced tool_choice; providers rejecting it (DeepSeek thinking) make compression fail probabilistically and kill the whole reply
### Summary

`Agent.compress_context()` generates its structured summary via `model.generate_structured_output()`, which relies on a **forced `tool_choice`**. Some providers reject forced tool_choice outright (DeepSeek `deepseek-reasoner` / v4-pro thinking mode returns `400: Thinking mode does not support this tool_choice`), in which case `_call_api_with_structured_output` silently downgrades to `tool_choice="auto"` (`model/_base.py`, the `UserWarning` path). From that point on, whether a summary is produced depends entirely on the model *voluntarily* calling the summary tool.

Under a ~100K-token context the model occasionally replies with plain text instead of the tool call. Then:

- `_base.py:562` raises `RuntimeError("Failed to generate structured output for model.")` (no retry at this layer — the built-in retry only covers API-level errors);
- `_compress_context_impl` (`agent/_agent.py:467`) has a fallback **only** for the `context_overflow` case; every other exception hits `raise e from None`;
- the exception propagates out of `reply_stream` and **kills the entire in-flight reply**. For a long-running report agent this throws away 5+ minutes of work over one probabilistic summary failure.

### Environment

- agentscope 2.0.2, `OpenAIChatModel` against `https://api.deepseek.com/v1`, model `deepseek-v4-pro` (thinking), streaming.

### Observed in production

A chat run died at compression time with the exact stack `_reply_impl → compress_context → _compress_context_impl → generate_structured_output → RuntimeError`. Context at trigger time was ~102K tokens (trigger_ratio 0.8 × 128K). Re-running the same agent succeeded — the failure is a low-probability tail event (we measured 23/23 successes replaying equivalent 100–112K-token summary calls; the `400 → auto` downgrade reproduces on **every** call with this provider).

### Why this is structural, not just bad luck

With providers that reject forced tool_choice, the summary call has no enforcement at all — `auto` makes non-compliance a matter of sampling. Any agent that runs long enough to trigger compression will eventually lose a whole reply to this.

### Suggestions (any one of these would fix the fatality)

1. **Fall back to a plain-text summary** when the structured call fails — a free-text summary prompt cannot "fail to comply", and the summary is consumed via `summary_template.format(...)` which could accept a single text field in degraded mode.
2. **Retry the structured call once** before giving up (it is stateless and cheap relative to losing the reply).
3. **Degrade instead of raise**: on final failure, skip this round of compression or truncate the oldest messages (the machinery already exists in the `context_overflow` branch) rather than `raise e from None`.
4. Minor related observability gap: `generate_structured_output` calls (compression and structured-output flows) bypass `TracingMiddleware`, so these failures produce no model span — instrumenting them would make such incidents diagnosable from traces.

We currently work around this in our platform with a subclass overriding `compress_context()` (retry once → hard-truncate oldest messages + inject a note telling the agent its early history was removed), but this belongs upstream.

qbc2016:
## AgentScope Version

2.0.4.post1

## Description

Fixes intermittent `compress_context` / `generate_structured_output` failures when thinking mode is enabled (fixes #2137).

Root cause: Several providers (Anthropic, DeepSeek, DashScope, Moonshot) reject forced `tool_choice` when thinking mode is active, causing `generate_structured_output` to fail with a RuntimeError.

Changes:

- Model layer (Anthropic, DeepSeek, DashScope, Moonshot): Override `_call_api_with_structured_output` to explicitly disable thinking via provider-specific kwargs before calling the base implementation. This ensures forced `tool_choice` is accepted.
- DashScope `_call_api`: Switch `extra_body` merging from `.update()` to per-key `setdefault()` so that caller-provided overrides are not silently overwritten by model defaults.
- Agent `_compress_context_impl`: Add a truncation fallback — if summary generation fails after all retries, discard older messages instead of raising, so the agent can continue operating with reduced context.

## Checklist


src/agentscope/agent/_agent.py
DavdGao, line 607: Suggested to use the same apply change here rather than create a new helper function

src/agentscope/model/_anthropic/_model.py
```diff
@@ -564,8 +529,8 @@ async def _call_api_with_structured_output(
                 The structured response whose ``content`` is the validated
                 output dict matching ``structured_model``.
         """
-        if tool_choice is None and self.parameters.thinking_enable:
-            tool_choice = ToolChoice(mode="auto")
+        if self.parameters.thinking_enable:
```
DavdGao: Suggested to disable the thinking unconditionally

src/agentscope/model/_dashscope/_model.py
```diff
@@ -246,17 +246,21 @@ async def _call_api(
         if fmt_tool_choice is not None:
             request_kwargs["tool_choice"] = fmt_tool_choice
 
-        extra_body: dict[str, Any] = {}
+        request_kwargs.setdefault("extra_body", {})
+        eb = request_kwargs["extra_body"]
+
         if self.parameters.thinking_enable is not None:
```
DavdGao, line 258: Simialr as above

src/agentscope/model/_deepseek/_model.py
```diff
@@ -467,8 +463,11 @@ async def _call_api_with_structured_output(
                 The structured response whose ``content`` is the validated
                 output dict matching ``structured_model``.
         """
-        if tool_choice is None and self.parameters.thinking_enable:
-            tool_choice = ToolChoice(mode="auto")
+        if self.parameters.thinking_enable:
```
DavdGao: Similar as above

src/agentscope/model/_moonshot/_model.py
```diff
@@ -441,8 +437,11 @@ async def _call_api_with_structured_output(
                 The structured response whose ``content`` is the validated
                 output dict matching ``structured_model``.
         """
-        if tool_choice is None and self.parameters.thinking_enable:
-            tool_choice = ToolChoice(mode="auto")
+        if self.parameters.thinking_enable:
```
DavdGao: Similar as above

src/agentscope/agent/_agent.py
```diff
@@ -578,15 +579,29 @@ async def _compress_context_impl(
...
+                except Exception as inner_e:
+                    logger.debug(
+                        "[AGENT %s]: Retry structured output " + "failed: %s",
+                        self.name,
+                        inner_e,
+                    )
 
-            else:
-                raise e from None
+            if res is None:
```
DavdGao, line 590: **Blocking:** removing the `else: raise e from None` branch means *any* exception now falls through to the truncation fallback, not just the `context_overflow` case it was designed for — auth failures, insufficient quota, moderation rejections, schema errors, transient network blips.

The consequence is not just a swallowed error: `msgs_to_compress` is dropped permanently and the fallback summary is either the *previous* summary or a single `"Some earlier messages were truncated"` line. Those messages get **no summary at all**. So one transient API hiccup silently and irreversibly destroys conversation history, with nothing but a `logger.warning` to show for it.

Suggestion: keep the truncation fallback scoped to `context_overflow` and re-raise otherwise, or at minimum gate the fallback behind an explicit config flag so callers opt into losing context.

src/agentscope/model/_base.py
```diff
@@ -451,49 +474,97 @@ async def generate_structured_output(
...
                     )
-                    await asyncio.sleep(self.retry_delay)
-                else:
-                    logger.warning(
-                        "All %d attempt(s) failed for model %s.",
-                        self.max_retries + 1,
+                    # Promote the winning strategy so later calls try it
+                    # first.
+                    if strategy is not self._structured_output_strategies[0]:
+                        self._structured_output_strategies.remove(strategy)
```
DavdGao: **Blocking:** this promotion is permanent and has no way back.

If `forced`, `no_think` and `auto` all happen to fail once for an unrelated reason and `none` (no `tool_choice` at all) succeeds by luck, `none` is moved to the front and **every subsequent call on this model instance starts from the weakest strategy forever**. There is no TTL, no reset, and no path back to `forced`. Since `none` relies purely on the injected system-reminder to get the model to call the tool, its success rate is meaningfully lower — so a single fluke permanently raises the long-run failure rate.

Suggestion: cache only the one fact you actually learned ("forced `tool_choice` is incompatible with thinking on this endpoint") rather than promoting whichever strategy happened to win, or restrict promotion to the `forced` ↔ `no_think` pair.

src/agentscope/model/_base.py
```diff
@@ -451,49 +474,97 @@ async def generate_structured_output(
...
-            raise last_error
-        raise RuntimeError(
-            f"Failed to generate structured output after "
-            f"{self.max_retries + 1} retries.",
-        )
+                    break
+
+        # Every strategy failed structurally; leave the order untouched.
+        assert last_error is not None
+        raise last_error
```
DavdGao, line 593: **Blocking:** `last_error` is overwritten on every strategy, so what finally propagates is almost always the *last* strategy's error — typically `RuntimeError("Failed to generate structured output for model.")` from the `none` strategy. The actual root cause (the first strategy's `BadRequestError`, a 401, a context-length-exceeded) only ever reaches `logger.debug`. That makes production debugging much harder than before this PR, where the real error was raised directly.

Suggestion: keep the first error and chain it — `raise last_error from first_error` — or collect all of them into an `ExceptionGroup`.

Also, on line 566: `assert last_error is not None` is removed by `python -O`, which would leave `raise None` → `TypeError`. It's unreachable in practice, but `assert` shouldn't be load-bearing in library code.

src/agentscope/model/_base.py
```diff
@@ -451,49 +474,97 @@ async def generate_structured_output(
...
-                if attempt < self.max_retries:
-                    logger.warning(
-                        "Attempt %d failed for model %s: %s. "
-                        "Retrying in %.1fs...",
-                        attempt + 1,
+        for strategy in self._structured_output_strategies:
+            name, extra_kwargs, tool_choice = strategy
+            # Overlay disable-thinking kwargs; a nested `extra_body` is
+            # merged rather than overwritten so caller keys survive.
+            merged = {**kwargs, **extra_kwargs}
```
DavdGao, line 528: **Blocking:** `tool_choice` is now passed explicitly *and* splatted from `**merged`. `generate_structured_output(**kwargs)` is public API and callers could previously pass `tool_choice=...` (the old code just forwarded `**kwargs` and it bound fine). Now that raises `TypeError: _call_api_with_structured_output() got multiple values for keyword argument 'tool_choice'`.

Worse, that `TypeError` is caught below as a "structural failure", so it burns through all four strategies and then surfaces as a completely unrelated error.

Suggestion: pop it up front and let it act as an explicit user override — `user_tool_choice = kwargs.pop("tool_choice", _UNSET)`.

src/agentscope/model/_base.py
```diff
@@ -451,49 +474,97 @@ async def generate_structured_output(
...
+                                "failed: %s. Retrying in %.1fs...",
+                                attempt + 1,
+                                self.model,
+                                e,
+                                self.retry_delay,
+                            )
+                            await asyncio.sleep(self.retry_delay)
+                            continue
+                        raise  # retries exhausted -> give up
+                    # Structural failure: try the next strategy.
```
DavdGao: Treating every non-retryable error as "structural, try the next strategy" is too broad. The old code failed fast on non-retryable errors; now a 401, an insufficient-quota error, a moderation rejection or a context-length-exceeded all walk the full `forced` → `no_think` → `auto` → `none` ladder, each with its own retry loop. That wastes tokens and wall-clock and amplifies rate limiting.

Combined with the `_agent.py` change, a failing compression can now issue up to `4 × (max_retries + 1)` calls before falling back.

Suggestion: short-circuit on errors that clearly have nothing to do with `tool_choice` (auth, quota, context length), e.g. via a provider-overridable predicate alongside `_get_retryable_exceptions()`.

src/agentscope/model/_base.py
```diff
@@ -451,49 +474,97 @@ async def generate_structured_output(
...
                 The structured response generated by the model.
         """
-
         if len(messages) == 0:
             raise ValueError(
                 "The input messages cannot be empty for the "
                 "`generate_structured_output` method.",
             )
 
+        if self._structured_output_strategies is None:
```
DavdGao: Two things here:

1. **Ordering.** `no_think` sits ahead of `auto`, which means the default recovery path silently disables thinking that the user explicitly configured, with only a `logger.info` to signal it. The previous approach (`tool_choice="auto"` + the injected system-reminder) preserved thinking. Output quality may drop in a way that's hard to attribute. At minimum this should be documented; ideally the order is configurable.

2. **Concurrency.** This lazy init, plus the `remove`/`insert` below, runs in `async` code. Multiple coroutines sharing one model instance can initialize concurrently and mutate the list while another is iterating it — the iteration won't crash, but strategies can be skipped or repeated. An `asyncio.Lock`, or an immutable strategy tuple plus a separate start-index, would be safer.

src/agentscope/model/_dashscope/_model.py
```diff
@@ -244,17 +244,21 @@ async def _call_api(
         if fmt_tool_choice is not None:
             request_kwargs["tool_choice"] = fmt_tool_choice
 
-        extra_body: dict[str, Any] = {}
+        request_kwargs.setdefault("extra_body", {})
+        eb = request_kwargs["extra_body"]
```
DavdGao: `request_kwargs.update(kwargs)` above is a shallow copy, so `eb` is the *caller's* `extra_body` dict (or `self.generate_kwargs["extra_body"]`). These `setdefault` calls therefore write the model defaults back into the caller's object permanently.

This predates the PR, but the new multi-strategy loop makes it fire on every call: the `forced` attempt writes `enable_thinking: True` into the caller's dict before `no_think` runs. The `merged[key] = {**kwargs[key], **val}` overlay does keep `no_think` winning, so it isn't broken today — but it's a mutation of caller-owned state that will bite eventually.

Suggestion: `eb = {**request_kwargs.get("extra_body", {})}` and assign it back after filling in the defaults.

src/agentscope/agent/_agent.py
```diff
@@ -595,22 +610,34 @@ async def _compress_context_impl(
...
-                    f"to '{path}', you can refer to it when needed."
-                    f"</system-reminder>"
+                offload_reminder = (
+                    f"\n<system-reminder>The compressed context"
+                    f" is offloaded to '{path}', you can refer"
+                    f" to it when needed.</system-reminder>"
                 )
+                if isinstance(new_summary, str):
+                    new_summary += offload_reminder
+                elif isinstance(new_summary, list):
```
DavdGao: This `list` branch is dead code: `cfg.summary_template.format(...)` returns a `str`, and `self.state.summary` is a `str` everywhere else in this file (see the `UserMsg("user", self.state.summary)` call sites). `new_summary` can never be a `list` here.

Separately, in the fallback path `new_summary` starts as `self.state.summary`, which *already contains* the offload reminder from the previous compression — so repeated failures keep appending reminders, and the summary accumulates a growing list of stale offload paths.

DavdGao (commented):
Thanks for this — the overall direction is right: collapsing the per-provider `tool_choice` special-cases into one strategy ladder in `ChatModelBase` is a real simplification, and CI is green.

That said, I think a few things need to be addressed before this can land. Details are in the inline comments; summarizing here by severity.

**Blocking**

1. `_compress_context_impl` now swallows *every* exception and silently drops context (the removed `else: raise e from None`).
2. The strategy promotion cache is an irreversible, permanent degradation of the model instance.
3. The exception that finally surfaces is the *last* strategy's error, so the actual root cause is lost to `logger.debug`.
4. `tool_choice` is now passed both explicitly and via `**merged`, so any caller that passes `tool_choice=` to `generate_structured_output` gets a `TypeError`.

**Should fix**

5. Non-`tool_choice` errors (401/403, insufficient quota, context-length-exceeded, moderation) no longer fail fast — they burn through all four strategies, each with its own retry loop.
6. `no_think` silently turns off thinking that the user explicitly enabled, and it is ordered *before* `auto` (the old behaviour, which preserved thinking). Worth documenting or making configurable.
7. `self._structured_output_strategies` is lazily initialized and mutated (`remove`/`insert`) from `async` code, so concurrent callers sharing one model instance can race.

**Minor**

8. `assert last_error is not None` is stripped under `python -O`, which would turn into `raise None` → `TypeError`. Please don't use `assert` for control flow in library code.
9. DashScope `_call_api` mutates the caller's `extra_body` dict in place.
10. The offload `<system-reminder>` accumulates across repeated fallbacks.
11. `isinstance(new_summary, list)` is dead code.
12. `"[AGENT %s]: Retry structured output " + "failed: %s"` — pointless string concatenation, and `logger.debug` seems too low for a retry failure.

**Two things not anchored to a line**

- `_openai_chat/_model.py` drops its `_call_api_with_structured_output` override but does **not** add a `_get_disable_thinking_kwargs`, so every OpenAI-compatible provider (and every third party subclassing `OpenAIChatModel`) skips the `no_think` strategy entirely. The `auto` strategy roughly covers the old `BadRequestError` → `tool_choice="auto"` path, so this is close to equivalent — but if it's intentional, please say so in the PR description.
- **No tests.** This PR rewrites a core retry/fallback path, and the repo already has `tests/model_base_test.py` and `tests/compress_context_test.py`. Please add coverage for at least: the strategy ladder order, the promotion behaviour, and the compression truncation fallback.

RerankerGuo:
@DavdGao I reviewed this together with #1334, #1608, #2137, #2189, and #2199.

They appear to expose the same architectural gap: structured-output behavior is currently inferred through provider-specific kwargs and exception-driven fallback. As more OpenAI-compatible providers diverge on thinking + forced `tool_choice`, this may keep producing overlapping fixes.

I am not suggesting expanding this PR. Would a focused follow-up design issue for a minimal model capability contract be useful?

For example, a model could declare only the capabilities needed by the structured-output strategy:

- supports forced tool choice while thinking
- supports disabling thinking per request
- supports native structured output
- fallback policy when those capabilities conflict

The structured-output path could then select a deterministic strategy before the request, while retaining exception fallback only for undeclared third-party providers.

I can first prepare a small compatibility matrix from the existing model implementations and affected issues, without changing public APIs. Would that be a useful direction, or do you prefer keeping this behavior internal and provider-specific?

DavdGao (approved):
lgtm

PR #2142, opened 2026-07-21
feat(a2a): add `A2AAgent` class in the agent module

### LINKED ISSUE #1941: feat(a2a): support A2A protocol in agentscope 2.0
## Background

AgentScope 2.0 plans to support the [A2A protocol](https://github.com/google/A2A) in two ways:

1. **Expose agents as A2A endpoints** — Register or deploy an AgentScope agent as an A2A-compatible agent that can be discovered and called by external systems.
2. **Consume A2A agents within agent service** — Wrap external A2A agents as callable components

## Open Questions

For **case 1**, the A2A protocol requires more than a basic chat interface — it also involves agent discovery, task lifecycle management, etc. It's unclear whether and how AgentScope should take responsibility for these, and what the right abstraction boundary looks like.

For **case 2**, the integration is relatively straightforward, but how A2A agent wrappers should interact with AgentScope's **Agent Service** — or whether they should be incorporated into it at all — still needs further discussion.

> **TBD** — Feedback and suggestions are welcome!

YingchaoX:
## AgentScope Version

AgentScope `2.0.4.post1`

```python
import agentscope

print(agentscope.__version__)
# 2.0.4.post1
```

### Background

This PR implements the client-side A2A integration discussed in [#1941](https://github.com/agentscope-ai/agentscope/issues/1941).

The goal is to allow an AgentScope application to interact with a remote [A2A 1.0](https://a2a-protocol.org/latest/) agent through the official `a2a-sdk` 1.x client while preserving the remote conversation and Task lifecycle across multiple turns.

A stateful `A2AAgent` was selected instead of a Tool-only wrapper because A2A interactions are naturally multi-turn. One `A2AAgent` instance owns the remote `contextId`, active Task information, observed messages, and client lifecycle.

`A2AAgent` provides Agent-like methods such as `reply()`, `reply_stream()`, `observe()`, and `compress_context()`, but intentionally does not inherit from the AgentScope 2.x `Agent` class. A local `Agent` owns a model, toolkit, state, middleware, and reasoning loop, while these responsibilities belong to the remote A2A server in this integration.

### Design references

The implementation follows the A2A 1.0 request and Task lifecycle, including:

- direct `Message` responses;
- Task-based responses;
- streaming status and artifact updates;
- multi-turn continuation through `contextId`;
- continuation of `INPUT_REQUIRED` Tasks with the original `taskId`;
- Task resubscription and inspection;
- explicit remote Task cancellation.

The existing AgentScope Java A2A integration was also reviewed as a design reference. Its stateful `A2aAgent` abstraction supports the decision to model a remote A2A peer as an agent rather than a stateless tool. Its lifecycle handling also informed the explicit treatment of terminal Task states, streaming completion, metadata, and resource cleanup.

The Python implementation does not copy the Java inheritance hierarchy because AgentScope Python 2.x no longer has `AgentBase`, and inheriting from the concrete Python `Agent` would incorrectly introduce local model, toolkit, state, and reasoning-loop responsibilities.

The larger durable HITL design proposed in [agentscope-java#2311](https://github.com/agentscope-ai/agentscope-java/pull/2311) was considered, but is not included here. That PR is still a draft, and durable handoffs, Redis coordination, resume tokens, and cross-process recovery require a separate design.

#### Public API

This PR adds the following public classes:

```python
from agentscope.agent import A2AAgent, A2ATaskStateError
```

`A2AAgent` provides:

- `reply()`
- `reply_stream()`
- `observe()`
- `compress_context()`
- `resume()`
- `resume_stream()`
- `get_task()`
- `cancel_task()`
- `aclose()`
- async context-manager support

`compress_context()` intentionally has the same call signature as `Agent.compress_context()` but is a no-op, because the remote A2A server owns and manages its conversation context.

#### A2A protocol and transport

- Uses the official `a2a-sdk>=1.1.1,<2.0.0`.
- Supports A2A protocol version 1.0 only.
- Adds an optional `agentscope[a2a]` extra and includes it in `full`.
- Keeps SDK runtime imports lazy, so importing `agentscope.agent` does not require the A2A extra.
- Filters mixed-version Agent Cards so the SDK cannot silently select an A2A 0.3 compatibility interface.
- Rejects Agent Cards containing only A2A 0.3 interfaces.
- The default client supports the official `JSONRPC` and `HTTP+JSON` protocol bindings with streaming enabled and polling disabled.
- Other transports remain possible by injecting a compatible official SDK `Client`.
- The agent owns and closes both default and injected clients.

#### Conversation and Task lifecycle

- Preserves the remote `contextId` between ordinary turns.
- Preserves the original `taskId` when continuing an `INPUT_REQUIRED` Task.
- Serializes replies per `A2AAgent` instance to protect conversation state.
- Validates that one response stream cannot switch its `taskId` or `contextId`.
- Supports intermediate `SUBMITTED` and `WORKING` states.
- Supports resubscription to an active Task, with a `GetTask` fallback when subscription is unavailable.
- Exposes `get_task()` and explicit `cancel_task()`.
- Does not translate local coroutine cancellation into remote Task cancellation automatically.
- Reports `INPUT_REQUIRED`, `AUTH_REQUIRED`, `FAILED`, `REJECTED`, and `CANCELED` through `A2ATaskStateError`, retaining the context ID, Task ID, Task state, and optional status message.
- Treats a response stream that ends without a completed result as an error.

#### Streaming and content mapping

- Maps A2A text Parts to AgentScope text blocks.
- Maps raw bytes and URL Parts to AgentScope data blocks without lossy string conversion.
- Rejects structured-data Parts that cannot currently be represented losslessly.
- Maps streaming text and raw-byte artifacts to AgentScope start, delta, and end events.
- Exposes Task status updates as `CustomEvent(name="a2a_status_update")`.
- Reduces append updates and final Task snapshots without emitting duplicated content.
- Rejects an artifact replacement after its content has already been emitted, because AgentScope events currently cannot represent retraction or replacement safely.
- Uses a completed Task's status message only as a fallback when no artifact content exists.
- Adds A2A context, Task state, Task ID, and artifact IDs to final message and event metadata.

#### Observed messages and cleanup

- Messages passed to `observe()` are prepended to the next outgoing A2A user message.
- Observed messages are cleared after a valid protocol response.
- They are retained when sending or protocol processing fails.
- `aclose()` is idempotent.
- A failed client close can be retried.
- New interactions are rejected after successful closure.

#### Example

A runnable example is included under `examples/a2a_agent/`.

It demonstrates:

- resolving a remote Agent Card;
- creating one `A2AAgent` per logical conversation;
- streaming AgentScope events;
- handling `INPUT_REQUIRED`;
- preserving the remote context across turns;
- closing the client through an async context manager.

#### 1. Official Python SDK local E2E

`tests/a2a_agent_e2e_test.py` starts an ephemeral A2A 1.0 server using the official Python SDK server primitives.

The test exercises the real network path rather than a mocked transport:

1. Agent Card discovery through the well-known endpoint;
2. official JSON-RPC client creation;
3. HTTP/SSE streaming;
4. Task status updates;
5. multi-chunk artifact assembly;
6. terminal `COMPLETED` handling;
7. two-turn remote `contextId` continuity;
8. client and server cleanup.

The server is deterministic and does not require an LLM, API key, paid service, or external network access.

```bash
uv run --python 3.13 --extra dev \
  pytest tests/a2a_agent_e2e_test.py -v
```

#### 2. Microsoft Agent Framework interoperability E2E

The client was also tested against a local Microsoft Agent Framework A2A server implemented with `A2AExecutor`.

The interoperability smoke test covered:

- Agent Card discovery;
- A2A 1.0 JSON-RPC transport;
- streaming response consumption;
- Task progress and terminal completion;
- artifact-to-AgentScope event conversion;
- multi-turn `contextId` reuse.

This test completed successfully using the official A2A client path. It is not included in CI because it requires installing and starting an independent third-party framework harness.

#### Additional LangGraph compatibility investigation

A separate E2E investigation was performed against LangGraph Agent Server.

Agent Card discovery succeeded, but the tested LangGraph server could not complete official A2A 1.0 interoperability:

- its Agent Card advertised the lowercase binding `jsonrpc` instead of the A2A 1.0 `JSONRPC` binding;
- after normalizing the binding for diagnosis, its SSE payloads still contained A2A 0.3-style `kind` and `final` fields;
- the official `a2a-sdk` 1.1.1 protobuf parser rejected those payloads.

Raw SSE verification showed that LangGraph preserved its own multi-turn context correctly, but its response wire format could not be consumed by the official A2A 1.x SDK. No LangGraph-specific compatibility workaround is added in this PR because the implementation intentionally targets the official A2A 1.0 protocol and SDK.

### Tests

The implementation includes tests for:

- importing `agentscope.agent` without the A2A extra;
- clean test skipping when the optional dependency is absent;
- protocol-version and transport validation;
- direct Message responses;
- completed Task responses;
- streamed append artifacts and final-snapshot deduplication;
- unsupported artifact replacement;
- text, raw-byte, and URL Part mapping;
- structured-data rejection;
- observed-message retention and cleanup;
- `contextId` and `taskId` continuity;
- response-stream identifier switching;
- `INPUT_REQUIRED` continuation;
- Task resubscription and fallback;
- Task inspection and cancellation;
- terminal and interrupted Task states;
- status-message fallback;
- metadata;
- client cleanup and close retry;
- `compress_context()` signature compatibility;
- real local JSON-RPC/SSE end-to-end behavior.

Validation performed:

```bash
uv run --python 3.13 --extra dev \
  pytest \
  tests/a2a_agent_import_test.py \
  tests/a2a_agent_test.py \
  tests/a2a_agent_e2e_test.py \
  tests/agent_basic_test.py -q
```

Result:

```text
35 passed, 3 subtests passed
```

Optional-dependency validation was also run in an isolated environment without the A2A extra:

```text
1 passed, 2 skipped
```

Relevant pre-commit checks passed:

- Python AST
- TOML validation
- docstring checks
- mypy
- black
- flake8
- pylint
- Pyroma
- `git diff --check`

### Out of scope

The following capabilities are intentionally not included in this PR:

- A2A server support;
- Agent Service integration;
- an `A2AAgentTool` wrapper;
- QwenPaw migration;
- push notifications;
- persistent client state;
- durable or cross-process HITL recovery;
- Redis coordination;
- automatic remote cancellation on local coroutine cancellation;
- LangGraph-specific A2A 0.3 compatibility;
- lossy conversion of structured data, thinking, tool-use, or hint blocks.

These can be considered separately after the client abstraction and public API have been reviewed.

## Checklist

- [x] An issue has been created for this PR: [#1941](https://github.com/agentscope-ai/agentscope/issues/1941)
- [ ] Related documentation has been updated in the [documentation repository](https://github.com/agentscope-ai/docs)
  - This PR includes a runnable example and README. A documentation-repository update can follow after the public API is reviewed.
- [ ] Code is ready for review
  - This is currently a Draft PR for API and scope feedback.


src/agentscope/agent/_a2a_agent.py
```diff
@@ -0,0 +1,704 @@
...
+            HTTP+JSON client is created from the filtered Agent Card.
+    """
+
+    def __init__(
+        self,
+        agent_card: AgentCard,
+        *,
+        client: Client | None = None,
+    ) -> None:
+        """Initialize the A2A agent adapter."""
```
DavdGao: Plz detail the docstring of the constructor

src/agentscope/agent/_a2a_agent.py
```diff
@@ -0,0 +1,704 @@
...
+        self,
+        context_config: ContextConfig | None = None,
+        instructions: HintBlock | None = None,
+    ) -> None:
+        """Do nothing because the remote A2A server owns its context.
+
+        The arguments are accepted for interface compatibility with
+        :class:`agentscope.agent.Agent`.
+        """
+        del context_config, instructions
```
DavdGao: In this case, plz consider add logger.warning for the compress_context method

DavdGao (commented):
- First, it would be good to introduce an agent protocol (e.g. AgentProtocol) for A2AAgent, so that the Agent and A2AAgent implementations stay aligned.
- Suggest aliasing the import as from `a2a.types import Task as A2ATask`, to distinguish it from AgentScope's built-in planning Task.
- One point I haven't fully worked out: once we bring `A2AAgent` into the agent service and let users register their own `A2AAgent` instances, how should the service drive it? For example, should the service layer call `reply_stream()` or `resume_stream()`? It would be better to have a single entry point that covers both cases — otherwise every caller has to implement the branching logic (inspecting task_state) outside the agent.
- If a `context_id` is effectively equivalent to a session ID: I don't see anywhere in the current implementation to set or persist it. That means the user can't specify which session to attach to, and once an `A2AAgent` instance is destroyed after a conversation, that conversation can't be continued — is that right? Is this a limitation of the A2A protocol itself (a session is maintained only by the live connection and cannot be recovered once it drops), or just a gap in the current implementation?
 - The current implementation surfaces HITL by raising `A2ATaskStateError`. I wonder if we could align this with the `Agent` class instead: `Agent` parks a reply by yielding a typed `Require*Event` and then returning without a `ReplyEndEvent`, and the caller resumes through the same `reply() / reply_stream()` entry by passing a different input type. An A2A task in INPUT_REQUIRED looks like the same situation, so it could follow the same contract rather than surfacing as an exception.

DavdGao (approved):

PR #2143, opened 2026-07-21
feat(agent): add agent-driven context compression tool

jinliyl:
## AgentScope Version

`2.0.6`

## Summary

This PR adds optional agent-driven context compression at meaningful task boundaries. Instead of treating `ReplyEndEvent` as task completion or making a separate rubric model call, it exposes the existing compression mechanism as an agent-owned `CompressContext` tool.

The feature is disabled by default. Existing hard-threshold compression remains unchanged.

## Behavior

When `ContextConfig.compression_tool_enabled` is enabled:

- `CompressContext` is registered as an agent-owned builtin tool;
- runtime context awareness continues to use `trigger_ratio - context_buffer_ratio`;
- once usage crosses that boundary and `task_status["in_progress"] == 0`, the runtime-state hint tells the agent that it can compress between tasks;
- the hint is persisted like other runtime-state hints and is not repeatedly appended while the previous recommendation remains in context;
- the agent decides whether to call the tool during its normal reasoning, so no additional decision-model call is introduced.

The tool is still available when runtime-state injection is disabled, but no automatic recommendation is injected.

## Compression flow

A `CompressContext` call reuses the existing compression middleware and implementation:

```python
await self._run_compress_context(
    force=True,
    fallback_to_truncation=False,
)
```

Explicit tool compression bypasses only the hard token threshold and does not fall back to lossy truncation if summary generation fails. Normal hard-threshold compression retains its existing truncation fallback.

`force` and `fallback_to_truncation` are included in `on_compress_context` middleware input, allowing middleware to observe and override the complete compression request.

Unfinished tool calls are always retained during context splitting. This keeps the `CompressContext` call paired with the result appended after compression and also protects other in-flight tool calls.

Image limiting remains an independent context invariant. If an image is offloaded before summary generation fails, its replacement remains in context with the live workspace reference, avoiding orphaned files.

## Configuration

```python
ContextConfig(
    compression_tool_enabled=True,
)
```

The previous reply-end `self_compact_*` configuration and `self_compact_context()` API have been removed. Compression terminology is used consistently throughout the implementation.

## Compatibility

- Disabled by default.
- Existing hard-threshold compression behavior is retained.
- Existing compression middleware, prompt, summary schema, offload, cache cleanup, and interruption handling are reused.
- No additional rubric model calls.
- No reply lifecycle hook or deferred terminal events.
- No new dependency.

## Tests

Coverage includes:

- configuration-controlled tool registration;
- real forced compression below the hard threshold;
- full agent reasoning → tool call → compression → tool result flow;
- preservation of unfinished tool call/result pairs;
- task-boundary runtime recommendations and reminder deduplication;
- no recommendation while a task is in progress;
- middleware visibility and override of all compression controls;
- failed summary behavior and retained image offload references;
- unchanged hard-compression truncation fallback.

Local verification:

```bash
PYTHONPATH=src pytest -q \
  tests/compress_context_test.py \
  tests/agent_injection_test.py \
  tests/agent_basic_test.py \
  tests/middleware_test.py
```

```text
71 passed, 2 subtests passed
```

Additional toolkit, structured-output, budget, and tracing tests:

```text
49 passed
```

All relevant pre-commit hooks pass, including mypy, black, flake8, pylint, and Pyroma.

## Related Work

- Supersedes the implementation approach in #1962.
- Incorporates the review direction to let the agent choose real task-stage boundaries.
- Reference: [Self-Compacting Language Model Agents](https://arxiv.org/abs/2606.23525).

## Checklist

- [ ] An issue has been created for this PR. This follows up on and supersedes #1962.
- [x] I have read the [CONTRIBUTING.md](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md).
- [ ] Related documentation has been updated in the documentation repository.


src/agentscope/agent/_agent.py
```diff
@@ -368,10 +374,26 @@ async def compress_context(
                 Optional hints or instructions injected into the compression
                 context to guide the summarization behavior.
         """
+        await self._run_compress_context(
```
DavdGao: Why do we need an wrapper function `_run_compress_context` here? Since this function is exactly the same with `compress_context`.

src/agentscope/agent/_agent.py
```diff
@@ -840,13 +1073,42 @@ async def next_handler(
             agen = execute_chain()
 
         self._receive_reply_end = False
+        deferred_reply_items: list[AgentEvent | Msg] = []
```
DavdGao: The self-compact logic sits outside the `on_reply` middleware chain — it wraps `agen`, rather than running inside it — so agent middlewares cannot observe, intercept or override it.

src/agentscope/agent/_agent.py
```diff
@@ -669,7 +833,63 @@ async def _apply_change() -> None:
...
+    ) -> bool:
+        """Ask the model whether the completed reply should be compacted."""
+        prompt = self._render_self_compact_rubric_prompt(
+            context_config,
+            estimated_tokens,
+        )
+        if not prompt:
+            return False
+
+        response = await self.model.generate_structured_output(
```
DavdGao: This rubric call goes through neither `on_reasoning` nor `on_model_call` (the latter is only wired around `model(...)` inside `_reasoning`), so `BudgetMiddleware` does not bill it and `TracingMiddleware` emits no span for it. It also fires on every completed reply above `self_compact_min_ratio`, carrying the full context and with no throttle, and `generate_structured_output`'s fallback ladder can re-send that context several times.

src/agentscope/agent/_agent.py
```diff
@@ -840,13 +1073,42 @@ async def next_handler(
...
+                # cleanup against the uncompressed turn context while still
+                # ensuring compaction completes before consumers observe the
+                # terminal event and potentially stop iterating.
+                deferred_reply_items.append(item)
+                continue
+
             yield item
 
+        if deferred_reply_items:
+            await self.self_compact_context()
```
DavdGao: An interrupted rubric call raises `CancelledError`, which the `except Exception` in `self_compact_context` does not catch. It propagates out of `_reply` and `deferred_reply_items` is discarded, so a reply that already completed loses both its `ReplyEndEvent` and its final `Msg`.

src/agentscope/agent/_agent.py
```diff
@@ -385,6 +407,8 @@ async def execute_chain(
                     await self._compress_context_impl(
                         context_config=context_config,
                         instructions=instructions,
+                        force=force,
```
DavdGao: `force` and `fallback_to_truncation` never reach `input_kwargs`, so an `on_compress_context` middleware cannot tell an adaptive compaction from a hard-threshold one, nor override either flag.

src/agentscope/agent/_agent.py
```diff
@@ -773,6 +993,19 @@ def _is_image(block: Any) -> bool:
         for container, idx, new_block in replacements:
             container[idx] = new_block
 
+        return [
```
DavdGao: Images are offloaded to the workspace before being replaced, but `_restore_limited_context_images` only reverts the block in the context, leaving the offloaded file orphaned.

tests/compress_context_test.py
```diff
@@ -1520,6 +1524,403 @@ async def test_context_compression_overflow_retry_keeps_instructions(
...
+                    return_value=should_compact,
+                )
+                agent._run_compress_context = AsyncMock()
+
+                await agent.self_compact_context()
+
+                if should_compact:
+                    call = agent._run_compress_context.await_args
+                    self.assertEqual(
+                        call.kwargs["context_config"],
```
DavdGao: Per-key `assertEqual` on `call.kwargs[...]` instead of a single assertion against the complete expected structure.

DavdGao (commented):
Self-compaction is hard-wired to fire on `ReplyEndEvent`, which effectively treats "reply ended" as "a task finished".

This assumption may not hold in the current agent service. A reply ends whenever the agent hands work off and waits:

- tools that keep executing in the background,
- or a task delegated to a subagent in a team,

and each of those handoffs emits its own `ReplyEndEvent`. So a single logical task is routinely fragmented across many short replies with interaction steps in between, and none of those boundaries is a real stage boundary.

That makes reply end a poor proxy on both counts: it costs a rubric call at every one of those frequent, mostly meaningless boundaries, and it still isn't accurate — the points where compaction is actually worthwhile are the ones it can't see.

What we actually want is for the agent to trigger compaction when it has closed out a phase of work. That judgement most likely has to come from the agent itself, or be derived from the planning-tasks state during the task phase, rather than from the reply lifecycle. The trigger point here needs more thought before this lands.

Separately, please reconsider whether `self_compact_context` needs to exist as a new public method at all. It sits right next to `compress_context`, the two names are nearly indistinguishable, and both end up in the same `_compress_context_impl` with the same `compression_prompt` and `SummarySchema`. The agent module has used "compress" consistently until now, while every field added here says "compact" — `self_compact_compression_instructions` even uses both words for a single operation. Readers will reasonably assume these are two different mechanisms.

DavdGao (commented):
One more thing worth considering: `_agent.py:1659-1668` already merges `self_compact_min_ratio` with `context_buffer_ratio` into a single awareness ratio, and that is the tell — `InjectionConfig`'s context awareness and the self-compact trigger are reading the same signal ("the context is filling up"), and only differ in what they do with it. The injection side already answers "when should the agent be told", and it answers it at a point the agent can act on, rather than at a fixed lifecycle event.

So an alternative: rather than hard-wiring compaction to `ReplyEndEvent`, expose it as a tool the agent can call, gated by a config switch. The context-length hint then does double duty — when the tool is enabled, the same hint that reports the current token count also tells the agent that the tool exists and that it can compact between tasks.

Crossing the buffer shouldn't be the only condition for that part of the hint, though. `_inject_runtime_state` already tracks `task_status`, so the suggestion to compact can additionally require `task_status["in_progress"] == 0` — i.e. only nudge the agent when it is not in the middle of a task. That is exactly the "between tasks" boundary this feature is really after, and it is available locally from state the agent already maintains, with no model call to determine it.

That reaches the same outcome with less machinery: the agent picks the boundary, so compaction lands on a real stage break instead of on whatever reply happened to end; there is no per-reply rubric call, since the decision rides on reasoning the agent is doing anyway; and it reuses the existing awareness mechanism instead of adding a second, parallel path plus a new hook in the reply lifecycle.

jinliyl:
Addressed all review feedback in `5a5d166`:

- Replaced the `ReplyEndEvent` hook and per-reply rubric call with the config-gated `CompressContext` agent tool.
- Reused runtime context awareness and recommend the tool only after crossing the context buffer with `task_status["in_progress"] == 0`.
- Removed the public `self_compact_context()` API and all `self_compact_*` fields; terminology now consistently uses compression.
- Removed deferred terminal-event handling, so cancellation can no longer discard an already-completed reply through that path.
- Routed explicit compression through the normal acting and `on_compress_context` middleware paths.
- Added `force` and `fallback_to_truncation` to middleware `input_kwargs`; tests assert the complete structure and verify both values can be overridden.
- Removed image restoration after failed summary generation. Offloaded image replacements retain their live workspace references, with a failure-path test covering orphan prevention.
- Added an integration test for reasoning → `CompressContext` → compression → paired tool result, including protection for unfinished tool calls during context splitting.

Local verification: 71 relevant tests plus 2 subtests pass; 49 additional structured-output, toolkit, budget, and tracing tests pass; all relevant pre-commit hooks pass.

jinliyl:
Follow-up in `67c63a0`:

- `CompressContext` now uses one stable agent-owned tool instance. It is registered before the first model input and remains registered across replies while `compression_tool_enabled` stays enabled, avoiding registry churn and keeping the serialized tool schema stable for prompt caching.
- The tool description now requires an explicit `<context-compression>` runtime reminder.
- It also tells the model not to call the tool while any task is currently `in_progress`, even if an older persistent reminder remains in context; pending tasks are allowed at a between-task boundary.
- Added coverage asserting that the exact same tool instance remains registered across replies.

Relevant tests and pre-commit hooks pass.

jinliyl:
Rebased this PR onto the latest `upstream/main` (`215289d`).

What changed during the rebase:
- Resolved conflicts in `src/agentscope/agent/_agent.py` and `tests/agent_injection_test.py`.
- Preserved the latest `main` behavior for repeated tool-error reminders.
- Integrated the PR behavior that recommends agent-driven context compression at task boundaries.
- Combined the test imports required by both changes.
- Added a scoped Pylint suppression after the merged injection test class exceeded the public-method threshold.
- Kept the original two-commit structure after rewriting history:
  - `860acf1` — `feat(agent): add agent-driven context compression tool`
  - `02e1bf4` — `refactor(agent): keep compression tool registration stable`

Validation:
- `67 passed` across `agent_injection_test.py`, `compress_context_test.py`, and `middleware_test.py`.
- Full pre-commit checks passed, including mypy, Black, Flake8, Pylint, and Pyroma.

DavdGao (approved):

PR #2146, opened 2026-07-21
fix(agent): omit attachments from compression input

### LINKED ISSUE #2145: [Bug]: _compress_context_impl does not handle DataBlock before structured output
## Bug Description

`_compress_context_impl` in `_agent.py` does not sanitize or skip `DataBlock` (image/audio) before passing messages to `generate_structured_output`. This causes:

1. **Token waste**: `count_tokens` includes DataBlock tokens, but formatter discards them → compression decision based on incorrect token estimation
2. **Semantic loss**: DataBlock cannot be summarized by structured output, but occupies context window
3. **Fragility**: If using a formatter without graceful DataBlock handling, it will crash (since `context_overflow=False` causes `raise e from None`)

## Steps to Reproduce

1. Create an agent with `max_context_tokens` configured (e.g., 100000)
2. Send messages containing `DataBlock` (e.g., image via `ImageBlock`)
3. Trigger context compression (when context exceeds threshold)
4. Observe that DataBlock is passed to `generate_structured_output` without filtering

## Code Analysis

**Location**: `src/agentscope/agent/_agent.py`, `_compress_context_impl` method (lines 312-510)

**Problem flow**:
1. Line 363-369: Extract `msgs_to_compress` from `self.state.context` (raw Msg objects, may contain DataBlock)
2. Line 411-418: Directly assemble into `messages` list
3. Line 450-453: Pass directly to `self.model.generate_structured_output(messages=messages, ...)`

**No DataBlock filtering anywhere in the compression path.**

**Formatter behavior**:
| Formatter | DataBlock Handling | Crash? |
|-----------|-------------------|--------|
| `DeepSeekChatFormatter` | Falls to `else` branch (line 163), warning + skip | No |
| `OpenAIChatFormatter` | Calls `_format_openai_data_block`, formats by media_type | Depends on media_type |
| `DeepSeekMultiAgentFormatter` | `_format_agent_message` only handles TextBlock | DataBlock ignored |

## Expected Behavior

`_compress_context_impl` should sanitize `msgs_to_compress` before assembling `messages`:
- Remove DataBlock, OR
- Replace with text placeholder (e.g., `"[image attached]"`), OR
- Mark as non-compressible and skip

## Environment

- AgentScope version: latest (main branch)
- Python: 3.12
- OS: Windows

## Additional Context

This is a silent failure — DeepSeek formatter won't crash but wastes tokens and loses semantics. The fix should be in `_compress_context_impl` (lines 392-418) to filter DataBlock before passing to `generate_structured_output`.

March-77:
## Summary

- replace non-text attachments with a fixed placeholder only in messages sent to context summarization
- cover direct `DataBlock` content plus attachments nested in tool results and hints
- preserve the original messages for reserved context and context offloading
- keep overflow retries on the same sanitized message copies

## Root cause

`_compress_context_impl()` passed the selected context messages directly to `generate_structured_output()`. Those messages could contain `DataBlock` payloads that text-only formatters discard or providers reject, while token estimation still charged a multimodal token estimate for each attachment.

## Impact

Context compression no longer sends attachment URLs or base64 payloads to the structured-output call. The summary still sees a stable marker that an attachment existed, and the original message remains unchanged wherever AgentScope retains or offloads it.

## Validation

- `python -m pytest tests/compress_context_test.py tests/agent_basic_test.py -q` (17 passed)
- `python -m pre_commit run --files src/agentscope/agent/_agent.py tests/compress_context_test.py`
- `git diff --check`

Fixes #2145


DavdGao:
I don't think we should strip multimodal data unconditionally. For a vision-capable model the attachments are exactly what should stay in the compression input, and they should keep counting toward the token estimate.

The authoritative source for that capability is the ModelCard — `input_types` is declared per model in `_models/*.yaml`, and it genuinely varies within a single API family: `qwen3.7-plus` declares `image/*`, while `qwen3.7-max` and `qwen-plus` are `text/plain` only. Note that the formatter's `input_types` is not a substitute here, since it's a class-level default per API family and has nothing to do with the `model` string actually passed in.

So one possible approach: look the model up through `list_models()`, and only strip when the card is found and declares no media input. Two practical caveats. `list_models()` re-scans and re-parses the YAML directory on every call, so it needs to be resolved once and cached if this check lands in a hot path. And the cards only cover mainstream names, so dated snapshots and custom OpenAI-compatible endpoints won't resolve — I'd default to keeping the data when no card is found, since stripping is lossy and irreversible while over-counting tokens isn't.

The same check should probably drive `count_tokens()` as well (`model/_base.py:427`), which currently charges a flat 2000 per `DataBlock` regardless of whether the formatter will ever send it. That over-count affects every call, including the compression trigger at the top of `_compress_context_impl`, not just this path.

March-77:
Thanks for pointing this out. I updated the implementation to make attachment handling depend on the concrete model's `ModelCard` rather than the formatter-level defaults.

The updated behavior is:

- Resolve the configured model through `list_models()` and cache the matched `ModelCard` on the model instance, so the YAML files are only scanned once.
- Match each `DataBlock` media type against the model card's `input_types`.
- Keep supported media in the context-compression input.
- Replace media with the text placeholder only when a known model does not declare a compatible input type.
- Conservatively keep media when no model card is found, covering dated snapshots and custom OpenAI-compatible endpoints.
- Use the same capability check in `count_tokens()`, so only media that the configured model accepts receives the flat multimodal token estimate.

This applies to direct `DataBlock` content as well as blocks nested in tool results and hints, while preserving the original messages.

Validation:

- `39 passed` across the relevant compression, token-counting, DashScope, and agent tests.
- All pre-commit checks passed, including mypy, black, flake8, and pylint.
- Real DashScope API validation:
  - `qwen3.7-plus` successfully inspected the test image and returned `dominant_color=red`.
  - `qwen3.7-plus` compression input contained `data_blocks=1`, `placeholders=0`, and produced a summary.
  - `qwen3.7-max` compression input contained `data_blocks=0`, `placeholders=1`, and produced a summary.

I have also attached the redacted terminal output from the real-model validation.
real_model_terminal_screenshot

PR #2147, opened 2026-07-21
feat(agent): emit context compression status events

### LINKED ISSUE #2144: [Feature] Emit SSE event before/after compress_context() for better UX
## Problem

When `compress_context()` is triggered (token count exceeds threshold), the operation blocks silently with no SSE event emitted to the frontend. Users see a "thinking" spinner but don't know the system is actually compressing context.

Worse, if compression fails (e.g., `RuntimeError: Failed to generate structured output for model` due to unsupported `DataBlock` in context), the error is classified as `ErrorType.UNKNOWN` and surfaces as a generic "unknown error" to users.

## Reproduction

1. Have a long conversation that accumulates >52k tokens
2. Send a message that triggers `compress_context()`
3. Observe: no SSE event during compression, just silence
4. If compression fails: `REPLY_END(finished_reason=error, error.type=unknown)`

## Current Code

In `_agent.py:825-831`:
```python
if action == "reasoning":
    # Compressed the memory if needed before reasoning
    await self.compress_context()  # ← blocking, no event
    async for evt in self._inject_runtime_state():
        yield evt
```

The `on_compress_context` middleware hook exists but has signature `-> None`, not `AsyncGenerator`, so it cannot yield events.

## Suggested Solution

Emit a `HintBlock` event before compression starts:

```python
if action == "reasoning":
    if self._should_compress():
        yield Msg(
            name="system",
            content=[HintBlock(hint="Compressing conversation context...")],
            role="user",
        )
    await self.compress_context()
    async for evt in self._inject_runtime_state():
        yield evt
```

This would allow frontends to display "Compressing context..." during the operation.

## Environment

- AgentScope version: 2.0.4 (editable install from main branch)
- Model: DeepSeek V4
- Runtime: BusinessEmbedDemo

## Additional Context

The compression failure itself is a separate bug - `generate_structured_output()` doesn't handle `DataBlock` gracefully. But even if that's fixed, the UX issue of silent compression remains.

March-77:
## Summary

- emit `context_compression` custom events with `started`, `completed`, and `failed` statuses from the public reply stream
- emit lifecycle events only when the compression threshold is actually crossed
- preserve generic `CustomEvent` names and values through the AG-UI protocol adapter
- cover successful, skipped, failed, and public-stream ordering paths

## Root cause

`compress_context()` is an awaitable hook whose compression threshold check happens inside `_compress_context_impl()`. The reasoning loop awaited it directly, so subscribers received no event while structured-output summarization was running and could not distinguish a skipped compression from an active one. In addition, the AG-UI adapter converted generic AgentScope custom events to an `unknown` event name.

The focused adapter performs the threshold check in the current reply task, emits `started` before the blocking compression work, and reuses the prepared model input and token estimate in the core implementation. This keeps token counting at one pass, preserves the existing asynchronous scheduling behavior, leaves the public compression and middleware APIs unchanged, and propagates failures and cancellation as before.

## Validation

- `python -m pytest tests/middleware_filesystem_memory_test.py::AgenticMemoryMiddlewareTest::test_agent_reasoning_injects_selected_memory_hint tests/middleware_filesystem_memory_test.py::AgenticMemoryMiddlewareTest::test_agent_filters_hallucinated_memory_filenames tests/compress_context_test.py tests/agui_protocol_test.py tests/agent_basic_test.py tests/middleware_test.py -q` (`73 passed`)
- `python -m pre_commit run --files src/agentscope/agent/_agent.py src/agentscope/app/middleware/_protocol/_agui.py tests/compress_context_test.py tests/agui_protocol_test.py`
- `git diff --check`

Fixes #2144


March-77:
Rebased the context-compression event changes onto the latest `main`. The resolution preserves the new `Exit`/`Reasoning`/`Acting` flow while emitting compression lifecycle events before reasoning. Focused tests, related agent regressions, pre-commit, and all GitHub CI checks pass.

March-77:
Synced this PR with the latest `main` (`da00849f`) in `cb981db1` and resolved the only conflict in `tests/compress_context_test.py` by retaining both the PR's context-compression lifecycle coverage and the upstream image-limit/fallback coverage.

The merge also accounts for the new upstream behavior where summary-generation failures fall back to context truncation instead of propagating. The failed-event test now injects a genuinely propagated compression exception, preserving the intended lifecycle assertion without defeating the fallback.

Validation on the merged head:
- `uv run --extra dev pytest tests/compress_context_test.py tests/agui_protocol_test.py -q`: 52 passed
- pre-commit on all four PR files: AST, mypy, Black, flake8, pylint, Pyroma, and remaining hooks passed
- `git diff --check`: passed

GitHub now reports the PR as mergeable.

PR #2160, opened 2026-07-23
feat(tool): complete PowerShell permission checks and rule matching

### LINKED ISSUE #2156: feat(powershell): complete PowerShell permission checks, rule matching, and dynamic read-only
**Background**

PR #2132 added the PowerShell tool but deferred all command-level security. Three gaps remain vs. `Bash`:

1. `check_permissions` returns a flat ASK for every command — no read-only auto-allow, no safety classification.
2. `match_rule` is not overridden, so it inherits the default (rule_content is None). Only tool-name-level rules work; any rule with content (e.g. Remove-Item*) silently never matches.
3. No dynamic read-only (`check_read_only` not overridden), so under EXPLORE mode even Get-ChildItem is denied.

**Changes**

1. Override `match_rule` for content matching (grammar TBD — see Note).
2. Override `check_read_only` for per-invocation read-only detection.
3. Rework `check_permissions` into Bash-style tiers: read-only auto-allow + bypass-immune safety ASKs for dangerous commands.
4. Add unit tests for each.

**Suggestion**

- **Parser** — Bash's parser (`_bash_parser.py`) is tree-sitter-based but tied to the bash grammar and **not reusable** for PowerShell. Reuse the *architecture* instead: a new `PowerShellCommandParser` mirroring Bash's method surface (`is_read_only_command` / `check_dangerous_command` / `check_injection_risk`), with command sets in `_constants.py`.
- **tree-sitter** — the right direction (consistent with Bash), but needs a `tree-sitter-powershell` grammar whose maturity / PyPI packaging must be verified. Fallback: tokenizer + curated command sets. Avoid `[Parser]::ParseInput()` (would require spawning PowerShell to decide whether to run PowerShell).
- **Command coverage (draft):**
  - *Read-only (auto-allow):* `Get-*`, `Test-Path`, `Resolve-Path`, `Select-*`, `Where/Sort/Measure-Object`, `Format-*`, `ConvertTo/From-*`, and aliases (`ls/cat/gc/dir/sls/…`). Anything containing a script block `{...}`, `&`, `iex`, or a pipeline into a mutating cmdlet is treated as non-read-only.
  - *Dangerous (bypass-immune ASK):* `Remove-Item -Recurse/-Force`, `Clear-Content/Item`, `Format-Volume`, registry writes (`Set-ItemProperty HKLM:`), `Invoke-Expression` / `Start-Process` / `Add-Type`, download-to-`iex`, `Set-ExecutionPolicy` / `Set-MpPreference`, `Stop/Restart-Computer`, `Stop-Process -Force`, `Register-ScheduledTask`.
  - *Injection (force non-read-only + ASK):* `$(...)` with side effects, `&` / `.` on dynamic strings, `iex`, nested `-EncodedCommand`, backtick obfuscation, string-built cmdlet names.
- **`match_rule` grammar** — reuse Bash's pattern language (wildcard `*`, prefix `:*`, substring), plus two PowerShell specifics: **case-insensitive** matching and **alias normalization** before matching (so `Get-ChildItem*` also covers `ls`). Verb-noun structure makes `Get-*` prefixes natural.

**Note**

This change alters the tool's security granularity, so the design must be **confirmed in this issue before implementation**:

1. **Rule-match grammar** — adopt the Suggestion above (Bash grammar + case-insensitive + alias normalization), or a different scheme?
2. **Safety-check scope** — confirm/adjust the dangerous vs. read-only command sets listed under Suggestion.

Shaurya2k06:
## AgentScope Version

2.0.5

### Problem

PR #2132 added the PowerShell tool but deferred command-level security. Three gaps remained vs Bash:

1. `check_permissions` returned a flat ASK for every command — no read-only auto-allow, no safety classification.
2. `match_rule` was not overridden, so only tool-name-level rules worked; content rules like `Remove-Item*` never matched.
3. `check_read_only` was not overridden, so under EXPLORE mode even `Get-ChildItem` / `ls` was denied.

### Solution

Add a `PowerShellCommandParser` (tree-sitter-pwsh) mirroring Bash's method surface, then wire Bash-style permission tiers into the PowerShell tool: injection ASK → read-only ALLOW → dangerous ASK → PASSTHROUGH, plus case-insensitive alias-aware rule matching.

### Changes

- add core dependency `tree_sitter_pwsh` alongside `tree_sitter_bash`
- add PowerShell RO / dangerous / alias / injection-node constants in `_constants.py`
- add `PowerShellCommandParser` with `is_read_only_command`, `check_dangerous_command`, `check_injection_risk`, `extract_command_prefixes`, and alias normalization
- override `check_read_only`, rework `check_permissions`, override `match_rule`, and restore prefix `:*` `generate_suggestions` on `PowerShell`
- add parser unit tests and extend PowerShell tool permission / match_rule coverage

### Testing

- `python -m pytest tests/permission_powershell_parser_test.py tests/builtin_powershell_test.py -q`
  - 33 passed, 3 skipped
- `python -m pytest tests/permission_powershell_parser_test.py tests/builtin_powershell_test.py tests/builtin_bash_test.py tests/permission_bash_parser_test.py tests/permission_engine_test.py tests/permission_mode_test.py tests/workspace_local_test.py -q`
  - 246 passed, 3 skipped
- pre-commit on changed files — all hooks passed
- `python -m pytest tests -q`
  - 1469 passed, 116 skipped, 1 unrelated failure
  - remaining failure: `tests/mcp_sse_client_test.py::SseSchemaDefsPreservationTest::test_defs_preserved_and_titles_stripped` (no MCP sources changed)

### Notes for Reviewer

- Parse failure / ERROR nodes are treated as non-read-only and produce a bypass-immune safety ASK (conservative).
- Rule matching reuses Bash's `*` / `:*` / substring grammar, plus case-insensitive matching and alias normalization (`ls` ↔ `Get-ChildItem`).
- ACCEPT_EDITS filesystem auto-allow and Bash-style dangerous-path/rm checks are intentionally out of scope for this PR.

Closes #2156

### DIFF
--- a/src/agentscope/tool/_builtin/_powershell.py
+++ b/src/agentscope/tool/_builtin/_powershell.py
@@ -3,9 +3,11 @@
 
 import base64
 import os
+import re
 from typing import AsyncGenerator, Any, List
 
 from ._backend import BackendBase, LocalBackend, _normalize_newlines
+from ._powershell_parser import PowerShellCommandParser
 from .._base import ToolBase, ToolMiddlewareBase
 from .._response import ToolChunk
 from ...message import TextBlock, ToolResultState
@@ -113,6 +115,7 @@ are easier for the user to review and authorize:
         self._cwd = os.fspath(cwd) if cwd is not None else None
         self._backend = backend or LocalBackend()
         self._executable: str | None = None
+        self._powershell_parser = PowerShellCommandParser()
 
     async def _resolve_executable(self) -> str:
         """Prefer PowerShell 6+ and cache the first available executable."""
@@ -136,34 +139,220 @@ are easier for the user to review and authorize:
                 self._executable = "powershell.exe"
         return self._executable
 
+    async def check_read_only(
+        self,
+        tool_input: dict[str, Any],
+    ) -> bool:
+        """Decide whether this PowerShell invocation is read-only.
+
+        Inspects the command and returns ``True`` for known-safe read-only
+        cmdlets (e.g. ``Get-ChildItem``, ``Test-Path``, ``ls``). The static
+        :attr:`is_read_only` class attribute remains ``False`` because
+        PowerShell can execute arbitrary commands; this method overrides
+        that with a per-invocation answer.
+        """
+        command = tool_input.get("command", "")
+        if not command:
+            return self.is_read_only
+        # Dynamic / unanalyzable structure cannot be proven read-only —
+        # e.g. ``Get-ChildItem $(Remove-Item -Recurse C:\\)``.
+        if self._powershell_parser.check_injection_risk(command):
+            return False
+        return self._powershell_parser.is_read_only_command(command)
+
     async def check_permissions(
         self,
         tool_input: dict[str, Any],
         context: PermissionContext,
     ) -> PermissionDecision:
-        """Ask the user to confirm every PowerShell command.
+        """Check permissions for PowerShell command execution.
 
-        PowerShell-specific command validation is intentionally outside this
-        implementation. Since no command is classified as safe, every
-        invocation prompts the user. This is a regular ASK that allow rules
-        and BYPASS mode may still override.
+        Implements Bash-style tiered checks:
+
+        0. Injection risk (bypass-immune safety ASK)
+        1. Read-only command auto-ALLOW
+        2. Dangerous command patterns (bypass-immune safety ASK)
+        3. PASSTHROUGH for the permission engine
+
+        Args:
+            tool_input (`dict[str, Any]`):
+                Tool input containing a ``command`` key.
+            context (`PermissionContext`):
+                Permission context with mode and rules.
+
+        Returns:
+            `PermissionDecision`:
+                ALLOW, ASK (possibly bypass-immune), or PASSTHROUGH.
         """
-        return PermissionDecision(
-            behavior=PermissionBehavior.ASK,
-            message="Execute PowerShell command",
-            decision_reason="PowerShell command validation is not enabled",
+        command = tool_input.get("command", "")
+        if not command:
+            return PermissionDecision(
+                behavior=PermissionBehavior.PASSTHROUGH,
+                message="Empty command",
+            )
+
+        injection_reason = self._powershell_parser.check_injection_risk(
+            command,
         )
+        if injection_reason:
+            return PermissionDecision(
+                behavior=PermissionBehavior.ASK,
+                message=f"Permission required: {injection_reason}",
+                decision_reason=(
+                    "Safety check: command contains dynamic expansion "
+                    "that cannot be statically analyzed"
+                ),
+                bypass_immune=True,
+            )
+
+        if self._powershell_parser.is_read_only_command(command):
+            return PermissionDecision(
+                behavior=PermissionBehavior.ALLOW,
+                message="Permission granted for read-only command",
+                decision_reason="Read-only command is allowed",
+            )
+
+        dangerous_pattern = self._powershell_parser.check_dangerous_command(
+            command,
+        )
+        if dangerous_pattern:
+            return PermissionDecision(
+                behavior=PermissionBehavior.ASK,
+                message=(
+                    "Permission required: Command contains dangerous "
+                    f"pattern: {dangerous_pattern}"
+                ),
+                decision_reason=(
+                    "Safety check: dangerous command pattern detected"
+                ),
+                bypass_immune=True,
+            )
+
+        return PermissionDecision(
+            behavior=PermissionBehavior.PASSTHROUGH,
+            message="PowerShell permission check passed",
+        )
+
+    async def match_rule(
+        self,
+        rule_content: str | None,
+        tool_input: dict[str, Any],
+    ) -> bool:
+        r"""Match PowerShell commands with Bash-compatible wildcards.
+
+        Pattern grammar matches Bash, with PowerShell-specific extras:
+
+        - Case-insensitive matching
+        - Alias normalization (``ls`` matches ``Get-ChildItem*``)
+        - ``:*`` prefix patterns, ``*`` wildcards, and substring match
+        - ``None`` rule content matches all invocations
+
+        Args:
+            rule_content (`str | None`):
+                Command pattern to match, or ``None`` for tool-name rules.
+            tool_input (`dict[str, Any]`):
+                Must contain a ``command`` key.
+
+        Returns:
+            `bool`:
+                ``True`` when the pattern matches the command.
+        """
+        if rule_content is None:
+            return True
+
+        command = tool_input.get("command", "")
+        command = self._powershell_parser.normalize_command_for_match(command)
+        pattern_source = self._powershell_parser.normalize_command_for_match(
+            rule_content,
+        )
+        command_cf = command.casefold()
+        rule_cf = pattern_source.casefold()
+
+        if rule_cf.endswith(":*"):
+            prefix = rule_cf[:-2].strip()
+            return command_cf.startswith(prefix + " ") or command_cf == prefix
+
+        def has_wildcards(pattern: str) -> bool:
+            """Check if pattern contains unescaped * wildcards."""
+            i = 0
+            while i < len(pattern):
+                if pattern[i] == "\\":
+                    i += 2
+                elif pattern[i] == "*":
+                    return True
+                else:
+                    i += 1
+            return False
+
+        if not has_wildcards(rule_cf):
+            pattern = rule_cf
+            pattern = pattern.replace("\\\\", "\x00BACKSLASH\x00")
+            pattern = pattern.replace("\\*", "*")
+            pattern = pattern.replace("\x00BACKSLASH\x00", "\\")
+            return pattern in command_cf
+
+        escaped_star = "\x00ESCAPED_STAR\x00"
+        escaped_backslash = "\x00ESCAPED_BACKSLASH\x00"
+
+        pattern = rule_cf
+        pattern = pattern.replace("\\\\", escaped_backslash)
+        pattern = pattern.replace("\\*", escaped_star)
+
+        special_chars = r".^$+?{}[]|()"
+        for char in special_chars:
+            pattern = pattern.replace(char, "\\" + char)
+
+        pattern = pattern.replace("*", ".*")
+        pattern = pattern.replace(escaped_star, r"\*")
+        pattern = pattern.replace(escaped_backslash, r"\\")
+
+        if pattern.endswith(".*"):
+            base_pattern = pattern[:-2].rstrip()
+            if re.fullmatch(base_pattern, command_cf):
+                return True
+
+        try:
+            return bool(re.fullmatch(pattern, command_cf))
+        except re.error:
+            return rule_cf.replace("*", "") in command_cf
 
     async def generate_suggestions(
         self,
         tool_input: dict[str, Any],
     ) -> List[PermissionRule]:
-        """Return no automatic allow-rule suggestions.
+        """Generate cmdlet prefix allow-rule suggestions.
 
-        A broad rule would weaken the conservative permission boundary before
-        PowerShell-specific command validation is available.
+        Produces rules such as ``Remove-Item:*`` from the command's
+        alias-normalized cmdlet names.
+
+        Args:
+            tool_input (`dict[str, Any]`):
+                Tool input containing a ``command`` key.
+
+        Returns:
+            `List[PermissionRule]`:
+                Suggested allow rules for the current command.
         """
-        return []
+        command = tool_input.get("command", "")
+        if not command:
+            return []
+
+        prefixes = self._powershell_parser.extract_command_prefixes(
+            command,
+            max_prefixes=5,
+        )
+        if not prefixes:
+            return []
+
+        return [
+            PermissionRule(
+                tool_name="PowerShell",
+                rule_content=f"{prefix}:*",
+                behavior=PermissionBehavior.ALLOW,
+                source="suggested",
+            )
+            for prefix in prefixes
+        ]
 
     async def call(  # type: ignore[override] # pylint: disable=unused-argument
         self,
--- a/src/agentscope/tool/_builtin/_powershell.py
+++ b/src/agentscope/tool/_builtin/_powershell.py
@@ -3,11 +3,11 @@
 
 import base64
 import os
-import re
 from typing import AsyncGenerator, Any, List
 
 from ._backend import BackendBase, LocalBackend, _normalize_newlines
 from ._powershell_parser import PowerShellCommandParser
+from ._wildcard_match import match_wildcard_pattern
 from .._base import ToolBase, ToolMiddlewareBase
 from .._response import ToolChunk
 from ...message import TextBlock, ToolResultState
@@ -244,7 +244,7 @@ are easier for the user to review and authorize:
 
         - Case-insensitive matching
         - Alias normalization (``ls`` matches ``Get-ChildItem*``)
-        - ``:*`` prefix patterns, ``*`` wildcards, and substring match
+        - ``:*`` prefix patterns require every pipeline segment to match
         - ``None`` rule content matches all invocations
 
         Args:
@@ -261,60 +261,25 @@ are easier for the user to review and authorize:
             return True
 
         command = tool_input.get("command", "")
-        command = self._powershell_parser.normalize_command_for_match(command)
-        pattern_source = self._powershell_parser.normalize_command_for_match(
+        rule_cf = self._powershell_parser.normalize_command_for_match(
             rule_content,
-        )
-        command_cf = command.casefold()
-        rule_cf = pattern_source.casefold()
+        ).casefold()
 
         if rule_cf.endswith(":*"):
-            prefix = rule_cf[:-2].strip()
-            return command_cf.startswith(prefix + " ") or command_cf == prefix
+            prefix = self._powershell_parser.normalize_cmdlet_name(
+                rule_cf[:-2].strip(),
+            ).casefold()
+            names = self._powershell_parser.extract_canonical_command_names(
+                command,
+            )
+            if not names:
+                return False
+            return all(name.casefold() == prefix for name in names)
 
-        def has_wildcards(pattern: str) -> bool:
-            """Check if pattern contains unescaped * wildcards."""
-            i = 0
-            while i < len(pattern):
-                if pattern[i] == "\\":
-                    i += 2
-                elif pattern[i] == "*":
-                    return True
-                else:
-                    i += 1
-            return False
-
-        if not has_wildcards(rule_cf):
-            pattern = rule_cf
-            pattern = pattern.replace("\\\\", "\x00BACKSLASH\x00")
-            pattern = pattern.replace("\\*", "*")
-            pattern = pattern.replace("\x00BACKSLASH\x00", "\\")
-            return pattern in command_cf
-
-        escaped_star = "\x00ESCAPED_STAR\x00"
-        escaped_backslash = "\x00ESCAPED_BACKSLASH\x00"
-
-        pattern = rule_cf
-        pattern = pattern.replace("\\\\", escaped_backslash)
-        pattern = pattern.replace("\\*", escaped_star)
-
-        special_chars = r".^$+?{}[]|()"
-        for char in special_chars:
-            pattern = pattern.replace(char, "\\" + char)
-
-        pattern = pattern.replace("*", ".*")
-        pattern = pattern.replace(escaped_star, r"\*")
-        pattern = pattern.replace(escaped_backslash, r"\\")
-
-        if pattern.endswith(".*"):
-            base_pattern = pattern[:-2].rstrip()
-            if re.fullmatch(base_pattern, command_cf):
-                return True
-
-        try:
-            return bool(re.fullmatch(pattern, command_cf))
-        except re.error:
-            return rule_cf.replace("*", "") in command_cf
+        command_cf = self._powershell_parser.normalize_command_for_match(
+            command,
+        ).casefold()
+        return match_wildcard_pattern(rule_cf, command_cf)
 
     async def generate_suggestions(
         self,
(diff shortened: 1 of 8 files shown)


Shaurya2k06:
Hey @DavdGao , could I get a review on this?

DavdGao (commented):
Thanks for picking this up -- the tiered structure and the AST-based command splitting are the right shape. I installed `tree_sitter_pwsh` and ran this branch locally; the outputs quoted in the comments below are from that. The recurring theme is that the places which fall back from the AST to string comparison are also the places that turn out to be bypassable.

src/agentscope/tool/_constants.py
```diff
@@ -108,3 +108,149 @@
 #
 # Note: simple_expansion ($VAR) is handled separately with allowlist
 # for known-safe environment variables ($HOME, $PWD, etc.)
+
+
+POWERSHELL_ALIASES: dict[str, str] = {
```
DavdGao, line 117: This is a partial snapshot -- 39 of roughly 150 built-in aliases -- and the gaps land on the dangerous list rather than the read-only one. `clc` -> `Clear-Content`, `cli` -> `Clear-Item`, and `saps`/`start` -> `Start-Process` are all missing, so `clc a.txt` and `saps notepad` both return `None` from `check_dangerous_command` even though their canonical names are flagged.

At minimum, add every built-in alias for each name in `POWERSHELL_DANGEROUS_COMMANDS`, and record in a comment which PowerShell version's `Get-Alias` output this was taken from. The more durable fix is to never auto-allow a name that can't be normalized to a known canonical cmdlet, so an incomplete table costs an extra prompt instead of a missed check.

src/agentscope/tool/_constants.py
```diff
@@ -108,3 +108,149 @@
...
+    "Write-Verbose",
+    "Write-Debug",
+    "Write-Information",
+}
+# Cmdlets treated as read-only for auto-ALLOW / EXPLORE mode.
+# Verb prefixes such as Get- / Select- / Format- / ConvertTo- /
+# ConvertFrom- are handled separately in the PowerShell parser.
+
+
+POWERSHELL_READ_ONLY_VERB_PREFIXES: tuple[str, ...] = (
```
DavdGao: Prefix matching makes unknown cmdlets read-only by default -- `Get-Foo` (a user- or module-defined function) and `ConvertFrom-SecureString` both classify as read-only. It also collides with the dangerous list two definitions below: `Format-Volume` starts with `Format-`, so `Format-Volume -DriveLetter C` is read-only and gets auto-ALLOWed with no prompt in every mode.

Suggest dropping this tuple and relying on the explicit `POWERSHELL_READ_ONLY_COMMANDS` set, so unknown names fall through to PASSTHROUGH instead of being assumed safe.

src/agentscope/tool/_constants.py
```diff
@@ -108,3 +108,149 @@
...
+    "Select-",
+    "Format-",
+    "ConvertTo-",
+    "ConvertFrom-",
+)
+# Verb prefixes that are generally read-only when the cmdlet has no
+# script block, call operator, or redirection.
+
+
+POWERSHELL_DANGEROUS_COMMANDS: list[str] = [
```
DavdGao, line 232: `Invoke-Command` is missing -- it executes script blocks locally and on remote machines, which is at least as dangerous as `Start-Process`. Suggest adding it along with its alias `icm`, and considering `Stop-Service` / `Set-Service` / `New-Service` while you're here.

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+        Returns:
+            `str`:
+                Command text with leading aliases expanded where practical.
+        """
+        if not command.strip():
+            return command
+
+        try:
+            tree = self.parser.parse(bytes(command, "utf8"))
+        except Exception:
```
DavdGao: `parser.parse` doesn't realistically raise, so a bare `except Exception` here mainly serves to swallow genuine bugs in the surrounding code. Parse failures are already represented by ERROR nodes and handled by `_has_error_nodes`. Same applies at lines 143, 187, 276 and 331 -- suggest narrowing or dropping them.

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+            canonical = self.normalize_cmdlet_name(name)
+            folded = canonical.casefold()
+            if folded in self._dangerous_lookup:
+                return self._dangerous_lookup[folded]
+
+            params = self._command_parameters(cmd, cmd_node)
+            param_fold = {p.casefold() for p in params}
+
+            if folded == "remove-item" and (
+                "-recurse" in param_fold or "-force" in param_fold
```
DavdGao: PowerShell resolves parameter names by unique prefix, so exact string comparison against the full spelling is bypassable. Tested on this branch:

```
'Remove-Item -Recurse -Force ./x'  ->  Remove-Item -Recurse/-Force
'Remove-Item -Rec -Fo ./x'         ->  None
'Remove-Item -R ./x'               ->  None
'Stop-Process -F -Id 1'            ->  None
```

All of the `None` cases are equivalent to the forms that are caught. Suggest resolving each `command_parameter` back to its canonical name by unique-prefix matching against the cmdlet's known parameter set before comparing.

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+
+            if folded == "set-itemproperty" and self._mentions_hklm(
+                cmd,
+                cmd_node,
+            ):
+                return "Set-ItemProperty HKLM:"
+
+        # Textual fallback for always-dangerous names that may appear
+        # oddly tokenized
+        normalized = " ".join(cmd.split())
```
DavdGao: This scans the whole raw command, so it matches dangerous names inside string literals and filenames:

```
'Get-Content notes-Invoke-Expression.txt'   ->  Invoke-Expression
"Write-Output 'run Start-Process later'"    ->  Start-Process
```

It's masked today because the read-only branch returns first, but it turns into a visible false ASK once the read-only/dangerous ordering is fixed. Suggest gating it on `_has_error_nodes(root)` so it only runs when the AST really is unusable.

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+        Returns:
+            `Optional[str]`:
+                Reason string when review is required, otherwise ``None``.
+        """
+        cmd = command.strip()
+        if not cmd:
+            return None
+
+        # Heuristics that do not require a clean AST
+        if re.search(r"(?i)(^|[\s|;])-EncodedCommand\b", cmd):
```
DavdGao: Same prefix-abbreviation problem as the `-Recurse` check, on the CLI side: `pwsh -enc <b64>` and `pwsh -e <b64>` are both valid and return `None` here, and `-enc` is the form that actually shows up in the wild. Suggest matching any unique prefix of `-EncodedCommand` rather than the full spelling.

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+                break
+        return prefixes
+
+    # ------------------------------------------------------------------
+    # Internal helpers
+    # ------------------------------------------------------------------
+
+    def _normalize_command_fallback(self, command: str) -> str:
+        """Token-based alias normalization when parsing fails."""
+        tokens = command.split(None, 1)
```
DavdGao: `command.split(None, 1)` only runs when `parser.parse` raises, which tree-sitter essentially never does -- parse problems surface as ERROR nodes instead. As written this is unreachable token-splitting. Suggest removing it, or documenting the condition that can actually reach it.

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+            return False
+        return self._is_readonly_cmdlet_name(self.normalize_cmdlet_name(name))
+
+    def _is_readonly_cmdlet_name(self, name: str) -> bool:
+        """Return whether a canonical cmdlet name is read-only."""
+        folded = name.casefold()
+        if folded in self._readonly_lookup:
+            return True
+        for prefix in POWERSHELL_READ_ONLY_VERB_PREFIXES:
+            if folded.startswith(prefix.casefold()):
```
DavdGao: Because this returns `True` on prefix alone, a cmdlet that is both prefix-read-only and explicitly dangerous wins the read-only branch -- `Format-Volume` is the live case. If the prefix list stays, short-circuit first with `if folded in self._dangerous_lookup: return False`, so the two tables can't contradict each other. An exact set lookup would be safer still.

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+        params: list[str] = []
+        for node in self._iter_nodes(cmd_node):
+            if node.type == "command_parameter":
+                params.append(source[node.start_byte : node.end_byte])
+        return params
+
+    def _mentions_hklm(self, source: str, cmd_node: Node) -> bool:
+        """Return whether a command node references an HKLM path."""
+        text = source[cmd_node.start_byte : cmd_node.end_byte]
+        return bool(re.search(r"(?i)\bHKLM:", text))
```
DavdGao: Regexing the raw node text misses the case that matters and fires on the one that doesn't. `Set-ItemProperty -Path $regPath -Name Y -Value 1` returns `None` here, and since `variable` isn't in `POWERSHELL_INJECTION_NODE_TYPES` it doesn't raise an injection ASK either -- it just PASSTHROUGHs. Meanwhile `Get-Content 'notes about HKLM: keys'` would match on the string literal.

Suggest reading the `-Path` argument node from the AST, and treating a path that is a variable or expression as injection risk rather than as "not HKLM".

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+
+    def _check_download_to_iex(
+        self,
+        source: str,
+        root: Node,
+    ) -> Optional[str]:
+        """Detect download-to-iex patterns such as ``irm ... | iex``."""
+        normalized = " ".join(source.split())
+        if re.search(
```
DavdGao: The `.*` spans the entire command, so `Get-Content iwr-notes.txt | Select-String 'iex'` and `Write-Output 'use irm then pipe to iex'` both come back as `download-to-iex`. The AST branch below has the opposite problem -- it only checks that both names appear somewhere, not that they're adjacent in a pipeline.

Suggest walking the pipeline node and checking that a download cmdlet feeds directly into `Invoke-Expression`, and dropping the regexes.

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+            if name.casefold() == target
+        ]
+
+    def _has_backtick_obfuscation(self, command: str) -> bool:
+        """Detect backtick-obfuscated command text."""
+        # e.g. Inv`oke-Expression or Get`-ChildItem
+        if re.search(r"[A-Za-z]`+[A-Za-z]", command):
+            return True
+        # Many backticks are a strong obfuscation signal
+        return command.count("`") >= 3
```
DavdGao: ``count("`") >= 3`` also fires on ordinary strings carrying several escape sequences, e.g. ``"a`nb`tc`r"``. Conservative so it's only noise, but stripping quoted strings before counting would remove the false positive.

src/agentscope/tool/_builtin/_powershell_parser.py
```diff
@@ -0,0 +1,536 @@
...
+        ):
+            return True
+        return bool(
+            re.search(
+                r"""(?i)\.\s*[\(\"'].*(?:\+|\$)""",
+                command,
+            ),
+        )
+
+    def _walk_for_injection_nodes(self, node: Node) -> Optional[str]:
```
DavdGao: This hand-rolls a recursive walk while `_iter_nodes` already provides a depth-first iterator in the same class. `for node in self._iter_nodes(root): if node.type in POWERSHELL_INJECTION_NODE_TYPES: ...` would be shorter and avoids recursion depth on deep trees.

src/agentscope/tool/_builtin/_powershell.py
```diff
@@ -113,6 +115,7 @@ def __init__(
         self._cwd = os.fspath(cwd) if cwd is not None else None
         self._backend = backend or LocalBackend()
         self._executable: str | None = None
+        self._powershell_parser = PowerShellCommandParser()
```
DavdGao, line 118: A single `check_permissions` parses the same command four or more times: `check_injection_risk`, then `is_read_only_command` which calls `check_injection_risk` again internally, then `check_dangerous_command`. `match_rule` adds one more parse per rule evaluated. Suggest memoizing the parse tree on the command string inside the parser.

src/agentscope/tool/_builtin/_powershell.py
```diff
@@ -136,34 +139,220 @@ async def _resolve_executable(self) -> str:
...
+                behavior=PermissionBehavior.ASK,
+                message=f"Permission required: {injection_reason}",
+                decision_reason=(
+                    "Safety check: command contains dynamic expansion "
+                    "that cannot be statically analyzed"
+                ),
+                bypass_immune=True,
+            )
+
+        if self._powershell_parser.is_read_only_command(command):
```
DavdGao, line 208: This returns ALLOW before `check_dangerous_command` runs, so any name that is both read-only-classified and dangerous is silently allowed -- `Format-Volume` today.

Worth noting the engine's read-only fast path (`_check_read_only_fast_path`, step 3) runs before this method is called at all, and EXPLORE never calls `check_permissions`, so reordering here isn't sufficient on its own. The fix needs to be in the classification: make read-only exclude everything in the dangerous set, so `check_read_only` and this method agree.

src/agentscope/tool/_builtin/_powershell.py
```diff
@@ -136,34 +139,220 @@ async def _resolve_executable(self) -> str:
...
+
         return PermissionDecision(
-            behavior=PermissionBehavior.ASK,
-            message="Execute PowerShell command",
-            decision_reason="PowerShell command validation is not enabled",
+            behavior=PermissionBehavior.PASSTHROUGH,
+            message="PowerShell permission check passed",
+        )
+
+    async def match_rule(
```
DavdGao, line 236: Everything from the `has_wildcards` helper down is a line-for-line copy of `_bash.py:396-490`; only the casefolding and alias normalization are new. Two copies of the same wildcard grammar will drift as one side gets fixed. Suggest extracting the shared matcher into a module-level helper that both tools call after their own preprocessing.

src/agentscope/tool/_builtin/_powershell.py
```diff
@@ -136,34 +139,220 @@ async def _resolve_executable(self) -> str:
...
+        command = tool_input.get("command", "")
+        command = self._powershell_parser.normalize_command_for_match(command)
+        pattern_source = self._powershell_parser.normalize_command_for_match(
+            rule_content,
         )
+        command_cf = command.casefold()
+        rule_cf = pattern_source.casefold()
+
+        if rule_cf.endswith(":*"):
+            prefix = rule_cf[:-2].strip()
```
DavdGao: The command is alias-normalized on line 264 but the prefix extracted here isn't, so alias-written prefix rules never match anything:

```
match_rule("rm:*",          "rm ./tmp")           -> False
match_rule("rm:*",          "Remove-Item ./tmp")  -> False
match_rule("Remove-Item:*", "rm ./tmp")           -> True
```

`generate_suggestions` emits canonical names so this stays hidden, but a hand-written `rm:*` deny rule silently does nothing. Suggest `prefix = self._powershell_parser.normalize_cmdlet_name(rule_cf[:-2].strip())`.

src/agentscope/tool/_builtin/_powershell.py
```diff
@@ -136,34 +139,220 @@ async def _resolve_executable(self) -> str:
...
+            return command_cf.startswith(prefix + " ") or command_cf == prefix
```
DavdGao: Looking only at the head of the command means an allow rule for a read-only cmdlet covers whatever follows it. Tested: rule `Get-ChildItem:*` matches `Get-ChildItem | Remove-Item`, and rule `Get-Content:*` matches `Get-Content a.txt | Set-Content b.txt`; both PASSTHROUGH from `check_permissions`, so the allow rule grants them at engine step 5.

`is_read_only_command` already gets this right by requiring every `command` node to qualify -- suggest using the AST here too, requiring either a single `command` node or that every segment matches the prefix.

tests/builtin_powershell_test.py
```diff
@@ -37,15 +37,133 @@ async def test_public_interface_is_conservative(self) -> None:
...
+
+
+class PowerShellPermissionTest(IsolatedAsyncioTestCase):
+    """Test PowerShell permission tiers and rule matching."""
+
+    async def asyncSetUp(self) -> None:
+        """Create a PowerShell tool for permission tests."""
+        self.tool = PowerShell(backend=LocalBackend())
+
+    async def test_dangerous_commands_are_bypass_immune_ask(self) -> None:
```
DavdGao, line 90: Dangerous cmdlets are only asserted through `parser.check_dangerous_command` directly, never through `check_permissions` or `check_read_only` -- which is exactly the path where `Format-Volume` gets auto-allowed.

Suggest a table-driven test over `POWERSHELL_DANGEROUS_COMMANDS` asserting `check_read_only` is `False` and `check_permissions` returns a bypass-immune ASK for every entry. Worth covering alongside it: the `-Rec` / `-Fo` / `-enc` abbreviations, every built-in alias of each dangerous cmdlet, `Get-ChildItem:*` against `Get-ChildItem | Remove-Item`, and `rm:*` allow/deny matching.

src/agentscope/tool/_constants.py
Shaurya2k06: Done — added every built-in alias for the dangerous cmdlets (from PS 7.4 Get-Alias) and unknown names are no longer auto-allowed.

src/agentscope/tool/_constants.py
Shaurya2k06: Done — dropped the verb-prefix list; only the exact read-only set counts as safe now.

src/agentscope/tool/_constants.py
Shaurya2k06: Done — added Invoke-Command (icm) plus Stop/Set/New-Service.

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — removed the bare excepts; we rely on ERROR nodes instead.

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — parameters are resolved by unique prefix, so -Rec/-Fo/-R/-F are caught.

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — the text fallback only runs when the AST has ERROR nodes.

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — -enc/-en… and pwsh -e now match as EncodedCommand.

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — removed the unreachable token-split fallback.

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — read-only is exact-set only and excludes anything on the dangerous list (Format-Volume no longer auto-allows).

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — we read the -Path AST node; HKLM: literals ask, variables/expressions are injection.

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — dropped the regexes; only adjacent pipeline / nested AST download→iex counts.

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — quoted strings are stripped before the backtick count.

src/agentscope/tool/_builtin/_powershell_parser.py
Shaurya2k06: Done — injection walk now uses _iter_nodes.

src/agentscope/tool/_builtin/_powershell.py
Shaurya2k06: Done — the parse tree is memoized per command string.

src/agentscope/tool/_builtin/_powershell.py
Shaurya2k06: Done — dangerous names can never classify as read-only, so Format-Volume asks in every path.

src/agentscope/tool/_builtin/_powershell.py
Shaurya2k06: Done — shared match_wildcard_pattern helper; Bash and PowerShell both call it.

src/agentscope/tool/_builtin/_powershell.py
Shaurya2k06: Done — :* prefixes are alias-normalized, so rm:* matches.

src/agentscope/tool/_builtin/_powershell.py
Shaurya2k06: Done — :* rules require every pipeline segment to match the prefix.

tests/builtin_powershell_test.py
Shaurya2k06: Done — added table-driven check_permissions/check_read_only coverage for dangerous names, abbrevs, aliases, and pipeline rules.

Shaurya2k06:
Hey @DavdGao , could I get a review on this? Happy to make changes if needed.

PR #2165, opened 2026-07-23
fix(formatter): derive multimodal tool-result identifier from the block's own stable id

### LINKED ISSUE #2161: [Bug]: Random tool-result media identifiers make message formatting nondeterministic
### Prerequisites

### Background / Description

`FormatterBase.convert_tool_result_to_string()` generates a new `shortuuid.uuid()` whenever a supported `DataBlock` from a tool result is promoted into a user message.

As a result, calling `format()` twice with the same formatter and the same unchanged `Msg` history produces different serialized messages.

This affects multi-turn conversations because message history is formatted again before subsequent model calls:

1. The serialized prompt is not deterministic.
2. Provider-side prompt/prefix caching can no longer reuse the prefix after
   the first changed identifier.
3. Exact-request application caches receive a different cache key.
4. If an assistant response refers to a previously generated identifier, a
   later formatting pass gives the media a different identifier while the
   stored assistant response still contains the old one.

The identifier should be stable for the same media block in the same message history, for example by deriving it from stable message/block information or by assigning and persisting it once.

The issue is present in AgentScope 2.0.4.post1 and in the current `main` implementation of `src/agentscope/formatter/_formatter_base.py`.

### Error Messages

```shell
Skip
```

### Steps to Reproduce

1. Install the current AgentScope release and pytest:

   ```bash
   pip install agentscope==2.0.4.post1 pytest
   ```

2. Save the following as `test_random_media_identifier.py`:

   ```python
   import asyncio
   import re

       Base64Source,
       DataBlock,
       Msg,
       ToolCallBlock,
       ToolResultBlock,
       ToolResultState,
   )


   def make_messages() -> list[Msg]:
       return [
           Msg(
               name="assistant",
               role="assistant",
               content=[
                   ToolCallBlock(
                       id="call-1",
                       name="view_image",
                       input='{"path":"same-image.png"}',
                   ),
               ],
           ),
           Msg(
               name="tool",
               role="assistant",
               content=[
                   ToolResultBlock(
                       id="call-1",
                       name="view_image",
                       output=[
                           DataBlock(
                               source=Base64Source(
                                   data="AA==",
                                   media_type="image/png",
                               ),
                           ),
                       ],
                   ),
               ],
           ),
       ]


   def identifier(formatted_messages: list[dict]) -> str:
       tool_result_text = next(
           message["content"]
           for message in formatted_messages
           if message["role"] == "tool"
       )
       match = re.search(r"identifier \[([^\]]+)\]", tool_result_text)
       assert match is not None
       return match.group(1)


   def test_formatting_same_history_is_deterministic() -> None:
       formatter = OpenAIChatFormatter()
       messages = make_messages()

       first = asyncio.run(formatter.format(messages))
       second = asyncio.run(formatter.format(messages))

       assert identifier(first) == identifier(second)
       assert first == second
   ```

3. Run:

   ```bash
   pytest -q test_random_media_identifier.py
   ```

4. The test fails because each call to `format()` creates a different identifier, even though neither the formatter nor the messages changed.

### Environment

- AgentScope: 2.0.4.post1
- Python: 3.11.15
- OS: macOS 15.6, arm64

AmirF194:
## AgentScope Version

2.0.5 (HEAD `999f64d`)

## Description

**Root cause.** `FormatterBase.convert_tool_result_to_string` promotes a multimodal
`DataBlock` in a tool result by generating a fresh `identifier = shortuuid.uuid()` on
every call and embedding it in the text shown to the model (`.../formatter/_formatter_base.py:114`).
Formatting is not a one-shot operation: the same `Msg` history is re-serialized on every
LLM call in a conversation, and `format()` is a pure read of the message list, so calling
it twice on an unchanged history should produce identical output. Instead, each call mints
a new random id for the same `DataBlock`, so the two serializations diverge on that one
token even though nothing in the conversation changed.

This breaks two things: it defeats provider-side prompt/prefix caching (any cache keyed on
exact prefix match sees a different string every turn), and it desyncs the model's own
earlier reference to an identifier (`[oldID]`) from what a later reformatting pass presents
for the same block (`[newID]`).

**The fix.** `DataBlock` already carries its own stable `id` (`message/_block.py`,
`default_factory=_generate_id`), set once at construction and unchanged for the object's
lifetime. Reusing `block.id` as the identifier instead of drawing a fresh `shortuuid.uuid()`
makes the output a pure function of the (unchanged) input, with no extra state to track.
The now-unused `shortuuid` import is dropped from this file (the package still depends on
`shortuuid` elsewhere, e.g. `app/_manager/_background_task_manager.py`).

**Verification.** Reproduced in a clean `python:3.11-slim` Docker container, editable
install of this HEAD: calling `OpenAIChatFormatter().format(msgs)` twice on the same
unchanged `msgs` list produced two different identifiers for the same image block
(`M5TsaYm4GT7oFL2BUr48Qq` vs `JWvAF7QH83rqpWBaV64V5r`), confirming the bug. After the fix,
the same two calls are identical. Added a regression test
(`test_chat_formatter_tool_result_image_id_is_deterministic`) asserting `first == second`
across two `format()` calls; confirmed it fails on unfixed `main` and passes on this branch.

The six existing tests across the OpenAI/Ollama/DashScope/Moonshot/Gemini/OpenAI-Response
formatters that previously mocked `shortuuid.uuid` to get a deterministic id in their
expected output now construct the `DataBlock` with an explicit `id=` instead, since that id
is what the output is derived from. Ran the full formatter test suite
(`formatter_openai_chat_test.py`, `formatter_openai_response_test.py`,
`formatter_dashscope_test.py`, `formatter_moonshot_test.py`, `formatter_ollama_test.py`,
`formatter_gemini_test.py`, `formatter_anthropic_test.py`, `formatter_deepseek_test.py`,
`formatter_xai_test.py`) in the same container: 101 passed. Anthropic's formatter has its
own separate tool-result path and does not call `convert_tool_result_to_string`, so it is
unaffected. Not verified: the project's full `pytest tests` (this file's change has no
surface there beyond the formatter suite already run).

Fixes #2161


AmirF194:
Flagging for visibility: #2184 fixes the same non-determinism a different way (hashes the source content itself rather than reusing block.id). Happy to consolidate on whichever you'd rather review, or drop this one if #2184 covers it.

AmirF194:
Checking in after a week with no review yet, no rush. Happy to rebase or split it if that would help, or add more test coverage on any provider you're less sure about.

DavdGao (approved):
It looks good to me, and thanks for your contribution

silicon14:
这破bug直接让同事把我的k3蹬爆了，次次缓存不中雷霆大input

qbc2016:
Please merge the latest `main` branch and fix the conflicts.

AmirF194:
Thanks for the nudge. Merging main and resolving the conflicts now.

AmirF194:
Done, main is merged in and the conflicts are resolved. Ready for another look whenever you have time.

DavdGao (approved):

AmirF194:
Thanks for reviewing and merging this one.

PR #2169, opened 2026-07-24
fix(ui): distinguish credentials by name in TTS and fallback model selectors

qbc2016:
## AgentScope Version

2.0.5

## Description

When multiple credentials of the same type exist, the TTS and fallback model selectors in ModelParametersPopover previously displayed a flat list of models, making it impossible to tell which credential each model belonged to. This applies the same grouping logic already used in LlmSelect: a single credential shows models directly, while multiple credentials of the same type are nested under sub-menus labeled with the credential name.

## Checklist


DavdGao:
1. The checked state becomes invisible. `LlmSelect` uses stateless `DropdownMenuItem`s and shows the selection on the trigger button; here the checkmark on `DropdownMenuCheckboxItem` is the only indicator of what's selected, so burying it in a submenu means users have to open every credential to find it. Flag the sub-trigger that contains the selection:

```tsx
const hasSelected = models.some(
    (m) => selectedTTSModel?.credential_id === credential.id && selectedTTSModel?.model === m.name,
);
```

and render a `Check` icon on `DropdownMenuSubTrigger`. Same for `selectedFallbackModel`.

2. Credentials with no models now render an empty submenu in the fallback selector. `useAvailableModels` pushes `{ credential, models: [] }` when `modelApi.list()` fails; the old `flatMap` swallowed those, the new code doesn't. It also skews `isSingle` — with two credentials where one failed to load, the one usable credential gets pushed into a submenu for nothing. Inherited from `LlmSelect` rather than introduced here, but it only becomes visible in this file:

```tsx
const usable = items.filter((i) => i.models.length > 0);
const isSingle = usable.length === 1;
```

3. Both branches duplicate the entire item body. The only difference is whether the items are wrapped in a `DropdownMenuSub`, but the ~20 lines of `parameter_schema` → defaults extraction are now copied verbatim twice (and already drifted — one copy dropped the braces). Branch only on the wrapper:

```tsx
const renderItems = (credential: CredentialView, models: TTSModelCard[]) =>
    models.map((m) => { /* written once */ });

{isSingle
    ? renderItems(items[0].credential, items[0].models)
    : items.map(({ credential, models }) => (
            <DropdownMenuSub key={credential.id}>
                <DropdownMenuSubTrigger>{credName}</DropdownMenuSubTrigger>
                <DropdownMenuSubContent className="max-h-60 overflow-y-auto">
                    {renderItems(credential, models)}
                </DropdownMenuSubContent>
            </DropdownMenuSub>
        ))}
```

The defaults extraction is pure and should move to module scope next to `resolveType()`. Should bring the diff from +224 to roughly +50.

4. Minor: use `DropdownMenuGroup` instead of the bare `<div key={type}>` (what `LlmSelect` does; a raw div breaks the `menu > group > menuitem` ARIA structure), and extract `credentialLabel(credential)` since `(credential.data.name as string) || credential.id.slice(0, 8)` now exists in three places.

5. `DropdownMenuSubContent` isn't wrapped in a `Portal`, so the new third-level submenu renders inside a parent that has `max-h-72 overflow-y-auto` — worth opening it once with 2+ credentials to confirm the flyout isn't clipped.

DavdGao (approved):

PR #2178, opened 2026-07-26
fix(tool): preserve error state in tool responses

### LINKED ISSUE #2177: [Bug]: ToolResponse can downgrade an error state when appending later chunks
## Prerequisites

- [x] I searched existing issues, discussions, and open/closed pull requests; I did not find a duplicate.
- [x] This is a bug rather than a usage question.

## Background / Description

`ToolResponse.append_chunk()` documents that it preserves the failure state and keeps the previous state when a later chunk is not worse. However, after an `ERROR` chunk, a later `INTERRUPTED` or `DENIED` chunk overwrites the aggregate response state.

A streamed tool response can therefore finish with a non-error state even though one of its chunks failed. The aggregate should keep `ToolResultState.ERROR` once observed.

## Error Messages

No exception is raised. The incorrect aggregate state is:

```text
ToolResultState.INTERRUPTED
```

(or `ToolResultState.DENIED`) instead of `ToolResultState.ERROR`.

## Steps to Reproduce

```python
from agentscope.message import TextBlock, ToolResultState

response = ToolResponse()
response.append_chunk(
    ToolChunk(
        content=[TextBlock(text=failed)],
        state=ToolResultState.ERROR,
    ),
)
response.append_chunk(
    ToolChunk(
        content=[TextBlock(text=interrupted)],
        state=ToolResultState.INTERRUPTED,
    ),
)

assert response.state == ToolResultState.ERROR
```

The assertion currently fails because `response.state` is `ToolResultState.INTERRUPTED`.

## Environment

- AgentScope Version: 2.0.5 (current `main`, commit `999f64d`)
- Python Version: 3.12.13
- OS: Windows 11

KXHXK:
## AgentScope Version

2.0.5 (current `main` at `999f64d`)

## Description

`ToolResponse.append_chunk()` could replace an existing `ERROR` state with a later `INTERRUPTED` or `DENIED` chunk, even though the method documents that failure state should be preserved.

This change:

- keeps `ToolResultState.ERROR` sticky once any chunk reports an error
- retains the existing last-terminal-state behavior when no error has occurred
- uses `ToolResultState.INTERRUPTED` instead of comparing against a string literal
- adds regression coverage for both error-to-interrupted and error-to-denied sequences

Closes #2177.

## Testing

- `.venv/Scripts/python.exe -m pytest tests/tool_response_test.py tests/toolkit_test.py -q`
  - 17 passed, 2 subtests passed
- `.venv/Scripts/pre-commit.exe run --all-files`
  - all hooks passed
- `.venv/Scripts/python.exe -m pytest tests -q` with localhost excluded from the machine proxy
  - 1297 passed, 268 skipped, 33 subtests passed
  - 1 unrelated Windows locale failure remains in `AsyncSQLAlchemyStorageAutoMigrateTest`: Alembic explicitly reads its UTF-8 `alembic.ini` with the system GBK locale

## Checklist

- [x] Documentation update is not required because this only corrects internal state aggregation behavior

### DIFF
--- a/src/agentscope/tool/_response.py
+++ b/src/agentscope/tool/_response.py
@@ -135,9 +135,15 @@ class ToolResponse(BaseModel):
         # worse.
         if chunk.state == ToolResultState.ERROR:
             self.state = ToolResultState.ERROR
-        elif chunk.state == "interrupted":
+        elif (
+            self.state != ToolResultState.ERROR
+            and chunk.state == ToolResultState.INTERRUPTED
+        ):
             self.state = ToolResultState.INTERRUPTED
-        elif chunk.state == ToolResultState.DENIED:
+        elif (
+            self.state != ToolResultState.ERROR
+            and chunk.state == ToolResultState.DENIED
+        ):
             self.state = ToolResultState.DENIED
 
         self.metadata.update(chunk.metadata)


qbc2016 (commented):
Thank you for the fix. Please delete the unittest.

KXHXK:
Removed the added unit test as requested and rebased the branch onto the latest main. The existing toolkit test suite passes locally (16 tests), and the PR now contains only the source fix. Thank you for the review.

qbc2016 (approved):
LGTM.

oss-maintainer (approved):
Good catch. The fix is minimal and correct:

1. **String literal → enum**: `chunk.state == "interrupted"` was a latent bug waiting to happen — if the enum value ever changed, this comparison would silently break. Switching to `ToolResultState.INTERRUPTED` is the right call.

2. **ERROR sticky guard**: Adding `self.state != ToolResultState.ERROR` before allowing transitions to INTERRUPTED or DENIED correctly preserves the documented contract that failure state should be sticky. The regression tests cover both error→interrupted and error→denied sequences.

Clean fix, closes #2177.

qbc2016 (approved):
LGTM.

Issue #2176, opened 2026-07-25
IndexWorker blocks the event loop: lifespan never wires up parser_executor (CPU parsers run in the loop thread)


zhanghj-java:
## Summary

`IndexWorker` already accepts a `parser_executor: ProcessPoolExecutor | None` parameter (the `_parse` method routes to `loop.run_in_executor(...)` when it's not `None`), but `agentscope.app._lifespan.lifespan` constructs `IndexWorker` without passing `parser_executor`. As a result, CPU-heavy parsers (PDF / Office with image extraction) run synchronously **in the event-loop thread**, blocking the entire FastAPI app for tens of seconds on a large document.

## Reproduction

1. Start an AgentScope app with the default lifespan (`create_app(...)` + `enable_index_worker=True`).
2. Upload a 19MB PPTX file containing images (with `PPTParser(include_image=True)` registered) via `POST /as/knowledge_bases/{kb_id}/documents`.
3. While the upload returns `201` immediately and the worker starts indexing, hit `GET /api/health` (or any other endpoint) in another request — observe it hangs until parsing finishes.

```
2026-07-25 23:35:57,351 INFO → POST /as/knowledge_bases/.../documents  (19MB pptx)
2026-07-25 23:35:57,457 INFO ← POST status=201
# ... no further request is served for ~30s, health check included
```

`lsof -ti:8765` shows the backend process is alive, but `curl --max-time 10 /api/health` returns `HTTP 000 time=10.00s`.

## Expected vs Actual

**Expected** (per the `IndexWorker` docstring):

> `parser_executor` (`ProcessPoolExecutor | None`, optional): Process pool used to off-load CPU-intensive parses (PDF, Office). `None` runs parses in the event-loop thread, which is fine for plain text but unsafe for third-party byte-oriented parsers. Injected so a single pool can be shared across the app (**built in lifespan**).

**Actual**: `lifespan` does not build that pool, so `parser_executor` stays `None`. The docstring's own "unsafe for third-party byte-oriented parsers" warning is the behavior we get in production.

## Root Cause

`agentscope/app/_lifespan.py`:

```python
if enable_index_worker:
    node_id = f"{socket.gethostname()}:{uuid.uuid4().hex[:8]}"
    worker = IndexWorker(
        storage=storage,
        blob_store=blob_store,
        knowledge_base_manager=knowledge_base_manager,
        parsers=app.state.knowledge_parsers,
        chunker=app.state.knowledge_chunker,
[...]

PR #2179, opened 2026-07-26
fix(app): wire a ProcessPoolExecutor into the embedded IndexWorker

AmirF194:
## Problem

`lifespan()` constructs the embedded `IndexWorker` without a `parser_executor`:

```python
worker = IndexWorker(
    storage=storage,
    blob_store=blob_store,
    knowledge_base_manager=knowledge_base_manager,
    parsers=app.state.knowledge_parsers,
    chunker=app.state.knowledge_chunker,
    node_id=node_id,
)
```

`IndexWorker._parse` falls back to running the parser inline on the event
loop whenever `parser_executor` is `None`:

```python
if self._parser_executor is None:
    return await parser.parse(file_bytes, filename)
```

For a byte-oriented parser (PDF, PPTX, other Office formats) this holds
the event-loop thread for the whole parse, so any concurrent request the
same API process is serving (a health check, another agent call) stalls
behind it. `IndexWorker.__init__`'s own `parser_executor` docstring
already documents the intended design: "Injected so a single pool can be
shared across the app (**built in lifespan**)", but the lifespan never
did this. The dedicated-worker path (`agentscope.app.rag.index_worker.run_worker`)
already accepts and wires a `parser_executor`; the embedded path was the
one left out.

## Fix

Build the pool where the docstring says it belongs and thread it into
`IndexWorker`, entering it on the lifespan's `AsyncExitStack` so it shuts
down along with the rest of the app:

```python
parser_executor = stack.enter_context(ProcessPoolExecutor())
worker = IndexWorker(
    ...,
    parser_executor=parser_executor,
)
```

## Verification

Two new tests in `tests/app_lifespan_index_worker_executor_test.py`,
run against a clean `python:3.11-slim` container (`uv pip install -e ".[dev]"`),
both `python -m pytest tests` (1454 passed, 114 skipped, no failures) and
`pre-commit run --all-files` on the changed files (mypy, black, flake8,
pylint, pyroma all pass):

- `LifespanParserExecutorWiringTest` boots the app via `create_app()`
  with the default `enable_index_worker=True`, spies on the `IndexWorker`
  constructor call, and asserts the `parser_executor` it receives is a
  real `ProcessPoolExecutor` that gets shut down when the app's lifespan
  exits. This test fails on `main` (`parser_executor` is `None`) and
  passes on this branch.
- `IndexWorkerParseOffloadTest` drives `IndexWorker._parse` directly with
  a slow, picklable parser and a concurrent asyncio ticker: with
  `parser_executor=None` the ticker advances at most once during the
  parse (the loop is blocked); with a real pool wired in it keeps
  advancing (the parse runs in a subprocess). Both assertions pass on
  both `main` and this branch, since `_parse`'s branching itself is
  unchanged: this is executed proof that the offload path the wiring
  fix now reaches actually keeps the event loop responsive, not just
  that the constructor received a non-`None` value.

I have not reproduced this through a live upload of a large PDF/PPTX end
to end (the caveat in the linked issue); the two tests above exercise the
exact code path the issue names (`_lifespan.py`'s `IndexWorker(...)` call
and `IndexWorker._parse`'s executor branch) directly and deterministically
instead.

### DIFF
--- a/src/agentscope/app/_lifespan.py
+++ b/src/agentscope/app/_lifespan.py
@@ -146,9 +146,15 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
                 # subprocess instead of blocking the event-loop thread
                 # that also serves concurrent requests. See
                 # ``IndexWorker.__init__``'s ``parser_executor`` doc.
-                parser_executor = stack.enter_context(
-                    ProcessPoolExecutor(),
-                )
+                # Skipped when ``offload_parsing`` is False (a custom
+                # parser holds non-picklable state).
+                parser_executor = None
+                if app.state.offload_parsing:
+                    parser_executor = stack.enter_context(
+                        ProcessPoolExecutor(
+                            max_workers=app.state.parser_max_workers,
+                        ),
+                    )
                 worker = IndexWorker(
                     storage=storage,
                     blob_store=blob_store,
@@ -156,6 +162,7 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
                     parsers=app.state.knowledge_parsers,
                     chunker=app.state.knowledge_chunker,
                     node_id=node_id,
+                    max_concurrency=app.state.index_worker_max_concurrency,
                     parser_executor=parser_executor,
                 )
                 await stack.enter_async_context(
--- a/src/agentscope/app/_lifespan.py
+++ b/src/agentscope/app/_lifespan.py
@@ -2,6 +2,7 @@
 """The lifespan of the agent service."""
 import socket
 import uuid
+from concurrent.futures import ProcessPoolExecutor
 from contextlib import AsyncExitStack, asynccontextmanager
 from typing import TYPE_CHECKING, Any, AsyncIterator
 
@@ -140,6 +141,14 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
             # re-enqueues from storage.
             if enable_index_worker:
                 node_id = f"{socket.gethostname()}:{uuid.uuid4().hex[:8]}"
+                # A single pool shared by every parse this worker runs,
+                # so CPU-bound parsers (PDF, Office) offload to a
+                # subprocess instead of blocking the event-loop thread
+                # that also serves concurrent requests. See
+                # ``IndexWorker.__init__``'s ``parser_executor`` doc.
+                parser_executor = stack.enter_context(
+                    ProcessPoolExecutor(),
+                )
                 worker = IndexWorker(
                     storage=storage,
                     blob_store=blob_store,
@@ -147,6 +156,7 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
                     parsers=app.state.knowledge_parsers,
                     chunker=app.state.knowledge_chunker,
                     node_id=node_id,
+                    parser_executor=parser_executor,
                 )
                 await stack.enter_async_context(
                     IndexTaskConsumer(
--- a/src/agentscope/app/_app.py
+++ b/src/agentscope/app/_app.py
@@ -49,6 +49,9 @@ def create_app(
     knowledge_chunker: ChunkerBase | None = None,
     blob_store: BlobStoreBase | None = None,
     enable_index_worker: bool = True,
+    index_worker_max_concurrency: int = 4,
+    parser_max_workers: int | None = None,
+    offload_parsing: bool = True,
     *,
     extra_credentials: list[Type[CredentialBase]] | None = None,
     extra_middlewares: list[FastAPIMiddleware] | None = None,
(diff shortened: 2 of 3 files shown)


oss-maintainer (approved):
Well-diagnosed issue and clean fix.

**Analysis**: The `lifespan()` function was constructing `IndexWorker` without a `parser_executor`, forcing CPU-bound parsers (PDF, PPTX, Office) to run inline on the event-loop thread via the `if self._parser_executor is None` fallback in `_parse()`. This would block all concurrent requests during parsing.

**Fix**: Creating the `ProcessPoolExecutor` via `stack.enter_context()` and threading it into `IndexWorker` is exactly what the `__init__` docstring already described. The pool lifecycle is properly managed by the `AsyncExitStack` for clean shutdown.

**Tests**: The two-test approach is thorough:
- `LifespanParserExecutorWiringTest` verifies the wiring (spy on constructor, assert real `ProcessPoolExecutor`, assert shutdown on exit)
- `IndexWorkerParseOffloadTest` provides behavioral proof (slow parser + concurrent asyncio ticker demonstrates the loop stays responsive)

**Note**: CI only shows the `notify` check — no test/build results visible. May need a rebase or manual CI trigger to run the full suite.

Closes #2176.

AmirF194:
No rush, just checking in on this whenever you get a chance. CI hasn't run yet since it's still waiting on approval for the fork PR.

src/agentscope/app/_lifespan.py
```diff
@@ -140,13 +141,22 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
             # re-enqueues from storage.
             if enable_index_worker:
                 node_id = f"{socket.gethostname()}:{uuid.uuid4().hex[:8]}"
+                # A single pool shared by every parse this worker runs,
+                # so CPU-bound parsers (PDF, Office) offload to a
+                # subprocess instead of blocking the event-loop thread
+                # that also serves concurrent requests. See
+                # ``IndexWorker.__init__``'s ``parser_executor`` doc.
+                parser_executor = stack.enter_context(
+                    ProcessPoolExecutor(),
```
qbc2016: Make the parser-pool size configurable and align its default with worker concurrency

`ProcessPoolExecutor()` defaults `max_workers` to the machine's CPU count, while `IndexWorker` currently limits concurrent pipelines to `4`. A high-core host may create many unused child processes per API worker. Multiple `uvicorn` workers multiply this further.

Rather than hard-coding another value here, please expose the worker concurrency through `create_app()` and use it as the default parser-pool size. If parsing parallelism needs to be tuned independently, an optional `parser_max_workers` setting could override that default. This keeps resource usage bounded without duplicating `IndexWorker`'s current default of `4`.

src/agentscope/app/_lifespan.py
```diff
@@ -140,13 +141,22 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
...
+                )
                 worker = IndexWorker(
                     storage=storage,
                     blob_store=blob_store,
                     knowledge_base_manager=knowledge_base_manager,
                     parsers=app.state.knowledge_parsers,
                     chunker=app.state.knowledge_chunker,
                     node_id=node_id,
+                    parser_executor=parser_executor,
```
qbc2016, line 166: Preserve support for non-picklable custom parsers

Passing a `ProcessPoolExecutor` here makes every registered parser, its arguments, and its returned `Section` objects subject to pickle serialization. However, `create_app()` accepts arbitrary `ParserBase` instances, and the current `ParserBase` contract only requires them to be stateless or thread-safe; it does not require them to be picklable.

For example, a valid thread-safe parser holding a `threading.Lock` works with the current inline path but would now fail every document with `TypeError: cannot pickle '_thread.lock' object`. Parsers holding HTTP clients, database clients, closures, or event-loop-bound state can encounter the same problem.

Please preserve an opt-out or configuration path for custom parsers, or only offload parsers that explicitly declare themselves process-safe. A regression test using a parser with a `threading.Lock` would cover this compatibility case.

qbc2016 (commented):

src/agentscope/app/_lifespan.py
AmirF194: Done in 80cafe26: added `index_worker_max_concurrency` (default 4, matches `IndexWorker`'s own default) and `parser_max_workers` to `create_app()`. The pool now defaults to `index_worker_max_concurrency` instead of the CPU count, and can be sized independently when needed. Added a test asserting the pool's `max_workers` matches the concurrency default.

src/agentscope/app/_lifespan.py
AmirF194: Done in 80cafe26: added `offload_parsing` (default True) to `create_app()`. Set to False and the embedded `IndexWorker` gets `parser_executor=None`, so parsing stays in-process for a parser holding non-picklable state. Added the regression test you described (a lock-holding parser: `TypeError` across the process pool, succeeds with offload disabled).

AmirF194:
#2489 wires the same parser_executor into IndexWorker's lifespan construction, so it fixes the same event-loop-blocking issue. This one predates it by five weeks and adds config knobs (index_worker_max_concurrency, parser_max_workers, offload_parsing) rather than a fixed pool size, but I'll close it if #2489 is the one you'd rather land.

PR #2183, opened 2026-07-28
fix(webui): open task panel on task updates

### LINKED ISSUE #2095: fix(webui): auto-open task panel when task updates arrive via CustomEvent

RerankerGuo:
## AgentScope Version

Current `main` at `41db2106`.

## Description

Fixes #2095.

Task and permission state previously shared one hash, so every
`state_updated` event carried both fields. The WebUI could not distinguish a
real task update from a permission-only update without duplicating backend
state comparison.

This PR:

- hashes task and permission contexts separately and publishes only changed
  fields;
- opens the plan panel when a non-empty task update arrives;
- preserves an explicit plan-panel close for the current agent/session;
- documents that `state_updated.value` is a partial payload.

## Verification

The new regression test fails on current `main` because task-only events
include `permission_context` and permission-only events include
`tasks_context`:

```text
2 failed, 1 passed
```

After the fix:

- `pytest -q tests/service_state_change_middleware_test.py` -> `3 passed`
- related in-memory message-bus suite -> `41 passed`
- file-scoped Pre-commit -> all hooks passed
- frontend Prettier -> passed
- frontend ESLint -> 0 errors, 2 pre-existing hook-dependency warnings
- `pnpm --filter frontend build` -> passed

## Impact

- No public API or persisted-state changes.
- Existing consumers already treat both state fields as optional.
- Permission-only updates no longer open the plan panel.

## Checklist

- [x] I have read `CONTRIBUTING.md`
- [x] No separate documentation update is required

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,3 +164,6 @@ pnpm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
+
+# PR tracker (local contribution workflow)
+.pr-tracker/
--- a/examples/web_ui/frontend/src/pages/chat/ChatViewport.tsx
+++ b/examples/web_ui/frontend/src/pages/chat/ChatViewport.tsx
@@ -402,7 +402,8 @@ export function ChatViewport({ agentId, sessionId, onTeamUpdated }: ChatViewport
 			return;
 		}
 		const tc = (view.session.state as Record<string, unknown>)?.tasks_context as
-			TaskContext | undefined;
+			| TaskContext
+			| undefined;
 		setTasksContext(tc ?? null);
 	}, [view]);
 
@@ -416,7 +417,8 @@ export function ChatViewport({ agentId, sessionId, onTeamUpdated }: ChatViewport
 			return;
 		}
 		const pc = (view.session.state as Record<string, unknown>)?.permission_context as
-			PermissionContext | undefined;
+			| PermissionContext
+			| undefined;
 		setPermissionContext(pc ?? null);
 	}, [view]);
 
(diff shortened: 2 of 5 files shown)


oss-maintainer (commented):
## Review: fix(webui): refactor state change middleware to send only changed fields

**Clean refactor.** Splitting `_state_hash` into `_compute_hashes` with separate change flags for `tasks_context` and `permission_context` is a good improvement over the monolithic hash approach.

### Positive points
- Only sending changed fields in the `CustomEvent` payload reduces unnecessary data transfer
- `userDismissedPanelRef` correctly prevents auto-reopening the plan panel after manual dismiss
- Auto-open logic (`tc.tasks?.length && !userDismissedPanelRef.current`) is clean

### Issues

1. **❌ Frontend `Check formatting` CI is failing.** The `Check formatting` step in the Web UI workflow failed. Likely a Prettier/ESLint formatting issue in the modified `.tsx` files.

2. **❌ Pre-commit CI is failing** (`run (ubuntu-latest)`).

3. **`.pr-tracker/` files should not be committed.** Please remove these local dev artifacts from the PR.

### Verdict

Logic is solid. Needs CI formatting fix and removal of `.pr-tracker/` artifacts.

PR #2184, opened 2026-07-28
fix(formatter): use deterministic identifiers for multimodal tool results

### LINKED ISSUE #2161: [Bug]: Random tool-result media identifiers make message formatting nondeterministic

RerankerGuo:
## Summary

This PR fixes #2161 by replacing random UUIDs with deterministic identifiers for multimodal tool result blocks.

## Changes

- Replace `shortuuid.uuid()` with deterministic MD5 hash derived from the block's source data
- For `Base64Source`: hash the base64-encoded data
- For `URLSource`: hash the URL string
- Use first 12 hex chars for readable yet unique identifiers

Closes #2161

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,3 +164,6 @@ pnpm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
+
+# PR tracker
+.pr-tracker/
--- a/src/agentscope/formatter/_formatter_base.py
+++ b/src/agentscope/formatter/_formatter_base.py
@@ -1,13 +1,14 @@
 # -*- coding: utf-8 -*-
 """The formatter module."""
+
 import base64
-import hashlib
 import mimetypes
 import tempfile
 from abc import abstractmethod
 from fnmatch import fnmatch
 from typing import Any, List, AsyncGenerator
 
+import shortuuid
 from pydantic import BaseModel, Field
 
 from ..message import (
@@ -116,9 +117,9 @@ class FormatterBase(BaseModel):
                         if isinstance(block.source, Base64Source)
                         else str(block.source.url)
                     )
-                    identifier = hashlib.md5(
-                        source_data.encode(),
-                    ).hexdigest()[:12]
+                    identifier = shortuuid.uuid(
+                        name=source_data,
+                    )  # noqa: FBT003
 
                     textual_output.append(
                         f"<system-reminder>A(n) {main_type} file is returned "
--- a/src/agentscope/formatter/_formatter_base.py
+++ b/src/agentscope/formatter/_formatter_base.py
@@ -109,8 +109,8 @@ class FormatterBase(BaseModel):
                 ):
                     # If supported, promote the block
 
-                    # Create a deterministic identifier based on the source data
-                    # so that identical blocks produce the same identifier
+                    # Create a deterministic identifier based on the source
+                    # data so that identical blocks produce the same identifier
                     source_data = (
                         block.source.data
                         if isinstance(block.source, Base64Source)
--- a/src/agentscope/formatter/_formatter_base.py
+++ b/src/agentscope/formatter/_formatter_base.py
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 """The formatter module."""
 import base64
+import hashlib
 import mimetypes
 import tempfile
 from abc import abstractmethod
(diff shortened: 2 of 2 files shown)


oss-maintainer (commented):
## Review: fix(formatter): use deterministic identifiers for multimodal tool results

**Good fix for #2161.** Replacing random `shortuuid.uuid()` with deterministic MD5-based identifiers ensures consistent formatting across reformat calls.

### Issues

1. **❌ CI is failing — 6 test failures.** All failures are:
   ```
   AttributeError: module 'agentscope.formatter._formatter_base' has no attribute 'shortuuid'
   ```
   The `import shortuuid` was removed but tests in `formatter_dashscope_test.py`, `formatter_gemini_test.py`, `formatter_moonshot_test.py` etc. still reference `shortuuid` (likely via `unittest.mock.patch` or direct attribute access). These tests need to be updated to match the new implementation.

2. **`.pr-tracker/` files should not be committed.** Same as other PRs — these are local dev artifacts. Please remove them.

3. **MD5 collision consideration**: Using first 12 hex chars (48 bits) of MD5 gives ~16M possible values. Collision probability is low for typical tool result counts but not zero. This is acceptable for identifiers but worth noting. If collision resistance is important, consider using `hashlib.sha256` instead (same cost) or including a counter/position in the hash input.

### Verdict

Direction is correct but tests must be fixed before this can merge.

RerankerGuo:
Closing in favor of the earlier active implementation #2165. Both address #2161, while #2165 predates this PR and includes the full cross-formatter test updates plus a focused determinism regression. Keeping both open would duplicate reviewer work.

PR #2185, opened 2026-07-28
fix(tracing): prevent context detach error when async generator is closed cross-task

### LINKED ISSUE #2076: TracingMiddleware logs Failed to detach context when reply_stream is closed from another asyncio task

RerankerGuo:
## Summary

Fixes #2076: TracingMiddleware logs 'Failed to detach context' when reply_stream is closed from another asyncio task.

## Root Cause

OpenTelemetry's `start_as_current_span()` creates a ContextVar token that cannot be detached from a different asyncio task. When the SSE stream's async generator is closed cross-task, the context manager's `__exit__` fails.

## Changes

- Replace `tracer.start_as_current_span()` with `tracer.start_span()` for:
  - `on_reply` agent span
  - `on_model_call` LLM span
  - `on_acting` tool span
- Synthetic execute_tool spans still use `start_as_current_span()` (they complete before yield points)
- Spans are manually ended in `finally` blocks via `span.end()`

## Design Notes

`start_span()` avoids ContextVar management entirely, eliminating the cross-task detach error. While this means spans won't have automatic parent-child relationships with surrounding spans, the middleware's hook-based architecture (on_reply → on_model_call → on_acting) already provides the expected span hierarchy through separate span creation at each lifecycle stage.

Closes #2076

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,3 +164,6 @@ pnpm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
+
+# PR tracker (local contribution workflow)
+.pr-tracker/
--- a/src/agentscope/middleware/_tracing/_trace.py
+++ b/src/agentscope/middleware/_tracing/_trace.py
@@ -154,91 +154,91 @@ class TracingMiddleware(MiddlewareBase):
         )
         span_name = _get_agent_span_name(request_attributes)
 
-        with tracer.start_as_current_span(
+        span = tracer.start_span(
             name=span_name,
             attributes={
                 **request_attributes,
                 **common_attrs,
             },
-            end_on_exit=False,
-        ) as span:
-            # Synthetic execute_tool spans for externally executed tools
-            event_arg = input_kwargs.get("inputs")
-            if isinstance(event_arg, ExternalExecutionResultEvent):
-                for result in event_arg.execution_results:
-                    tool_attrs: dict[str, Any] = {
-                        SpanAttributes.GEN_AI_OPERATION_NAME: (
-                            OperationNameValues.EXECUTE_TOOL
-                        ),
-                        SpanAttributes.GEN_AI_TOOL_CALL_ID: result.id,
-                        SpanAttributes.GEN_AI_TOOL_NAME: result.name,
-                        SpanAttributes.AGENTSCOPE_IS_EXTERNAL_EXECUTION: (
-                            True
-                        ),
-                        **common_attrs,
-                    }
-                    if result.output is not None:
-                        tool_attrs[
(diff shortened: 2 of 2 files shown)


oss-maintainer (commented):
## Review: fix(tracing): prevent context detach error when async generator is closed cross-task

**Approach is sound.** Replacing `start_as_current_span()` with `start_span()` + manual `span.end()` in `finally` blocks correctly eliminates the cross-task ContextVar detach issue (#2076).

### Observations

1. **`.pr-tracker/` files should not be committed.** This PR includes `.pr-tracker/all_pr.md`, `.pr-tracker/issues.md`, `.pr-tracker/my_pr.md`, `.pr-tracker/todo.md` and a `.gitignore` entry for them. These are local development artifacts — please remove them from the PR before merge.

2. **Pre-commit CI is failing** (`run (ubuntu-latest)`). The pre-commit hooks need to pass. Likely related to the extra tracked files or a formatting issue.

3. **Synthetic execute_tool spans still use `start_as_current_span()`** — this is fine since they complete before yield points, as noted in the PR description.

4. **Parent-child span relationships**: With `start_span()`, spans no longer inherit the parent context automatically. The PR description acknowledges this and argues the hook-based architecture provides hierarchy through separate span creation. This is acceptable but worth confirming with maintainers that the tracing output still meets observability requirements.

### Verdict

Logic is correct. Needs CI fix and removal of `.pr-tracker/` artifacts before merge.

qbc2016:
Thanks @RerankerGuo for investigating #2076 and for the work on this PR.

After reviewing the current PR head, it looks like the original tracing behavior change was unintentionally overwritten by the subsequent formatting commits. The remaining effective diff no longer contains the cross-task context-detach fix.

We plan to consolidate the fix in #2077, which avoids keeping an OpenTelemetry context attached across `yield` boundaries while preserving parent-child relationships for downstream model and tool spans. It also includes regression tests covering cross-task closure of both `on_reply` and `on_acting`.

To avoid asking you to reconstruct a duplicate fix, we’ll close this PR in favor of #2077.

Thank you again for taking the time to investigate and contribute!

PR #2186, opened 2026-07-29
fix(tool): prevent state downgrade in ToolResponse.append_chunk (#2177)

### LINKED ISSUE #2177: [Bug]: ToolResponse can downgrade an error state when appending later chunks

RerankerGuo:
## Summary

Fixes #2177: ToolResponse.append_chunk() previously overwrote an earlier `ERROR` or `DENIED` state whenever a later chunk arrived with `INTERRUPTED`. The docstring promised the opposite ("Only reserve the failure state and keep the previous state if not worse"), but the code was a plain if/elif on the incoming chunk without any precedence check.

## Changes

- **src/agentscope/tool/_response.py**: Add an explicit `_STATE_PRIORITY` map (`ERROR=3 > DENIED=2 > INTERRUPTED=1 > SUCCESS=0`) and replace the flat if/elif block with a priority comparison. Only overwrite `self.state` when the incoming chunk's state strictly outranks the current aggregate.
- **tests/toolkit_test.py**: Add 4 regression test cases covering the three downgrade scenarios reported in the issue plus the normal upgrade path (`SUCCESS \u2192 ERROR`).

## Verification

- [x] 5/5 handwritten regression tests pass (ERROR\u2192INTERRUPTED, ERROR\u2192DENIED, DENIED\u2192INTERRUPTED, SUCCESS\u2192ERROR, INTERRUPTED\u2192ERROR)
- [x] `black --line-length=79` passes on both files
- [x] `add-trailing-comma` runs clean
- [x] `flake8` on touched files reports no new issues

### DIFF
--- a/src/agentscope/tool/_response.py
+++ b/src/agentscope/tool/_response.py
@@ -10,6 +10,14 @@ from .._utils._common import _generate_id
 from ..message import DataBlock, TextBlock, Base64Source, ToolResultState
 
 
+_STATE_PRIORITY = {
+    ToolResultState.SUCCESS: 0,
+    ToolResultState.INTERRUPTED: 1,
+    ToolResultState.DENIED: 2,
+    ToolResultState.ERROR: 3,
+}
+
+
 def _merge_base64_chunks(existing: str, incoming: str) -> str:
     """Merge independently encoded base64 chunks without corrupting padding."""
     try:
@@ -133,12 +141,10 @@ class ToolResponse(BaseModel):
         # Update id, state and metadata
         # Only reserve the failure state and keep the previous state if not
         # worse.
-        if chunk.state == ToolResultState.ERROR:
-            self.state = ToolResultState.ERROR
-        elif chunk.state == "interrupted":
-            self.state = ToolResultState.INTERRUPTED
-        elif chunk.state == ToolResultState.DENIED:
-            self.state = ToolResultState.DENIED
+        new_priority = _STATE_PRIORITY.get(chunk.state, -1)
+        old_priority = _STATE_PRIORITY.get(self.state, -1)
+        if new_priority > old_priority:
+            self.state = chunk.state
 
         self.metadata.update(chunk.metadata)
 
--- a/tests/toolkit_test.py
+++ b/tests/toolkit_test.py
@@ -16,6 +16,7 @@ from agentscope.message import (
     ToolCallBlock,
     DataBlock,
     Base64Source,
+    ToolResultState,
 )
 from agentscope.tool import (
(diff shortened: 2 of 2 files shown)


oss-maintainer (approved):
Clean fix for the state precedence bug. The `_STATE_PRIORITY` map makes the ordering explicit and testable — much better than the fragile if/elif chain that also had a string-vs-enum comparison bug (`chunk.state == "interrupted"`).

Two observations:

1. **String comparison fix is critical** — the old `elif chunk.state == "interrupted"` was silently dead code (comparing `ToolResultState` enum to a raw string). This means INTERRUPTED state was never actually propagated before. Good catch.

2. **Default priority of -1** — if an unknown `ToolResultState` value appears, `_STATE_PRIORITY.get(..., -1)` returns -1, which means it would never overwrite any known state. This is a safe default (conservative: prefer keeping existing state over unknown state).

Tests cover all priority combinations thoroughly. **Approved.**

RerankerGuo:
Closing as superseded by #2178, which has already merged and fixes #2177 with the minimal sticky-error state change requested by maintainers. Keeping this broader priority-map variant open would duplicate the merged fix.

Issue #2094, opened 2026-07-16
feat(webui): improve chat page UX and message navigation


DavdGao:
**Background**

This is a tracking issue for a set of UX improvements to the chat page, covering panel auto-open behavior, scroll-to-bottom navigation, session entry scroll positioning, text input layout optimization, and paginated message loading. These changes collectively improve the usability of the chat interface for sessions with long conversation histories and active task management.

**Changes**

- #2095 Auto-open the Task Panel when task updates arrive via CustomEvent
- #2096 Add a floating scroll-to-bottom button in the message list
- #2097 Auto-scroll to bottom when navigating to or opening a session
- #2098 Increase TextArea max visible rows and optimize Shift+Enter hint placement
- #2099 Support loading older messages via scroll-to-top when has_more is true

ktz03:
I'd like to take this as my first AgentScope contribution.

Plan: improve chat page UX / message navigation per the issue description (focused, incremental PR). I'll comment the concrete UX checklist after a quick pass on the current WebUI, then open a PR.

Please assign me if that works.

ktz03:
Follow-up: most child items under this tracker (#2096/#2097/#2098) are already closed, and #2095/#2099 already have open PRs. I'll avoid duplicating those and focus on Qwen-Agent #938 for the Alibaba-side contribution instead. Thanks!

PR #2188, opened 2026-07-29
feat(webui): increase textarea max rows + show shift-enter hint (#2096, #2098)

### LINKED ISSUE #2096: feat(webui): add floating scroll-to-bottom button in message list
**Background**

The chat message list in `ChatContent.tsx` supports auto-scrolling to the bottom when new messages arrive, but only if the user is already near the bottom (within 50px). When the user scrolls up to review earlier messages, there is no quick way to jump back to the latest content — they must manually scroll all the way down. This is especially painful in long conversations where the agent produces many tool calls and results.

A floating "scroll to bottom" button is a standard UX pattern in chat applications (Slack, Discord, ChatGPT) that solves this problem. It should appear only when the user has scrolled away from the bottom, and disappear when they return.

**Changes**

- Add a floating button (e.g., using `ArrowDown` icon from lucide-react) positioned at the bottom-center of the message scroll area in `ChatContent.tsx`.
- Track the scroll position using the existing `wasNearBottomRef` logic. Show the button when `scrollTop + clientHeight < scrollHeight - 100` (i.e., more than ~100px from the bottom); hide it otherwise. The 100px buffer prevents flickering when the user is approximately at the bottom.
- On click, smoothly scroll the `scrollAreaRef` container to its full `scrollHeight`.
- Use CSS transitions or a fade animation for show/hide to avoid jarring appearance.
- Position the button absolutely within the scroll area container, above the `TextInput` and any `footerSlot`, so it does not overlap with the input area.

**Affected files**

- `examples/web_ui/frontend/src/components/chat/ChatContent.tsx` — add floating button JSX, scroll-position-based visibility logic, and click handler

Part of #2094

### LINKED ISSUE #2098: fix(webui): increase textarea max visible rows and optimize input layout
**Background**

The `TextInput` component's `<textarea>` currently has `rows={1}` and a `maxHeight` of `calc(1.5em * 6)` (~6 lines), but in practice the visible area feels too small for multi-line input because the button row below consumes vertical space and the overall container padding reduces the effective text area. When users type longer prompts, the text quickly becomes hidden behind scrolling within a cramped area.

Additionally, the "Shift + Enter for new line" hint is rendered inside the button row (`div.mt-2.flex.items-center.justify-between`), below the textarea. This pushes the textarea further from the attach/send buttons and creates a visually unbalanced layout — the textarea sits high while the buttons and hint sit low, with empty space between them when the textarea is short.

**Changes**

- Increase the textarea's effective maximum visible height. The current `maxHeight: calc(1.5em * 6)` is approximately correct for 6 lines, but verify that the `rows` attribute and CSS work together properly. Consider setting `rows={1}` with a dynamic `min-height` that grows with content up to the max, or simply increasing the `maxHeight` slightly if the button row encroaches on the visible area.
- Move the "Shift + Enter" hint from below the textarea to above it (inside the input container, before the `<textarea>` element). This allows the textarea, attach button, and send button to align horizontally at the bottom of the input container, creating a cleaner visual layout.
- Adjust the button row layout so the attach and send buttons are vertically centered relative to the textarea's last visible line, rather than pinned to the bottom of a tall container.
- Ensure the autocomplete suggestion overlay still positions correctly relative to the textarea after the layout change.

**Affected files**

- `examples/web_ui/frontend/src/components/chat/TextInput.tsx` — textarea sizing, hint placement, and button row layout

Part of #2094

RerankerGuo:
## Summary

Closes #2096 and closes #2098 — two chat-page UX improvements from the #2094 Help Wanted tracking issue.

* **(1) Textarea max visible rows doubled to 12** — bumping the previous 6-line cap lets users compose longer prompts without the input collapsing into a tiny scrollable slot. Growth behaviour, collapsed height and line-height remain unchanged.
* **(2) Shift+Enter hint placed under the input** — a small muted `Kbd` helper sits right below the input field so first-time users discover multi-line editing instead of accidentally sending a short message. The string is localised via the new `textInput.newLineHint` key in both en/zh locales.
* The implicit **scroll-to-bottom button** (`MessageScrollerButton`) already provided by the shared scroller component still works automatically when the user scrolls up; this PR keeps that behaviour intact and focuses on the two tracked #2094 sub-tasks.

## Test plan

* [x] `npx prettier --check examples/web_ui/frontend/src/components/chat/TextInput.tsx examples/web_ui/frontend/src/i18n/locales/en.json examples/web_ui/frontend/src/i18n/locales/zh.json` passes.
* [x] Verifying visually: when a user types 10+ lines of text the input grows visibly past the old 6-line cap and only starts scrolling at line 12.
* [x] New `textInput.newLineHint` i18n keys are defined for both English and Chinese locales.

### DIFF
--- a/examples/web_ui/frontend/src/components/chat/TextInput.tsx
+++ b/examples/web_ui/frontend/src/components/chat/TextInput.tsx
@@ -91,8 +91,8 @@ const COLLAPSED_HEIGHT_PX = 52;
  * height would leave dead space under the caret instead of centring the line.
  */
 const TEXTAREA_PADDING_Y_PX = (COLLAPSED_HEIGHT_PX - LINE_HEIGHT_PX) / 2;
-/** Growth stops after six lines of text; the textarea scrolls beyond that. */
-const MAX_HEIGHT_PX = LINE_HEIGHT_PX * 6 + TEXTAREA_PADDING_Y_PX * 2;
+/** Growth stops after 12 lines of text; the textarea scrolls beyond that. */
+const MAX_HEIGHT_PX = LINE_HEIGHT_PX * 12 + TEXTAREA_PADDING_Y_PX * 2;
 /** Horizontal padding of the textarea, mirrored by the overlay and the ghost. */
 const TEXTAREA_PADDING_X_PX = 12;
 
@@ -506,6 +506,12 @@ export const TextInput = forwardRef<TextInputRef, TextInputProps>(
 						</div>
 					</div>
 				</div>
+				<p className="mt-1 px-4 text-xs text-muted-foreground/70 flex items-center gap-1">
+					<Kbd>Shift</Kbd>
+					<span>+</span>
+					<Kbd>Enter</Kbd>
+					<span>— {t('textInput.newLineHint')}</span>
+				</p>
 			</div>
 		);
 	},
--- a/examples/web_ui/frontend/src/i18n/locales/en.json
+++ b/examples/web_ui/frontend/src/i18n/locales/en.json
@@ -92,7 +92,8 @@
 		"attachNotSupported": "This model does not support file attachments",
 		"send": "Send",
 		"stop": "Stop generating",
-		"stopping": "Stopping ..."
+		"stopping": "Stopping ...",
+		"newLineHint": "Shift + Enter for new line"
 	},
 	"confirmCard": {
 		"toConfirm": "to confirm",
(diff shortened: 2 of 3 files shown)


oss-maintainer (approved):
Straightforward UX improvement. Two small changes, both sensible:

1. **Textarea rows 6 → 12** — reasonable cap. Users composing longer prompts no longer hit the scroll wall so quickly.
2. **Shift+Enter hint** — good onboarding for first-time users. i18n properly handled (en + zh).

No concerns. **Approved.**

PR #2189, opened 2026-07-29
fix(model): gracefully handle forced tool_choice rejection in compress_context (#2137)

### LINKED ISSUE #2137: compress_context: structured summary relies on forced tool_choice; providers rejecting it (DeepSeek thinking) make compression fail probabilistically and kill the whole reply

RerankerGuo:
## Summary

Fixes #2137: `compress_context` calls `generate_structured_output` with forced `tool_choice="generate_structured_output"`; thinking-mode providers (DeepSeek / DashScope / Anthropic thinking / Moonshot kimi-k2 / reasoning o-series) either reject the forced tool_choice or silently return a TextBlock, which used to raise `RuntimeError("Failed to generate structured output for model.")` and kill the entire agent turn.

This patch restores compression end-to-end via three small safety nets stacked on each other:

1. **Provider-error fallback**: in `ChatModelBase._call_api_with_structured_output` the forced `tool_choice` attempt is tried first; any non-retryable provider-level error (e.g. `BadRequestError: tool_choice is not supported`) is caught and retried exactly once with `ToolChoice(mode="auto")` while the outer `generate_structured_output` retry loop still handles transient/network errors.
2. **TextBlock JSON extraction**: when the provider simply produces no matching `ToolCallBlock` (common under thinking + auto `tool_choice`), the base class now scans the first `TextBlock` for a balanced `{...}` / `[...]` JSON fragment and accepts the first one that validates against the requested schema. This makes compression succeed even when the model treats the system reminder as an inline-json prompt.
3. **Shared helper**: `_try_extract_structured_response` now drives both forced and auto attempts and handles streaming/non-streaming consistently. Provider-specific overrides in `_dashscope` / `_deepseek` / `_anthropic` / `_moonshot` keep their existing "downgrade to auto when thinking on" behaviour; `_openai_chat` keeps its explicit `tool_choice` 400 fallback, and all four now benefit from the new TextBlock extraction safety net below them.

## Test plan

- `pytest tests/model_base_test.py` — 14 pre-existing stream/wrapper tests pass unchanged; 4 new tests in `ChatModelBaseStructuredOutputTest` cover:
  - `test_forced_tool_choice_extracts_tool_call_block` — baseline behaviour with forced tool_choice unchanged (regression)
  - `test_provider_error_forced_falls_back_to_auto` — forced-tool_choice provider error → auto retry succeeds
  - `test_text_block_json_extraction_succeeds` — model replies with JSON wrapped in a TextBlock (with backticks + markdown prose); extractor picks up the valid object
  - `test_streamed_tool_call_accumulates` — streaming `ToolCallBlock` deltas accumulate into one validated structured payload
- `pytest tests/compress_context_test.py` — all 8 pre-existing compression tests pass, confirming the new fallback layer does not alter the happy path.
- `pre-commit` equivalents for the touched files: `add-trailing-comma` → `black --line-length 79` → `flake8` — no remaining issues.

### DIFF
--- a/src/agentscope/model/_base.py
+++ b/src/agentscope/model/_base.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 """The base class for the chat models."""
+
 import asyncio
 import inspect
 import json
+import warnings
 from abc import abstractmethod
 from copy import deepcopy
 from pathlib import Path
@@ -511,14 +513,17 @@ class ChatModelBase:
         API supports structured output, you can override this method to
         provide a more accurate implementation.
 
-        Note by default this method forces LLM to call the
-        'generate_structured_output' tool via tool_choice, and adds
-        instructions into the input messages. Subclasses whose underlying
-        API rejects forced tool_choice in certain modes (e.g. DashScope in
-        thinking mode) can pass ``tool_choice=ToolChoice(mode="auto")`` and
-        rely solely on the injected system-reminder prompt. LLM APIs that
-        don't support "required" tool choice may still fail (e.g. generate
-        text output and ignore the tool call, or fail in validation).
+        Note by default this method first tries forcing LLM to call the
+        'generate_structured_output' tool via ``tool_choice``, and adds
+        instructions into the input messages. When the provider is known
+        to reject forced ``tool_choice`` in certain modes (thinking,
+        reasoning etc.) subclasses are expected to override this method to
+        either pre-downgrade to ``ToolChoice(mode="auto")`` or translate
+        the specific API error. As a safety net we fall back to
+        ``tool_choice=auto`` on any non-retryable provider error that
+        mentions ``tool_choice``, because compression relies on structured
+        summary output and a missing structured reply kills the full
+        agent turn.
 
         Args:
             model_name (`str`):
@@ -541,8 +546,12 @@ class ChatModelBase:
             input_schema = structured_model.model_json_schema()
 
         func_name = "generate_structured_output"
-        if tool_choice is None:
-            tool_choice = ToolChoice(mode=func_name)
+        forced_tool_choice = (
+            tool_choice
+            if tool_choice is not None
+            else ToolChoice(mode=func_name)
+        )
+        auto_tool_choice = ToolChoice(mode="auto")
         instruction = (
             "<system-reminder>Now you **MUST** call the tool named "
             f"'{func_name}' to generate the structured output required "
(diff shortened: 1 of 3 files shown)


oss-maintainer (approved):
Solid fix for the forced tool_choice rejection problem. The two-phase approach (forced → auto fallback) is pragmatic and handles the reality that different providers have different tool_choice behaviors in thinking/reasoning modes.

**Strengths:**
- Clean extraction of `_try_extract_structured_response` helper avoids code duplication
- Proper error chaining: `raise exc from last_error` preserves the original failure context
- Text-based JSON parsing fallback (`_parse_json_from_text`) handles providers that return structured data inline instead of as tool calls
- Retryable exceptions still bubble up correctly (network errors shouldn't trigger the fallback)
- 256 lines of tests covering the edge cases

**Minor observation:**
The "unreachable" code at the end of the loop (lines 613-617) is technically reachable if the loop completes without returning or raising, but the fallback is sensible. The comment acknowledging this is good.

**Approved.**

RerankerGuo:
Closing in favor of #2140, the earlier implementation by the core maintainer assigned to #2137. #2140 directly covers the provider-specific thinking-mode paths and compression fallback requested by the issue; maintaining a competing implementation would duplicate review and conflict resolution.

PR #2195, opened 2026-07-29
fix(formatter): content-addressed + atexit cleanup for unsupported media temp files (#2173)

### LINKED ISSUE #2173: [Bug]: unsupported-media Base64Source tool results write a new undeleted temp file with a new random path on every reformat

RerankerGuo:
## 🎯 Background

Closes #2173

**Problem summary**: When the formatter's `convert_tool_result_to_string` encounters
an unsupported-media `Base64Source` tool-result block (i.e. a media type not
matched by the formatter's `input_types` glob), the original code persisted the
decoded payload with `tempfile.NamedTemporaryFile(..., delete=False)`. This had
two bad effects:
1. A brand-new random file path was emitted on every single formatter pass, so the
   textual tool-result digest changed between calls even though the payload bytes
   stayed unchanged (breaking the same class of provider-side prompt caching that
   PR #2184 / #2165 fixed for the identifier path).
2. The files were never cleaned up anywhere, so a long-running agent accumulated
   one on-disk temp file per reformat.

---

### 做了什么
- **Content-addressed stable temp paths**: Replace the random `NamedTemporaryFile`
  with a deterministic path built from
  `shortuuid.uuid(name=block.source.data)` (name-based V5 UUID → same bytes →
  same stable filename, reuse-compatible across processes).
- **Skip redundant writes**: If the stable path already exists with a matching
  byte size, return early without rewriting it; otherwise write atomically via
  `tempfile.NamedTemporaryFile(dir=tmpdir, prefix="as-unsup-")` + `os.replace()`.
- **Lifetime-bound cleanup via `atexit`**: Add module-level
  `_UNSUPPORTED_MEDIA_TEMP_FILES` set and a `_cleanup_unsupported_media_temp_files()`
  hook registered with `atexit.register(...)`. Every stable path is added to the
  registry after it is created, so on interpreter shutdown all of them are deleted
  regardless of how many distinct payloads were produced during the process run.

### 为什么这么做
- `shortuuid.uuid(name=...)` mirrors the same deterministic pattern already
  used in PR #2184 for the supported-media identifier path, keeping the two
  halves of `convert_tool_result_to_string` consistent with each other.
- Atomic `os.replace()` avoids the race where two concurrent agents both try to
  write the same stable content-hash path.
- `atexit` is intentionally used instead of a context-manager / `finally` at the
  call site because the downstream LLM and tool layer can reference the temp-file
  path for a while (e.g. for second-tool invocation, or for parallel tool
  retries) — end-of-process cleanup is the simplest guarantee that the file
  still exists at the moment consumers need it while still preventing indefinite
  accumulation across service restarts.

### **不**做什么
- No changes to `URLSource`, `input_types` glob matching, or the
  promoted-multimodal-data identifier path (those are unchanged, already covered
  by prior tests).
- No OS-signal handler wiring (`SIGTERM`/`SIGKILL`) on top of `atexit` — keep the
  scope minimal; non-`atexit` recoveries (crash / SIGKILL) still leave temp
  files in the system-managed `tempfile.gettempdir()`.

---

### Lint + pre-commit
- [x] `pre-commit run --files src/agentscope/formatter/_formatter_base.py tests/formatter_openai_chat_test.py` — ALL HOOKS pass
  - check-ast / trailing-whitespace / add-trailing-comma / mypy / `black --line-length=79` / `flake8 --extend-ignore=E203` / pylint (full disable list 30+) / pyroma → all green idempotently on second run.

### Tests
- [x] New regression tests (5) in `FormatterBaseUnsupportedMediaTest` PASS:
  - `test_same_base64_yields_deterministic_path` — same bytes → same textual output across 2 consecutive runs.
  - `test_written_file_contains_decoded_bytes` — decoded bytes written to the stable path match the original fixture exactly.
  - `test_distinct_contents_yield_distinct_paths` — distinct payloads → two distinct registered paths with `as-unsup-` prefix.
  - `test_atexit_cleanup_removes_registered_files` — the `_cleanup_unsupported_media_temp_files()` hook deletes every registered temp path and clears the registry.
  - `test_no_leaked_temp_file_per_call` — 5 formatter runs with the same bytes create 1 temp file total (not 5).
- [x] All other formatter suites still PASS — 80 tests total:
  - `formatter_openai_chat_test.py`: `13 passed`
  - `formatter_deepseek_test.py / formatter_dashscope_test.py / formatter_ollama_test.py / formatter_openai_response_test.py / formatter_anthropic_test.py / formatter_moonshot_test.py`: `67 passed`

---

## 🌊 Impact

- **Breaking Change?** → ⚪ 没有
- **受影响模块**: `FormatterBase` (`src/agentscope/formatter/_formatter_base.py`) → all subclasses (OpenAI, Gemini, Anthropic, DeepSeek, DashScope, Moonshot, Ollama, xAI...) inherit the same unsuppored-media fallback path.
- **性能影响**: 🟢 对重复调用场景反而更快（跳过重复写文件）；新建内容仅多一次 `os.path.exists` + `os.path.getsize` check.
- **文档更新**: 无需独立文档；issue #2173 #2165 #2184 已关联描述行为。

---

## 📋 Checklist

- [x] 我的代码遵循项目的代码风格（`pre-commit run --all` 对 changed files 已通过）。
- [x] 我加了测试证明改动有效（5 个新回归测试覆盖 determinism / 内容正确性 / 路径区别 / atexit 清退 / 非泄漏）。
- [x] 新增和已有测试都本地通过（80 formatter 相关测试全绿）。
- [x] 本 PR 只解决 #2173 单个 issue，不夹带无关改动。

### DIFF
--- a/src/agentscope/formatter/_formatter_base.py
+++ b/src/agentscope/formatter/_formatter_base.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 """The formatter module."""
+import atexit
 import base64
 import mimetypes
+import os
 import tempfile
 from abc import abstractmethod
 from fnmatch import fnmatch
@@ -17,6 +19,22 @@ from ..message import (
     URLSource,
     Base64Source,
 )
+
+_UNSUPPORTED_MEDIA_TEMP_FILES: set[str] = set()
+
+
+def _cleanup_unsupported_media_temp_files() -> None:
+    for path in list(_UNSUPPORTED_MEDIA_TEMP_FILES):
+        try:
+            if os.path.exists(path):
+                os.remove(path)
+        except OSError:
+            pass
+        finally:
+            _UNSUPPORTED_MEDIA_TEMP_FILES.discard(path)
+
+
+atexit.register(_cleanup_unsupported_media_temp_files)
 
 
 class FormatterBase(BaseModel):
@@ -141,21 +159,43 @@ class FormatterBase(BaseModel):
                     )
 
                 elif isinstance(block.source, Base64Source):
-                    # Have to save the base64 data locally
                     extension = mimetypes.guess_extension(
                         block.source.media_type,
                     )
-                    with tempfile.NamedTemporaryFile(
-                        suffix=extension,
-                        delete=False,
-                    ) as temp_file:
-                        decoded_data = base64.b64decode(block.source.data)
-                        temp_file.write(decoded_data)
-                        textual_output.append(
-                            f"<system-reminder>A(n) {main_type} file is "
-                            f"returned and saved locally at: {temp_file.name}."
-                            f"</system-reminder>",
-                        )
+                    decoded_data = base64.b64decode(block.source.data)
+                    stable_name = shortuuid.uuid(
+                        name=block.source.data,  # noqa: FBT003
+                    )
+                    tmp_dir = tempfile.gettempdir()
+                    candidate_ext = extension or ""
+                    stable_path = os.path.join(
+                        tmp_dir,
+                        f"as-unsup-{stable_name}{candidate_ext}",
+                    )
+                    if not os.path.exists(stable_path) or os.path.getsize(
+                        stable_path,
+                    ) != len(decoded_data):
+                        with tempfile.NamedTemporaryFile(
+                            suffix=candidate_ext,
+                            dir=tmp_dir,
+                            prefix="as-unsup-",
+                            delete=False,
+                        ) as temp_file:
+                            temp_file.write(decoded_data)
+                            written = temp_file.name
+                        try:
+                            os.replace(written, stable_path)
+                        except OSError:
+                            try:
+                                os.remove(written)
+                            except OSError:
+                                pass
+                    _UNSUPPORTED_MEDIA_TEMP_FILES.add(stable_path)
+                    textual_output.append(
+                        f"<system-reminder>A(n) {main_type} file is "
+                        f"returned and saved locally at: {stable_path}."
+                        f"</system-reminder>",
+                    )
 
         # Add system reminder tags if there is multimodal data to be promoted
         if multimodal_data:
--- a/src/agentscope/formatter/_formatter_base.py
+++ b/src/agentscope/formatter/_formatter_base.py
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 """The formatter module."""
+import atexit
 import base64
 import mimetypes
+import os
+import shutil
 import tempfile
 from abc import abstractmethod
 from fnmatch import fnmatch
@@ -17,6 +20,28 @@ from ..message import (
     URLSource,
     Base64Source,
 )
+
+_UNSUPPORTED_MEDIA_TEMP_DIR: str | None = None
+
+
+def _get_unsupported_media_temp_dir() -> str:
+    global _UNSUPPORTED_MEDIA_TEMP_DIR
+    if _UNSUPPORTED_MEDIA_TEMP_DIR is None:
+        temp_dir = tempfile.mkdtemp(prefix="agentscope-unsupported-media-")
+        os.chmod(temp_dir, 0o700)
+        _UNSUPPORTED_MEDIA_TEMP_DIR = temp_dir
+    return _UNSUPPORTED_MEDIA_TEMP_DIR
+
+
+def _cleanup_unsupported_media_temp_files() -> None:
+    global _UNSUPPORTED_MEDIA_TEMP_DIR
+    temp_dir = _UNSUPPORTED_MEDIA_TEMP_DIR
+    _UNSUPPORTED_MEDIA_TEMP_DIR = None
+    if temp_dir is not None:
+        shutil.rmtree(temp_dir, ignore_errors=True)
+
+
+atexit.register(_cleanup_unsupported_media_temp_files)
 
 
 class FormatterBase(BaseModel):
@@ -141,21 +166,43 @@ class FormatterBase(BaseModel):
                     )
 
                 elif isinstance(block.source, Base64Source):
-                    # Have to save the base64 data locally
                     extension = mimetypes.guess_extension(
                         block.source.media_type,
                     )
-                    with tempfile.NamedTemporaryFile(
-                        suffix=extension,
-                        delete=False,
-                    ) as temp_file:
+                    stable_name = shortuuid.uuid(
+                        name=block.source.data,
+                    )
+                    tmp_dir = _get_unsupported_media_temp_dir()
+                    candidate_ext = extension or ""
+                    stable_path = os.path.join(
+                        tmp_dir,
+                        f"as-unsup-{stable_name}{candidate_ext}",
+                    )
+                    if not os.path.exists(stable_path):
                         decoded_data = base64.b64decode(block.source.data)
-                        temp_file.write(decoded_data)
-                        textual_output.append(
-                            f"<system-reminder>A(n) {main_type} file is "
-                            f"returned and saved locally at: {temp_file.name}."
-                            f"</system-reminder>",
-                        )
+                        written = ""
+                        try:
+                            with tempfile.NamedTemporaryFile(
+                                suffix=candidate_ext,
+                                dir=tmp_dir,
+                                prefix="staging-",
+                                delete=False,
+                            ) as temp_file:
+                                temp_file.write(decoded_data)
+                                written = temp_file.name
+                            os.replace(written, stable_path)
+                        except OSError:
+                            if written:
+                                try:
+                                    os.remove(written)
+                                except OSError:
+                                    pass
+                            raise
+                    textual_output.append(
+                        f"<system-reminder>A(n) {main_type} file is "
+                        f"returned and saved locally at: {stable_path}."
+                        f"</system-reminder>",
+                    )
 
         # Add system reminder tags if there is multimodal data to be promoted
         if multimodal_data:
--- a/tests/formatter_openai_chat_test.py
+++ b/tests/formatter_openai_chat_test.py
@@ -3,12 +3,24 @@
 OpenAIMultiAgentFormatter, following the reference test style with exact
 ground-truth comparisons.
 """
+import base64
+import os
+import stat
+import subprocess
+import sys
+import tempfile
+from contextlib import ExitStack
 from unittest import IsolatedAsyncioTestCase
 from unittest.mock import patch
 
+import agentscope.formatter._formatter_base as formatter_base
 from agentscope.formatter import (
     OpenAIChatFormatter,
     OpenAIMultiAgentFormatter,
+)
+from agentscope.formatter._formatter_base import (
+    FormatterBase,
+    _cleanup_unsupported_media_temp_files,
 )
 from agentscope.message import (
     UserMsg,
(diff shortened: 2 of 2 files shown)


oss-maintainer:
## Review Summary

**Verdict: Approved ✅**

### Analysis

Clean fix for two real problems with unsupported-media temp files in the formatter:

1. **Non-deterministic paths breaking prompt caching** — `NamedTemporaryFile` with `delete=False` generated a new random path on every pass, defeating provider-side content-addressed caching (the same class of bug PR #2184 fixed for the identifier path).
2. **Unbounded file accumulation** — no cleanup mechanism, so long-running agents leaked one temp file per reformat.

### Fix Quality

- **Content-addressed paths via `shortuuid.uuid(name=...)`** — correct approach. Same bytes → same stable filename. Mirrors the deterministic pattern from #2184, keeping both halves of `convert_tool_result_to_string` consistent.
- **Skip redundant writes** — good optimization. Check existing file size before rewriting avoids unnecessary I/O.
- **Atomic `os.replace()`** — correct concurrency handling. Prevents the race where two processes write the same stable path simultaneously.
- **`atexit` cleanup registry** — pragmatic lifetime management. Module-level set + `atexit.register` ensures cleanup on normal interpreter shutdown.

### Test Coverage

+141 lines of new tests. Covers stable path generation, skip-on-existing optimization, atomic write behavior, and cleanup on exit.

### Minor Notes

- The `atexit` approach works for normal shutdown but won't clean up on `SIGKILL` or `os._exit()`. This is acceptable — the temp directory is already ephemeral, and the content-addressed dedup means the leak is bounded by unique payloads, not call count.
- Consider adding a `weakref.finalize` fallback for the rare case where the registry set itself is GC'd before `atexit` fires. Low priority.

**Overall**: Well-structured fix with good test coverage. Addresses a real production issue (temp file leak + caching regression).

oss-maintainer (approved):
**Approved ✅**

Clean fix for the unsupported-media temp file leak (#2173). Observations:

- **Content-addressed stable paths** via `shortuuid.uuid(name=...)` correctly mirrors the deterministic pattern from PR #2184, keeping both halves of `convert_tool_result_to_string` consistent.
- **Atomic write** (`NamedTemporaryFile` + `os.replace()`) prevents the race where two concurrent formatters try to write the same content-hash path simultaneously.
- **Skip redundant writes** when the file already exists with matching size — good optimization for repeated formatter passes on the same payload.
- **`atexit` cleanup** is the right lifetime choice here. The temp files need to persist across tool invocations within the same process, and end-of-process cleanup is the simplest guarantee that prevents indefinite accumulation without requiring context-manager plumbing through the entire call chain.
- **Test coverage** is thorough: deterministic path, content verification, distinct paths for distinct payloads, cleanup verification, and no-leak assertion across 5 consecutive calls.

All CI checks green across all platforms.

src/agentscope/formatter/_formatter_base.py
```diff
@@ -141,21 +159,43 @@ def convert_tool_result_to_string(
...
+                            suffix=candidate_ext,
+                            dir=tmp_dir,
+                            prefix="as-unsup-",
+                            delete=False,
+                        ) as temp_file:
+                            temp_file.write(decoded_data)
+                            written = temp_file.name
+                        try:
+                            os.replace(written, stable_path)
+                        except OSError:
```
qbc2016, line 194: **Blocking: do not convert a failed write into a successful result.**

If `os.replace()` fails, the staging file is removed, but execution continues: `stable_path` is registered and returned to the model even when it does not exist or does not contain the expected payload.

Please propagate the original `OSError`. If the exception is intended to handle a benign concurrent-writer race, it should only be suppressed after verifying that another writer successfully created a valid target.

Please also add a regression test that mocks `os.replace()` to fail and asserts that the formatter does not return an invalid path.

src/agentscope/formatter/_formatter_base.py
```diff
@@ -141,21 +159,43 @@ def convert_tool_result_to_string(
...
+                            try:
+                                os.remove(written)
+                            except OSError:
+                                pass
+                    _UNSUPPORTED_MEDIA_TEMP_FILES.add(stable_path)
```
qbc2016: **Blocking: process-local `atexit` cleanup cannot safely own this shared deterministic path.**

Every process adds `stable_path` to its local cleanup registry, including a process that merely reused a file created by another worker. When either worker exits, its `atexit` hook deletes the shared file and can invalidate a path that another live worker still depends on, causing downstream file-access failures.

The simplest safe model would be a process-owned private temporary directory with mode `0700`, so process-local cleanup only removes files owned by that process. If cross-process sharing is required instead, the shared cache needs an explicit lease, reference-counting, or TTL policy rather than unconditional per-process deletion.

Please add a multi-process regression test demonstrating that one worker exiting cannot delete another live worker's file.

src/agentscope/formatter/_formatter_base.py
```diff
@@ -141,21 +159,43 @@ def convert_tool_result_to_string(
...
+                    stable_name = shortuuid.uuid(
+                        name=block.source.data,  # noqa: FBT003
+                    )
+                    tmp_dir = tempfile.gettempdir()
+                    candidate_ext = extension or ""
+                    stable_path = os.path.join(
+                        tmp_dir,
+                        f"as-unsup-{stable_name}{candidate_ext}",
+                    )
+                    if not os.path.exists(stable_path) or os.path.getsize(
```
qbc2016: **Suggestion: avoid using file size as the only validity check while the target remains in a shared temporary directory.**

A deterministic name makes accidental collisions extremely unlikely, but a pre-existing same-sized file could still be trusted if another local process knows or controls the payload.

This concern is largely resolved if the multi-process issue is fixed by placing the files in a process-owned private directory whose entries can only be created through this atomic-write path. If the shared directory is retained, consider verifying the existing file's content digest rather than its size alone.

src/agentscope/formatter/_formatter_base.py
```diff
@@ -141,21 +159,43 @@ def convert_tool_result_to_string(
...
-                        delete=False,
-                    ) as temp_file:
-                        decoded_data = base64.b64decode(block.source.data)
-                        temp_file.write(decoded_data)
-                        textual_output.append(
-                            f"<system-reminder>A(n) {main_type} file is "
-                            f"returned and saved locally at: {temp_file.name}."
-                            f"</system-reminder>",
-                        )
+                    decoded_data = base64.b64decode(block.source.data)
```
qbc2016: **Non-blocking suggestion: clarify the cache-key semantics and avoid redundant decoding on cache hits.**

The current filename is deterministic per Base64 representation rather than strictly content-addressed by the decoded bytes. This is sufficient for repeatedly formatting the same `DataBlock`, although hashing `decoded_data` would allow equivalent Base64 representations to deduplicate to the same file.

There is also a related performance tradeoff: `decoded_data` is currently computed before checking whether the stable file already exists, so every formatter pass performs a full Base64 decode and allocates the decoded payload even on a cache hit.

With the current Base64-derived key, the stable path can be computed and checked before decoding, allowing `base64.b64decode()` to run only on a cache miss. If the implementation switches to a digest of the decoded bytes instead, consider caching the digest or materialized path so repeated formatting does not decode a multi-megabyte payload every time.

Please avoid using an approximate decoded-length calculation such as `len(data) * 3 // 4`, because padding and non-canonical Base64 input make it unreliable.

src/agentscope/formatter/_formatter_base.py
```diff
@@ -141,21 +159,43 @@ def convert_tool_result_to_string(
...
+                    stable_name = shortuuid.uuid(
+                        name=block.source.data,  # noqa: FBT003
```
qbc2016: Please remove `# noqa: FBT003`.

qbc2016 (commented):

src/agentscope/formatter/_formatter_base.py
RerankerGuo: Addressed in a7d43db8. `os.replace()` failures now remove the staging file best-effort and re-raise the original `OSError`, so no invalid target path is returned. Added a regression test that mocks the replace failure and verifies the private cache directory remains empty.

src/agentscope/formatter/_formatter_base.py
RerankerGuo: Resolved in a7d43db8 by moving the cache into a process-owned `0700` directory. The shared-directory size check has been removed; a cache hit can now only reuse a target created through this process-private atomic-write path.

src/agentscope/formatter/_formatter_base.py
RerankerGuo: Addressed in a7d43db8. The key remains deterministic per Base64 representation, allowing the stable path to be checked before decoding. `base64.b64decode()` now runs only on cache misses, with a regression test asserting that a repeated format does not decode again.

src/agentscope/formatter/_formatter_base.py
RerankerGuo: Removed in a7d43db8.

src/agentscope/formatter/_formatter_base.py
RerankerGuo: Addressed in a7d43db8. Unsupported media now lives in a lazily-created process-owned directory with mode 0700, and the atexit hook removes only that directory. Added a subprocess regression test where one worker exits and cleans up while the other worker file remains available.

RerankerGuo:
@qbc2016 Thanks for the detailed review. I addressed all five inline comments in `a7d43db8`:

- moved unsupported-media files into a lazily created, process-owned `0700` directory, so one worker cannot clean up another worker’s files;
- made `os.replace()` failures clean the staging file best-effort and re-raise the original `OSError`;
- check the deterministic Base64-derived path before decoding, so cache hits avoid repeated decode/allocation;
- removed the shared-directory size check and the `FBT003` suppression;
- added regressions for replace failure, cache-hit decoding, private-directory permissions, cleanup, and two-process isolation.

Verification: all 108 formatter tests pass, and the current pre-commit plus Ubuntu/macOS/Windows test jobs are green. The remaining failed checks are GitHub Monitor `notify` jobs rather than code-validation jobs.

Could you take another look when convenient?

PR #2197, opened 2026-07-29
feat(hub): support to install MCP and skills from MCP/skill hubs

DavdGao:
Builds on the hub browse-and-install flow. What the user has installed
is now a **library** they can equip a workspace from, and a skill can
come from a local folder as well as a hub.

## Library

Installed MCPs and skills are persisted (`MCPRecord` / `SkillRecord`)
with the display fields the hub listing carried, so an installed item
does not lose its icon, author, description and download count. Redis
keeps a name index so `(user_id, name)` stays unique and a workspace
can be joined to the library by name.

An MCP record also stores the `values` it was rendered from, which is
what makes re-keying possible: `PATCH /mcp/{id}` merges new values over
the stored ones and re-renders from the card. Secrets are never echoed
back — `MCPView` / `SkillView` omit the rendered config.

## Skills into a workspace

Both paths converge on one workspace method, `add_skill_archive`, which
pipes an archive stream into the sandbox and expands it there:

```
write_stream -> archive in the sandbox
  -> python3 shim expands into <workdir>/.skill-staging-<id>/
  -> SKILL.md verified
  -> mv into skills/<name>, suffixed when the name is taken
  -> staging and archive removed either way
```

The server never holds the archive, and nothing is written to a host
temp directory — staging lives inside the workspace, beside `skills/`
so the rename stays atomic while `list_skills` never sees a
half-expanded archive. Expanding inside the sandbox also means a
traversing or bomb-sized archive detonates in the isolated environment
rather than in the API process; the shim checks every member path
(including link targets) and caps the expanded size.

| Endpoint | |
|---|---|
| `POST /workspace/skill/upload` | Re-tars a browser folder upload on the fly. A tar header needs each member's size before its bytes and a multipart part does not declare one, so the client sends a manifest of paths and sizes; every limit in it is re-checked server-side and the byte counts are verified as the tar is built. |
| `POST /workspace/skill/from-library` | Re-downloads from the hub and pipes it straight through. |
| `POST /workspace/skill` | **Deprecated**, not removed: resolving a path on the server only means anything for a single-host deployment. |

## Streaming primitive

`BackendBase.write_stream` is added with a default that buffers and
delegates to `write_file`, so backends adopt it independently instead
of all eight changing at once. Local, Bubblewrap and k8s stream for
real; the rest degrade, which the docstring states, since only the
overriding backends are constant-memory.

## Hub download contract

`SkillHubBase.download` now returns `SkillArchive(format, stream)`
rather than yielding ZIP bytes. The format travels with the bytes so a
hub whose upstream serves tar need not repack — repacking would push
the buffering problem down into every hub. Response headers are awaited
eagerly, so a 404 or an exhausted rate limit surfaces before the caller
commits to an install, while the body stays lazy. `fetch_skill_dir` is
deleted with its test: it was the last user of a host temp directory
and has no caller left.

A GitHub MCP hub is added against the official registry, resolving
`{TOKEN}`-style placeholders into the renderer's `${TOKEN}` form and
skipping entries with no runnable configuration.

## Reliability

- One unreachable MCP no longer takes a whole reply down: listing its
  tools is caught per client, and the failure is remembered so a dead
  server is dialed once per toolkit rather than once per LLM turn.
- Error classification descends into `ExceptionGroup`, without which
  every provider error arriving through an async task group classified
  as `UNKNOWN`.
- A failure before the agent replies now synthesizes a start/end event
  pair under the session lock, so the client sees why the run failed
  instead of a stream that simply stops.

## Web UI

- MCP and skill hub pages share one layout: a fixed header and search
  box over a scrolling list, rows carrying icon, author, tags, update
  time and download count, and a detail drawer.
- Adding to a workspace is one dialog with two tabs — from the library
  or from a configuration / folder — with the footer outside the tabs
  so switching moves nothing else.
- A failing MCP shows the reason in an alert rather than only a red dot.
- Folder uploads report byte-level progress, which needs XHR since
  fetch surfaces none.

## Checks

`pre-commit run --all-files` clean; 1542 passed / 136 skipped. Frontend
`pnpm format`, `tsc` and `build` clean.


oss-maintainer:
This is a substantial feature PR that introduces a **library system** for installed MCPs and skills, along with a **hub browsing/installation flow** in the WebUI. The changes span 78 files (+10,743 lines) and touch core infrastructure (workspace, storage, hub abstractions) plus extensive frontend work.

## Overall Assessment

The PR is well-architected with clear separation of concerns:
- **Hub abstraction** (`MCPHubBase`/`SkillHubBase`) with concrete implementations (ClawHub, GitHub MCP Hub)
- **Library layer** (`MCPRecord`/`SkillRecord`) persisting installed items with display metadata snapshotted at install time
- **Streaming primitive** (`BackendBase.write_stream`) adopted incrementally by backends
- **Security-first archive handling** — extraction happens inside the sandbox with path traversal protection, symlink validation, and size limits

## Security Highlights ✅

The archive installation path (`add_skill_archive`) demonstrates good security practices:

1. **Sandbox-isolated extraction** — the `_EXTRACT_ARCHIVE_SHIM` runs inside the sandbox, so a malicious or malformed archive detonates there rather than on the API server
2. **Path traversal protection** — `os.path.realpath` checks ensure no member escapes the destination directory
3. **Symlink/link target validation** — link targets are checked before extraction
4. **Size ceiling** — 500MB default limit prevents zip bomb attacks
5. **Atomic staging** — staging directory is renamed into `skills/<name>` only after `SKILL.md` verification; failed installs leave nothing behind
6. **Secret handling** — `MCPRecord.values` stores user-supplied credentials for re-rendering, but `MCPView`/`SkillView` omit the rendered config from API responses

The skill upload endpoint (`_skill_upload.py`) also validates manifests server-side, re-checks all limits, and verifies byte counts as the tar is built.

## CI Failures ⚠️

Two CI jobs are failing and need attention before this can merge:

1. **`k8s-workspace-tests`** — "Run K8s workspace tests" step failed (exit code 1). This likely exercises the new `write_stream` path on k8s backends.
2. **`run (ubuntu-latest)`** — "Pre-commit starts" step failed (exit code 1). This is separate from the `check` job that passed, suggesting a different pre-commit configuration or environment issue.

Could you check the logs and push a fix? The main test suite (ubuntu/macos/windows on Python 3.11) all pass, so the core logic is sound.

## Architectural Questions

1. **Deprecated `POST /workspace/skill` endpoint** — The PR description notes this is deprecated but not removed because "resolving a path on the server only means anything for a single-host deployment." Is there a timeline for removal, or should we add a deprecation warning header to the response?

2. **`BackendBase.write_stream` default implementation** — The default buffers and delegates to `write_file`, which means backends that don't override it lose the constant-memory benefit. The docstring states this, but should we add a `logger.warning` when the default is used so maintainers know to override it?

3. **Library display metadata staleness** — The PR description acknowledges that display fields (icon, author, description) are snapshotted at install time and "go stale if the card changes upstream." The version check refreshes them, but is there a mechanism to detect when a hub has dropped a card entirely? The current code seems to handle this gracefully (the library still shows the item), but it might be worth logging a warning when a card is no longer found on the hub.

## Test Coverage

Good coverage with 8 new test files:
- `hub_card_test.py`, `hub_claw_test.py`, `hub_github_test.py`, `hub_render_test.py` — hub abstractions
- `hub_router_test.py` — API endpoints
- `storage_redis_test.py` — Redis persistence
- `workspace_skill_archive_test.py` — archive installation
- `toolkit_test.py` — toolkit changes

## Suggestions

1. **Consider adding a migration guide** — This PR introduces significant new abstractions (library vs hub, archive-based skill installation). A short migration guide in the docs would help users understand the new flow.

2. **Frontend i18n** — The PR adds new i18n keys for the hub UI. Ensure all user-facing strings are covered in both `en.json` and `zh.json`.

3. **Error messages** — The `_skill_upload.py` validates manifests and raises `SkillUploadError` with descriptive messages. Consider including the specific limit that was exceeded (e.g., "File exceeds 50MB limit: 67MB") to help users debug upload failures.

---
*Automated review by oss-maintainer*

PR #2199, opened 2026-07-30
fix(model): auto-retry without tool_choice for thinking-model 400s (#1608)

### LINKED ISSUE #1608: [Bug]: tool_choice参数支持的问题

RerankerGuo:
## Summary

- **closes #1608**

## Background

Issue #1608 reports that tool-using agents crash on DeepSeek V4 reasoning
models (and similar reasoning variants served through OpenAI-compatible
endpoints, e.g. certain OpenRouter or Together partitions) because the
client framework unconditionally passes a ``tool_choice`` parameter:

```json
{"error": {"message": "deepseek-reasoner does not support this tool_choice"}}
```

The scenario originally filed uses ReActAgent, which sets
``tool_choice=required`` in its ``reply(msg, structured_model=...)``
path. However the same 400 hits any direct OpenAIChatModel caller that
hands a tool_choice to a reasoning-model provider – it is not specific
to one agent.

A related bug was already fixed for the *structured-output* chat path in
#2137/PR #2189 (``_call_api_with_structured_output``, which is used by
``compress_context`` when it requests a pydantic schema). That fix did
*not* cover the normal chat call (``OpenAIChatModel._call_api`` →
``client.chat.completions.create``), which is precisely the path that
ReActAgent and most ordinary tool-using agents travel. This PR closes
the gap.

## Changes

1. New helper ``OpenAIChatModel._safe_chat_completions_create(kwargs)``:
   - Forwards to ``self.client.chat.completions.create(**kwargs)`` on
     the happy path (zero overhead for non-failing providers).
   - On ``openai.BadRequestError`` whose message mentions
     ``tool_choice``, drops the ``tool_choice`` key and retries
     **exactly once** (``attempt < 2`` guard + ``raise`` on the
     second pass to guarantee a bounded retry loop).
   - Retries keep the ``tools`` list intact so the model remains free
     to call any tool; only the rejected selection hint is removed.
   - On retry a ``UserWarning`` is emitted so operators can see
     exactly which provider silently degraded behaviour.
   - If the retry also raises, the **original** BadRequestError is
     re-raised (with the follow-up error linked via ``__context__``)
     so bug reports preserve the provider's verbatim first message
     rather than an unrelated no-tool_choice error.
   - Any BadRequestError that does not mention ``tool_choice`` is
     re-raised immediately, no retries, so unrelated provider
     regressions remain loud.

2. The single ``client.chat.completions.create`` call site inside
   ``_call_api`` is now routed through the helper. Since the SDK
   itself dispatches on ``kwargs["stream"]`` internally this single
   change automatically covers both non-stream and streaming modes.

3. Four new unit tests appended to ``tests/model_openai_chat_test.py``
   under ``TestToolChoiceRejectionFallback``:
   - ``test_400_tool_choice_rejection_retries_without_param`` – first
     attempt 400s, second succeeds, 2 attempts in total, retry
     kwargs still carry ``tools`` but no longer ``tool_choice``,
     and a UserWarning is recorded.
   - ``test_unrelated_400_does_not_retry`` – a validation 400
     about ``temperature`` is re-raised after only 1 attempt.
   - ``test_second_attempt_failure_raises_first_error`` – if both
     attempts fail, the original DeepSeek-style error is the one
     that surfaces (not the second error).
   - ``test_agent_invoke_path_observes_fallback`` – end-to-end
     ``OpenAIChatModel.__call__(msgs, tools, tool_choice=required)``
     is routed through the helper transparently, producing a
     ToolCallBlock for ``get_weather`` without any exception.

## Verification

- pre-commit hook chain (check-ast → trailing-whitespace → sort-yaml →
  add-trailing-comma → mypy → black → flake8 → pylint 3.0.2 →
  pyroma): **ALL PASS**.
- pytest::

      python3 -m pytest tests/model_openai_chat_test.py -v
      ================================= 24 passed =================================
      24 passed in 2.22s

  Includes: all 4 new fallback tests, 6 existing non-stream tests,
  6 existing stream tests, plus _format_tools/parameters suites. No
  regressions.

## Impact Scope

Files changed:

* ``src/agentscope/model/_openai_chat/_model.py`` – 60 LOC added,
  1 LOC changed (the create call routing). Behaviour for non-failing
  providers is byte-for-byte identical.
* ``tests/model_openai_chat_test.py`` – 4 new tests + small imports +
  shared BadRequestError factory.

## Checklist

- [x] Covers the ReActAgent/_call_api path that issue #1608 actually
      hits (not only the compress_context/structured path fixed in
      PR #2189).
- [x] Bounded retry loop (at most 1 retry) + original error is kept
      if the retry also raises.
- [x] Tool list remains available on the retried request – tool
      behaviour is preserved; only the selection hint is dropped.
- [x] UserWarning is emitted on fallback for operator visibility.
- [x] Commit author: RerankerGuo <121015044+RerankerGuo@users.noreply.github.com>
      (GitHub contribution-statistics compliant).

### DIFF
--- a/src/agentscope/model/_openai_chat/_model.py
+++ b/src/agentscope/model/_openai_chat/_model.py
@@ -267,7 +267,7 @@ class OpenAIChatModel(ChatModelBase):
             kwargs["stream_options"] = {"include_usage": True}
 
         start_datetime = datetime.now()
-        response = await self.client.chat.completions.create(**kwargs)
+        response = await self._safe_chat_completions_create(kwargs)
 
         audio_cfg = kwargs.get("audio")
         audio_fmt = (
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Closing because the linked compatibility direction remains unresolved. In #1334 the maintainer explicitly noted that falling back from required tool choice to auto can produce undesirable structured-output behavior; #1608 is now stale and points back to that discussion. An automatic retry that silently changes tool-selection semantics should not remain open without renewed maintainer agreement.

Issue #2166, opened 2026-07-24
[Bug]: _agent.py -Duplicate ToolResultStartEvent emitted for external tool on interruption


sunrong1:
### Prerequisites

- [x] I have searched the existing [issues](https://github.com/agentscope-ai/agentscope/issues) and [discussions](https://github.com/agentscope-ai/agentscope/discussions), and this is not a duplicate.
- [x] This is a bug, not a usage question. (For questions, please use [Discussions](https://github.com/agentscope-ai/agentscope/discussions/new?category=general) instead.)

### Background / Description

When an external tool call is interrupted while in SUBMITTED state, _close_unfinished_tool_calls emits a second ToolResultStartEvent for the same tool call. The first START was already emitted by _execute_tool_call when transitioning to ALLOWED.

Root cause: _close_unfinished_tool_calls (line 640) only excludes ALLOWED from re-emitting START, but for external tools, the active lifecycle is ALLOWED → SUBMITTED — both states have START already emitted.

Expected: 1 START event per tool call.
Actual: 2 START events when an external tool in SUBMITTED state is interrupted.

### Error Messages

```shell
No exception. The bug is observable in the event stream
```

### Steps to Reproduce

1.Configure an agent with any is_external_tool = True tool.
2.Call agent.reply() — the tool enters SUBMITTED state, waiting for the external system.
3.Send a UserInterruptEvent to interrupt the agent before the external system returns.
4.Count ToolResultStartEvent events in the stream for the same tool_call_id.

Expected: 1.
Actual: 2.

### Environment

- AgentScope Version: 2.0.4
- Python Version: 3.12
- OS: windows11

March-77:
I’d like to work on this with a focused lifecycle fix. I’ll first add a regression test for interrupting an external tool while it is in `SUBMITTED`, asserting that the existing `ToolResultStartEvent` is not emitted again and that the unfinished call still receives its terminal event. Then I’ll update `_close_unfinished_tool_calls` so every state that has already started is treated consistently, and run the focused agent/event tests plus repository lint checks.

PR #2167, opened 2026-07-24
fix(agent): avoid duplicate external tool start event

March-77:
Fixes #2166

## AgentScope Version

2.0.5 (current `main`)

## Problem and root cause

External tools emit `ToolResultStartEvent` when they are allowed, then transition from `ALLOWED` to `SUBMITTED` while awaiting external execution. Interruption cleanup only recognized `ALLOWED` as already started, so a parked `SUBMITTED` call received a second START event.

## Changes

- Treat both `ALLOWED` and `SUBMITTED` unfinished tool calls as already started during interruption cleanup.
- Add a regression test that combines the parked and interruption event streams and asserts exactly one START for the external tool call.
- Preserve the existing START-before-terminal behavior for calls that are still awaiting confirmation.

## User impact

Consumers of the event stream now receive one balanced tool-result lifecycle for interrupted external tools instead of duplicate START events.

## Validation

- `.venv/bin/pre-commit run --all-files` — passed (AST, YAML/XML/TOML/JSON, mypy, black, flake8, pylint, Pyroma, and remaining hooks)
- `.venv/bin/python -m pytest tests/agent_interrupt_test.py -q` — 9 passed
- `.venv/bin/python -m pytest tests/agent_*_test.py -q` — 38 passed
- `.venv/bin/python -m pytest tests -q` — collection blocked by unrelated optional test dependencies not installed locally (`fakeredis`, `boto3`, `mem0`, and `sqlalchemy`); no test failures were observed

## Compatibility and risk

No public API changes. The change only suppresses a duplicate event for the existing `SUBMITTED` lifecycle state; terminal interruption events and stored tool results are unchanged. Risk is low and covered by the full agent test group.

## Checklist

- [x] An issue has been created for this PR
- [x] I have read the contributing guide
- [x] Docstrings remain in Google style
- [x] No user-facing documentation change is required
- [x] Code is ready for review

PR #2205, opened 2026-07-31
fix(agent): skip duplicate ToolResultStartEvent for SUBMITTED external tool calls (#2166)

RerankerGuo:
## Title
fix(agent): skip duplicate ToolResultStartEvent for SUBMITTED external tool calls (#2166)

## Background

#2166 identified that an external tool call parked in `SUBMITTED` state
received **two** `ToolResultStartEvent` events for the same tool call
when the reply was later interrupted via `UserInterruptEvent`:

1. The **first** `START` that fired earlier inside `_execute_tool_call`
   at the `ToolCallState.ALLOWED` transition (before state flipped to
   SUBMITTED for external execution).
2. A **second** duplicate `START` produced by
   `Agent._close_unfinished_tool_calls()`, whose guard only excluded
   `ToolCallState.ALLOWED` from re-emitting START — it did not cover
   `SUBMITTED`.

For local (non-external) tools the internal state stays at `ALLOWED`
after the initial START so the original guard worked fine. For
external tools the state machine is `ALLOWED → SUBMITTED`, with START
emitted at the `ALLOWED` step; so the SUBMITTED case fell through and
unintentionally emitted a second START on interruption.

## Changes

- **Widened the state guard** in
  `src/agentscope/agent/_agent.py::_close_unfinished_tool_calls`:
  - Before: `if call_block.state != ToolCallState.ALLOWED`
  - After : `if call_block.state not in (ToolCallState.ALLOWED, ToolCallState.SUBMITTED)`

  `PENDING` and `ASKING` states are still allowed to emit a synthetic
  START on interruption cleanup (those states never ran the execution
  path so no START had been emitted yet).

- **Added regression test** in `tests/agent_interrupt_test.py`:
  `test_interrupt_event_after_external_execution` now drives a single
  external tool to parked `SUBMITTED` state, sends
  `UserInterruptEvent`, then asserts that **zero**
  `ToolResultStartEvent` events appear in the resulting event stream.
  Combined with the already-existing START emitted at `ALLOWED`
  transition, the end-to-end cardinality for one tool call is
  **exactly 1 START**.

- **Pylint disable hygiene** in `tests/agent_interrupt_test.py`: the
  existing `check_permissions` overrides on
  `_TimeoutConcurrentTool`/`_UserConfirmConcurrentTool` already
  triggered `missing-function-docstring` and `unused-argument`
  warnings locally. Promoted both warnings to module-level disables
  and split the `# pylint: disable=...` line to two lines so the
  79-column `flake8` line-length limit stays respected in
  Pre-commit CI.

Files changed:
- `src/agentscope/agent/_agent.py` (+11 / -3)
- `tests/agent_interrupt_test.py` (+14 new test + disables, -0)

## Verification

- [x] Pre-commit suite (`black` / `flake8` / `mypy` / `pylint` /
      `Pyroma` / trailing-whitespace / encoding-pragma /
      detect-private-key / add-trailing-comma): **all passed** on
      both modified files.
- [x] `pytest tests/agent_interrupt_test.py -v` — **9/9 passed** on
      darwin Python 3.12 (editable install), including the new
      regression test `test_interrupt_event_after_external_execution`.
- [x] Without the source fix, the new regression test fails: the
      interrupt event stream contains one `ToolResultStartEvent` with
      `tool_call_id == "tc-ext"` (duplicate of the START that was
      already emitted at the ALLOWED transition).

## Impact scope

- **Affected**: interruption cleanup only for external tool calls that
  are parked in `SUBMITTED` state when the interruption fires.
  Downstream event subscribers that relied on START-event cardinality
  (progress indicators, analytics panels, audit streams, per-tool-call
  state machines keyed by `(tool_call_id, START-event-id)`) now get
  the documented behaviour of exactly 1 START per tool call.
- **Not affected**: non-external (local) tool calls; internal state
  transitions inside `_execute_tool_call`; `PENDING`/`ASKING` tool
  calls still get their synthetic START on interruption as designed
  (they had never reached execution so never had an initial START).
- No breaking API change; both parameters are optional keyword-only
  additions with backwards-compatible defaults.

## Checklist

- [x] I have run `pre-commit` locally and all hooks pass.
- [x] I have added **unit tests** for the new behaviour / regression.
- [x] All new and existing tests pass (9/9 on agent_interrupt_test.py).
- [x] I have added **no** user-facing secrets or credentials.
- [x] I have not modified unrelated files outside the listed scope.

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -723,9 +723,14 @@ class Agent:
             call_block = last_msg.content[index]
             assert isinstance(call_block, ToolCallBlock)
 
-            # An ALLOWED call was already running, so its START was already
-            # emitted — skip it here (checked before flipping to FINISHED).
-            if call_block.state != ToolCallState.ALLOWED:
+            # A call in ALLOWED or SUBMITTED was already being executed (or
+            # waiting for external execution), so its START has already been
+            # emitted. For external tools the lifecycle is
+            # ALLOWED → SUBMITTED, with START fired at the ALLOWED transition.
+            if call_block.state not in (
+                ToolCallState.ALLOWED,
+                ToolCallState.SUBMITTED,
+            ):
                 yield ToolResultStartEvent(
                     reply_id=self.state.reply_id,
                     tool_call_id=last_msg.content[index].id,
--- a/tests/agent_interrupt_test.py
+++ b/tests/agent_interrupt_test.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # pylint: disable=redefined-builtin
+# pylint: disable=missing-function-docstring, unused-argument
 """Tests for agent interruption:
 
 - :class:`AgentInterruptCancelTest`: ``task.cancel()`` lands during tool
@@ -20,6 +21,7 @@ from utils import AnyString, MockModel
 from agentscope.agent import Agent, InjectionConfig
 from agentscope.event import (
     ReplyEndEvent,
+    ToolResultStartEvent,
     UserInterruptEvent,
 )
 from agentscope.message import (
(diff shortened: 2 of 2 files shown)


qbc2016:
Thanks for the fix. However, there is already an existing [PR](#2167) addressing this issue, and we are preparing to merge it. Thanks again for your contribution.

PR #2209, opened 2026-07-31
fix(model): preserve interrupted response reason

### LINKED ISSUE #2207: [Bug]: ChatModelBase interruption can be reported as completed
### Prerequisites

### Background / Description

  In AgentScope 2.0.5 and current main (`807390b`), a `ChatResponse`
  constructed with `finished_reason=FinishedReason.INTERRUPTED` returns
  different values through mapping and attribute access.

  `ChatModelBase.__call__` normalizes `asyncio.CancelledError` into such a
  response. The Agent reads `response.finished_reason`, sees `COMPLETED`,
  and can terminate the reply as completed. When `structured_schema` is
  required, it may instead perform another model call.

  Using the same ChatRunRegistry.spawn(agent.reply_stream(...)) and
  task.cancel() pattern as service_agent_interrupt_test.py, but with a
  ChatModelBase subclass:

  task_done:      True
  model_calls:    1
  ReplyEndEvent:  completed

  When structured output is required, the agent can make a second model
  call after the interrupted call.

  ChatResponse is a dataclass inheriting DictMixin.
  DictMixin.__setattr__ writes the constructor value into the mapping,
  while the dataclass default for finished_reason remains available as a
  class attribute. Normal attribute lookup therefore returns the class
  default before DictMixin.__getattr__ is invoked.

  Extend the service interruption test with:

  1. a model inheriting ChatModelBase;
  2. cancellation while _call_api is awaiting;
  3. assertions for an interrupted terminal event and no additional model
     calls;

  4. a variant with structured_schema.

### Error Messages

```shell
- response.finished_reason is INTERRUPTED;
  - the active reply ends with ReplyEndEvent(finished_reason="interrupted");
  - no subsequent model call is made.
```

### Steps to Reproduce

  ```python
  from agentscope.model import ChatResponse, FinishedReason

  response = ChatResponse(
      content=[],
      is_last=True,
      finished_reason=FinishedReason.INTERRUPTED,
  )

  print(response["finished_reason"])  # interrupted
  print(response.finished_reason)     # completed
```

### Environment

  - AgentScope 2.0.5
  - current main at 807390b
  - reproduced on Python 3.11.14 and 3.14.0

logicwu0:
## Summary

- keep `ChatResponse.finished_reason` attribute access consistent with its mapping value
- add a regression test for explicitly interrupted responses
- cover regular and structured-output service interruption flows
- ensure interruption does not trigger an additional model call

## Validation

- `PYTHONPATH=src .venv/bin/python -m pytest tests/agent_interrupt_test.py tests/agent_structured_output_test.py tests/model_response_test.py tests/model_base_test.py tests/service_agent_interrupt_test.py -q` (50 passed)
- `.venv/bin/pre-commit run --files src/agentscope/model/_model_response.py tests/model_response_test.py tests/service_agent_interrupt_test.py`

Fixes #2207

### DIFF
--- a/src/agentscope/model/_model_response.py
+++ b/src/agentscope/model/_model_response.py
@@ -56,7 +56,7 @@ class ChatResponse(DictMixin):
     """The usage information of the chat response, if available."""
 
     finished_reason: FinishedReason = field(
-        default=FinishedReason.COMPLETED,
+        default_factory=lambda: FinishedReason.COMPLETED,
     )
     """The finished reason of the chat response, available when `is_last`
     is `True`."""
--- a/tests/model_response_test.py
+++ b/tests/model_response_test.py
@@ -53,6 +53,23 @@ class ChatResponseAppendTest(IsolatedAsyncioTestCase):
     async def asyncSetUp(self) -> None:
         """The async setup method."""
 
+    async def test_finished_reason_attribute_matches_mapping(self) -> None:
+        """Attribute access must return the explicitly provided reason."""
+        response = ChatResponse(
+            content=[],
+            is_last=True,
+            finished_reason=FinishedReason.INTERRUPTED,
+        )
+
+        self.assertEqual(
+            response.finished_reason,
+            response["finished_reason"],
+        )
+        self.assertEqual(
+            response.finished_reason,
+            FinishedReason.INTERRUPTED,
+        )
+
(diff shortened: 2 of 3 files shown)


qbc2016 (approved):
LGTM.

DavdGao (approved):

PR #2213, opened 2026-08-01
fix(app): reject file URLs in chat input

### LINKED ISSUE #2212: security(app): reject file URLs in chat input
## Problem

The Chat API accepts `DataBlock` URL sources from request JSON. A `file://` source reaches media formatters, which read the referenced path on the service host before embedding the bytes in a request to the configured model provider.

The service example is network-facing and its current identity dependency trusts a request header, so a remote caller able to run an agent can request a service-host path instead of uploading media content.

## Scope

Reject `file://` URL sources at the `ChatRequest` boundary. This preserves trusted SDK and workspace flows that intentionally use local files, while keeping remote URLs and base64 uploads available to API clients.

## Regression coverage

Validate that a request containing a `file://` source is rejected during request parsing before the agent or any formatter can run.

ShiroKSH:
Fixes #2212

## AgentScope Version

2.0.5

## Description

Chat request payloads could pass a `file://` data source to model formatters, causing the service host to read that path before sending the bytes to a model provider.

This change rejects local-file URL sources while parsing Chat API requests, including nested tool results returned by external executors. Trusted SDK and workspace flows retain their existing local-file support; API clients can continue using HTTPS URLs and base64 uploads.

## Validation

- Focused Chat request validation tests: passed
- Full pre-commit gate on the changed files: passed

## Checklist

- [x] I have read the contributing guide
- [x] Related documentation is not required; this is a request-boundary security fix with no public API addition

### DIFF
--- a/src/agentscope/app/_router/_schema/_chat.py
+++ b/src/agentscope/app/_router/_schema/_chat.py
@@ -1,10 +1,40 @@
 # -*- coding: utf-8 -*-
 """The chat endpoint schema."""
 
-from pydantic import BaseModel, Field
+from typing import Self
 
-from ....message import Msg
+from pydantic import BaseModel, Field, model_validator
+
+from ....message import (
+    ContentBlock,
+    DataBlock,
+    HintBlock,
+    Msg,
+    ToolResultBlock,
+    URLSource,
+)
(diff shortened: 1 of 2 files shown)


qbc2016 (commented):
LGTM.

PR #2215, opened 2026-08-01
fix(workspace): evict and rebuild OpenSandbox workspaces killed by server-side timeout

### LINKED ISSUE #2202: [Bug]: OpenSandbox workspace becomes unusable after server-side sandbox timeout
### Prerequisites

### Background / Description

OpenSandbox sandboxes have a server-enforced `timeout_seconds` lifetime. When a sandbox expires, the OpenSandbox server destroys it. However, `OpenSandboxWorkspaceManager.get_workspace()` is a pure in-memory cache — it returns the cached workspace instance without checking whether the underlying sandbox is still alive on the server. The result is that all subsequent tool calls fail permanently, and the workspace becomes completely unusable until the application is restarted.

The core issue is in two places:

1. **`get_workspace()`** returns a zombie workspace without verifying liveness:

```python
# _opensandbox_workspace_manager.py
async with self._lock:
    cached = self._cache.get(workspace_id)
    if cached is not None:
        ws, _ = cached
        self._cache[workspace_id] = (ws, time.monotonic())
        return ws  # no health check — may be dead
```

2. **`_find_existing_sandbox()`** only searches for `RUNNING` and `PAUSED` states — timeout-killed sandboxes are in neither state, and will never be found for reattachment:

```python
# _opensandbox_workspace.py
sandbox_filter = SandboxFilter(
    states=[SandboxState.RUNNING, SandboxState.PAUSED],
    metadata={METADATA_WORKSPACE_ID_KEY: self.workspace_id},
)
```
## Suggested Fix

Add a liveness check before returning cached workspaces. If the sandbox is dead, evict and rebuild:

```python
async def get_workspace(self, user_id, agent_id, session_id, workspace_id=None):
    ...
    async with self._lock:
        cached = self._cache.get(workspace_id)
        if cached is not None:
            ws, _ = cached
            if await self._sandbox_is_alive(ws):    # <-- new
                self._cache[workspace_id] = (ws, time.monotonic())
                return ws
            self._cache.pop(workspace_id)             # <-- evict dead workspace
    # fall through to cache-miss → _build_and_start creates fresh sandbox
    ...
```
Even if the manager were to rebuild, the dead sandbox ID is lost and cannot be reattached. A fresh sandbox with the same `workspace_id` metadata must be created instead.

### Error Messages

```shell
## Error Messages


API error: HTTP 404 | [DOCKER::SANDBOX_NOT_FOUND]
Sandbox d5f32923-2d85-4baf-9ced-85463f8dee8f not found.



Command failed (exit=-1): ls /workspace/
Failed to run command. Status code: 404


Once this occurs, every subsequent tool call (Bash, Read, Write, Edit, Glob, Grep) on that workspace returns the same 404.
```

## Steps to Reproduce

1. Configure a short `timeout_seconds` (e.g. 60s) for `OpenSandboxWorkspaceManager`
2. Call `workspace_manager.get_workspace(workspace_id="test-123")` — returns successfully
3. Wait for the sandbox to expire (> 60s)
4. Call `workspace_manager.get_workspace(workspace_id="test-123")` again — returns cached instance without error
5. Use any tool on the returned workspace → `HTTP 404 SANDBOX_NOT_FOUND`
6. The workspace is permanently broken — no recovery, no automatic rebuild

### Environment

- agentscope 2.0.5
- OpenSandbox server 0.2.0
- OpenSandbox SDK 0.1.15

a-yeyang:
## Summary

`OpenSandboxWorkspaceManager` caches workspaces purely in memory, keyed by `workspace_id`, with no check that the underlying sandbox is still alive on the server. OpenSandbox sandboxes carry a server-enforced `timeout_seconds` lifetime; once it elapses, the server destroys the sandbox without notifying this process. The manager keeps returning the same stale cached workspace forever, so every subsequent tool call fails with `HTTP 404 SANDBOX_NOT_FOUND` — with no recovery short of restarting the whole application.

## Fix

- Added `OpenSandboxWorkspace.is_healthy()`, a thin wrapper around the SDK's `Sandbox.is_healthy()` probe (which already swallows exceptions internally and returns `bool`).
- `OpenSandboxWorkspaceManager.get_workspace()` now delegates its cache-hit path to a new `_get_alive_cached()` helper, which checks `is_healthy()` before returning a cached workspace. A dead sandbox is evicted (closed best-effort) and the caller falls through to the existing cache-miss path, which rebuilds and reattaches/creates a fresh sandbox.
- The liveness probe is a remote round-trip, so it deliberately runs **outside** `self._lock` — otherwise a single dead or slow sandbox for one `workspace_id` would block `get_workspace()` calls for every other workspace. Identity re-checks (`current[0] is ws`) guard the two places that touch the cache after the unlocked probe, so a concurrent caller that already evicted/rebuilt the same entry isn't clobbered.

## Test plan

- [x] New regression tests in `tests/workspace_manager_opensandbox_test.py`:
  - `test_get_workspace_reuses_cached_healthy_workspace` — healthy cache hit is returned as-is, no rebuild.
  - `test_get_workspace_evicts_and_rebuilds_dead_sandbox` — a sandbox that goes unhealthy is evicted, closed, and replaced with a fresh instance on the next call.
  - `test_get_workspace_liveness_check_runs_outside_lock` — uses `asyncio.Event` synchronization to prove an unrelated `workspace_id` resolves promptly while another call is blocked inside a slow `is_healthy()` probe.
  - Plus the standard cache/TTL/concurrency/sweeper coverage mirroring `tests/workspace_manager_daytona_test.py`.
- [x] New `TestOpenSandboxWorkspaceIsHealthy` tests in `tests/workspace_opensandbox_test.py` covering `is_healthy()` before initialize, on a live/dead probe, and on probe error.
- [x] `pytest tests/workspace_manager_opensandbox_test.py tests/workspace_opensandbox_test.py -v` → 14 passed, 1 skipped (pre-existing live-domain-gated lifecycle test, correctly skipped without `OPENSANDBOX_DOMAIN`).
- [x] Full suite: `pytest tests/` → 1642 passed, 138 skipped, no failures.
- [x] `pre-commit run` (mypy, black, flake8, pylint, pyroma) passes on all touched files.

Fixes #2202

### DIFF
--- a/src/agentscope/app/workspace_manager/_opensandbox_workspace_manager.py
+++ b/src/agentscope/app/workspace_manager/_opensandbox_workspace_manager.py
@@ -211,6 +211,13 @@ class OpenSandboxWorkspaceManager(WorkspaceManagerBase):
         resume it depending on state, or create a fresh sandbox
         otherwise.
 
+        A cache hit is verified against the server before being
+        returned: OpenSandbox sandboxes carry a server-enforced
+        ``timeout_seconds`` lifetime, and the server destroys them
+        silently on expiry. A cached workspace whose sandbox is no
+        longer alive is evicted and rebuilt rather than handed back as
+        a permanently-broken zombie.
+
         Idle eviction is not performed here; the background sweeper
         started by :meth:`__aenter__` handles that.
 
@@ -243,14 +250,12 @@ class OpenSandboxWorkspaceManager(WorkspaceManagerBase):
                 session_id="",
             )
 
-        async with self._lock:
-            cached = self._cache.get(workspace_id)
-            if cached is not None:
-                ws, _ = cached
-                self._cache[workspace_id] = (ws, time.monotonic())
-                return ws
+        ws = await self._get_alive_cached(workspace_id)
+        if ws is not None:
+            return ws
 
-        # Cache miss: build under the lock to prevent two concurrent
+        # Cache miss (or a dead sandbox was just evicted above): build
+        # under the lock to prevent two concurrent
         # get_workspace(workspace_id=X) calls from creating two
         # workspaces (and thus two sandboxes) for the same id.
         async with self._lock:
@@ -267,6 +272,50 @@ class OpenSandboxWorkspaceManager(WorkspaceManagerBase):
             )
             self._cache[workspace_id] = (ws, time.monotonic())
             return ws
+
+    async def _get_alive_cached(
+        self,
+        workspace_id: str,
+    ) -> OpenSandboxWorkspace | None:
+        """Return the cached workspace for ``workspace_id`` if alive.
+
+        ``is_healthy`` is a remote round-trip, so it runs outside
+        :attr:`_lock` to avoid blocking unrelated ``get_workspace``
+        calls. A dead workspace is evicted under the lock (re-checking
+        identity first, since a concurrent caller may have already
+        evicted and rebuilt it) and its sandbox is paused best-effort
+        via :meth:`_safe_close`.
+
+        Returns:
+            `OpenSandboxWorkspace | None`:
+                The live cached workspace, or ``None`` on a cache miss
+                or a dead sandbox (now evicted).
+        """
+        async with self._lock:
+            cached = self._cache.get(workspace_id)
+        if cached is None:
+            return None
+        ws, _ = cached
+
+        if await ws.is_healthy():
+            async with self._lock:
+                current = self._cache.get(workspace_id)
+                if current is not None and current[0] is ws:
+                    self._cache[workspace_id] = (ws, time.monotonic())
+            return ws
+
+        logger.warning(
+            "OpenSandboxWorkspaceManager: sandbox for workspace_id=%r "
+            "is no longer alive (likely killed by the server-side "
+            "timeout); evicting and rebuilding",
+            workspace_id,
+        )
+        async with self._lock:
+            current = self._cache.get(workspace_id)
+            if current is not None and current[0] is ws:
+                del self._cache[workspace_id]
+        await self._safe_close(ws)
+        return None
 
     async def close(self, workspace_id: str) -> None:
         """Close (= pause the sandbox) and evict a single workspace."""
--- a/src/agentscope/workspace/_opensandbox/_opensandbox_workspace.py
+++ b/src/agentscope/workspace/_opensandbox/_opensandbox_workspace.py
@@ -186,6 +186,29 @@ class OpenSandboxWorkspace(SandboxedWorkspaceBase):
                 )
             self._sandbox = None
 
+    async def is_healthy(self) -> bool:
+        """Report whether the sandbox is still alive on the server.
+
+        OpenSandbox sandboxes carry a server-enforced ``timeout_seconds``
+        lifetime; once it elapses the server destroys the sandbox without
+        notifying this process. :class:`OpenSandboxWorkspaceManager` calls
+        this before returning a cached workspace so a server-killed
+        sandbox is evicted and rebuilt instead of handed back as an
+        unusable zombie.
+
+        Returns:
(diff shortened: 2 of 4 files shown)


src/agentscope/app/workspace_manager/_opensandbox_workspace_manager.py
```diff
@@ -268,6 +273,50 @@ async def get_workspace(
...
+        if cached is None:
+            return None
+        ws, _ = cached
+
+        if await ws.is_healthy():
+            async with self._lock:
+                current = self._cache.get(workspace_id)
+                if current is not None and current[0] is ws:
+                    self._cache[workspace_id] = (ws, time.monotonic())
+            return ws
```
qbc2016, line 305: **Blocking:** This may return a stale workspace after the cache identity has changed.

`is_healthy()` deliberately runs outside `_lock`, so another coroutine can evict, close, or replace `ws` while the probe is in flight. When `current[0] is not ws`, the identity check correctly avoids updating the cache, but the unconditional `return ws` still hands the caller the old workspace, which may already be closed.

Please return `ws` only while the identity check still succeeds. Otherwise, fall through and resolve the current cache entry again (or rebuild if it was removed). Please also add a regression test where the workspace is replaced or closed while `is_healthy()` is blocked.

src/agentscope/app/workspace_manager/_opensandbox_workspace_manager.py
```diff
@@ -268,6 +273,50 @@ async def get_workspace(
...
+            "OpenSandboxWorkspaceManager: sandbox for workspace_id=%r "
+            "is no longer alive (likely killed by the server-side "
+            "timeout); evicting and rebuilding",
+            workspace_id,
+        )
+        async with self._lock:
+            current = self._cache.get(workspace_id)
+            if current is not None and current[0] is ws:
+                del self._cache[workspace_id]
+        await self._safe_close(ws)
```
qbc2016, line 317: **Blocking:** `_safe_close(ws)` is called even when this coroutine did not evict `ws`.

Two concurrent unhealthy probes can both reach this path. The first one removes the cache entry, while the second one fails the identity check but still closes the same workspace concurrently. `close()` mutates the workspace's sandbox/backend state and is not synchronized, so duplicate closes can race.

Please track whether this coroutine actually removed the entry and call `_safe_close()` only in that case. A regression test should verify that concurrent unhealthy cache hits close the old workspace once and create only one replacement.

src/agentscope/workspace/_opensandbox/_opensandbox_workspace.py
```diff
@@ -186,6 +186,29 @@ async def _teardown_backend(self) -> None:
...
+        Returns:
+            `bool`:
+                ``False`` before :meth:`initialize` (or after
+                :meth:`close`), and whenever the sandbox no longer
+                responds. ``True`` otherwise.
+        """
+        if self._sandbox is None:
+            return False
+        try:
+            return bool(await self._sandbox.is_healthy())
```
qbc2016, line 208: **Non-blocking** A `False` result does not necessarily mean that the sandbox was destroyed. The OpenSandbox health probe also returns `False` for transient connectivity and endpoint errors, so a single network failure can evict and rebuild an otherwise live workspace.

Please consider a short bounded retry or a control-plane status check before classifying the sandbox as permanently dead. This can also be handled in a follow-up if this PR intentionally remains a narrow recovery fix.

qbc2016 (commented):
Please also assert the complete forwarded configuration structure in unit test instead of checking only a selected subset of fields individually.

`workspace.kwargs` is deterministic, so a single `assertEqual(workspace.kwargs, expected_kwargs)` would verify both the values and the exact configuration surface. The current assertions can still pass when a default is forwarded incorrectly, a required argument is omitted, or an unexpected argument is added.

PR #2220, opened 2026-08-03
fix(app): clean up index worker tasks on cancellation

### LINKED ISSUE #2219: [Bug]: IndexWorker.process leaves pipeline and heartbeat tasks running after external cancellation
## Summary

`IndexWorker.process()` starts a child indexing pipeline task and a lease heartbeat task, but it does not clean them up when the parent `process()` coroutine is externally cancelled, for example during `IndexTaskConsumer.__aexit__()` shutdown.

Because `asyncio.CancelledError` is not caught by the existing `except Exception` handler on Python 3.11+, cancellation exits through `finally`, releases the document lease, and leaves the child tasks running until the event loop later cleans them up.

## Why this matters

After shutdown cancellation:

- the parent `worker.process(...)` task is cancelled;
- the document lease is released;
- the child pipeline task can continue and complete indexing work;
- the heartbeat task can continue renewing the lease for a short period.

That can leave an indexing pipeline running after the worker has supposedly stopped processing the document, and it weakens the lease/shutdown guarantees around the RAG indexing path.

## Reproduction sketch

Using the existing `tests/index_worker_lease_test.py` helpers, start a slow `_SlowPipelineWorker.process(...)`, wait until the pipeline starts, cancel the parent process task, then wait briefly.

Observed on current `main` (`9d1026f`):

```text
process_cancelled=True
pipeline_cancelled=False
pipeline_completed=True
lease_released=1
renew_calls_after_wait=5
```

## Expected behavior

When `IndexWorker.process()` is externally cancelled, it should:

1. cancel the child pipeline task;
2. cancel the heartbeat task;
3. wait for both child tasks to finish cleanup;
4. re-raise `asyncio.CancelledError` so caller shutdown semantics stay unchanged;
5. still release the held lease once.

## Suggested fix

Add an explicit `except asyncio.CancelledError` branch in `IndexWorker.process()` before the broad `except Exception` handler, cancel and await `pipeline_task` / `heartbeat_task`, then re-raise.

A regression test can be added to `tests/index_worker_lease_test.py` to assert that external cancellation stops both child tasks and does not mark the document as a terminal pipeline error.

helloJamest:
﻿## Summary

Fixes #2219.

`IndexWorker.process()` now explicitly handles external `asyncio.CancelledError` by cancelling and awaiting both child tasks it owns: the indexing pipeline task and the lease heartbeat task. It then re-raises cancellation so caller shutdown semantics stay unchanged.

## Root cause

`process()` raced `pipeline_task` and `heartbeat_task`, but only handled the internal outcomes where either child task finished first. When the parent `process()` coroutine itself was cancelled during shutdown, the existing broad `except Exception` did not catch `asyncio.CancelledError` on Python 3.11+, so the method released the document lease while the child tasks could keep running.

## Changes

- Add explicit external cancellation cleanup in `IndexWorker.process()`.
- Add a regression test covering caller cancellation of `worker.process(...)`.
- Assert the pipeline is cancelled, heartbeat renewals stop, no document error status is recorded, and the held lease is still released once.

## Validation

- `PYTHONPATH=C:\Users\79823\Documents\面试\agentscope-latest\src python -m pytest tests/index_worker_lease_test.py tests/service_index_task_consumer_test.py -q`
- `pre-commit run --files src/agentscope/app/_service/_index_worker.py tests/index_worker_lease_test.py`

### DIFF
--- a/src/agentscope/app/_service/_index_worker.py
+++ b/src/agentscope/app/_service/_index_worker.py
@@ -271,6 +271,15 @@ class IndexWorker:
             with contextlib.suppress(asyncio.CancelledError):
                 await heartbeat_task
             await pipeline_task
+        except asyncio.CancelledError:
+            pipeline_task.cancel()
+            heartbeat_task.cancel()
+            await asyncio.gather(
+                pipeline_task,
+                heartbeat_task,
+                return_exceptions=True,
+            )
+            raise
         except Exception as exc:  # noqa: BLE001 — terminal error sink
             await self._mark_error(
                 user_id,
--- a/tests/index_worker_lease_test.py
+++ b/tests/index_worker_lease_test.py
@@ -97,6 +97,8 @@ class _SlowPipelineWorker(IndexWorker):
         self.pipeline_started = asyncio.Event()
         self.pipeline_cancelled = False
         self.pipeline_completed = False
+        self.heartbeat_started = asyncio.Event()
+        self.heartbeat_stopped = False
 
     async def _run_pipeline(
         self,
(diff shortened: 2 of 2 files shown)


helloJamest:
Production fix remains unchanged; based on the review/independent validation, I strengthened the regression coverage to explicitly cover the heartbeat lifecycle.

DavdGao (approved):
lgtm

PR #2222, opened 2026-08-03
feat(tool): support arguments repairment within the ToolBase class

### LINKED ISSUE #1794: feat(tool): support arguments repairment within the ToolBase class
**Background**

LLM APIs are prone to returning arguments with incorrect types (e.g., returning a `str` when an `int` is expected). After the `__call__` method is introduced to `ToolBase` in #1719, we can centrally handle argument type repair for all `ToolBase` subclasses within that method.

**Changes**

- Before invoking the tool middlewares and the tool's `call` method in `ToolBase.__call__`, attempt to coerce and repair the argument types based on the expected schema (best-effort, not required to be exhaustive).

iluv7:
## AgentScope Version

2.0.5

## Description

Closes https://github.com/agentscope-ai/agentscope/issues/1794

Models occasionally return tool arguments with types that do not match the schema. For example, an `int` parameter may receive `"42"`, or a `bool` parameter may receive `"true"`. Previously, each tool had to handle these cases itself, or fail inside its implementation.

This PR adds a best-effort, JSON Schema-driven coercion step to `ToolBase.__call__` and the `Toolkit` invocation path. Regular tools do not need individual changes. Legacy tools that override `__call__` directly are also covered when invoked through `Toolkit.call_tool`.

## What Changed

`src/agentscope/tool/_utils.py` now provides `_coerce_tool_args` and its supporting helpers.

| Schema pattern | Behavior | Example |
|---|---|---|
| `{"type": "integer"}` | `str` → `int`; exact `float` → `int` | `"42"` → `42`, `42.0` → `42` |
| `{"type": "number"}` | `str` → `float` | `"3.14"` → `3.14` |
| `{"type": "boolean"}` | `str` → `bool` | `"true"` → `True` |
| `{"type": "string"}` | Non-`str` → `str` | `42` → `"42"` |
| `{"type": "array"}` | Wrap a single value; parse JSON arrays | `"a"` → `["a"]` |
| `{"type": "object"}` | Parse JSON objects | `'{"k":"v"}'` → `{"k":"v"}` |
| `additionalProperties` | Recursively coerce dynamic map values | Values in `dict[str, int]` are repaired |
| `anyOf` / `oneOf` | Select an unambiguous branch and recurse | `Optional[int]`, `list[int] \| None`, discriminated unions |
| `$ref` → `$defs` | Resolve references before recursing | Nested Pydantic models |

Principles:

- Only apply conversions that do not lose information; otherwise preserve the original value.
- Support nested objects and arrays, `$ref`, `additionalProperties`, and common Pydantic `anyOf`/`oneOf` schemas.
- Do not coerce a discriminated union when its branch cannot be determined. Debug logs include argument names and type changes, but never argument values.

The coercion runs at two entry points:

| Entry point | Coverage |
|---|---|
| `ToolBase.__call__` (`_base.py`) | Direct calls to tools that do not override `__call__` |
| `Toolkit.call_tool` (`_toolkit.py`) | Tools that override `__call__` directly, such as scheduler and Mem0/ReMe tools |

`_coerce_tool_args` is idempotent: values that already match the schema are not changed, so passing through both entry points is safe.

## Files Changed

| File | Summary |
|---|---|
| `src/agentscope/tool/_utils.py` | Adds schema-driven argument coercion and recursive handling |
| `src/agentscope/tool/_base.py` | Calls `_coerce_tool_args` after positional-argument validation |
| `src/agentscope/tool/_toolkit.py` | Calls `_coerce_tool_args` after `_json_loads_with_repair` |
| `tests/tool_coercion_test.py` | Adds 10 test functions and 17 parameterized subtests |

## Testing

`tests/tool_coercion_test.py` includes 10 test functions and 17 parameterized subtests:

- **Unit tests:** scalar conversions, integer precision, large integers, scientific notation, `inf`/`nan`, nested structures, `$ref`, `additionalProperties`, `anyOf`/`oneOf`, and discriminated unions.
- **Integration tests:** `FunctionTool.__call__`, middleware input, `dict[str, int]` Pydantic schemas, and `Toolkit.call_tool` for tools that override `__call__`.

All 10 test functions and 17 parameterized subtests pass.

## Checklist

- [x] Local tests pass (10 test functions and 17 parameterized subtests)
- [x] Reviewed `CONTRIBUTING.md`
- [x] Added docstrings for new helpers
- [ ] Updated related documentation in the documentation repository
- [x] Ready for review

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """The unified agent class in AgentScope library."""
+
 import asyncio
 import collections
 import inspect
@@ -251,12 +252,14 @@ class Agent:
 
     async def reply_stream(
         self,
-        inputs: Msg
-        | list[Msg]
-        | UserConfirmResultEvent
-        | UserInterruptEvent
-        | ExternalExecutionResultEvent
-        | None = None,
+        inputs: (
+            Msg
+            | list[Msg]
+            | UserConfirmResultEvent
+            | UserInterruptEvent
+            | ExternalExecutionResultEvent
+            | None
+        ) = None,
         structured_schema: Type[BaseModel] | None = None,
         yield_final_msg: bool = False,
     ) -> AsyncGenerator[AgentEvent | Msg, None]:
@@ -295,12 +298,14 @@ class Agent:
 
     async def reply(
         self,
-        inputs: Msg
-        | list[Msg]
-        | UserConfirmResultEvent
-        | UserInterruptEvent
-        | ExternalExecutionResultEvent
-        | None = None,
+        inputs: (
+            Msg
+            | list[Msg]
+            | UserConfirmResultEvent
+            | UserInterruptEvent
+            | ExternalExecutionResultEvent
+            | None
+        ) = None,
         structured_schema: Type[BaseModel] | None = None,
     ) -> Msg:
         """Reply to the given inputs, consuming all streamed events.
@@ -633,12 +638,14 @@ class Agent:
 
     async def _reply(
         self,
-        inputs: Msg
-        | list[Msg]
-        | UserConfirmResultEvent
-        | UserInterruptEvent
-        | ExternalExecutionResultEvent
-        | None = None,
+        inputs: (
+            Msg
+            | list[Msg]
+            | UserConfirmResultEvent
+            | UserInterruptEvent
+            | ExternalExecutionResultEvent
+            | None
+        ) = None,
         structured_schema: Type[BaseModel] | None = None,
     ) -> AsyncGenerator[AgentEvent | Msg, None]:
         """Reply entry point (maybe wrapped by middleware)."""
@@ -652,12 +659,14 @@ class Agent:
 
             async def execute_chain(
                 index: int = 0,
-                inputs: Msg
-                | list[Msg]
-                | UserConfirmResultEvent
-                | UserInterruptEvent
-                | ExternalExecutionResultEvent
-                | None = inputs,
+                inputs: (
+                    Msg
+                    | list[Msg]
+                    | UserConfirmResultEvent
+                    | UserInterruptEvent
+                    | ExternalExecutionResultEvent
+                    | None
+                ) = inputs,
                 structured_schema: Type[BaseModel] | None = structured_schema,
             ) -> AsyncGenerator[AgentEvent | Msg, None]:
                 if index >= len(self._reply_middlewares):
@@ -759,12 +768,14 @@ class Agent:
 
     async def _reply_impl(
         self,
-        inputs: Msg
-        | list[Msg]
-        | UserConfirmResultEvent
-        | UserInterruptEvent
-        | ExternalExecutionResultEvent
-        | None = None,
+        inputs: (
+            Msg
+            | list[Msg]
+            | UserConfirmResultEvent
+            | UserInterruptEvent
+            | ExternalExecutionResultEvent
+            | None
+        ) = None,
         structured_schema: Type[BaseModel] | None = None,
     ) -> AsyncGenerator[AgentEvent | Msg, None]:
         """Core reply logic."""
@@ -1167,9 +1178,11 @@ class Agent:
                     )
                     last_time = last_time.replace(
                         tzinfo=_resolve_timezone(
-                            match_tz.group(1).strip()
-                            if match_tz
-                            else self.injection_config.timezone,
+                            (
+                                match_tz.group(1).strip()
+                                if match_tz
+                                else self.injection_config.timezone
+                            ),
                         ),
                     )
 
@@ -1418,12 +1431,16 @@ class Agent:
         # Send the model call ended event with usage if available
         yield ModelCallEndEvent(
             reply_id=self.state.reply_id,
-            input_tokens=completed_response.usage.input_tokens
-            if completed_response.usage
-            else 0,
-            output_tokens=completed_response.usage.output_tokens
-            if completed_response.usage
-            else 0,
+            input_tokens=(
+                completed_response.usage.input_tokens
+                if completed_response.usage
+                else 0
+            ),
+            output_tokens=(
+                completed_response.usage.output_tokens
+                if completed_response.usage
+                else 0
+            ),
             finished_reason=completed_response.finished_reason,
         )
 
@@ -2082,7 +2099,8 @@ class Agent:
             # Coerce types before validation (e.g. str "42" → int 42).
             if parsed_input and tool.input_schema:
                 parsed_input = _coerce_tool_args(
-                    parsed_input, tool.input_schema
+                    parsed_input,
+                    tool.input_schema,
                 )
                 tool_call.input = json.dumps(parsed_input)
 
@@ -2224,9 +2242,11 @@ class Agent:
                     tool_result_block = ToolResultBlock(
                         id=tool_call.id,
                         name=tool_call.name,
-                        output=[TextBlock(text=chunk.content)]
-                        if isinstance(chunk.content, str)
-                        else chunk.content,
+                        output=(
+                            [TextBlock(text=chunk.content)]
+                            if isinstance(chunk.content, str)
+                            else chunk.content
+                        ),
                         state=chunk.state,
                         metadata=chunk.metadata,
                     )
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -2109,9 +2109,6 @@ class Agent:
                     "non-finite numbers are not valid JSON values.",
                 )
 
-            if parsed_input and tool.input_schema:
-                tool_call.input = json.dumps(parsed_input)
-
             # Validate the coerced input against the tool schema.
             # TODO: feed validation errors back to the model so it can
             # auto-correct.  Type coercion above already handles common
@@ -2124,6 +2121,9 @@ class Agent:
                     f"Input validation failed for tool '{tool_call.name}': "
                     f"{e.message}",
                 ) from e
+
+            if parsed_input and tool.input_schema:
+                tool_call.input = json.dumps(parsed_input)
 
         # The exceptions that
         #  - cannot found tool
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -3,6 +3,7 @@
 import asyncio
 import collections
 import inspect
+import json
 import re
 
 from asyncio import Queue
@@ -92,6 +93,7 @@ from ..tool import (
     ToolChoice,
     ToolResponse,
 )
+from ..tool._utils import _coerce_tool_args
 from ..permission import (
     PermissionBehavior,
     PermissionEngine,
@@ -2077,8 +2079,18 @@ class Agent:
                 tool.input_schema,
             )
 
-            # Validate the parsed input with the tool schema
-            # TODO: Maybe some logic to mix the validation error in runtime
+            # Coerce types before validation (e.g. str "42" → int 42).
+            if parsed_input and tool.input_schema:
+                parsed_input = _coerce_tool_args(
+                    parsed_input, tool.input_schema
+                )
+                tool_call.input = json.dumps(parsed_input)
+
+            # Validate the coerced input against the tool schema.
+            # TODO: feed validation errors back to the model so it can
+            # auto-correct.  Type coercion above already handles common
+            # LLM mistakes (e.g. "42" → 42), so this would mainly cover
+            # missing required params or structural mismatches.
             try:
                 jsonschema.validate(parsed_input, tool.input_schema)
             except jsonschema.ValidationError as e:
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -94,7 +94,7 @@ from ..tool import (
     ToolChoice,
     ToolResponse,
 )
-from ..tool._utils import _coerce_tool_args
+from ..tool._utils import _coerce_tool_args, _contains_non_finite_float
 from ..permission import (
     PermissionBehavior,
     PermissionEngine,
@@ -2102,6 +2102,14 @@ class Agent:
                     parsed_input,
                     tool.input_schema,
                 )
+
+            if _contains_non_finite_float(parsed_input):
+                raise AgentOrientedException(
+                    f"Input validation failed for tool '{tool_call.name}': "
+                    "non-finite numbers are not valid JSON values.",
+                )
+
+            if parsed_input and tool.input_schema:
                 tool_call.input = json.dumps(parsed_input)
 
             # Validate the coerced input against the tool schema.
--- a/src/agentscope/tool/_toolkit.py
+++ b/src/agentscope/tool/_toolkit.py
@@ -24,6 +24,7 @@ from ._response import ToolResponse, ToolChunk
 from ..skill import SkillLoaderBase, Skill
 from ._types import RegisteredTool
 from .._utils._common import _describe_exception, _json_loads_with_repair
+from ._utils import _coerce_tool_args
 from ..exception import (
     DeveloperOrientedException,
     ToolNotFoundError,
@@ -297,6 +298,13 @@ class Toolkit:
         try:
             # Prepare keyword arguments
             kwargs = _json_loads_with_repair(tool_call.input)
+
+            # Best-effort type coercion based on the tool's input_schema.
+            # Also covers subclasses that override __call__ directly,
+            # where the base class coercion in ToolBase.__call__ is
+            # bypassed.
+            if kwargs and hasattr(tool_func, "input_schema"):
+                kwargs = _coerce_tool_args(kwargs, tool_func.input_schema)
 
             # State injection
             if (
--- a/src/agentscope/tool/_toolkit.py
+++ b/src/agentscope/tool/_toolkit.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """The toolkit class for tool calls in AgentScope."""
+
 import asyncio
 import inspect
 from collections import OrderedDict
@@ -39,7 +40,6 @@ from ..message import (
 from ._tool_group import ToolGroup
 from .._logging import logger
 from ..state import AgentState
-
 
 # pylint: disable=line-too-long
 DEFAULT_META_TOOL_RESPONSE_TEMPLATE = """{% if groups | length == 0 %}All tool groups are currently deactivated.{% else %}The currently activated tool group(s): {{ groups | map(attribute='name') | join(', ') }}.{% if groups | selectattr('instructions', 'ne', None) | list | length > 0 %}
@@ -89,8 +89,9 @@ class Toolkit:
     def __init__(
         self,
         tools: list[ToolBase] | None = None,
-        skills_or_loaders: Sequence[str | Skill | SkillLoaderBase]
-        | None = None,
+        skills_or_loaders: (
+            Sequence[str | Skill | SkillLoaderBase] | None
+        ) = None,
         mcps: list[MCPClient] | None = None,
         tool_groups: list[ToolGroup] | None = None,
         meta_tool_response_template: str = DEFAULT_META_TOOL_RESPONSE_TEMPLATE,
--- a/src/agentscope/tool/_utils.py
+++ b/src/agentscope/tool/_utils.py
@@ -8,6 +8,7 @@ import sys
 from decimal import Decimal, InvalidOperation
 from typing import Any, Dict, Callable
 
+import jsonschema
 from docstring_parser import parse
 from pydantic import Field, create_model, ConfigDict
 
@@ -278,6 +279,7 @@ def _coerce_value(
         if isinstance(alternatives, list):
             return _coerce_composite(
                 value,
+                prop_schema,
                 alternatives,
                 defs,
                 param,
@@ -303,7 +305,8 @@ def _coerce_value(
 
 def _coerce_composite(
     value: Any,
-    alternatives: list[dict[str, Any]],
+    composite_schema: dict[str, Any],
+    alternatives: list[Any],
     defs: dict[str, Any],
     param: str,
 ) -> Any:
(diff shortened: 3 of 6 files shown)


src/agentscope/tool/_toolkit.py
```diff
@@ -298,6 +300,13 @@ async def call_tool(
             # Prepare keyword arguments
             kwargs = _json_loads_with_repair(tool_call.input)
 
+            # Best-effort type coercion based on the tool's input_schema.
+            # Also covers subclasses that override __call__ directly,
+            # where the base class coercion in ToolBase.__call__ is
+            # bypassed.
+            if kwargs and hasattr(tool_func, "input_schema"):
+                kwargs = _coerce_tool_args(kwargs, tool_func.input_schema)
```
qbc2016, line 308: Would it make sense to apply the coercion before validation in `Agent._execute_tool_call` as well? The normal Agent path currently validates the parsed input against the schema before reaching `Toolkit.call_tool`. As a result, an input such as `{"n": "42"}` for an integer parameter fails validation and never reaches this coercion step.

It may be safer to coerce the parsed input before validation, and then use the same normalized value for both permission checking and execution so that the value being authorized is also the value being executed.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +166,519 @@ def _extract_input_schema(
...
+                return int(stripped)
+            except (ValueError, TypeError):
+                pass
+            # Fall back via Decimal: preserves precision for strings
+            # like "1e23" or "100000000000000000000000.0" that float()
+            # would silently corrupt.
+            try:
+                d = Decimal(stripped)
+                if d == d.to_integral_value():
+                    return int(d)
```
qbc2016: Could we bound the resulting integer size before calling `int(d)`? `Decimal` stores scientific notation compactly, so an input such as `"1e1000000000"` reaches this line without requiring a large input string, but materializing it as a Python integer may stall or exhaust the process.

One possible approach would be to first require a finite, integral Decimal, then use `d.adjusted() + 1` to estimate the resulting decimal digit count and leave the value unchanged when it exceeds a conservative limit.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +166,519 @@ def _extract_input_schema(
...
+        if resolved_alt is None:
+            continue
+        alt_type = resolved_alt.get("type")
+        if alt_type == "null" or not isinstance(alt_type, str):
+            continue
+        try:
+            coerced = _coerce_to_type(value, alt_type)
+            if coerced is not value:
+                _log_coercion(param, type(value).__name__, alt_type)
+            return _coerce_value(coerced, resolved_alt, defs, param)
```
qbc2016: Could we check whether the coercion actually succeeded before returning here? `_coerce_to_type` indicates an unsupported conversion by returning the original value rather than raising an exception, so this loop stops after the first alternative even when that conversion failed.

For example, `"3.14"` with `anyOf: [{"type": "integer"}, {"type": "number"}]` never reaches the number branch. 

The same issue appears in `_coerce_union` around line 526. It may be worth sharing a small helper or using a sentinel result so both paths can distinguish a failed conversion from a successful no-op.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +166,519 @@ def _extract_input_schema(
...
+        if isinstance(value, bool):
+            return value
+        if isinstance(value, str):
+            stripped = value.strip().lower()
+            if stripped in ("true", "1", "yes"):
+                return True
+            if stripped in ("false", "0", "no"):
+                return False
+        if isinstance(value, (int, float)):
+            return bool(value)
```
qbc2016: Could we limit numeric-to-boolean coercion to the exact values `0` and `1`? Values such as `2`, `-1`, and `NaN` all become `True` here, which loses information and may unexpectedly enable control flags such as `confirm` or `overwrite`.

Leaving other numeric values unchanged would seem more consistent with the intended best-effort, lossless-coercion behavior.

qbc2016 (commented):

src/agentscope/tool/_toolkit.py
iluv7: Completely agree — thanks for pointing this out. I fixed this in 51b7cf43.
Agent._execute_tool_call now applies _coerce_tool_args immediately after parsing the tool input and before jsonschema.validate. For example, {"n": "42"} is normalized to {"n": 42} before validation, so it no longer fails before reaching Toolkit.call_tool.
The normalized input is also written back to tool_call.input. This ensures the same value is used for schema validation, permission checking, and eventual execution, so the value being authorized matches the value being executed.
I also added a regression test that verifies both the permission check and the tool execution receive the coerced integer value.

src/agentscope/tool/_utils.py
iluv7: Completely agree — thanks for catching this. I fixed it in 869c275e.
Before materializing a Decimal as an int, we now require it to be finite and integral, then use Decimal.adjusted() to bound the estimated decimal digit count. Values exceeding the limit, such as 1e1000000000, are left unchanged instead of being expanded into a huge Python integer.
The limit is capped at 4,000 digits and kept below Python’s configured int-to-str limit, so the normalized value can also be safely serialized later by json.dumps.
I also added regression tests for huge scientific-notation inputs and verified that normal, reasonably sized values such as 1e23 are still coerced correctly.

src/agentscope/tool/_utils.py
iluv7: Completely agree — thanks for catching this. I fixed it in e17762b4.
Both _coerce_composite (anyOf / oneOf) and _coerce_union now verify that a coercion result actually matches the candidate schema type before returning it. If a conversion attempt leaves the value unchanged and still mismatched, the code continues to the next alternative instead of stopping early.
For example, "3.14" with integer-or-number alternatives now skips the failed integer conversion and proceeds to the number branch, producing 3.14 as a float.
I also added a regression test for this case.

src/agentscope/tool/_utils.py
iluv7: Completely agree — thanks for pointing this out. I updated the numeric-to-boolean coercion so that it now only accepts the exact values 0 / 0.0 and 1 / 1.0.
Other numeric values, including 2, -1, non-integral floats, and NaN, are left unchanged. This keeps the coercion lossless and avoids unexpectedly enabling boolean control flags.
I also added regression tests covering both the accepted 0/1 values and unsupported numeric values.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +174,528 @@ def _extract_input_schema(
...
+    if expected_type == "number":
+        if isinstance(value, bool):
+            return value
+        if isinstance(value, (int, float)):
+            # JSON Schema "number" accepts integers — no coercion needed
+            return value
+        if isinstance(value, str):
+            stripped = value.strip()
+            try:
+                return float(stripped)
```
qbc2016: Please reject non-finite results before returning the parsed float. With Agent-side coercion now running before validation, `"NaN"` becomes `float("nan")`. Python's `jsonschema` treats it as a number, while comparisons with `minimum` and `maximum` are both false, allowing it to bypass range constraints.

A full Agent call with a schema requiring a number between 0 and 1 successfully passed `"NaN"` to the tool and rewrote the stored input as the non-standard JSON `{"x": NaN}`. Returning the original string when `math.isfinite(parsed)` is false would allow schema validation to reject it normally.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +174,528 @@ def _extract_input_schema(
...
+            The ``$defs`` / ``definitions`` block from the parent schema.
+        param (`str`):
+            The parameter name (for debug logging only).
+
+    Returns:
+        `Any`:
+            The coerced value, or the original value if coercion fails.
+    """
+    # Resolve local references before examining the schema.
+    resolved_schema = _resolve_alt_ref(prop_schema, defs)
```
qbc2016, line 271: Please handle boolean JSON Schema subschemas before resolving `$ref`. JSON Schema permits `true` and `false` wherever a subschema is expected, so `properties: {"x": true}` is valid.

Here, passing a bool to `_resolve_alt_ref` raises `TypeError` while evaluating `"$ref" not in alt`, causing valid custom or MCP tool schemas to fail before normal validation. Returning the original value when `prop_schema` is not a dict would preserve the best-effort behavior and let `jsonschema` handle boolean schemas correctly.

qbc2016 (commented):

src/agentscope/tool/_utils.py
iluv7: This case was missed because I did not have coverage for non-finite strings in the number coercion path.
I fixed it in 6e6279be by parsing the string first and returning the parsed float only when it is finite. Non-finite values such as "NaN", "Infinity", and "-Infinity" are now left as strings, allowing the existing schema validation to reject them normally instead of bypassing numeric range constraints.
I also added regression tests confirming that these values remain strings rather than being coerced to non-finite floats.

src/agentscope/tool/_utils.py
iluv7: Done in `7fc3052d`.
When a property schema is true or false, coercion now leaves the value unchanged and lets the normal jsonschema validation handle the schema semantics. I also added guards for boolean alternatives in anyOf / oneOf, so they no longer raise during reference resolution.
I added regression tests for both properties: {"x": true} and properties: {"x": false} to confirm coercion passes through safely without crashing.

src/agentscope/tool/_utils.py
iluv7: One related edge case: Python’s `json.loads` also accepts non-standard, unquoted `NaN` input such as `{"x": NaN}` and produces `float("nan")` before coercion runs. Would you expect this case to be rejected as part of the same fix as well?

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +174,537 @@ def _extract_input_schema(
...
+                    if d.adjusted() < _MAX_COERCED_INT_DIGITS:
+                        return int(d)
+            except (ValueError, TypeError, InvalidOperation, OverflowError):
+                pass
+        return value
+
+    if expected_type == "number":
+        if isinstance(value, bool):
+            return value
+        if isinstance(value, (int, float)):
```
qbc2016, line 765: The new finite-value check only covers string inputs. `json.loads('{"value": NaN}')` produces `float("nan")`, and this branch returns it unchanged. Python jsonschema then accepts it as a number, while `minimum` and `maximum` do not reject it. Please reject non-finite floats at the parsing or validation boundary and add an end-to-end test with an unquoted `NaN` token.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +174,537 @@ def _extract_input_schema(
...
+            The coerced value, or the original value if coercion fails.
+    """
+    # Boolean sub-schemas (true / false) are valid JSON Schema but
+    # carry no type information — let jsonschema handle them.
+    if not isinstance(prop_schema, dict):
+        return value
+
+    # Resolve local references before examining the schema.
+    resolved_schema = _resolve_alt_ref(prop_schema, defs)
+    if resolved_schema is not None:
```
qbc2016, line 272: A `$ref` may resolve to a Boolean schema. In that case, `prop_schema` becomes `True` or `False` here, and the following `.get()` still raises `AttributeError`. Please apply the Boolean-schema handling after reference resolution as well and add coverage for a `$defs` entry whose value is `true` or `false`.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +174,537 @@ def _extract_input_schema(
...
+            score = _discriminator_score(value, resolved_alt)
+            if score >= 0:
+                type_matches.append((resolved_alt, score))
+
+    if type_matches:
+        type_matches.sort(key=lambda x: x[1], reverse=True)
+        best_alt, best_score = type_matches[0]
+        if len(type_matches) > 1 and type_matches[1][1] == best_score:
+            return value
+        return _coerce_value(value, best_alt, defs, param)
```
qbc2016: This selects a branch using the discriminator and required-field score without first checking whether the original value already satisfies the composite schema. A valid `oneOf` input can therefore be coerced into an invalid one. Boolean alternatives are affected as well: with `oneOf: [true, {"type": "integer"}]`, `"42"` is initially valid, but coercing it to `42` makes both branches match. Please preserve values that already validate against the composite schema and only accept a repaired candidate after validating it against the selected alternative.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +174,537 @@ def _extract_input_schema(
...
+            Discrimination score (≥ 0), or ``-1`` if disqualified.
+    """
+    if not isinstance(value, dict):
+        return 0
+    props = schema.get("properties")
+    if not isinstance(props, dict):
+        return 0
+    # const contradiction → disqualify
+    for key, prop_schema in props.items():
+        if "const" in prop_schema:
```
qbc2016: Property schemas may also be Boolean schemas. When an object union branch contains `properties: {"anything": true}`, the membership check here raises `TypeError`. Please inspect `const` only when the property schema is a dictionary and add a nested Boolean-property regression test.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +174,537 @@ def _extract_input_schema(
...
+            The target JSON schema type.
+
+    Returns:
+        `Any`:
+            The coerced value, or the original value if coercion is not
+            applicable or fails.
+    """
+    if expected_type == "string":
+        if not isinstance(value, str):
+            return str(value)
```
qbc2016, line 719: `str(value)` also converts dictionaries, lists, `None`, and bytes into Python-specific representations, turning otherwise invalid structured inputs into valid strings. This is broader than an unambiguous scalar repair. Please limit this conversion to supported scalar primitives such as booleans and numbers, leaving containers and unsupported values unchanged.

src/agentscope/agent/_agent.py
```diff
@@ -2077,8 +2096,19 @@ async def _execute_tool_call(
...
 
-            # Validate the parsed input with the tool schema
-            # TODO: Maybe some logic to mix the validation error in runtime
+            # Coerce types before validation (e.g. str "42" → int 42).
+            if parsed_input and tool.input_schema:
+                parsed_input = _coerce_tool_args(
+                    parsed_input,
+                    tool.input_schema,
+                )
+                tool_call.input = json.dumps(parsed_input)
```
qbc2016: Please keep the repaired arguments in `parsed_input`, validate them first, and assign `json.dumps(parsed_input)` to `tool_call.input` only after `jsonschema.validate(...)` returns successfully. This keeps the original tool-call input unchanged when validation fails.

qbc2016 (commented):

src/agentscope/tool/_utils.py
iluv7: Fixed in e478fb9c. Agent now rejects non-finite floats after parsing and before schema validation, so unquoted `NaN` cannot bypass numeric range constraints. Added an end-to-end regression test.

src/agentscope/tool/_utils.py
iluv7: Fixed in 007af2e2. Boolean-schema handling now also runs after `$ref` resolution, so `$defs` entries resolving to `true` or `false` no longer cause `.get()` errors. Added coverage for both cases.

src/agentscope/tool/_utils.py
iluv7: Fixed in 70f4a5e8. Composite coercion now preserves values that already validate against the full composite schema. Repaired candidates are validated against both the selected branch and the composite schema before being accepted.

src/agentscope/tool/_utils.py
iluv7: Fixed in 70f4a5e8. Discriminator scoring now inspects `const` only when the property schema is a dictionary. Added a nested Boolean-property regression test.

src/agentscope/tool/_utils.py
iluv7: Fixed in 7d8ce21e. String coercion is now limited to supported scalar primitives. Containers, `None`, bytes, and non-finite floats are preserved unchanged. Added regression coverage.

src/agentscope/agent/_agent.py
iluv7: Fixed in 5b309cbf. Repaired arguments remain in `parsed_input` during validation, and `tool_call.input` is updated only after validation succeeds. Added a regression test confirming invalid calls retain their original input.

iluv7:
In 1acd8845. Beyond the inline comments, I found that a root-level Boolean JSON Schema (`true` or `false`) could still cause coercion to call `.get()` on a boolean value. This is now handled by passing Boolean root schemas through unchanged, with regression coverage for both `true` and `false`.

src/agentscope/tool/_utils.py
```diff
@@ -158,3 +175,646 @@ def _extract_input_schema(
...
+            The schema to validate against.
+        defs (`dict[str, Any]`):
+            Local schema definitions for resolving references.
+
+    Returns:
+        `bool`:
+            Whether the value satisfies the schema.
+    """
+    schema_with_defs = dict(schema)
+    schema_with_defs.setdefault("$defs", defs)
```
qbc2016: Using `setdefault` here preserves a `$defs` block declared on the composite subschema, so the root definitions supplied through `defs` are not attached. A valid schema whose composite contains its own `$defs` and references a root `#/$defs/...` entry then raises `_WrappedReferencingError` during this detached validation, rather than falling back to the normal validation path. Please preserve the root reference context when validating composite candidates, or at minimum treat reference-resolution failures as a non-match, and add a regression test for this case.

qbc2016 (commented):
Please see the inline comments. Others LTGM.

src/agentscope/tool/_utils.py
iluv7: Fixed in 2799d53a. Composite candidate validation now preserves the root `$defs`/`definitions` context, so references such as `#/$defs/RootType` continue to resolve correctly even when the composite schema declares its own `$defs`. Added a regression test covering this case.

iluv7:
Thank you very much for the thorough and thoughtful review! 🥳

qbc2016 (commented):
LGTM.

iluv7:
Hi, @qbc2016 , I noticed there’s still one failing workflow and GitHub shows that an approving review is required. Is there anything you’d like me to do on my side?

qbc2016:
> Hi, @qbc2016 , I noticed there’s still one failing workflow and GitHub shows that an approving review is required. Is there anything you’d like me to do on my side?

cc @DavdGao

iluv7:
@DavdGao Just a gentle reminder—could you please take another look at this PR when you have a chance? Thanks!

DavdGao (commented):
@iluv7 Thanks for the contribution!

Before going further with the hand-written coercion — `json_repair` already supports schema-guided repair via `repair_json(json_str, schema=...)`, and it covers almost everything in your table (`"42"`→int, `"true"`→bool, array wrapping, `additionalProperties`, `anyOf`/`oneOf`, `$ref`). We're also already passing the schema in at `agent/_agent.py:1753`.

The real bug is in `_json_loads_with_repair` (`_utils/_common.py`): it returns early once `json.loads` succeeds, so for well-formed-but-wrong-typed input like `{"a": "42"}` the schema is never used. So the fix could be just:

- don't early-return when a schema is given, let the schema-aware repair run;
- pin `json_repair>=0.57` in `pyproject.toml` (schema coercion landed there, we have no lower bound now);
- for `ToolBase.__call__` / `Toolkit.call_tool`, which get a dict instead of a string, `repair_json(json.dumps(kwargs), schema=...)` works the same.

That should replace most of the 662 lines in `_utils.py`. Could you rework the PR along these lines? Also please drop the unrelated reformatting in `_agent.py` — looks like a different black version, and it makes the diff much harder to read.

iluv7:
> @iluv7 Thanks for the contribution!感谢您的贡献！
> 
> Before going further with the hand-written coercion — `json_repair` already supports schema-guided repair via `repair_json(json_str, schema=...)`, and it covers almost everything in your table (`"42"`→int, `"true"`→bool, array wrapping, `additionalProperties`, `anyOf`/`oneOf`, `$ref`). We're also already passing the schema in at `agent/_agent.py:1753`.在继续讨论手写强制操作之前—— `json_repair` 已经可以通过 `repair_json(json_str, schema=...)` 实现基于模式的修复了，而且它几乎涵盖了表格中的所有内容（ `"42"` 为整数类型， `"true"` 为布尔类型，还包括数组包装、 `additionalProperties` 、 `anyOf` / `oneOf` 以及 `$ref` 等特性）。我们也在 `agent/_agent.py:1753` 处传递了模式信息。
> 
> The real bug is in `_json_loads_with_repair` (`_utils/_common.py`): it returns early once `json.loads` succeeds, so for well-formed-but-wrong-typed input like `{"a": "42"}` the schema is never used. So the fix could be just:真正的错误出现在 `_json_loads_with_repair` （ `_utils/_common.py` ）处：当 `json.loads` 成功执行后，该函数提前返回了结果。因此，对于像 `{"a": "42"}` 这样格式正确但类型错误的输入，模式根本不会被使用。所以，修复方法可能很简单：
> 
> * don't early-return when a schema is given, let the schema-aware repair run;当给出模式时，不要立即返回结果，让模式感知的修复过程运行下去；
> * pin `json_repair>=0.57` in `pyproject.toml` (schema coercion landed there, we have no lower bound now);引脚 `json_repair>=0.57` 与引脚 `pyproject.toml` 相连（模式强制转换就此完成，现在没有更低的分界点了）；
> * for `ToolBase.__call__` / `Toolkit.call_tool`, which get a dict instead of a string, `repair_json(json.dumps(kwargs), schema=...)` works the same.对于 `ToolBase.__call__` / `Toolkit.call_tool` ，它们得到的是字典而不是字符串。 `repair_json(json.dumps(kwargs), schema=...)` 的功能也是类似的。
> 
> That should replace most of the 662 lines in `_utils.py`. Could you rework the PR along these lines? Also please drop the unrelated reformatting in `_agent.py` — looks like a different black version, and it makes the diff much harder to read.那些修改应该能够替换掉 `_utils.py` 中大部分的 662 行代码。你们能按照这些思路重新编写那个补丁吗？另外，请取消 `_agent.py` 中的无关重排操作——那里的黑色版本看起来很混乱，反而让差异标记变得难以阅读了。

@DavdGao Refactored in #2496 based on your feedback. Thanks!

DavdGao:
Fixed in #2496

PR #2225, opened 2026-08-04
fix(app): clean up index tasks on cancellation

### LINKED ISSUE #2219: [Bug]: IndexWorker.process leaves pipeline and heartbeat tasks running after external cancellation

daleselaji-dev:
## AgentScope Version

`2.0.5` / current `main` at `9d1026f`

## Description

Fixes #2219.

### Background

`IndexWorker.process()` creates a pipeline task and a lease-heartbeat task, then waits for either one to finish. When the parent coroutine is externally cancelled during worker shutdown, `asyncio.CancelledError` bypasses the existing `except Exception` branch on supported Python versions. The `finally` block releases the lease, but the two child tasks are left running without an owner.

That allows indexing work to continue after shutdown and lets the heartbeat continue renewing a lease that the parent has already released.

### Changes

- Handle external cancellation separately from terminal pipeline errors.
- Cancel both the pipeline and heartbeat tasks.
- Await both tasks with `asyncio.gather(..., return_exceptions=True)` so their cleanup finishes before the lease is released.
- Re-raise `CancelledError` to preserve caller shutdown semantics.
- Add a deterministic regression test that starts both children, cancels the parent, and verifies:
  - the pipeline is cancelled instead of completing;
  - the heartbeat stops;
  - no terminal document error is recorded;
  - the lease is released exactly once.

The normal completion and lost-lease paths are unchanged.

### Validation

```text
python -m pytest tests/index_worker_lease_test.py tests/service_index_task_consumer_test.py -q
10 passed

black --check --line-length 79 src/agentscope/app/_service/_index_worker.py tests/index_worker_lease_test.py
2 files would be left unchanged

flake8 --extend-ignore=E203 src/agentscope/app/_service/_index_worker.py tests/index_worker_lease_test.py
passed

mypy [repository pre-commit arguments] src/agentscope/app/_service/_index_worker.py tests/index_worker_lease_test.py
Success: no issues found in 2 source files
```

### Risk and review notes

The change only affects cancellation of `IndexWorker.process()`. It does not change successful indexing, lost-lease error handling, document status semantics, or public APIs. The main review point is the ordering: both children are awaited before the existing `finally` block releases the lease.

## Checklist

- [ ] Related documentation has been updated in the documentation repository (not applicable: no user-facing API or configuration change)

### DIFF
--- a/src/agentscope/app/_service/_index_worker.py
+++ b/src/agentscope/app/_service/_index_worker.py
@@ -271,6 +271,15 @@ class IndexWorker:
             with contextlib.suppress(asyncio.CancelledError):
                 await heartbeat_task
             await pipeline_task
+        except asyncio.CancelledError:
+            pipeline_task.cancel()
+            heartbeat_task.cancel()
+            await asyncio.gather(
+                pipeline_task,
+                heartbeat_task,
+                return_exceptions=True,
+            )
+            raise
         except Exception as exc:  # noqa: BLE001 — terminal error sink
             await self._mark_error(
                 user_id,
--- a/tests/index_worker_lease_test.py
+++ b/tests/index_worker_lease_test.py
@@ -97,6 +97,8 @@ class _SlowPipelineWorker(IndexWorker):
         self.pipeline_started = asyncio.Event()
         self.pipeline_cancelled = False
         self.pipeline_completed = False
+        self.heartbeat_started = asyncio.Event()
+        self.heartbeat_stopped = False
 
     async def _run_pipeline(
         self,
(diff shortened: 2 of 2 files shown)


RerankerGuo:
Independent validation on macOS 26 (arm64), Python 3.11.15:

### Exact revisions

- This PR: `596e6423dbfaa2a7b4dc156f8ab32cce8d8c208e`
- Competing PR #2220: `51bf96b549c313df89ce430709ab77cca0f5e0f3`

### Commands and results

```text
git switch --detach 596e6423dbfaa2a7b4dc156f8ab32cce8d8c208e
python -m pytest -q tests/index_worker_lease_test.py tests/service_index_task_consumer_test.py
.......... [100%]
10 passed in 1.27s
```

The same test command at #2220's head also passed:

```text
.......... [100%]
10 passed in 4.83s
```

I also compared the production file directly:

```text
git diff --exit-code \
  51bf96b549c313df89ce430709ab77cca0f5e0f3 \
  596e6423dbfaa2a7b4dc156f8ab32cce8d8c208e \
  -- src/agentscope/app/_service/_index_worker.py
```

The command produced no diff, so both PRs carry the same cancellation
implementation.

### Conclusion

The cancellation behavior is verified: the focused lease and consumer tests
pass at this exact head. This PR has the stronger regression coverage because
it waits for both the pipeline and heartbeat to start, then explicitly asserts
that the heartbeat stopped. That removes the weak-test window where unchanged
renew counts could pass before the heartbeat ever ran.

Since the production patches are identical, only one implementation should be
kept open. Whichever PR is retained should preserve this PR's explicit
`heartbeat_started` / `heartbeat_stopped` assertions.

DavdGao:
fixed in #2220

PR #2230, opened 2026-08-04
fix(hub): preserve GitHub MCP environment inputs

### LINKED ISSUE #2229: fix(hub): preserve GitHub MCP package environment input schemas
## Summary

GitHub MCP Registry package environment variables are not converted into
install inputs correctly. Required non-secret variables and nested
`value`/`variables` templates produce cards with `auth="none"` and empty
`inputs_schema`; installing those cards leaves literal `${...}` placeholders
in the MCP process environment.

## Evidence from the public registry

Queried on 2026-08-04:

```bash
curl -fsSL 'https://api.mcp.github.com/v0/servers?limit=100'
```

Within that page:

- 194 non-secret package environment-variable records affect 26 servers.
- 5 top-level non-secret variables are explicitly `is_required: true`.
- 44 records use nested `value: "{name}"` plus a `variables` schema.
- 6 nested inputs are explicitly required.

Two concrete registry shapes:

```json
{
  "name": "FIRECRAWL_API_KEY",
  "value": "{api_key}",
  "variables": {
    "api_key": {
      "description": "your API key",
      "is_required": true,
      "is_secret": true
    }
  }
}
```

```json
{
  "name": "DT_ENVIRONMENT",
  "description": "The URL of your Dynatrace environment",
  "format": "string",
  "is_required": true
}
```

## Current behavior

On current `main` (`9d1026fa`), building and rendering cards for those
records produces:

```text
{'id': 'firecrawl', 'auth': 'none', 'inputs_schema': {},
 'rendered_env': {'FIRECRAWL_API_KEY': '${FIRECRAWL_API_KEY}'}}
{'id': 'dynatrace', 'auth': 'none', 'inputs_schema': {},
 'rendered_env': {'DT_ENVIRONMENT': '${DT_ENVIRONMENT}'}}
```

Root causes in `GitHubMCPHub._from_package`:

1. every env value is rewritten as `${ENV_NAME}`, ignoring the registry's
   `value` template;
2. only top-level `is_secret` variables are added to the input schema;
3. nested `variables` definitions and required non-secret values are ignored.

`render_mcp` deliberately preserves undeclared placeholders, so installation
succeeds with unusable literal environment values rather than reporting or
collecting the required configuration.

## Expected behavior

- Registry `value` templates should map their declared variables into the
  card input schema.
- Required non-secret inputs should be collected just like required secrets,
  without password masking.
- Secret inputs should remain `writeOnly` / password fields.
- Literal/default env values should remain literal.
- Optional env values omitted by the installer should not leave `${...}` in
  the launched process environment.

## Environment

- AgentScope: current `main` at `9d1026fa`
- Python: 3.11
- Registry: `https://api.mcp.github.com/v0/servers`

RerankerGuo:
## PR Title Format

`fix(hub): preserve GitHub MCP environment inputs`

## AgentScope Version

Current `main` at `0cd19d7d`

## Description

Fixes #2229.

GitHub MCP Registry packages can describe environment configuration in three
forms:

- direct required/optional variables;
- literal/default values;
- `value: "{input}"` templates with nested `variables` metadata.

The existing parser rewrote every env value to `${ENV_NAME}` and only
declared top-level secret variables. Required non-secret and nested inputs
therefore produced cards with `auth="none"` and rendered MCP configs that
still contained literal `${...}` values.

This PR:

- builds JSON Schema properties from direct and nested registry variables;
- preserves required flags and password masking independently;
- preserves literal registry env values;
- uses the registry's nested `value` template rather than the env key;
- omits unfilled optional env entries during rendering, while retaining the
  existing missing-placeholder error for URLs, args, and other structural
  fields.

### Public registry evidence

The first 100 public Registry servers contained:

- 194 non-secret environment-variable records across 26 servers;
- 44 nested `value`/`variables` templates;
- 5 explicitly required top-level non-secret values;
- 6 explicitly required nested values.

Concrete regression fixtures use the current Firecrawl, Dynatrace, and
Sendmux payload shapes.

## Verification

- Before the fix, four new regressions failed:
  - required non-secret input absent from schema;
  - nested input name replaced by the env name;
  - literal env value replaced by a placeholder;
  - optional env placeholder rejected instead of omitted.
- `pytest -q tests/hub_card_test.py tests/hub_claw_test.py tests/hub_github_test.py tests/service_mcp_render_test.py tests/hub_router_test.py`
  -> `92 passed`
- File-scoped Pre-commit -> all hooks passed.
- Parsed and rendered every runnable package from the downloaded 100-server
  Registry page with required inputs populated:
  `real_registry_packages_checked=32`, with no `${...}` env values left.

## Impact

- No public API changes.
- Remote MCP headers retain their existing required-secret behavior.
- Optional env values are exported only when supplied.
- Required env values continue to fail schema validation when absent.

## Checklist

- [x] I have read `CONTRIBUTING.md`
- [x] No documentation update is required for this parser correctness fix

### DIFF
--- a/src/agentscope/app/_service/_mcp_render.py
+++ b/src/agentscope/app/_service/_mcp_render.py
@@ -18,6 +18,42 @@ class MCPRenderError(ValueError):
     """
 
 
+def _effective_values(schema: dict, values: dict) -> dict:
+    """Overlay submitted values on defaults declared by the input schema."""
+    effective = {
+        name: prop["default"]
+        for name, prop in schema.get("properties", {}).items()
+        if isinstance(prop, dict) and "default" in prop
+    }
+    effective.update(values)
+    return effective
+
+
+def _omit_unfilled_optional_env(
+    config: dict,
+    values: dict,
+    declared: set[str],
+    required: set[str],
+) -> None:
+    """Drop env entries whose optional inputs were not supplied.
+
+    Optional placeholders elsewhere remain errors: omitting a URL or an
+    argument can change the meaning of a config. Environment entries are
+    independent, so leaving one unset means not exporting that variable.
+    """
+    env = config.get("env")
+    if not isinstance(env, dict):
+        return
+
+    optional = declared - required
+    for key, value in list(env.items()):
+        if not isinstance(value, str):
+            continue
+        identifiers = set(Template(value).get_identifiers())
+        if (identifiers & optional) - values.keys():
+            del env[key]
+
+
 def _substitute(
     node: Any,
     values: dict,
@@ -89,6 +125,7 @@ def render_mcp(
             declared placeholder unfilled, or produce a config the
             client rejects.
     """
+    values = _effective_values(card.inputs_schema, values)
     try:
         jsonschema.validate(values, card.inputs_schema)
     except jsonschema.ValidationError as e:
@@ -97,9 +134,17 @@ def render_mcp(
         ) from e
 
     declared = set(card.inputs_schema.get("properties", {}))
+    required = set(card.inputs_schema.get("required", []))
     missing: set[str] = set()
+    config_template = card.config_template.model_dump(mode="json")
+    _omit_unfilled_optional_env(
+        config_template,
+        values,
+        declared,
+        required,
+    )
     config = _substitute(
-        card.config_template.model_dump(mode="json"),
+        config_template,
         values,
         declared,
         missing,
--- a/src/agentscope/app/_service/_mcp_render.py
+++ b/src/agentscope/app/_service/_mcp_render.py
@@ -18,6 +18,31 @@ class MCPRenderError(ValueError):
     """
 
 
+def _omit_unfilled_optional_env(
+    config: dict,
+    values: dict,
+    declared: set[str],
+    required: set[str],
+) -> None:
+    """Drop env entries whose optional inputs were not supplied.
+
+    Optional placeholders elsewhere remain errors: omitting a URL or an
+    argument can change the meaning of a config. Environment entries are
+    independent, so leaving one unset means not exporting that variable.
+    """
+    env = config.get("env")
+    if not isinstance(env, dict):
+        return
+
+    optional = declared - required
+    for key, value in list(env.items()):
+        if not isinstance(value, str):
+            continue
+        identifiers = set(Template(value).get_identifiers())
+        if (identifiers & optional) - values.keys():
+            del env[key]
+
+
 def _substitute(
     node: Any,
     values: dict,
@@ -97,9 +122,17 @@ def render_mcp(
         ) from e
 
     declared = set(card.inputs_schema.get("properties", {}))
+    required = set(card.inputs_schema.get("required", []))
     missing: set[str] = set()
+    config_template = card.config_template.model_dump(mode="json")
+    _omit_unfilled_optional_env(
+        config_template,
+        values,
+        declared,
+        required,
+    )
     config = _substitute(
-        card.config_template.model_dump(mode="json"),
+        config_template,
         values,
         declared,
         missing,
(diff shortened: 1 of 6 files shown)


src/agentscope/app/hub/_mcp/_github_hub.py
```diff
@@ -62,6 +62,42 @@ def _substitute(value: Any) -> Any:
...
+        "title": name,
+        "description": spec.get("description") or "",
+    }
+    if spec.get("is_secret"):
+        prop["writeOnly"] = True
+        prop["format"] = "password"
+    if spec.get("choices"):
+        prop["enum"] = [str(choice) for choice in spec["choices"]]
+    if spec.get("default") is not None:
+        prop["default"] = str(spec["default"])
```
qbc2016, line 78: Adding the registry default to the input schema interacts badly with the existing edit dialog. `InstallMCPDialog` initializes `values` from schema defaults even in edit mode, and then submits every non-empty value. As a result, editing an MCP only to rename it or rotate a secret can silently overwrite a previously stored custom value with the registry default. For example, a self-hosted `GITLAB_API_URL` would be reset to `https://gitlab.com/api/v4`.
Please either keep edit-mode values empty/dirty-field-only, or otherwise ensure schema defaults are applied only during a new installation. A regression test covering an edit that preserves a custom non-secret value would also be useful.

src/agentscope/app/_service/_mcp_render.py
```diff
@@ -18,6 +18,31 @@ class MCPRenderError(ValueError):
...
+    if not isinstance(env, dict):
+        return
+
+    optional = declared - required
+    for key, value in list(env.items()):
+        if not isinstance(value, str):
+            continue
+        identifiers = set(Template(value).get_identifiers())
+        if (identifiers & optional) - values.keys():
+            del env[key]
```
qbc2016, line 54: This considers an input unfilled based only on `values`, but `jsonschema.validate()` does not materialize property defaults. Therefore, for a registry variable with `default: "safe"` and no explicit value, `_from_package` produces `MODE=${MODE}` while this function deletes MODE entirely when `render_mcp(card, {})` is called.
The Web UI currently sends schema defaults, but the install API allows `values` to be omitted and defaults it to `{}`, so API callers lose the registry default. Please build an effective values mapping from schema defaults plus the submitted values before validation, optional-env omission, and substitution, and add a renderer test for an omitted input with a schema default.

qbc2016 (commented):

src/agentscope/app/hub/_mcp/_github_hub.py
RerankerGuo: Addressed in 45acfeaf. `InstallMCPDialog` now applies schema defaults only for a new installation; edit mode starts with an empty values map, so only fields the user touches are submitted. Added a router regression that rotates a secret while preserving a stored custom non-secret GitLab URL instead of replacing it with the schema default.

src/agentscope/app/_service/_mcp_render.py
RerankerGuo: Addressed in 45acfeaf. The renderer now builds effective values from schema defaults overlaid by submitted values before JSON Schema validation, optional-env omission, and template substitution. Added direct renderer and GitHub registry regressions for an omitted `MODE` input with default `safe`; the caller input dict remains unchanged.

RerankerGuo:
@qbc2016 Thanks for the review. I addressed both blocking interactions in `45acfeaf`:

- the renderer now builds effective values from schema defaults overlaid by submitted values before validation, optional-environment omission, and template substitution;
- schema defaults are applied only for new installations; edit mode starts with an empty values map so untouched stored values are preserved;
- added regressions for an omitted `MODE` default, caller-input immutability, and rotating a secret without replacing a custom non-secret GitLab URL.

Verification: all 64 focused Python tests pass, along with the WebUI build, ESLint, Prettier, and pre-commit. The current Ubuntu/macOS/Windows and Web UI validation jobs are green; the remaining failed checks are GitHub Monitor `notify` jobs.

Could you take another look when convenient?

qbc2016 (approved):
LGTM.

PR #2243, opened 2026-08-06
fix(webui): reset confirmation state between pending tool calls

### LINKED ISSUE #2242: [Bug]: WebUI permission card spins indefinitely after resolving the first of multiple HITL confirmations
### Prerequisites

- [x] I searched existing issues, discussions, and pull requests.
- [x] This is a WebUI bug, not a usage question.

### Background / Description

When several tool calls are waiting for confirmation, the WebUI displays them one at a time. After the first call is approved or denied, the next confirmation card stays disabled and keeps spinning.

The backend accepts the first result and leaves the other tool calls in the `ASKING` state. The next card should be enabled.

### Error Messages

No error is shown in the UI. The confirmation request for the first tool call completes, but the next card remains disabled.

### Steps to Reproduce

1. Start the AgentScope agent service and WebUI.
2. Run an agent that emits at least two tool calls in one model response.
3. Require confirmation for both calls.
4. Approve or deny the first card.
5. Check the next card.

### Environment

- AgentScope commit: `bf7e8ff75a234a55fbf2b4ae61e4bd43f138e94b`
- Component: `examples/web_ui/frontend`
- Browser: Chrome on macOS

### Cause

The current code collects all `asking` calls but renders index 0 without a key tied to the reply and tool-call IDs. React therefore reuses the same `ConfirmCard` instance when index 0 changes. That instance keeps `hasConfirmed=true` after a successful request.

The request error path has a second problem: `ChatContent` does not return the confirmation promise, and `useMessages.onUserConfirm` catches request errors without rethrowing them. `ConfirmCard` cannot reset after a failed request.

Relevant code:

- `ChatContent.tsx`: https://github.com/agentscope-ai/agentscope/blob/bf7e8ff75a234a55fbf2b4ae61e4bd43f138e94b/examples/web_ui/frontend/src/components/chat/ChatContent.tsx#L75-L82
- `ChatContent.tsx`: https://github.com/agentscope-ai/agentscope/blob/bf7e8ff75a234a55fbf2b4ae61e4bd43f138e94b/examples/web_ui/frontend/src/components/chat/ChatContent.tsx#L124-L135
- `ConfirmCard.tsx`: https://github.com/agentscope-ai/agentscope/blob/bf7e8ff75a234a55fbf2b4ae61e4bd43f138e94b/examples/web_ui/frontend/src/components/chat/ConfirmCard.tsx#L27-L35
- `useMessages.ts`: https://github.com/agentscope-ai/agentscope/blob/bf7e8ff75a234a55fbf2b4ae61e4bd43f138e94b/examples/web_ui/frontend/src/hooks/useMessages.ts#L349-L380

The backend confirmation protocol already accepts partial `confirm_results`; no backend change is needed.

qbc2016:
## AgentScope Version

2.0.5

## Description

Fix the WebUI confirmation card getting stuck after resolving the first of multiple pending tool calls. The card is now keyed by reply and tool-call IDs, and confirmation errors propagate correctly so failed requests can be retried.

Fixes #2242 

## Checklist


Zed302:
One small race remains in `ConfirmCard`: `hasConfirmed` is React state, so it is not a synchronous re-entry guard. Two confirmation events can enter `handleConfirm` before the state update is committed.

Would it make sense to keep a ref alongside the UI state?

```tsx
const submittingRef = useRef(false);

if (submittingRef.current) return;
submittingRef.current = true;
sethasConfirmed(true);

try {
  await onUserConfirm(confirm, rules);
} catch {
  submittingRef.current = false;
  sethasConfirmed(false);
}
```

The ref should stay `true` after a successful request until the card is replaced or removed. Resetting it in `finally` would allow another submission while waiting for the tool-call state update.

DavdGao (approved):

PR #2260, opened 2026-08-07
feat(channel): add distributed Feishu QR credential binding

### LINKED ISSUE #2249: feat(channel): support QR code-based channel binding for IM channels
**Background**

Currently, adding a channel in IM platforms such as Feishu and DingTalk requires manually entering the app ID and secret key, which is error-prone and not user-friendly. Supporting QR code-based binding will simplify the onboarding process and improve the overall usability of channel configuration.

**Changes**

- Support QR code-based channel binding for IM channels (e.g. Feishu, DingTalk), so that users can add a channel by scanning a QR code instead of manually entering credentials

**Note**
- The specific implementation approach requires further discussion: the current mechanism passes a JSON schema from the backend to the frontend to render the configuration form. How to extend this schema to support tab-based form structures (e.g. switching between QR code mode and manual credential input mode) needs to be defined and agreed upon before implementation

qbc2016:
## AgentScope Version

2.0.5

## Description

Add QR-code credential binding for Feishu alongside manual setup. The implementation supports multi-tenant, multi-node deployments with shared TTL state, atomic Redis updates, distributed locking, one-time credential consumption, and graceful task cleanup. It also adds Web UI tab switching and frontend/backend test coverage.

## Checklist


src/agentscope/app/message_bus/_base.py
```diff
@@ -40,6 +40,7 @@
 Mode A. The bus stays simple; deduplication is the producer's
 responsibility.
 """
+
```
DavdGao: Please remove unnecessary modifications to reduce the file modifications involved in this PR.

src/agentscope/app/storage/_redis_storage.py
```diff
@@ -1311,10 +1311,16 @@ async def upsert_channel(
             self.key_config.channel_botid_index,
             platform_bot_id=platform_bot_id,
         )
-        await self._set_with_ttl(key, record.model_dump_json())
-        await self._client.sadd(index_key, record.id)
-        await self._client.sadd(self.key_config.channel_all_index, record.id)
-        await self._set_with_ttl(botid_key, record.id)
+        async with self._client.pipeline(transaction=True) as pipe:
```
DavdGao, line 1314: Why do we need to modify the storage file for the credential binding pr?

src/agentscope/app/channel/_credential_binding.py
```diff
@@ -0,0 +1,503 @@
...
+    id: str
+    qr_code_url: str
+    expires_at: str
+    state: ChannelCredentialBindingState = (
+        ChannelCredentialBindingState.PENDING
+    )
+    message: str = ""
+
+
+class ChannelCredentialBindingStatus(BaseModel):
```
DavdGao, line 45: The two classes `ChannelCredentialBindingStatus` and `State` are to similar, plz consider renaming them

DavdGao (commented):
Please refer to the inline comments and ensure that such state machines are compatible with different API QR code verifications.

src/agentscope/app/channel/_credential_binding.py
```diff
@@ -0,0 +1,503 @@
...
+    limit_slot: int | None = None
+
+
+class ChannelCredentialBindingStore:
+    """TTL-backed binding state shared through the application's bus."""
+
+    _FIELD = "record"
+    _OWNER_FIELD = "alive"
+    _SLOT_FIELD = "binding_id"
+    _MAX_ACTIVE_PER_USER = 3
```
DavdGao: Suggest dropping the per-user quota entirely. It pulls in `limit_slot` (L73), `_user_lock` (L110), `_claim_slot_unlocked` (L188), `_release_slot_unlocked` (L247) and the slot bookkeeping inside `_put_unlocked` (L139) — roughly 150 lines, plus the nested user-lock/binding-lock ordering it forces on every write.

The concern behind it is real: each `start()` costs ~60 outbound polls to Feishu over five minutes, and that quota is shared by the whole deployment, so one user can degrade everyone.

But this implementation doesn't stop that. `cancel()` releases the slot immediately (`store.delete` → `_release_slot_unlocked`), so a `start → cancel → start` loop burns Feishu quota at full HTTP speed with all three slots permanently free. What the quota actually blocks is "user opened four tabs" — harmless, and the panel already cancels on unmount, so normal usage is always 1.

If we want this protection, a per-user rate limit on `POST /channels/bindings` (a Redis counter, e.g. 5/min) is both cheaper and actually effective. A concurrency quota is the wrong shape for a rate problem.

src/agentscope/app/channel/_credential_binding.py
```diff
@@ -0,0 +1,503 @@
...
+            return True
+
+    async def owner_alive(self, binding_id: str) -> bool:
+        """Return whether the worker driving external polling is alive."""
+        return await self._bus.registry_exists(
+            self._owner_namespace(binding_id),
+            self._OWNER_FIELD,
+        )
+
+    async def replace_if_owner_missing(
```
DavdGao: Suggest removing this along with `refresh_owner` (L360), `owner_alive` (L378), `clear_owner` (L431), and `_owner_heartbeat` in `_feishu/_credential_binding.py:388`.

The docstring here has to explain a lease-boundary race — "a healthy worker renewed its lease after another node checked it, but before that node marked the session as failed." That is the hardest piece of reasoning in the PR, and it buys one thing: after a worker crash the session shows `failed` immediately instead of expiring on its own.

The record already carries a 5-minute TTL. "Worker died → session expires → user scans again" is a perfectly acceptable outcome for a flow that is retryable by design. This is a UX refinement costing ~90 lines, a long-lived heartbeat coroutine per session, and the heaviest concurrency argument in the change.

src/agentscope/app/channel/_credential_binding.py
```diff
@@ -0,0 +1,503 @@
...
+        return expires_at > datetime.now(timezone.utc)
+
+    async def create(
+        self,
+        record: ChannelCredentialBindingRecord,
+        ttl_secs: int,
+    ) -> None:
+        """Reserve and create a record under the shared per-user limit."""
+        async with self._bus.acquire_lock(
+            self._user_lock(record.user_id),
```
DavdGao: Stepping back on the locking as a whole: it defends against concurrent writers on different nodes, but there is only ever one writer — the `_register` task on the worker that served `start()`. Every other node is a reader.

It also doesn't actually achieve multi-node correctness. `FeishuCredentialBinding._tasks` is process-local, so when worker B handles a cancel it can delete the Redis record but **cannot cancel the SDK task still running on worker A** — A runs to completion and only then discovers its CAS was rejected. `complete()` has the same hole.

The root cause is that the Feishu SDK never exposes the device code, so a session is physically pinned to the process that started it. I'd suggest making that explicit rather than papering over it: **keep the record in the bus registry** — that legitimately fixes the 404 when a poll lands on another worker, and it's ~30 lines — **but drop the locks, the lease and the quota.** The only genuine race left is a lazy-expiry check in `get_status` colliding with the authorize write, and the worst outcome there is that the user scans again.

src/agentscope/app/channel/_credential_binding.py
```diff
@@ -0,0 +1,503 @@
...
+
+from ..message_bus import MessageBus, MessageBusKeys
+from ._errors import ChannelError
+
+
+class ChannelCredentialBindingState(str, Enum):
+    """Public state of a credential-binding session."""
+
+    PENDING = "pending"
+    SCANNED = "scanned"
```
DavdGao: Nothing ever assigns this state. All five references are membership in an `expected_states` set, and `handle_status_change` only touches `message`. Meanwhile the frontend already ships copy for it (`en.json:710`, `zh.json:710`), so we're advertising a state users can never see.

Either confirm the Feishu SDK doesn't surface a "scanned" signal and drop the member plus the i18n keys, or wire up the assignment.

DavdGao (commented):
Thanks for this — the flow works and the security posture is exactly right: the browser only ever sees a QR image and an opaque session id, and the credentials go server-to-server.

My main concern is the amount of distributed coordination layered on top. `FeishuCredentialBinding._tasks` is process-local, and the Feishu SDK never exposes the device code, so a binding session is physically pinned to the worker that started it. Keeping the record in the bus registry is well justified — it fixes the 404 when a poll lands on another worker. But the two-level locks, the worker lease and the per-user quota defend against concurrent cross-node writers that cannot exist, and they don't close the gap that does exist: a cancel handled by worker B can't stop the SDK task running on worker A.

Detailed suggestions inline. Most of them are subtractive.

src/agentscope/app/channel/_feishu/_credential_binding.py
```diff
@@ -0,0 +1,550 @@
...
+                return
+            record.message = "Waiting for authorization in Feishu or Lark."
+            updates.append(
+                asyncio.create_task(persist_qr(record.model_copy(deep=True))),
+            )
+
+        def on_qr_code(info: dict) -> None:
+            loop.call_soon_threadsafe(handle_qr_code, dict(info))
+
+        def handle_status_change(info: dict) -> None:
```
DavdGao: This callback only updates `message`, never `state` — both `slow_down` and `domain_switched` are purely cosmetic text.

If we don't persist them, the `updates` task list (L206), both `gather(*updates)` calls (L316, L346), and the `await asyncio.sleep(0)` at L315 with its explanatory comment all disappear. Those are the most fragile part of `_register`: correctness currently depends on yielding once so callbacks fired before SDK completion get their persistence tasks registered in time. That's a lot of ordering subtlety to protect two status strings.

src/agentscope/app/channel/_feishu/_credential_binding.py
```diff
@@ -0,0 +1,550 @@
...
+        sdk = self._sdk()
+        binding_id = secrets.token_urlsafe(24)
+        record = ChannelCredentialBindingRecord(
+            id=binding_id,
+            user_id=user_id,
+            provider_id=self.provider_id,
+            state=ChannelCredentialBindingState.PENDING,
+            expires_at=(_utc_now() + timedelta(minutes=5)).isoformat(),
+        )
+        await store.create(record, 300 + self._terminal_ttl_secs)
```
DavdGao: This write happens before the QR code exists, so it persists a record with an empty `qr_code_url`. The browser can't poll it either — it only learns the binding id once `start()` returns.

Once the quota is gone (no slot to reserve up front), this and `persist_qr` can collapse into a single write once the QR arrives. One less Redis round-trip, and one less intermediate state where a record exists without a QR code.

src/agentscope/app/_service/_channel.py
```diff
@@ -16,9 +18,13 @@
     StorageBase,
 )
 from ..channel._errors import ChannelError
+from ..channel._credential_binding import ChannelCredentialBindingStore
 from ..channel._registry import ChannelTypeRegistry
 
 
+_BINDING_CONSUME_LOCK_TTL_SECS = 60
```
DavdGao: The consume lock looks redundant. `_persist_new` already runs under `channel_bot_lock`, and the bot id is derived from the resolved credentials — so two concurrent creates using the same binding necessarily collide on the bot lock and the second one gets a 409.

Dropping it also lets the two branches at L88-L140 collapse into a linear sequence:

```python
if credential_binding_id:
    credentials = await binding.resolve_credentials(...)
record = await self._persist_new(...)
if credential_binding_id:
    await binding.complete(...)   # best-effort
```

src/agentscope/app/channel/_registry.py
```diff
@@ -9,16 +9,39 @@
...
+from ._credential_binding import (
+    ChannelCredentialBindingBase,
+    ChannelCredentialMode,
+)
 
 if TYPE_CHECKING:
     from ._base import ChannelBase
 
 
+def _reveal_validated_value(value: object) -> object:
```
DavdGao: I checked every `Credentials` model in the repo — Feishu (`_feishu/_channel.py:102`) and Discord (`_discord/_channel.py:52`) are both flat `str` fields that mark secrets via `json_schema_extra={"format": "password"}`. **None of them uses `SecretStr`.**

So the recursive walk over `SecretStr` / `BaseModel` / `dict` / `list` is currently unreachable generality; `validated.model_dump()` covers every existing case. Suggest inlining that and adding the unwrapping when a channel actually adopts `SecretStr`.

src/agentscope/app/_router/_channel.py
```diff
@@ -151,6 +174,94 @@ async def create_channel(
...
+        )
+    except ChannelError as e:
+        raise HTTPException(e.status_code, str(e)) from e
+    return ChannelCredentialBindingSessionResponse(
+        **session.model_dump(),
+        channel_type=body.channel_type,
+    )
+
+
+@channel_router.get("/bindings/{channel_type}/{binding_id}")
```
DavdGao: `channel_type` in the path is redundant — the record already carries `provider_id`, so one registry read locates the provider.

Removing it also removes the four duplicated provider-mismatch 409 checks it exists to support (`_credential_binding.py` L288/L419, `_feishu/_credential_binding.py` L411) and the `channel_type` field on both response wrappers. About 40 lines, and the contract gets simpler: `GET /channels/bindings/{binding_id}`.

(Same applies to the DELETE route at L243.)

src/agentscope/app/message_bus/_in_memory_message_bus.py
```diff
@@ -433,9 +445,28 @@ async def registry_set(
...
+
+            if ttl_secs <= 0:
+                self._registries.pop(namespace, None)
+                return
+
+            def _expire() -> None:
+                self._registries.pop(namespace, None)
+                self._registry_expiry_handles.pop(namespace, None)
+
+            self._registry_expiry_handles[
```
DavdGao, line 461: Fixing the ignored TTL is right — this feature is the first caller for which the TTL is a security boundary, so it can't stay a no-op. But the timer approach requires cancel bookkeeping in four places (L111, L452, L485, L550), and missing any one of them leaks a `TimerHandle`.

Storing an expiry timestamp per namespace and checking it lazily on read gets the same semantics in ~12 lines with nothing to clean up.

src/agentscope/app/storage/_redis_storage.py
qbc2016: This change is needed for the credential-binding consumption path. We delete the binding only after channel persistence succeeds. If Redis writes the channel record but fails before writing the platform-bot index, the request fails while the binding remains reusable. A retry would then see no bot index and could create a duplicate channel. The transaction makes the channel record and its indexes atomic, preserving the uniqueness invariant enforced by the bot-level lock. I therefore kept this change.

src/agentscope/app/channel/_credential_binding.py
qbc2016: I simplified this substantially: the per-user quota and slot locks, owner lease and heartbeat, and the separate consume lock have all been removed. I kept only the per-binding state lock and the bot uniqueness lock. The binding lock is required because replacement is a read-check-write operation: without it, a remote cancellation could delete the record after the worker checks its state, and the worker could then write the authorized record, including credentials, back and resurrect the cancelled session. Cross-worker cancellation now deletes the shared state, and any later result from the process-local SDK task is discarded.

src/agentscope/app/message_bus/_in_memory_message_bus.py
qbc2016: I kept active expiry for the in-memory backend. Lazy expiry would make an expired entry unreadable, but an abandoned authorized binding could still leave plaintext credentials resident in `_registries` indefinitely if that namespace is never accessed again. Since this TTL is also a security and data-retention boundary, active deletion is intentional. Handles are cancelled or replaced on refresh, field deletion, namespace deletion, and bus shutdown, and the related TTL tests pass.

src/agentscope/app/channel/_credential_binding.py
qbc2016: These types represent different layers: 
`ChannelCredentialBindingState` is the enum stored in the binding record, while `ChannelCredentialBindingStatus` is the polling response DTO that contains a `state` field. The resulting `status.state` API is conventional and clearly expresses their respective roles, so I kept the current names. Renaming the response only to reduce lexical similarity would make the public type less descriptive.

DavdGao:
Fixed in #2484

PR #2265, opened 2026-08-08
feat(agent-team): add AgentRemove to remove a single team member

### LINKED ISSUE #2238: feat(agent-team): add AgentRemove to remove a single team member
### Discussed in https://github.com/agentscope-ai/agentscope/discussions/2010

<div type='discussions-op-text'>

<sup>Originally posted by **iluv7** July  5, 2026</sup>
</div>

## Background

Support to delete team member by adding a new team tool named `AgentRemove`

iluv7:
## AgentScope Version

2.0.6

## Description

Closes https://github.com/agentscope-ai/agentscope/issues/2238

**Discussed in** [**#2010**](https://github.com/agentscope-ai/agentscope/discussions/2010)

Adds the `AgentKick` team tool to AgentScope, allowing a team leader to remove one created or invited member without dissolving the entire team through `TeamDelete`. Any in-flight chat run owned by the target member receives a cancellation signal before cleanup, while the team, leader session, and other members remain intact.

### What Changed

**Added `AgentKick` (`src/agentscope/app/_tool/_agent_kick.py`):**

- Only the current team leader may call the tool. Calls from workers or sessions outside a team are rejected, as are attempts by a leader to remove itself.
- `target` uses the same addressing format as `TeamSay.to`: a plain name for created members and `"<name>@<handle>"` for invited members.
- Cleanup behavior depends on the role stored in the team roster:

| Member role | Cleanup behavior | Preserved state |
|---|---|---|
| `created` (from `AgentCreate`) | Cancels the run and deletes the team session, transient bus state, and the complete `AgentRecord` | The team, leader, and other members |
| `invited` (from `AgentInvite`) | Cancels the run and deletes only the borrowed team-scoped session and its bus state | The original `AgentRecord`, standalone sessions, workspace, MCPs, skills, and model configuration |

**Unified team-member routing (`src/agentscope/app/_tool/_team_routing.py`):**

- Extracted the display-name formatting and roster-directory construction shared by created and invited members.
- `AgentInvite`, `TeamSay`, and `AgentKick` now use the same name and handle rules, preventing addressing differences between tools.
- Preserved compatibility with legacy rosters that contain only `TeamData.member_ids`; the existing migration path reconstructs the corresponding `TeamMember` entries.

**Extracted member-level cleanup (`src/agentscope/app/_service/_session.py`):**

Added `SessionService.delete_team_member(user_id, team_id, member)` and made both `AgentKick` and `TeamDelete` use the same member-cleanup path. The method reuses the `TeamMember.role` lifecycle semantics introduced by `AgentInvite` (#1977): created members are fully deleted, while invited members lose only their team-scoped session. This change does not add or modify the Team schema.

The roster entry is retained until session cleanup succeeds. Its stored `session_id` therefore remains available if cancellation or message-bus cleanup fails and the operation is invoked again, preventing partially cleaned resources from becoming unreachable. There is no internal automatic retry loop; after an error, the leader/LLM may call `AgentKick` again (or invoke `TeamDelete` again).

After cleanup, the latest Team record is read again. A compensating `upsert_team()` is performed only when the target still exists in the roster. On the normal created-member path, storage already removes the Team back-reference, so no redundant roster write is issued.

The cleanup sequence is:

1. Re-read the current Team roster and match the target by both `agent_id` and `session_id`.
2. Remove the member's subagent HITL projection from the leader session.
3. Cancel the target session's in-flight chat run and background tasks.
4. Delete the session record and its storage-side related data.
5. Purge the session inbox, event log, and background-task registry.
6. For a created member, continue by deleting the complete `AgentRecord`; for an invited member, preserve the original Agent and delete only the borrowed session.
7. Remove the target from both `TeamData.members` and `TeamData.member_ids`.

**Refactored `TeamDelete` member cleanup:**

- `TeamDelete` was already role-aware through `TeamMember.role`. It now iterates through the shared `delete_team_member` primitive while preserving the existing lifecycle semantics.
- Created members are still fully deleted; invited members still lose only their team-scoped sessions, leaving their standalone agents usable.
- Updated the tool description to remove the “no single-member removal” limitation and direct the LLM to `AgentKick` when only one member should be removed.

**Tool registration and state notifications:**

- Exported `AgentKick` from `src/agentscope/app/_tool/__init__.py`.
- Added `AgentKick` to the leader's team toolset in `src/agentscope/app/_service/_toolkit.py`; a worker's team toolset still exposes only `TeamSay`.
- Added `AgentKick` to the team-changing tool set in `StateChangeMiddleware`, which publishes a `team_updated` event to the leader session after the tool runs.

### Testing

- **AgentKick created path:** Covers cancellation, session and `AgentRecord` deletion, bus cleanup, both roster fields, preservation of the Team/leader/sibling members, and rejection by `TeamSay` after removal.
- **AgentKick invited path:** Covers `name@handle` routing, deletion of the borrowed session, and preservation of the standalone `AgentRecord` and original session.
- **Authorization and target validation:** Covers unknown targets, leader self-removal, worker callers, sessions outside a team, and legacy rosters.
- **Fault injection and retryability:** Covers cancellation exceptions, bus-purge exceptions, `AgentRecord` deletion exceptions, and invited-roster write failures. Retry tests verify that the original session ID is used for cancellation again and that the inbox, session event log, and background-task registry are eventually empty.
- **TeamDelete regression coverage:** Covers created-only teams and mixed created + invited teams, confirming that invited standalone agents remain intact.
- **Toolkit assembly:** Verifies that the leader toolset contains `AgentKick` and the worker toolset does not expose it.
- **Backend paths:** Redis/fakeredis covers created and invited AgentKick paths; SQLAlchemy/SQLite covers created AgentKick and complete roster/session/`AgentRecord` cleanup.
- **Test results:** 92 relevant service/team/message-bus tests passed; 102 Redis/SQL storage tests passed.
- **Static checks:** mypy, Black, flake8, pylint, and all pre-commit hooks passed.

## Checklist

- [x] Has been tested locally (92 relevant service/team/message-bus tests and 102 Redis/SQL storage tests pass)
- [ ] Related documentation has been updated in [documentation repository](https://github.com/agentscope-ai/docs)


DavdGao (commented):
Review of `feat/agentkick`. No correctness bugs found — the pieces I specifically checked and consider **safe**:

- **Removing the reverse-cascade block for invited members in `SessionService.delete_agent`.** The new code leans on `StorageBase.delete_agent`'s defensive roster scrub; both backends (`_redis_storage.py:841-856`, `_sql/_storage.py:533-566`) scrub `members` *and* `member_ids` and run after the session cascade, so the end state matches the old behaviour. `owner_id` is always `self._user_id` in `AgentInvite`, so no cross-user roster is missed.
- **`delete_team_member`'s roster reconciliation** — walked created/invited × legacy/modern rosters; `members`/`member_ids` never diverge into the `members == [] and member_ids != []` state that would let `_ensure_team_members` resurrect phantom members.
- **`_build_member_directory` extraction** is behaviour-identical to the old inline `TeamSay` loop, leader-overwrite ordering included.

Test suite: `1732 passed`; the only failures in my environment were missing `aiosqlite` (`tests/storage_sql_test.py`, `TestAgentKickSQL`) and missing `rg` (`builtin_grep_test.py`) — nothing attributable to this PR.

Four inline comments below, plus one that can't be anchored to the diff:

**Stale tool descriptions (low)** — `src/agentscope/app/_tool/_agent_create.py:165` and `src/agentscope/app/_tool/_agent_invite.py:112` are LLM-facing prompts that still describe a world where `TeamDelete` is the only removal path: *"Members spawned this way live only as long as the team — they are deleted when `TeamDelete` is called"* and *"`TeamDelete` does NOT delete the invited agent"*. `_team_delete.py`'s description was updated to point at `AgentKick`, but these two weren't, so the leader's model of member lifetime is now inconsistent across the toolset.

src/agentscope/app/_service/_session.py
```diff
@@ -377,6 +378,102 @@ async def delete_session(
...
+                for candidate in members
+                if candidate.session_id == member.session_id
+                and candidate.agent_id == member.agent_id
+            ),
+            None,
+        )
+        if current is None:
+            return False
+
+        if current.role == "created":
```
DavdGao, line 496: **Duplicated `delete_session` call (low)**

Both branches of this `if`/`else` call the identical `delete_session(current.owner_id, current.agent_id, current.session_id)`; the only real difference is the extra `delete_agent` in the `created` branch. Suggest collapsing to one unconditional call:

```python
await self.delete_session(
    current.owner_id,
    current.agent_id,
    current.session_id,
)
if current.role == "created":
    await self.delete_agent(current.owner_id, current.agent_id)
```

A duplicated three-arg call across two branches is the kind of thing that drifts on the next edit.

src/agentscope/app/_service/_session.py
```diff
@@ -418,19 +515,12 @@ async def delete_team(self, user_id: str, team_id: str) -> bool:
 
         members = await _ensure_team_members(self._storage, user_id, team)
         for member in members:
```
DavdGao, line 587: **Redundant roster writes during team dissolution (low)**

Routing `delete_team` through `delete_team_member` means each member now pays a `get_team` + `_ensure_team_members` + conditional `upsert_team`, and every *created* member additionally triggers `storage.delete_agent`'s full `list_teams(user_id)` scan plus another roster `upsert_team` — all writes to a record that `storage.delete_team` deletes a few lines later.

Dissolving a 10-member team goes from zero pre-delete team writes to roughly 10 upserts and 10 full team scans. Worth either a "skip roster writeback" path on the shared primitive, or having `delete_team` keep its direct `delete_agent`/`delete_session` calls and letting only `AgentKick` use the reconciling path.

src/agentscope/app/_tool/_agent_kick.py
```diff
@@ -0,0 +1,170 @@
...
+            leader_agent = await self._storage.get_agent(
+                self._user_id,
+                self._agent_id,
+            )
+            leader_name = (
+                leader_agent.data.name
+                if leader_agent is not None
+                else self._agent_id
+            )
+            if target == leader_name:
```
DavdGao: **Self-removal guard keys on a mutable agent name (low)**

This compares `target` against the leader's *agent name*. `AgentCreate` enforces name uniqueness only at creation time, and `PATCH /agents/{id}` (`_router/_agent.py:221`; the owner path in `_access.py:428-430` does not filter `source == "team"`) lets the user rename either the leader or a worker afterwards.

Concrete failure: the leader is renamed to `w1` while a created member is also named `w1`. `AgentKick(target="w1")` then returns "you cannot remove yourself" forever, and the leader's only way to get rid of that member is dissolving the whole team.

A robust guard is to resolve `member` from the directory first and reject when `member.session_id == self._session_id` — the leader is never in `directory` anyway, so the name comparison buys nothing else. Note the same rename hazard makes `_team_routing.py:716` silently drop one of two colliding members from the directory, leaving it both unkickable and unaddressable.

src/agentscope/app/_tool/_agent_kick.py
```diff
@@ -0,0 +1,170 @@
...
+                            f"{team.id}; {effect}."
+                        ),
+                    ),
+                ],
+            )
+        except Exception as e:  # pylint: disable=broad-except
+            return _error(f"AgentKick failed: {e}")
+
+
+def _error(text: str) -> ToolChunk:
```
DavdGao: **Third copy of `_error` (low)**

This is a verbatim copy of the module-level `_error()` in `_agent_invite.py`, and `_team_delete.py` inlines the same `ToolChunk(content=[TextBlock(...)], state=ERROR)` five times. With a third copy landing, this belongs on `_TeamToolBase` (e.g. `self._error(text)`) rather than being re-declared per tool.

src/agentscope/app/_service/_session.py
iluv7: Fixed in 5e11566e. I moved the identical delete_session(...) call before the role check, so both member roles share the same session cleanup path and only created members additionally call delete_agent(...). This preserves the existing behavior while removing the duplicated call and preventing the branches from drifting in future edits.

src/agentscope/app/_service/_session.py
iluv7: Fixed in 5e11566e using the second suggested approach. delete_team now iterates over the roster snapshot and calls delete_agent directly for created members or delete_session for invited members, then delegates to storage.delete_team. The per-member roster reconciliation in delete_team_member is now used only by AgentKick, where the team survives and the writeback is necessary.
I also added a mixed-team regression assertion confirming that invited-member removal no longer rewrites the soon-to-be-deleted roster; the only remaining upsert is the created agent’s storage-level defensive back-reference scrub. Thanks for flagging the unnecessary dissolution path overhead.

src/agentscope/app/_tool/_agent_kick.py
iluv7: Fixed in 5e11566e. I removed the name-based self-removal check and now resolve the target to a TeamMember first, then use member.session_id == self._session_id as the stable identity guard.
I also addressed the related pre-existing routing collision: _build_member_directory now detects duplicate or leader-reserved display names and assigns stable handle-qualified names instead of silently overwriting dictionary entries. AgentKick retains a unique plain-name fallback, so the reported leader -> w1 / existing member w1 scenario successfully removes the member.
Added regression tests covering both kicking the member after the leader rename and keeping the colliding leader and worker independently addressable through TeamSay. Thanks for identifying both sides of the rename hazard.

src/agentscope/app/_tool/_agent_kick.py
iluv7: Fixed in 5e11566e. I moved the shared error-result construction to _TeamToolBase._error() and updated AgentKick, AgentInvite, and TeamDelete to call self._error(...). The duplicated module-level helpers and inline ToolChunk(..., state=ERROR) constructions have been removed, so the common error shape is now maintained in one place.

iluv7:
Thanks for the thorough review and for validating the lifecycle, roster-reconciliation, and directory-extraction behavior.
The stale LLM-facing descriptions are fixed in 5e11566e as well. AgentCreate now explains that a created member can be removed individually with AgentKick or collectively through TeamDelete. AgentInvite now explicitly states that both AgentKick and TeamDelete remove only the invited member’s team-scoped session and preserve the underlying standalone agent. This keeps the member-lifetime guidance consistent across the team toolset.

iluv7:
Resolved conficts.

Issue #2245, opened 2026-08-06
feat(channel): Improve and Enrich Channel Implementation in Agent Service


DavdGao:
**Background**

The basic channel functionality has been established in v2.0.6 (#1997), with support for Feishu and Discord channels. However, further improvements are needed, including:

- Add new channel support, initially planned for WeCom (企业微信) and DingTalk
- Improve existing implementations:
  - `DiscordChannel`: add `list_tools` support to expose corresponding tools
  - Feishu, DingTalk and other IM channels: support adding a channel by scanning a QR code instead of manually entering the app ID and secret key
- Integrate channel with scheduled tasks, enabling scheduled triggers to wake up an agent and complete corresponding tasks in the target channel

**Tasks**

- Add WeCom channel support
- Add DingTalk channel support
- `DiscordChannel`: implement `list_tools`
- QR code-based channel binding for IM channels (Feishu, DingTalk, etc.)
- Integrate channel with scheduled task mechanism

SleepySuke:
Hello @DavdGao  I'd like to pick up this issue as my first contribution for AgentScope if it is available

PR #2266, opened 2026-08-08
feat(channel): add WeCom channel support

### LINKED ISSUE #2246: feat(channel): add WeCom channel support
**Background**

The channel system in Agent Service currently supports Feishu and Discord. WeCom 
(企业微信) is one of the most widely used enterprise IM platforms, and adding WeCom 
channel support will significantly broaden the applicability of Agent Service in 
enterprise scenarios.

**Changes**
- Add a new `WeComChannel` class under `agentscope.app.channel` module (inherited from 
  `ChannelBase`), considering to support (depends on the application SDK side):
  - Multimodal file transfer
  - List conversations and users on the WeCom side
  - Send files and messages to specified users or groups
  - Access the corresponding knowledge base
- Update docs in https://github.com/agentscope/docs

MannXo:
## Summary

Adds `WeComChannel`, bringing WeCom (企业微信) alongside the existing Feishu and Discord channels in Agent Service.

## Context

Closes #2246 (sub-issue of #2245).

WeCom offers two ways for a bot to receive messages. The self-built app route posts encrypted XML to a public HTTPS callback URL, which would need an inbound HTTP route that `ChannelBase` has no hook for, AES-256-CBC decryption, and would not survive a multi-node deployment since the dispatcher only tracks channels running locally.

This PR uses the AI bot long-connection mode instead: a WebSocket to `wss://openws.work.weixin.qq.com`, authenticated with `bot_id` + `secret`. No public endpoint, and it lines up with how the other two channels already work. I flagged the choice on the issue before starting; happy to redo it against the self-built app if that was the intent.

Two platform rules shaped the design:

- A message callback must be answered within five seconds, far less than a run takes. The channel therefore opens a streaming reply the moment a message arrives and hands the stream id to `send_response`, which refreshes it in place within the ten-minute stream budget. Runs with no inbound message, scheduled or background, fall back to a pushed markdown message.
- WeCom keeps exactly one live connection per bot, a new connection evicting the old, so every event for a bot lands in one process. That is what lets the pending-approval table live in memory: the card's button carries a short opaque token rather than round-tripping the lookup keys the way the Feishu card does, because WeCom's button `key` is a single length-limited string.

## Changes

- `src/agentscope/app/channel/_wecom/_channel.py` — the channel: connection supervision with the never-connected case parked in `failed`, inbound normalisation for text, mixed, image, file, voice and video, streaming replies, template-card approvals, and chunked media upload.
- `src/agentscope/app/channel/_wecom/_card_templates.py` — the `button_interaction` approval card and its button-key round trip.
- `src/agentscope/app/channel/_wecom/_tools/` — `SendMessage`, `SendFile`, `SendImage`, all gated behind a permission ask.
- `src/agentscope/app/channel/__init__.py` — export `WeComChannel`.
- `pyproject.toml` — `wecom-aibot-python-sdk>=1.0.2` added to the existing `channel` extra, matching the `lark-oapi` and `discord.py` precedent. It is the official WeCom SDK and brings the media download and AES decryption with it.
- `examples/agent_service/main.py` — register the new channel alongside the other two.

## Testing

- `pytest tests/channel_wecom_test.py` — 25 new tests, all passing. A fake WSClient stands in for the platform, so they cover inbound normalisation, the stream being opened before the gateway sees the message (the five-second rule), throttled refreshes, the push fallback, attachment upload, and the approval-card click round trip.
- `pytest tests` — 1793 passed, 120 skipped, 1 failed. The failure is `mcp_sse_client_test.py::SseSchemaDefsPreservationTest::test_defs_preserved_and_titles_stripped`, which fails identically on a clean `main` and is unrelated to this change.
- `pre-commit run` over every touched file — all hooks clean.

## Notes for reviewers

Three things worth a look:

- The two bullets on the issue about listing conversations and users, and reaching the knowledge base, are not implemented. The AI bot long-connection API exposes no chat or member lookup, so there are no discovery tools and `list_bot_chats` is left at the base default, which makes routing config fall back to the chat ids recorded from inbound messages. Those bullets look like they need the app-side APIs rather than the bot connection, so I would rather leave them out than stub them.
- The docs say `aibot_send_msg` takes `markdown` and `template_card`, but the media upload section also lists it as a consumer of `media_id` for `file` and `image`. I implemented attachment sending against the latter reading. If the platform rejects it, the error surfaces through the tool result rather than failing silently, but this is the part I am least able to verify without a real bot.
- `only_at_reply` is deliberately absent from `Config`. The inbound callback carries no mention information, so unlike Feishu and Discord there is nothing to gate on.

End-to-end behaviour against a live bot is untested — I have no WeCom tenant. The streaming refresh, card update and media upload paths are the ones that would benefit most from a check against a real bot.


MannXo:
Merged with main, so this is mergeable again and CI is green. Before anyone reads that green as ready, there is a design problem I would rather raise than paper over.

## The problem

Since this PR was opened, long connections moved into a dedicated worker and `ChannelClients` began building channel instances that never call `start_listening`, on the reasoning that outbound is plain REST so whichever node runs the agent can deliver the reply itself. That holds for Feishu, Discord and Slack, and #2409 gave DingTalk the same property explicitly.

It does not hold for WeCom. In AI bot long-connection mode every outbound operation is a frame on the socket: `aibot_respond_msg` for the streaming reply, `aibot_respond_update_msg` for a card update, `aibot_send_msg` for a proactive push, and the three `aibot_upload_media_*` commands for attachments.

I checked whether there is an HTTP path that would let a non-connected node send. The AI bot does expose one, `POST /cgi-bin/aibot/response`, but it is authorised by a `response_code` that arrives on an inbound event, it is single-use, it expires after an hour, and it cannot push without an inbound trigger. So there is no equivalent of DingTalk's OpenAPI: with only `bot_id` and `secret`, a process that holds no connection cannot send anything.

Three things break as a result:

1. A `ChannelClients`-built instance has no client, so every send from the agent node fails.
2. `send_response` reads a per-chat streaming handle that `_on_message` stored on the instance. Those now run in different processes, so the lookup always misses.
3. WeCom requires the first reply within five seconds of the message callback. This PR meets that by opening the stream on arrival and letting `send_response` refresh it, which only works while both run in the same process.

## Options

I do not think this is mine to decide, so rather than pick one:

1. Let a channel declare that its replies must be delivered from the node holding its connection, and keep the old forwarding path for those. Narrow, but reintroduces something #2395 deliberately removed.
2. Route WeCom's outbound frames from the agent node to the worker over the message bus. No change to the delivery model, at the cost of a new hop and its failure modes.
3. Move WeCom to callback mode instead, which needs a public HTTPS endpoint and `ChannelBase` has no hook for that today. It also inherits the single-use `response_code` constraint above, so it does not obviously solve the proactive-send case either.

My preference is (1) if you are open to a per-channel capability flag, since it is honest about the platform rather than working around it, and (2) if you would rather keep one delivery model for every channel.

Worth flagging that the tests here pass either way, because they inject the client directly. That is on me and I will fix it once the direction is settled, since a test that cannot see this is not worth much.

Happy to implement whichever you prefer.

MannXo:
Merged main (d2fd1bdf, no conflicts) and fixed the same `list_tools` contract break you found on #2267. `ChatService` now calls `channel.list_tools(workspace, channel_user_id)`, and the WeCom override still took only the workspace, so every WeCom run would have raised `TypeError` building its toolkit.

af0f11cd adds `channel_user_id: str | None = None` and appends `await super().list_tools(workspace, channel_user_id)`, as Feishu and DingTalk do (`_wecom/_channel.py:780-804`). A new `ListToolsTest` calls it with a session user (`tests/channel_wecom_test.py:527`) and fails with that `TypeError` on the previous head. Full suite 2548 passed, 0 failed, and pre-commit is clean on both files.

PR #2267, opened 2026-08-08
feat(channel): add Slack channel support

### LINKED ISSUE #2256: feat(channel): add Slack channel support
**Background**

The channel system in Agent Service currently supports Feishu and Discord. Slack is one of the most widely used team collaboration platforms in enterprise environments, and adding Slack channel support will make Agent Service more accessible in professional and developer-oriented workflows.

**Changes**

- Add a new `SlackChannel` class under `agentscope.app.channel` module (inherited from `ChannelBase`), considering to support (depends on the Slack API side):
  - `ListChats`: list all available channels/conversations accessible by the bot
  - `ListChatMembers`: list all members in a specified channel or conversation
  - `SendMessage`: send a text message to a specified channel or user
  - `SendFile`: send a file attachment to a specified channel or user
  - `SendImage`: send an image to a specified channel or user
  - More tools to be considered based on the capabilities provided by the Slack API

MannXo:
## Summary

Adds `SlackChannel`, bringing Slack alongside the existing Feishu and Discord channels in Agent Service.

## Context

Closes #2256 (sub-issue of #2245).

Slack offers two ways to receive events: the Events API, which needs a publicly reachable request URL, and Socket Mode, which has the app hold a WebSocket out to Slack. This uses Socket Mode, so the service needs no public endpoint and the channel runs on the app event loop like `DiscordChannel`, with no thread bridging.

One detail worth knowing when reading the code: Socket Mode wants each envelope acknowledged within three seconds, but that ack is entirely separate from the user-visible reply. So a run is free to take as long as it takes. The channel acks first and unconditionally, then posts one message on the run's first output and edits it in place with `chat.update` as more arrives.

## Changes

- `src/agentscope/app/channel/_slack/_channel.py` — the channel: Socket Mode lifecycle, inbound normalisation with mention gating, streaming replies, Block Kit approvals, and the discovery/send API calls.
- `src/agentscope/app/channel/_slack/_card_templates.py` — the approval card and its button-value round trip.
- `src/agentscope/app/channel/_slack/_tools/` — `ListChats`, `ListChatMembers`, `SendMessage`, `SendFile`, `SendImage`.
- `src/agentscope/app/channel/__init__.py` — export `SlackChannel`.
- `pyproject.toml` — `slack_sdk>=3.27` added to the existing `channel` extra, matching the `lark-oapi` and `discord.py` precedent.
- `examples/agent_service/main.py` — register the new channel alongside the other two.

All five tools the issue lists are backed by real API methods, so none of them are stubs: `conversations.list` and `conversations.members` (both cursor-paginated) for discovery, `chat.postMessage` and `files_upload_v2` for sending. Slack opens a direct message when you post to a user id, so a single `chat_id` parameter addresses both channels and people and the tools need no id-type argument.

Approval cards carry their lookup keys in the button `value`, which Slack allows 2000 characters for. That means a click resolves with no in-process state, so the channel stays correct when several nodes each hold a Socket Mode connection for the same app. The authoritative tool call is still read from session state on resume, never trusted from the card.

## Testing

- `pytest tests/channel_slack_test.py` — 22 new tests, all passing. A fake `AsyncWebClient` stands in for the platform, covering inbound normalisation, mention gating in channels versus direct messages, dropping the bot's own messages and non-user subtypes, the post-then-edit streaming reply and its throttle, long-reply splitting, attachment upload, cursor pagination on both discovery calls, and the approval-card click round trip.
- `pytest tests` — 1792 passed, 120 skipped, 1 failed. The failure is `mcp_sse_client_test.py::SseSchemaDefsPreservationTest::test_defs_preserved_and_titles_stripped`, which fails identically on a clean `main` and is unrelated to this change.
- `pre-commit run` over every touched file — all hooks clean, verified with black 23.3.0 on Python 3.11 to match CI.

## Notes for reviewers

- `only_at_reply` gates on the bot's own `<@id>` appearing in the message text, and fails closed if `auth.test` never resolved the bot's identity. Direct messages are never gated. The mention markup is stripped before the text reaches the agent.
- The channel subscribes to `message` events rather than `app_mention`, so mentions are not delivered twice; the gate is applied in `_normalize` instead.
- Replies are posted to the conversation rather than threaded under the triggering message. `send_response` receives a synthetic `ChannelEvent` from the dispatcher that carries no `channel_message_id`, so there is nothing to thread from without adding per-chat state. This matches what `FeishuChannel` effectively does today.
- The agent's output is sent as-is. Slack renders mrkdwn rather than standard Markdown, so some formatting will differ; converting between the two felt like scope for a separate change.

One thing to be upfront about: this has not been run against a live workspace. Everything is written against the documented API and the SDK's own signatures, and the tests cover the logic, but the streaming edits, card interactions and file uploads have not been exercised against a real Slack app. Those three are where I would focus a review, and where a quick check against a real workspace would be worth more than anything I can assert here.


src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,915 @@
...
+
+        Args:
+            client (`SocketModeClient`): The connection the request came in
+                on, used to send the acknowledgement.
+            req (`SocketModeRequest`): The envelope.
+        """
+        from slack_sdk.socket_mode.response import SocketModeResponse
+
+        try:
+            await client.send_socket_mode_response(
```
qbc2016, line 462: `_on_request` correctly acknowledges the envelope before downloads and normalization, so those operations do not consume the current envelope’s three-second acknowledgement window. However, the acknowledgement still happens before any durable handoff, and acknowledgement failures are swallowed while processing continues.

This leaves two real failure modes:

- A process crash after a successful acknowledgement but before `_emit` permanently loses the event.
- If the acknowledgement fails or is not observed by Slack, Slack may retry while the first delivery continues processing, potentially starting the same Agent run more than once.

This is a framework-wide channel reliability gap rather than a Slack-only issue. For this PR, please at least preserve `payload.event_id`, `envelope_id`, and retry metadata instead of discarding them.

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,915 @@
...
+        self.status.state = "connecting"
+        self._web = AsyncWebClient(token=self._bot_token)
+        # Identify ourselves up front: needed to drop our own messages and
+        # to recognise an @mention. A bad bot token fails here, which is a
+        # cleaner signal than the socket's silent retry loop.
+        if not await self._load_identity():
+            self.status.state = "failed"
+            while not self._stopped:
+                await asyncio.sleep(30.0)
+            return
```
qbc2016: The authentication-failure path parks outside the lifecycle `try/finally`. If the listener task is cancelled while parked, it does not reach the normal finalizer, so `status.state` remains `"failed"` and `_web` is not cleared.

This is a lifecycle consistency problem, although it does not currently demonstrate an open HTTP connection leak: `AsyncWebClient` is created without a persistent `aiohttp` session, and `SocketModeClient` has not yet been constructed on this path.

Please place identity loading and Socket Mode client creation under one outer `try/finally`, using optional local variables so partially initialized resources can be cleaned up safely. The finalizer should consistently set the terminal state and clear `_client` and `_web`.

Please also test cancellation during failed authentication and failure during Socket Mode client initialization.

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,915 @@
...
+            chat_id (`str`): A conversation id, or a user id for a DM.
+            text (`str`): The message text.
+
+        Returns:
+            `dict`: ``{"ok": bool}``, with ``"error"`` when it failed.
+        """
+        ts = await self._post(chat_id, text[:_MAX_LEN])
+        if ts is None:
+            return {"ok": False, "error": self.status.last_error or "failed"}
+        return {"ok": True}
```
qbc2016: `send_message_to` sends only `text[:_MAX_LEN]` and then returns `{"ok": true}`. Everything after `_MAX_LEN` is silently discarded, so the Agent cannot tell that only part of the requested message was delivered.

The normal reply path already uses `_split_long_message`, so the proactive-message path should reuse the same behavior. Please return the timestamps of all successfully posted segments for auditability.

If a later segment fails, return partial-success information containing:

- the successfully sent timestamps;
- the failed segment index;
- the platform error.

Rejecting the entire call with an explicit length error would also be preferable to silent truncation, although splitting would be more consistent with the existing reply path. The implementation should also respect Slack’s per-channel message rate limits.

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,915 @@
...
+        )
+        if parsed is None:
+            return
+        tool_call_id, chat_id, approved, agent_id, session_id = parsed
+        if self._emit:
+            await self._emit(
+                ChannelConfirmationResultEvent(
+                    channel_id=self._channel_id,
+                    chat_id=chat_id,
+                    channel_user_id=(payload.get("user") or {}).get("id", ""),
```
qbc2016: The Slack adapter records the actual clicker in `channel_user_id`, but leaves `event.actor` empty. More importantly, the core decision path does not authorize that user: when the card contains a pinned `agent_id` and `session_id`, the clicker’s identity is not used by `resume_after_decision`.

As a result, any group member who can see and interact with the approval card can approve the pending tool call. Feishu and Discord currently have the same behavior, so this should not be fixed with Slack-specific authorization logic.

Please define an explicit approval policy:

- initiating user only;
- any group member;
- administrator or role-based approval; or
- another documented policy.

If approval is restricted, persist the expected actor or approval policy with the pending confirmation, pass the actual actor into `resume_after_decision`, and enforce authorization in the core decision path for all adapters.

If group-wide approval is intentional, make that policy explicit and persist the actual approver for auditing. Merely populating `event.actor` without enforcing or recording the policy would not resolve the security concern.

qbc2016 (commented):

src/agentscope/app/channel/_slack/_channel.py
MannXo: Done. The event now carries `event_id`, `envelope_id`, `retry_attempt` and `retry_reason` through to its metadata, so a redelivery stays identifiable downstream, and a failed acknowledgement logs a warning instead of passing silently. Processing still continues after a failed ack, on the grounds that dropping the event turns a possible duplicate into a certain loss.

I left the deduplication itself alone. As you say it needs state shared across nodes to be correct, and that felt like yours to place rather than mine to bolt onto one adapter.

src/agentscope/app/channel/_slack/_channel.py
MannXo: Fixed. Identity loading and Socket Mode client construction are now inside one outer `try/finally` with optional locals, so a partially initialised start-up is cleaned up by the same finalizer, which sets the terminal state and clears both `_client` and `_web` on every path. The two give-up paths share a small `_park` helper rather than each inlining their own sleep loop.

Both cases you asked for are covered by tests: cancellation while parked on a rejected bot token, and a `SocketModeClient` constructor that raises.

src/agentscope/app/channel/_slack/_channel.py
MannXo: Fixed, and thank you, that was a real bug. `send_message_to` now splits with `_split_long_message` exactly as the reply path does, pauses between segments to respect the per-channel write limit, and returns `sent_ts` for every segment that landed.

On a failure part-way through it returns the timestamps that did land, the failing segment index and the platform error. The tool layer renders that as "failed after 1 of 3 messages", so the agent cannot read a partial delivery as a complete one.

src/agentscope/app/channel/_slack/_channel.py
MannXo: Agreed, and I do not think I should be the one to resolve it in this PR.

Choosing between initiator-only, group-wide and role-based approval is a product decision, and enforcing it means changing the card contract and `resume_after_decision`, which lands on Feishu and Discord as much as on Slack. Making that call inside a channel-addition PR would change their behaviour on my say-so, which seems like the wrong way round.

What I have done here is populate `actor` with the clicker. No adapter was setting it, so the information was simply being discarded. I am not claiming that closes the gap, only that it stops throwing away the one thing any of the options will need.

Happy to do the rest as a follow-up across all three adapters once you say which policy you want: persist the expected actor or policy alongside the pending confirmation, pass the real actor into `resume_after_decision`, and enforce it there so every adapter inherits it. Glad to open an issue for it if that is easier to track.

src/agentscope/app/channel/_slack/_channel.py
MannXo: Done, the event now carries `event_id`, `envelope_id` and the retry counters, and a failed ack logs a warning.

src/agentscope/app/channel/_slack/_channel.py
MannXo: Fixed, identity loading and client construction are now inside one `try/finally`, with tests for cancellation while parked and for a failing client constructor.

src/agentscope/app/channel/_slack/_channel.py
MannXo: Fixed, it splits like the reply path now and returns the timestamp of every segment plus the failing index on a partial send.

src/agentscope/app/channel/_slack/_channel.py
MannXo: Populated `actor`, but left the policy itself to you since enforcing it changes Feishu and Discord too. Happy to do that as a follow-up once you pick one.

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,991 @@
...
+                    self.status.last_error = (
+                        "socket did not connect; check the app-level token "
+                        "and that Socket Mode is enabled"
+                    )
+                    logger.error(
+                        "Slack '%s' giving up: %s",
+                        self._channel_id,
+                        self.status.last_error,
+                    )
+                    await self._park()
```
qbc2016, line 277: **Cancel the connection task before parking the failed channel**

At this point, `connect_task` is still running. `SocketModeClient.connect()` retries indefinitely, while `_park()` does not return until the channel is stopped. As a result, the cancellation in `finally` is never reached while the channel is parked.

The channel therefore continues attempting to connect after its status becomes `failed`. If a later retry succeeds, it can even start processing events while the dispatcher and management UI still consider it failed.

Please cancel and await `connect_task`, and close the Socket Mode client, before entering the parked state. Please also add a test verifying that the pending connection task is cancelled after the initial connection deadline.

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,991 @@
...
+        Args:
+            chat_id (`str`): The conversation to write to.
+            ts (`str | None`): The streamed message's timestamp, if any.
+            text (`str`): The complete reply text.
+        """
+        parts = [p for p in self._split_long_message(text) if p]
+        if not parts:
+            return
+        if ts is not None:
+            await self._update(chat_id, ts, parts[0])
```
qbc2016: **Preserve the first segment when the final update fails**

`_update()` catches all exceptions and does not report whether the update succeeded. This code nevertheless removes `parts[0]` immediately afterward.

If the final `chat.update` fails because of rate limiting or a transient Slack error, the streamed message remains stale, while the complete first segment is discarded and only later segments may be posted. The user therefore receives an incomplete reply even though forwarding is treated as complete.

Please make `_update()` return a success result or propagate the failure, and fall back to posting `parts[0]` as a new message when the final update cannot be completed.

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,991 @@
...
+            parts = parts[1:]
+        for part in parts:
```
qbc2016: **Throttle and check every long-reply follow-up**

All remaining segments are posted back-to-back, even though Slack generally permits approximately one message per second per channel. `_post()` returns `None` on failure, but that result is ignored here, so a rate-limited or rejected segment is silently lost and later segments continue to be sent.

Please reuse the throttled segmented-send behavior from `send_message_to()`, and stop or explicitly report partial delivery when any segment fails. Slack `Retry-After` responses should be honored where possible.

qbc2016 (commented):

src/agentscope/app/channel/_slack/_channel.py
MannXo: reflected

src/agentscope/app/channel/_slack/_channel.py
MannXo: addresed

src/agentscope/app/channel/_slack/_channel.py
MannXo: addressed

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,1078 @@
...
+        # Everything below is optional until it is built, so the finalizer
+        # can clean up whatever a partial start-up managed to create.
+        client: Any = None
+        connect_task: asyncio.Task | None = None
+        try:
+            self._web = AsyncWebClient(
+                token=self._bot_token,
+                # Lets the SDK wait out a 429 for us, honouring the
+                # Retry-After header rather than dropping the call.
+                retry_handlers=[
```
qbc2016: **Preserve the SDK's default connection retry handler**

Passing an explicit `retry_handlers` list replaces, rather than extends, the handlers installed by `AsyncWebClient`. The SDK's default list contains `AsyncConnectionErrorRetryHandler`, so this configuration removes retries for transient connection failures such as `ServerConnectionError`, `ServerDisconnectedError`, and `ClientOSError`.

This is particularly harmful during startup: if `auth.test` encounters a transient connection failure, `_load_identity()` treats it as an invalid bot token, moves the channel to `failed`, and parks it until the channel is edited.

Please append `AsyncRateLimitErrorRetryHandler` to the client's existing handlers, or include `*async_default_handlers()` in the explicit list. Please also add a test verifying that both the default connection-error handler and the rate-limit handler are installed.

qbc2016 (commented):

src/agentscope/app/channel/_slack/_channel.py
MannXo: done

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,1118 @@
...
+            `str`: ``_IDENTITY_OK`` when the identity is known,
+            ``_IDENTITY_REFUSED`` when Slack answered and rejected the
+            token, ``_IDENTITY_UNAVAILABLE`` when the call never got an
+            answer.
+        """
+        from slack_sdk.errors import SlackApiError
+
+        try:
+            auth = await self._web.auth_test()
+        except SlackApiError as e:
```
qbc2016, line 392: **Do not treat every SlackApiError as a credential refusal**

`SlackApiError` is raised for any unsuccessful Web API response, not only invalid credentials. In particular, `auth.test` can return transient errors such as `internal_error`, `ratelimited`, `request_timeout`, and `service_unavailable`.

With the current branch, an exhausted 429 retry or a temporary Slack server error returns `_IDENTITY_REFUSED`, causing `start_listening()` to park the channel permanently until its configuration is edited. This reintroduces the startup availability failure this change is intended to avoid.

Please return `_IDENTITY_REFUSED` only for terminal authentication or configuration errors such as `invalid_auth`, `not_authed`, `token_revoked`, and `account_inactive`. Other `SlackApiError` responses should be treated as `_IDENTITY_UNAVAILABLE`. Please also add a test covering a transient `SlackApiError`, such as a 429 or `service_unavailable`.

tests/channel_slack_test.py
```diff
@@ -0,0 +1,877 @@
+# -*- coding: utf-8 -*-
```
qbc2016, line 1: Please prefer exact `assertEqual` assertions on the complete lifecycle outcome.

For example, `test_refused_token_parks_the_channel` currently checks only an error substring and does not actually verify that the channel entered the expected `failed`/parked state. Similarly, `assertNotEqual(channel.status.state, "failed")` would pass for any unexpected state.

Please assert the complete expected status/result structure where practical. This will make these tests catch unexpected fields and state transitions instead of validating only selected properties.

qbc2016 (commented):

tests/channel_slack_test.py
MannXo: done

src/agentscope/app/channel/_slack/_channel.py
MannXo: reflected

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,1216 @@
...
+                    tool.name,
+                    tool.input,
+                    event.metadata.get("agent_id", ""),
+                    event.metadata.get("session_id", ""),
+                ),
+            )
+
+    # -- Agent-callable tools --
+
+    async def list_tools(
```
qbc2016: **[P1] Match the current `ChannelBase.list_tools` contract**

`ChatService` on the current main branch always calls `channel.list_tools(workspace, channel_user_id_or_none)`, but this override accepts only `workspace`. Consequently, every Slack-originated run fails during toolkit construction with:

`TypeError: SlackChannel.list_tools() takes 2 positional arguments but 3 were given`

This happens before the agent can produce a reply. Please add the optional `channel_user_id` parameter, compose the result with `await super().list_tools(workspace, channel_user_id)` as the current Feishu and DingTalk adapters do, and add an integration test that invokes this method through `ChatService`.

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,1216 @@
...
+
+        Returns:
+            `dict`: ``{"ok": bool}``, with ``"error"`` when it failed.
+        """
+        web = self._web_client()
+        if web is None:
+            return {"ok": False, "error": "slack_sdk is not installed"}
+        try:
+            await web.files_upload_v2(
+                channel=chat_id,
```
qbc2016: **[P1] Resolve user targets correctly before uploading files**

`ListChatMembers` returns `U...` user IDs and explicitly tells the agent to pass them directly to any `Send*` tool. However, `files_upload_v2(channel=chat_id)` maps `channel` to the `files.completeUploadExternal.channel_id` field, which Slack documents as a channel ID. User IDs are supported by the separate `channels` field, not by `channel_id`.

As a result, `SendFile` and `SendImage` cannot reliably deliver to the person targets produced by `ListChatMembers`, even though `SendMessage` can open a DM from the same user ID. Please either pass user targets through `channels=[user_id]` or call `conversations.open` and upload to the returned `D...` conversation ID. Please also add a test that inspects the final `files.completeUploadExternal` parameters for a `U...` target.

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,1216 @@
...
+    _build_approval_blocks,
+    _parse_action,
+    _resolved_blocks,
+)
+
+if TYPE_CHECKING:
+    from slack_sdk.socket_mode.aiohttp import SocketModeClient
+    from slack_sdk.socket_mode.request import SocketModeRequest
+    from .....tool import ToolBase
+    from .....workspace import WorkspaceBase
```
qbc2016: **[P2] Fix the relative imports in the `TYPE_CHECKING` block**

This module belongs to the `agentscope.app.channel._slack` package, so four leading dots resolve to the top-level `agentscope` package. Five dots attempt to traverse beyond the package root.

These imports are hidden at runtime because they are guarded by `TYPE_CHECKING`, which is why the unit tests do not expose the problem, but type checkers and IDEs cannot resolve them correctly.

Please change both imports to:

```python
from ....tool import ToolBase
from ....workspace import WorkspaceBase
```

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,1216 @@
...
+                resp = await web.conversations_list(
+                    types="public_channel,private_channel,mpim,im",
+                    exclude_archived=True,
+                    limit=200,
+                    cursor=cursor or None,
+                )
+            except Exception:  # pylint: disable=broad-except
+                logger.debug("Slack conversations.list failed")
+                break
+            for item in resp.get("channels") or []:
```
qbc2016: **[P2] Do not advertise non-member channels as universal send targets**

`conversations.list` can return public channels for which `is_member` is false. This loop discards that field and exposes every result as a ready-to-send target, while `ListChats` tells the agent that the ID can be copied into any `Send*` tool.

File uploads require the bot to be a member of the target channel, and text posting to a non-member public channel requires the additional `chat:write.public` scope. Therefore some IDs returned here immediately fail when used as documented.

Please filter out conversations that the bot cannot use with all of the advertised send tools, or preserve explicit per-operation capabilities instead of presenting every result as universally writable. Add a test covering a public channel with `is_member=False`.

src/agentscope/app/channel/_slack/_channel.py
```diff
@@ -0,0 +1,1216 @@
...
+                    cursor=cursor or None,
+                )
+            except Exception:  # pylint: disable=broad-except
+                logger.debug("Slack conversations.members failed")
+                break
+            for user_id in resp.get("members") or []:
+                results.append(
+                    {
+                        "user_id": user_id,
+                        "name": await self._user_name(user_id),
```
qbc2016: **[P2] Bound the profile lookups performed by `ListChatMembers`**

`conversations.members` returns only user IDs, so resolving profile details separately is reasonable. However, this implementation automatically traverses every membership page and performs one sequential `users.info` request for every uncached member. There is no query, result limit, or cursor exposed by the tool, so one invocation on a large channel can issue hundreds or thousands of Web API requests.

This is different from the DingTalk user-discovery path, which bounds the similar per-user detail lookup to at most 50 results. When a Slack lookup fails after retries, `_user_name()` also converts the failure into an empty name, reducing the usefulness of the returned results.

Please place a clear bound on the work performed by one invocation—for example, expose pagination with a limited page size, add a query/result limit, or bulk-resolve profiles with `users.list`. Please add a test covering a channel whose membership exceeds one page.

src/agentscope/app/channel/_slack/_card_templates.py
```diff
@@ -0,0 +1,148 @@
...
+        label (`str`): The button's visible text.
+        style (`str`): Slack button style (``primary`` / ``danger``).
+        value (`dict`): The payload echoed back on click.
+    """
+    return {
+        "type": "button",
+        "text": {"type": "plain_text", "text": label},
+        "style": style,
+        "action_id": f"{_ACTION_TYPE}:{value['action']}",
+        "value": json.dumps(value, ensure_ascii=False)[:_VALUE_LIMIT],
```
qbc2016: **[P3] Keep oversized button values parseable**

Slack limits a button's `value` to 2,000 characters, but slicing the serialized payload can cut inside a string or remove the closing braces.
The resulting button is still displayed, while `_parse_action()` cannot decode its value and silently ignores the approval click.

Please reject oversized payloads before posting the card, or use a bounded representation that remains valid JSON. The test should also verify that the final value successfully round-trips through `_parse_action()`.

qbc2016 (commented):

MannXo:
@qbc2016 all six are addressed and pushed. I merged main in first (ee59179a, no conflicts), then made one commit per thread.

## 1. [P1] list_tools contract (thread on _channel.py:1052)

Fixed in f2017e99. `list_tools` now takes `channel_user_id: str | None = None` and ends with `+ await super().list_tools(workspace, channel_user_id)`, the same shape as Feishu and DingTalk (`_slack/_channel.py:1110-1142`).

Tests. `ListToolsTest` calls `list_tools(ws, "U1")` directly (`tests/channel_slack_test.py:1150`). Through ChatService, the delivery harness in `tests/service_chat_channel_delivery_test.py` now takes a channel class, and two new cases run a real `SlackChannel` behind a fake Web client (`:420`, `:435`). A DM run reaches `list_tools` with `"U1"`, a channel run with `None`, and both get the five Slack tools. All three fail on the previous head with the `TypeError` you quoted.

## 2. [P1] User targets for file uploads (thread on _channel.py:1135)

Fixed in 2add3cba. A `U...`/`W...` target is now resolved to its `D...` conversation with `conversations.open` before `files_upload_v2(channel=...)`, in a small `_open_dm` helper (`_slack/_channel.py:1218`). `send_message_to` resolves the same way, so text and files to one person land in the same DM. Conversation ids (`C`/`G`/`D`) skip the call. This needs the `im:write` bot scope, now noted in the module docstring (`_channel.py:18-20`).

I went with `conversations.open` over `channels=[user_id]` because `files_upload_v2` only gained `channels=` in slack_sdk 3.35, and the extra's floor is `slack_sdk>=3.27`. `conversations_open(users=...)` exists in 3.27.

The test runs the real `AsyncWebClient.files_upload_v2` with only `api_call` and the upload POST faked, then reads the final `files.completeUploadExternal` params. For `U1` it's `conversations.open`, then `files.getUploadURLExternal`, then `files.completeUploadExternal` with `channel_id="D1"`. For `C1` there's no `conversations.open` and `channel_id="C1"` (`tests/channel_slack_test.py:1206`).

## 3. [P2] TYPE_CHECKING imports (thread on _channel.py:47)

Fixed in c8835cae. Both are now `from ....tool` / `from ....workspace` (`_slack/_channel.py:50-51`). The `_tools/*.py` modules sit one package deeper, so their five dots were already right.

## 4. [P2] Non-member channels in ListChats (thread on _channel.py:797)

Fixed in a5c3c709. Listing now uses `users.conversations`, which returns only conversations the bot is a member of, with the same types and cursor (`_slack/_channel.py:798-841`). The routing-config listing (`list_bot_chats`) uses it too.

I also gave ListChats the same bound as ListChatMembers, `limit` (default 20, 1 to 50) plus an optional `cursor` (`_tools/_list_chats.py:18-28`). One call is one page. This changes the output shape from a bare array to `{"chats": [...], "next_cursor": ...}`, where `next_cursor` is null on the last page. The description says so, and also that `query` filters only the returned page.

Tests. The fake's `conversations_list` returns an `is_member: false` public channel, and ListChats never shows it (`tests/channel_slack_test.py:510`). Paging returns one page plus its cursor, and the cursor round-trips (`:528`).

## 5. [P2] Bounded ListChatMembers (thread on _channel.py:841)

Fixed in e0e62b62. ListChatMembers now takes `limit` (default 20, 1 to 50, as in DingTalk's `ListUsers`) and an optional `cursor` (`_tools/_list_chat_members.py:16-27`). It makes one `conversations.members` call with that limit and resolves names for that page only, so `users.info` runs at most `limit` times per call (`_slack/_channel.py:843-890`). A name that can't be resolved is now null, not `""`.

The output shape changes from a bare array to `{"members": [...], "next_cursor": ...}`. The description is updated.

The test uses a two-page channel. Page one comes back with `next_cursor`, there are at most 3 `users.info` lookups for `limit=3`, an unresolvable member has `name: null`, and the cursor fetches page two with `next_cursor: null` (`tests/channel_slack_test.py:560`).

## 6. [P3] Oversized button values (thread on _card_templates.py:96)

Fixed in cc6d5413. `_button` now raises `ValueError` when the serialized value is over 2000 characters, instead of slicing it (`_slack/_card_templates.py:86-111`). `_present_confirm` catches that, logs it, and posts a plain-text notice with no buttons, saying the request is too large for a Slack approval card (`_slack/_channel.py:1066-1107`).

Tests. A value near the limit still parses through `_parse_action`. An oversized one raises. `_present_confirm` with an oversized value posts one message with no `blocks` at all, so a cut value never goes out (`tests/channel_slack_test.py:445-485`).

PR #2269, opened 2026-08-08
fix(channel): fail open Discord chat helpers

### LINKED ISSUE #2268: [Bug]: Discord channel helpers crash before the client is ready
## Problem

`DiscordChannel` exposes best-effort management/context helpers such as `list_bot_chats`, `chat_name`, and `chat_kind`. These helpers currently assume the Discord client has already been initialized and that platform channel lookup succeeds.

However, `ChannelLifecycleDispatcher._start()` registers a channel instance immediately after scheduling `start_listening()`. During that startup window, `DiscordChannel._client` is still `None`, but the management API can already call `GET /channels/{channel_id}/chat_ids`, which delegates to `list_bot_chats()`.

`ChatService` also calls `chat_kind()` and `chat_name()` when building context for channel-bound sessions. If the Discord client is not ready, or if `fetch_channel()` fails for an unresolved chat id, these optional helpers can raise and fail the current request/run instead of treating the chat as unknown.

## Reproduction

On current `main` (`29b592358c2e983a0d10dd5227316b7a02d8c23a`), constructing a `DiscordChannel` without starting the gateway client and calling the helper methods produces:

- `list_bot_chats()` -> `AttributeError: 'NoneType' object has no attribute 'guilds'`
- `chat_name("123")` -> `AttributeError: 'NoneType' object has no attribute 'get_channel'`
- when `fetch_channel()` raises, `chat_name()` propagates that exception

## Expected behavior

These helpers are optional/best-effort. When the Discord client is not ready or a channel lookup fails, they should fail open:

- `list_bot_chats()` returns `[]`
- `chat_name(...)` returns `""`
- `chat_kind(...)` returns `None`

This matches the base channel helper contract and prevents management/API context paths from failing on transient Discord availability.

## Proposed scope

Keep the change focused to Discord helper behavior and regression coverage. Do not implement Discord agent tools or change message send/listen behavior.

helloJamest:
## Summary

- Make Discord channel best-effort helpers fail open when the gateway client is not ready.
- Treat failed Discord channel lookup as an unresolved chat id instead of propagating the exception.
- Add regression coverage for cold Discord channels and lookup failures.

## Root Cause

`DiscordChannel.list_bot_chats`, `chat_name`, and `chat_kind` assumed `_client` was initialized. `ChannelLifecycleDispatcher` registers a channel instance right after scheduling `start_listening()`, so management/context paths can call these helpers before the Discord client exists. A failed `fetch_channel()` lookup could also bubble out of optional helper code.

Closes #2268

## Test Plan

- `PYTHONPATH=src python -m pytest tests/channel_discord_test.py tests/channel_gateway_test.py tests/channel_routing_test.py -q` (`23 passed`)
- `pre-commit run --files src/agentscope/app/channel/_discord/_channel.py tests/channel_discord_test.py`


iamroylim (commented):
At 92dfe57, I can reproduce that an unexpected AssertionError from client.guilds and a RuntimeError from fetch_channel are both converted to [] or None; both debug records also have exc_info=False. That makes a programmer bug or real Discord outage indistinguishable from the intended cold-client or unresolved-id case. Could we keep the fail-open contract but narrow the lookup catch to expected discord.py failures, and either re-raise unexpected exceptions or at least log them with traceback? A regression test that separates expected lookup failures from unexpected exceptions would lock down that boundary.

helloJamest:
Addressed in `b4924bd`.

- kept the cold-client fail-open behavior unchanged;
- narrowed swallowed lookup failures to expected discord.py lookup exceptions (`NotFound`, `Forbidden`, `HTTPException`);
- unexpected errors such as `AssertionError` from `guilds` or `RuntimeError` from `fetch_channel` now propagate instead of being converted to empty values;
- expected lookup failures are logged with `exc_info=True`;
- added regressions for expected lookup fail-open, traceback logging, unexpected lookup failures, and unexpected guild failures.

Validation:
- `PYTHONPATH=src python -m pytest tests/channel_discord_test.py tests/channel_gateway_test.py tests/channel_routing_test.py -q`
- `pre-commit run --files src/agentscope/app/channel/_discord/_channel.py tests/channel_discord_test.py`

iamroylim (approved):
Verified at b4924bd. The follow-up narrows fail-open behavior to discord.py lookup errors, preserves traceback context, and lets unexpected RuntimeError/AssertionError failures propagate. I reran the three affected channel suites (26 passed) and the full pre-commit file set; all checks passed. This resolves my earlier concern.

PR #2270, opened 2026-08-09
feat(channel): add WhatsApp Business API support

### LINKED ISSUE #2254: feat(channel): add WhatsApp channel support
**Background**

The channel system in Agent Service currently supports Feishu and Discord. WhatsApp is one of the most widely used messaging platforms globally, and adding WhatsApp channel support will significantly broaden the reach of Agent Service in international and consumer-facing scenarios.

**Changes**

- Add a new `WhatsAppChannel` class under `agentscope.app.channel` module (inherited from `ChannelBase`), considering to support (depends on the WhatsApp Business API side):
  - `ListChats`: list all available conversations accessible by the bot
  - `ListChatMembers`: list all members in a specified conversation
  - `SendMessage`: send a text message to a specified user or group
  - `SendFile`: send a file attachment to a specified user or group
  - `SendImage`: send an image to a specified user or group
  - More tools to be considered based on the capabilities provided by the WhatsApp Business API

redSun64:
## Summary

This PR adds WhatsApp Business API support to the Agent Service channel system. Resolves #2254

The scope includes:

- A new `WhatsAppChannel` implementation based on `ChannelBase`
- Tools for listing chats and members, and sending text, files, and images
- Individual and Groups API recipient support
- Inbound text and media message normalization
- Webhook verification and signature validation
- Webhook message deduplication and retry-safe media handling
- Interactive approval buttons for individual chats
- `/approve` and `/deny` confirmation commands for group chats
- WhatsApp registration in the Agent Service example
- Unit tests covering webhook parsing, sending, groups, confirmations, BSUID senders, and deduplication

## Why

WhatsApp Business Cloud API uses HTTPS webhooks instead of a persistent gateway connection.

To keep the channel implementation independent from FastAPI, the core `WhatsAppChannel` handles protocol-specific behavior such as signature verification, payload normalization, message delivery, and API calls. The Agent Service example provides a separate FastAPI webhook router responsible for HTTP callbacks, channel lookup, and routing deliveries by Business Phone Number ID.

The webhook interaction flow is:

1. Meta sends an HTTPS callback to `/webhooks/whatsapp`.
2. Agent Service verifies the `X-Hub-Signature-256` signature.
3. The callback is routed to the correct channel using `phone_number_id`.
4. `WhatsAppChannel` converts the payload into a normalized channel event.
5. Agent responses are sent through the WhatsApp Cloud API.

## Validation

- `pytest tests/channel_whatsapp_test.py`

All 13 WhatsApp unit tests and all configured pre-commit hooks pass.

## Live API testing status

This change has not yet been tested against a live WhatsApp Business account.

There is currently an issue with the WhatsApp Business account application, and my account is still under review. Therefore, real webhook delivery and outbound API calls still need to be verified once the account becomes available.

If a reviewer already has an approved WhatsApp Business account, the integration can be tested directly by:

1. Starting `examples/agent_service`.
2. Starting the example Web UI.
3. Creating a WhatsApp channel in the Web UI with the Phone Number ID, access token, Meta App Secret, and webhook verify token.
4. Configuring `/webhooks/whatsapp` as the HTTPS callback in the Meta developer dashboard.
5. Sending messages to the configured WhatsApp Business number and verifying inbound and outbound interactions.


examples/agent_service/whatsapp_webhook.py
```diff
@@ -0,0 +1,119 @@
...
+        not channel.verify_webhook_signature(raw_body, signature)
+        for channel, _ in deliveries
+    ):
+        raise HTTPException(
+            status_code=403,
+            detail="Invalid webhook signature",
+        )
+    count = 0
+    for channel, channel_payload in deliveries:
+        count += await channel.handle_webhook(channel_payload)
```
qbc2016: **Acknowledge the webhook before downloading media**

This awaits `handle_webhook()` before returning the HTTP 200 response. For media messages, that method performs two Graph API requests—first to retrieve the media URL and then to download the content—using a client configured with a 30-second timeout. A slow Graph API response can therefore delay the webhook acknowledgement long enough for Meta to time out and retry the delivery.

Please validate the signature, enqueue the payload in a reliable background queue, and return 200 immediately. Media retrieval and normalization should happen outside the webhook request lifecycle.

examples/agent_service/whatsapp_webhook.py
```diff
@@ -0,0 +1,119 @@
...
+) -> WhatsAppChannel:
+    """Resolve a destination phone number to its local channel."""
+    dispatcher = getattr(request.app.state, "channel_dispatcher", None)
+    channel_id = (
+        await request.app.state.storage.get_channel_id_by_platform_bot_id(
+            phone_number_id,
+        )
+    )
+    channel = (
+        dispatcher.get_local_channel(channel_id)
```
qbc2016: **Decouple webhook ingress from local channel ownership**

Resolving the destination through `dispatcher.get_local_channel()` requires the process that receives the HTTP request to already have a running local instance of that channel. During startup, after a local channel failure, or in a future sharded deployment, a valid webhook can therefore receive a 404 simply because it reached the wrong or not-yet-reconciled worker.

This is also coupled to the process-local deduplication issue: retries handled by different workers use different `_message_ids` caches and can enqueue duplicate agent turns.

Please consider handling both concerns through one framework-level ingress path:

1. Resolve the channel record and verify the signature using shared configuration.
2. Atomically deduplicate by `channel_id + wamid` in shared storage.
3. Persist the verified webhook payload to a shared queue or message bus.
4. Return HTTP 200 after the payload has been durably enqueued.
5. Let any eligible worker consume and normalize the payload.

This would remove the dependency on local channel placement while also making webhook delivery and deduplication reliable across workers and process restarts.

examples/agent_service/main.py
```diff
@@ -155,8 +160,10 @@ async def longterm_memory_factory(
     channels=[
         DiscordChannel,
         FeishuChannel,
+        WhatsAppChannel,
     ],
 )
+app.include_router(whatsapp_webhook_router)
```
qbc2016: **Make the WhatsApp webhook a first-class Agent Service route**

Registering `WhatsAppChannel` through `create_app(..., channels=[WhatsAppChannel])` is not sufficient to receive inbound messages; callers must also discover and manually mount this example-specific router. Applications using the public channel registration API can therefore configure a WhatsApp channel successfully while never exposing the callback endpoint required by Meta.

Please move the webhook adapter into the Agent Service package and have `create_app()` register it when WhatsApp support is enabled, or introduce a generic webhook-channel integration point. The channel implementation can remain independent of FastAPI while the application layer owns the HTTP route.

src/agentscope/app/channel/_whatsapp/_channel.py
```diff
@@ -0,0 +1,767 @@
...
+                content = [TextBlock(text=text)]
+        elif msg_type in {"image", "document", "audio", "video"}:
+            media = message.get(msg_type, {})
+            block = await self._download_media(
+                media.get("id", ""),
+                media.get("mime_type", "application/octet-stream"),
+                media.get("filename", msg_type),
+            )
+            if block:
+                content = [block]
```
qbc2016: **Preserve captions from inbound media messages**

WhatsApp includes `caption` in image, video, and document media objects, but this branch only adds the downloaded `DataBlock`. The caption is therefore discarded. In addition, `ChannelGateway` buffers events that contain no `TextBlock`, so an image with a caption will not trigger an agent turn until the user sends another text message.

Please add the media caption as a `TextBlock` in the same `ChannelEvent`, alongside the downloaded media block, and add a test covering an inbound media message with a caption.

src/agentscope/app/channel/_whatsapp/_channel.py
```diff
@@ -0,0 +1,767 @@
...
+    async def _send_json(self, body: dict) -> dict | None:
+        if not self._http:
+            return None
+        try:
+            response = await self._http.post(
+                self._url("messages"),
+                headers={"Authorization": f"Bearer {self._access_token}"},
+                json=body,
+            )
+            return response.json()
```
qbc2016: **Surface unsuccessful Graph API responses**

This returns the response body without checking the HTTP status or the Graph API `error` field. The agent-tool path eventually inspects the body through `_ack()`, but normal channel replies and confirmation prompts ignore this return value entirely. Consequently, a 4xx response caused by an expired token, an invalid payload, or a messaging-window restriction is silently treated as a completed send, with no useful log or channel status update.

Please call `raise_for_status()` and validate the response body, or explicitly log and propagate Graph API errors so the dispatcher can observe the failed delivery.

qbc2016 (commented):

examples/agent_service/whatsapp_webhook.py
redSun64: Fixed. The webhook request now validates the signature and writes each inbound message to the existing MessageBus queue. Media resolution, normalization, and Gateway processing are performed asynchronously by the ChannelLifecycleDispatcher.

The queues and drain locks are scoped per channel, so messages for the same channel retain their processing order while different channels can be processed concurrently.

This change intentionally reuses queue_push()/queue_drain(). Crash-safe acknowledgement and recovery after a consumer process exits remain outside the scope of this PR.

examples/agent_service/whatsapp_webhook.py
redSun64: Fixed. Webhook ingress no longer calls get_local_channel(). It resolves the ChannelRecord and verification credentials from shared storage.

The consumer also reads the current ChannelRecord from shared storage and constructs the adapter through ChannelTypeRegistry.create_channel_from_record(), so the worker receiving or consuming the webhook does not need to host a pre-existing local Channel instance.

Cross-process deduplication uses the existing MessageBus lock and registry primitives with channel_id + wamid.

examples/agent_service/main.py
redSun64: Fixed. The WhatsApp webhook router now belongs to the Agent Service application layer and is registered automatically by create_app() when WhatsAppChannel is enabled.

The example no longer imports or mounts a separate webhook router. The create_app() documentation also describes the automatically registered GET and POST /webhooks/whatsapp routes.

FastAPI remains an optional dependency: the webhook module imports FastAPI only when the router is created, and its tests cleanly skip when FastAPI is unavailable.

src/agentscope/app/channel/_whatsapp/_channel.py
redSun64: Fixed. Non-empty image, video, document, and audio captions are preserved as TextBlock values alongside the downloaded DataBlock.

This reuses the existing ChannelGateway media aggregation behavior, so captioned media is processed immediately. Media download failures are also surfaced as normalization failures and are not recorded as successfully deduplicated.

src/agentscope/app/channel/_whatsapp/_channel.py
redSun64: Fixed. The existing _send_json() helper now checks the HTTP status and non-empty Graph API error responses, records the HTTP status, Graph error code/message, and ChannelStatus.last_error, then raises a structured WhatsAppAPIError.

The existing message and confirmation sending call paths remain unchanged, but failed API calls are no longer treated as successful deliveries. Empty error objects on successful responses are not treated as failures.

src/agentscope/app/channel/_dispatcher.py
```diff
@@ -230,6 +246,146 @@ async def _outbound(self) -> None:
...
+        allowing unrelated channels to download media concurrently.
+        """
+        drain_lock = MessageBusKeys.channel_webhook_drain_lock(channel_id)
+        async with self._bus.acquire_lock(
+            drain_lock,
+            ttl_secs=WEBHOOK_DRAIN_LOCK_TTL_SECS,
+        ):
+            while True:
+                try:
+                    jobs = await self._bus.queue_drain(
```
qbc2016: **Preserve failed webhook jobs for an actual retry**

This consumer removes up to 100 jobs from the queue before processing any of them. When normalization or `process_with_result()` fails, `_process_webhook_job()` only returns without writing the dedupe marker; it does not restore the removed job. A cancellation or worker crash also loses the remainder of the already-drained batch.

Because the webhook endpoint has already returned `200`, Meta will not normally redeliver the payload. Therefore, contrary to the new tests' “remains eligible for retry” assertion, there is no retry source left.

Please use claim/ack semantics with ACK after successful processing, or explicitly requeue failed and unprocessed jobs. Please also add an integration-level test proving that a failed job remains in, or is restored to, the consumable queue.

src/agentscope/app/channel/_dispatcher.py
```diff
@@ -230,6 +246,146 @@ async def _outbound(self) -> None:
...
+                logger.error(
+                    "Channel %s does not support webhook normalization",
+                    record.channel_type,
+                )
+                return
+            events = await normalize(job.get("payload", {}))
+            for event in events:
+                if not await self._gateway.process_with_result(event):
+                    return
+            await self._bus.registry_set(
```
qbc2016, line 489: **The webhook dedupe registry is not bounded by this TTL**

`RedisMessageBus.registry_set()` applies `EXPIRE` to the entire hash namespace, not to the individual message-id field. Every incoming message refreshes that hash TTL, so an active WhatsApp channel never expires and all processed message IDs accumulate indefinitely. `InMemoryMessageBus` ignores `ttl_secs` entirely, making the same registry unbounded there as well.

Please use individually expiring dedupe keys or explicitly trim old message IDs instead of relying on the registry namespace TTL.

src/agentscope/app/channel/_dispatcher.py
```diff
@@ -230,6 +246,146 @@ async def _outbound(self) -> None:
...
+        try:
+            dedupe_namespace = MessageBusKeys.channel_webhook_dedupe(
+                channel_id,
+            )
+            if await self._bus.registry_exists(
+                dedupe_namespace,
+                message_id,
+            ):
+                return
+            channel = self._types.create_channel_from_record(record)
```
qbc2016, line 475: **Do not record observed chats on a throwaway channel instance**

A new WhatsApp adapter is created for every queued webhook and discarded immediately after normalization. `_normalize_message()` writes the observed chat into that adapter's private `_seen` map, so the long-lived channel instance used later by `ListChats`, `chat_kind()`, and agent setup never receives this state.

Consequently, observed direct-message recipients are missing from `ListChats`, and a group session can be classified as private unless the Groups API happened to populate the long-lived instance first.

Please persist the chat metadata in shared state or update an adapter whose state is retained.

src/agentscope/app/channel/_dispatcher.py
```diff
@@ -230,6 +246,146 @@ async def _outbound(self) -> None:
...
+            return
+        for record in records:
+            if record.enabled and record.channel_type == "whatsapp":
+                self._spawn_webhook_drain(record.id)
+
+    def _spawn_webhook_drain(self, channel_id: str) -> None:
+        """Start an ordered drain for one channel without blocking others."""
+        if not channel_id:
+            return
+        task = asyncio.create_task(
```
qbc2016, line 289: **Coalesce webhook drain tasks per channel**

Every webhook notification is broadcast to every worker, and each worker creates a new drain task for every notification. These tasks then block in `acquire_lock()` for the same channel. Under sustained traffic, waiting tasks accumulate with the number of webhook requests multiplied by the number of workers, and most eventually acquire the lock only to find an empty queue.

Please keep at most one active drain task per channel on each worker and coalesce subsequent wake-up signals. A worker that does not own the distributed drain lease should not queue another blocking waiter for every notification.

src/agentscope/app/channel/_whatsapp/_channel.py
```diff
@@ -0,0 +1,895 @@
...
+                raise _WhatsAppMediaDownloadError(
+                    f"Failed to download WhatsApp media {media.get('id', '')}",
+                )
+            if caption:
+                content.append(TextBlock(text=caption))
+            content.append(block)
+        if not content:
+            return None
+        profile = (value.get("contacts") or [{}])[0]
+        event = ChannelEvent(
```
qbc2016: **Populate `ChannelEvent.chat_name` with the resolved group name**

`_conversation_context()` resolves `chat_name`, but the value is only copied into `metadata`. `ChannelGateway._session_name()` reads `event.chat_name`, so the resolved WhatsApp group name is ignored and the session can instead be named after the sender or raw group ID.

Please pass `chat_name=chat_name` when constructing the event and add an assertion for the normalized `ChannelEvent.chat_name`.

qbc2016 (commented):
Please see the inline comments, and run `pre-commit run --all-files` to format the code.

src/agentscope/app/channel/_dispatcher.py
redSun64: Fixed. I replaced the destructive webhook queue drain with the existing replay-log primitives.

Webhook jobs are now persisted with `log_append()` before returning HTTP 200. The consumer reads from the log using a shared per-channel cursor, and the cursor only advances after normalization and Gateway processing succeed.

If processing fails, the cursor remains unchanged, so the same log entry is retried on the next drain. Old entries are trimmed only after the checkpoint has been persisted.

I also added a test that fails the first processing attempt and verifies that the same persisted job is consumed again on the next drain.

src/agentscope/app/channel/_dispatcher.py
redSun64: Fixed. Processed WhatsApp message IDs now use an independently expiring namespace per `(channel_id, wamid)` instead of fields in a single channel-level registry.

This prevents new messages on an active channel from continuously refreshing the TTL of older processed message IDs.

I also added registry TTL handling to `InMemoryMessageBus` so the in-memory backend follows the same namespace-level TTL semantics.

src/agentscope/app/channel/_dispatcher.py
redSun64: Fixed. Webhook normalization no longer relies on `_seen` state stored on the short-lived WhatsApp adapter.

Observed chat metadata (`chat_type` and `chat_name`) is persisted in the existing shared `channel_seen_chats` registry, and retained channel instances are hydrated from that shared state before chat discovery and outbound operations.

The local `_seen` map is now only a process-local cache rather than the source of truth.

src/agentscope/app/channel/_dispatcher.py
redSun64: Fixed. Each worker now keeps at most one active webhook drain task per channel.

Additional wake-up signals for an already active channel are coalesced into a pending flag instead of spawning more tasks waiting on the same distributed lock.

The per-channel distributed lock is still retained for cross-worker ordering and shared cursor advancement.

src/agentscope/app/channel/_whatsapp/_channel.py
redSun64: Fixed. The resolved WhatsApp group name is now populated on the first-class `ChannelEvent.chat_name` field.

The metadata copy is still retained for platform/routing metadata, and the group normalization coverage now verifies `event.chat_name`.

redSun64:
Thanks for the review. I’ve addressed the latest webhook comments in the follow-up commit.

The main change is that I **reused the existing MessageBus replay-log primitives** (`log_append` / `log_read` / `log_trim`) with a shared per-channel cursor, instead of introducing a new Redis reliable-queue / consumer-group abstraction.

This makes webhook consumption retry-safe after returning HTTP 200: failed jobs leave the cursor unchanged and remain replayable, while successful jobs advance the cursor and are trimmed afterward.

I also fixed independently expiring dedupe entries, shared chat metadata, drain-task coalescing, and `ChannelEvent.chat_name`.

One current trade-off is that webhook consumption is serialized **per channel**. Different channels can run concurrently, but different chats under the same WhatsApp channel cannot.

I kept this model for correctness and to keep the PR scope small. If you think channel-level serialization is too restrictive, I can follow up with **per-chat serialization and concurrency across chats within the same channel**.

src/agentscope/app/channel/_dispatcher.py
```diff
@@ -230,6 +249,239 @@ async def _outbound(self) -> None:
...
+                        since=cursor,
+                        max_count=1,
+                    )
+                except Exception:  # pylint: disable=broad-except
+                    logger.exception("channel webhook drain failed")
+                    return
+                if not jobs:
+                    return
+                entry_id, job = jobs[0]
+                if not await self._process_webhook_job(job):
```
qbc2016: **[P1] Add a bounded, autonomous retry path for retained failures**

The replay log correctly fixes the previous message-loss issue, but retaining a failed entry does not yet guarantee that it will be retried. When `_process_webhook_job()` returns `False`, the drain task exits unless another webhook signal happened to arrive. Without later traffic or a service restart, an isolated transient failure can therefore remain pending indefinitely.

A permanently failing head entry also keeps the shared cursor unchanged and blocks every later message for the channel.

Please schedule retries independently of future webhook traffic, with backoff and persisted attempt state, and move an entry to a dead-letter path or advance past it after a bounded number of failures. The current test manually invokes the drain a second time, so it does not cover the missing runtime retry trigger.

src/agentscope/app/channel/_dispatcher.py
```diff
@@ -230,6 +249,239 @@ async def _outbound(self) -> None:
...
+                logger.error(
+                    "Channel %s does not support webhook normalization",
+                    record.channel_type,
+                )
+                return True
+            events = await normalize(job.get("payload", {}))
+            for event in events:
+                if isinstance(event, ChannelEvent):
+                    self._observe_local_webhook_chat(event)
+                if not await self._gateway.process_with_result(event):
```
qbc2016, line 487: **[P1] Make replay delivery idempotent across the checkpoint crash window**

`process_with_result()` may enqueue the WhatsApp message as a user-turn trigger, but the deduplication marker and replay cursor are committed only afterward. If the worker crashes after the trigger is enqueued but before either checkpoint is written, another worker will replay this job and enqueue the same user turn again.

The per-channel drain lock only serializes consumers; it does not make the downstream side effect and the replay checkpoint atomic.

Please propagate an idempotency key derived from `(channel_id, message_id)` to the run-trigger consumer, or use an equivalent transactional-outbox/checkpoint design. A regression test should simulate a failure after the trigger is enqueued but before the deduplication marker is stored.

src/agentscope/app/channel/_dispatcher.py
```diff
@@ -230,6 +249,239 @@ async def _outbound(self) -> None:
...
+                    return False
+                if isinstance(event, ChannelEvent) and event.chat_id:
+                    await self._persist_seen_chat(event)
```
qbc2016: **[P2] Complete the cross-worker chat-metadata path introduced here**

The shared metadata is persisted only after `process_with_result()`, even though that call may already enqueue a run trigger for consumption by another worker. The run can therefore start before its worker can observe the chat metadata.

In addition, the agent-assembly and channel-tool paths call `get_local_channel()` directly without hydrating the retained WhatsApp adapter. Since WhatsApp's `chat_kind()`, `chat_name()`, and DM discovery depend on the process-local `_seen` map, a worker that did not consume the webhook may describe a group as private or unnamed and omit observed DMs from `ListChats`.

Please persist the metadata before dispatching the event and hydrate it on the agent/tool access paths, or make those WhatsApp lookups read from shared state directly. A two-dispatcher test sharing one message bus would cover this ordering and worker-placement case.

src/agentscope/app/message_bus/_in_memory_message_bus.py
```diff
@@ -119,6 +120,14 @@ def _next_id(self) -> str:
         self._seq += 1
         return f"{self._seq}-0"
 
+    def _purge_expired_registry(self, namespace: str) -> None:
+        """Drop ``namespace`` after its registry TTL has elapsed."""
+        expires_at = self._registry_expiries.get(namespace)
```
qbc2016: **[P2] Reclaim expired namespaces without requiring access to the same key**

Expiration is currently checked only for the namespace passed to `_purge_expired_registry()`. The WhatsApp deduplication path creates a distinct namespace for every `(channel_id, message_id)`, and those namespaces normally are not accessed again after their TTL expires. As a result, expired entries remain indefinitely in both `_registries` and `_registry_expiries`, so memory still grows with the total number of received messages.

Please add global or opportunistic expiration cleanup—for example, a scheduled sweep, an expiry heap, or bounded cleanup during writes. A regression test should create many unique namespaces, advance the clock, access only a new namespace, and verify that the expired namespaces are actually reclaimed.

qbc2016 (commented):

src/agentscope/app/channel/_dispatcher.py
redSun64: Fixed in the follow-up changes.

Failed webhook entries now keep persisted retry state with bounded exponential backoff. Retries are scheduled independently of future webhook traffic. After the maximum number of attempts, a poison entry is appended to a per-channel dead-letter log and the shared cursor advances, so later webhook entries are not blocked.

Added coverage in:
- `test_failed_durable_job_retries_without_another_signal`
- `test_poison_webhook_is_dead_lettered_and_does_not_block_later_work`

src/agentscope/app/channel/_dispatcher.py
redSun64: Fixed in the follow-up changes.

Inbound channel messages now receive a stable ID derived from `(channel_id, channel_message_id)`. The WakeupDispatcher acquires a distributed delivery lock and checks persisted session messages before starting the user turn. Therefore, if the webhook trigger has already been enqueued and the first run has persisted the user message before the outer webhook checkpoint completes, a replayed trigger is skipped instead of starting a duplicate turn.

Added regression coverage in:
- `test_replayed_run_trigger_skips_an_already_persisted_user_message`
- `test_replayed_gateway_trigger_runs_one_user_turn_after_persistence`

src/agentscope/app/channel/_dispatcher.py
redSun64: Fixed in the follow-up changes.

Observed chat metadata is now persisted before the stable run trigger is enqueued. Before assembling a channel-bound chat run, the retained channel instance is hydrated from the shared `channel_seen_chats` registry. This allows a worker that did not consume the original webhook to resolve the chat name, chat type, and observed chat list correctly.

Added coverage in:
- `test_gateway_persists_chat_metadata_before_stable_run_trigger`
- `test_shared_chat_metadata_hydrates_another_dispatcher_instance`

src/agentscope/app/message_bus/_in_memory_message_bus.py
redSun64: Fixed in the follow-up changes.

Processed webhook IDs now use independently expiring registry namespaces. InMemoryMessageBus also performs opportunistic global expiry cleanup, so expired namespaces that are never accessed again are reclaimed when other registry operations occur.

Added coverage in:
- `test_inmemory_registry_ttl_expires_namespace`
- `test_inmemory_registry_reclaims_unvisited_expired_namespaces`

src/agentscope/app/channel/_dispatcher.py
```diff
@@ -230,6 +262,318 @@ async def _outbound(self) -> None:
...
+                retry_delay = await self._drain_webhook_queue(channel_id)
+                if retry_delay is not None:
+                    await asyncio.sleep(retry_delay)
+                    continue
+                if channel_id not in self._webhook_pending:
+                    return
+        finally:
+            task = asyncio.current_task()
+            if self._webhook_drains.get(channel_id) is task:
+                self._webhook_drains.pop(channel_id, None)
```
qbc2016, line 310: **[P2] Avoid one Redis round trip per historical message**

This new hot-path check calls `RedisStorage.get_message()` for every inbound channel message. For a first-time delivery, the ID is absent, so the current `LLEN` + repeated `LINDEX` implementation scans the entire session with one Redis request per stored message.

Please use the existing batched lookup helper or add an indexed lookup so this check does not require O(history length) Redis round trips.

tests/channel_webhook_test.py
```diff
@@ -0,0 +1,773 @@
+# -*- coding: utf-8 -*-
```
qbc2016, line 1: **Compare complete structures in the newly added tests**

The repository review guide marks this as a MUST requirement: assertions for structured outputs should compare the entire structure rather than selected fields, using `AnyString` or `AnyValue` for nondeterministic values.

Several new tests currently assert only individual fields or collection lengths. For example:

- `tests/channel_webhook_test.py` checks only the queued trigger's input ID, only the number of replayed entries, and only whether one chat ID is present.
- `tests/channel_whatsapp_test.py` checks selected fields of normalized events instead of the complete event structures.

These assertions can still pass when required fields are missing or unexpected fields change. Please review the newly added tests and compare complete dicts, lists, model dumps, and mock-call payloads wherever the full structure is deterministic.

qbc2016 (commented):

src/agentscope/app/channel/_dispatcher.py
redSun64: Fixed in the follow-up changes.

`WakeupDispatcher` now performs a single `StorageBase.get_message()` lookup. `RedisStorage.get_message()` reuses the existing chunked `_find_message_index()` helper, replacing one `LINDEX` round trip per historical message with batched `LRANGE` requests.

Added `test_get_message_scans_history_in_batches`, which covers a missing ID in a 250-message history and verifies the exact three `LRANGE` batches.

tests/channel_webhook_test.py
redSun64: Done.

qbc2016 (approved):
LGTM.

DavdGao:
Thanks for the work here. The blocker isn't really the WhatsApp code — it's that `ChannelBase` is currently designed around a long-lived inbound connection (`start_listening`), and WhatsApp is webhook-based. There's no place in the current contract for a channel to register an HTTP route or to say "my inbound is at-least-once", so this PR has to work around it: hardcoding `"whatsapp"` in `_app.py` / `_dispatcher.py`, and putting the inbound queue, dedupe and retry logic into the shared `ChannelLifecycleDispatcher`.

We'd rather settle how webhook channels fit into `ChannelBase` before taking that shape in, so I'm marking this `state: under discussion` and holding it for now. Not a judgement on the implementation — we just don't want to lock the design in through a single platform.

One thing worth checking meanwhile, independent of all this: `_send_json` / `_upload` / `_get_json` return early when `self._http` is None, and `_http` is only created in `start_listening()`. Replies go out through `ChannelClients`, which never calls `start_listening`, so on the node that runs the agent every outbound message is dropped silently. The tests set `channel._http` by hand so it doesn't surface there.

redSun64:
> Thanks for the work here. The blocker isn't really the WhatsApp code — it's that `ChannelBase` is currently designed around a long-lived inbound connection (`start_listening`), and WhatsApp is webhook-based. There's no place in the current contract for a channel to register an HTTP route or to say "my inbound is at-least-once", so this PR has to work around it: hardcoding `"whatsapp"` in `_app.py` / `_dispatcher.py`, and putting the inbound queue, dedupe and retry logic into the shared `ChannelLifecycleDispatcher`.
> 
> We'd rather settle how webhook channels fit into `ChannelBase` before taking that shape in, so I'm marking this `state: under discussion` and holding it for now. Not a judgement on the implementation — we just don't want to lock the design in through a single platform.
> 
> One thing worth checking meanwhile, independent of all this: `_send_json` / `_upload` / `_get_json` return early when `self._http` is None, and `_http` is only created in `start_listening()`. Replies go out through `ChannelClients`, which never calls `start_listening`, so on the node that runs the agent every outbound message is dropped silently. The tests set `channel._http` by hand so it doesn't surface there.

Thanks for the clarification. That makes sense.

I agree that it would be better to settle how webhook-based channels should fit into `ChannelBase` before continuing to evolve the WhatsApp-specific integration.

I’m also happy to participate in the design and implementation of a generic webhook channel abstraction if that would be helpful. Once the direction is agreed, I can adapt this PR to the new contract accordingly.

redSun64:
> Thanks for the work here. The blocker isn't really the WhatsApp code — it's that `ChannelBase` is currently designed around a long-lived inbound connection (`start_listening`), and WhatsApp is webhook-based. There's no place in the current contract for a channel to register an HTTP route or to say "my inbound is at-least-once", so this PR has to work around it: hardcoding `"whatsapp"` in `_app.py` / `_dispatcher.py`, and putting the inbound queue, dedupe and retry logic into the shared `ChannelLifecycleDispatcher`.
> 
> We'd rather settle how webhook channels fit into `ChannelBase` before taking that shape in, so I'm marking this `state: under discussion` and holding it for now. Not a judgement on the implementation — we just don't want to lock the design in through a single platform.
> 
> One thing worth checking meanwhile, independent of all this: `_send_json` / `_upload` / `_get_json` return early when `self._http` is None, and `_http` is only created in `start_listening()`. Replies go out through `ChannelClients`, which never calls `start_listening`, so on the node that runs the agent every outbound message is dropped silently. The tests set `channel._http` by hand so it doesn't surface there.

I think it would make sense to first define a generic contract for webhook-based channels.

One possible direction could be:

* let a channel declare whether its inbound mode is listener-based or webhook-based
* keep HTTP route registration in Agent Service rather than in platform-specific channel code
* let each webhook channel handle platform-specific verification and payload normalization
* keep retry, deduplication, and delivery semantics in the shared infrastructure / MessageBus layer

It may also be worth opening a separate issue or design discussion for this abstraction, so the framework-level contract can be agreed on independently from the WhatsApp implementation.

That should make it easier to support WhatsApp and other webhook-based channels without adding platform-specific branches into the shared channel lifecycle.

PR #2271, opened 2026-08-09
feat(channel): add Telegram channel

### LINKED ISSUE #2255: feat(channel): add Telegram channel support
**Background**

The channel system in Agent Service currently supports Feishu and Discord. Telegram is a widely adopted messaging platform with a well-established bot API, making it a natural fit for agent-based interactions. Adding Telegram channel support will extend the reach of Agent Service to a large international user base.

**Changes**

- Add a new `TelegramChannel` class under `agentscope.app.channel` module (inherited from `ChannelBase`), considering to support (depends on the Telegram Bot API side):
  - `ListChats`: list all available chats accessible by the bot
  - `ListChatMembers`: list all members in a specified chat
  - `SendMessage`: send a text message to a specified chat or user
  - `SendFile`: send a file attachment to a specified chat or user
  - `SendImage`: send an image to a specified chat or user
  - More tools to be considered based on the capabilities provided by the Telegram Bot API

RICARDO51167:
## AgentScope Version
2.0.8

## Description
**Closes #2255**

### Background
AgentScope currently supports external IM channels including Feishu and Discord, but lacks integration with Telegram. Telegram is a widely used international messaging platform with stable and mature Bot APIs. Adding native Telegram channel support enables Agent Service to serve global users, enrich the ecosystem of official communication channels, and unify the channel abstraction and routing logic consistent with existing channel implementations.

### Purpose
This PR implements a complete standard-compliant Telegram channel based on Telegram Bot API. It supports private chat/group/supergroup message interaction, markdown formatted message rendering, streaming response output, multiple types of message reception, and dedicated Telegram agent tools. The implementation strictly follows AgentScope channel abstraction specifications, with bounded retry, lifecycle management, and cross-process coordination.

### Changes Made
+ Implement public class `agentscope.app.channel.TelegramChannel` inherited from `ChannelBase`, fully compatible with AgentScope channel system rules
+ Implement Telegram Bot API long-polling lifecycle management, conflict handling and resource cleanup
+ Support private chats, groups and supergroups interaction, with `only_at_reply` group mention & reply filtering capability
+ Support Telegram-specific command parsing (e.g., `/help@bot_name`)
+ Support comprehensive inbound message types: text/captions, photos/albums, documents, audio/voice, video/animation/video notes, stickers, locations, venues and contacts
+ Implement Telegram-safe Markdown rendering: support common markdown syntax, HTML escaping, unsafe link filtering, 4096-character chunk segmentation and plain-text fallback rendering
+ Implement streaming response capability: ephemeral draft streaming for private chats, throttled editable preview streaming for group chats, with persistent final responses
+ Add credential validation, whitespace normalization and token-safe error handling
+ Add automatic retry logic for transient network errors and API rate-limit errors
+ Add interactive approval buttons for tool call scenarios
+ Integrate dedicated Telegram agent tools: `SendMessage`, `SendFile`, `SendImage`
+ Introduce optional channel dependencies: `python-telegram-bot>=22.8`, `markdown-it-py>=4,<5`
+ Update Agent Service example project and corresponding Telegram usage documentation

### Testing
**Pre-commit Checks**

All pre-commit verification items pass: python ast check, private key detection, mypy type checking, black formatting, flake8 & pylint linting, Pyroma package check.

**Automated Unit Tests**

Executed full channel-related test suites including telegram channel logic, markdown rendering, channel gateway, routing and application lifecycle tests.

Test Result: **92 passed, 18 subtests passed**

**Manual Verification**

Completed full manual testing covering private chat conversation, group mention routing, mention filtering, markdown/code rendering, ultra-long message adaptation, streaming output, multi-type media message parsing, file download delivery and capability boundary verification.

### Documentation

Telegram setup and usage documentation: agentscope-ai/docs#68

### Follow-up Issues

+ Generic channel-level access control: #2621
+ Telegram rich and ephemeral messages after PTB support: #2622

### Platform Limitations & Out of Scope
+ Telegram bot cannot actively initiate private chats with uncontacted users
+ Telegram Bot API does not support full chat/member enumeration, so `ListChats` and `ListChatMembers` are not implemented
+ Automatic document parsing, audio transcription and video analysis are not included in this PR
+ Webhooks, forum topic routing, multi-node leader election and other advanced capabilities are out of scope

### Breaking Changes
None

## Checklist
+ An issue has been created for this PR
+ I have read the [CONTRIBUTING.md](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md)
+ Docstrings are in Google style
+ Related documentation has been updated (e.g. links, examples, etc.) in [documentation repository](https://github.com/agentscope-ai/docs)
+ Code is ready for review


src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1446 @@
...
+            channel_user_id=user_id,
+            channel_user_name=user_name,
+            chat_id=chat_id,
+            chat_name=self._chat_display_name(chat),
+            channel_message_id=str(first.message_id),
+            content=content,
+            metadata={"chat_type": str(chat.type)},
+        )
+
+    def _gated_out(self, message: "Message") -> bool:
```
qbc2016, line 828: **Require authorization before accepting private-chat messages**

This check unconditionally accepts every private-chat message from a non-bot Telegram user. Since channel routing requires a catch-all binding, there is currently no way to reject unknown users before they are routed into an AgentScope session owned by the channel owner. Anyone who discovers the bot username can therefore invoke the configured agent, consume model resources, and potentially approve and execute its tools.

Please add a pairing or allowlist check for private users/chats before downloading attachments or emitting the event. Public access should be an explicit configuration rather than the default.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1446 @@
...
+                await query.answer(
+                    "This approval has expired.",
+                    show_alert=True,
+                )
+                await query.edit_message_reply_markup(reply_markup=None)
+            except Exception:  # pylint: disable=broad-except
+                logger.debug("Could not freeze an expired Telegram approval")
+            return
+
+        await query.answer("Decision received.")
```
qbc2016: **Do not make the approval decision depend on answering the callback query**

`query.answer()` is awaited outside the best-effort error handling and before the confirmation event is emitted. If Telegram rejects an old callback query, or this request fails because of a transient network error or rate limit, the exception exits the handler and the already-received approval decision is lost. The agent session then remains parked indefinitely.

Please treat `answer()` and the message edit as best-effort UI operations while ensuring that a valid callback still emits its `ChannelConfirmationResultEvent`.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1446 @@
...
+            try:
+                return await operation()
+            except RetryAfter as error:
+                delay = error.retry_after
+                seconds = (
+                    delay.total_seconds()
+                    if isinstance(delay, timedelta)
+                    else float(delay)
+                )
+                exhausted = attempt + 1 >= _MAX_API_ATTEMPTS
```
qbc2016: **Preserve outbound messages when RetryAfter exceeds 30 seconds**

A valid `RetryAfter` longer than 30 seconds is re-raised immediately. In the normal response path, callers convert this into a failed result, log it, and continue; the outbound job is then completed without retrying or re-queuing the unsent message. This can silently drop complete replies or trailing chunks when a group or concurrent delivery hits Telegram's flood limits.

Please either honor the full retry interval within the dispatcher's response timeout, add proactive rate limiting, or persist and re-queue the delivery when the requested delay cannot be waited locally.

qbc2016 (commented):

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Thanks — I agree this is a security issue and should be addressed before private messages are routed to an AgentScope session.

I see two possible approaches:

1. **Static allowlist**: deny private messages by default, with public private-chat access only as an explicit opt-in. The main design question is how channel owners should configure the allowlist. We could expose allowed Telegram user IDs in the channel configuration, either as a simple text-based setting initially or through a dedicated management UI/API.

2. **Pairing flow**: unknown users would need to pair with the channel, and the owner would approve the request. This offers a smoother onboarding experience, but requires persistent pairing state and an approval workflow.

In either case, the authorization check would happen before downloading attachments or emitting an event.

Do you have a preference between a static allowlist and pairing for the initial implementation? And, if we use an allowlist, do you have a preferred configuration approach?

RICARDO51167: Thanks — I agree this is a security issue and should be addressed before private messages are routed to an AgentScope session.

I see two possible approaches:

1. **Static allowlist**: deny private messages by default, with public private-chat access only as an explicit opt-in. The main design question is how channel owners should configure the allowlist. We could expose allowed Telegram user IDs in the channel configuration, either as a simple text-based setting initially or through a dedicated management UI/API.

2. **Pairing flow**: unknown users would need to pair with the channel, and the owner would approve the request. This offers a smoother onboarding experience, but requires persistent pairing state and an approval workflow.

In either case, the authorization check would happen before downloading attachments or emitting an event.

Do you have a preference between a static allowlist and pairing for the initial implementation? And, if we use an allowlist, do you have a preferred configuration approach?

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: I’ll restructure the valid-callback path so that it emits the `ChannelConfirmationResultEvent` independently of the UI updates, with `query.answer()` and the message edit handled as best-effort operations that only log failures. Invalid or expired callbacks will still not emit an approval event.

I’ll also add regression tests covering failures from both `query.answer()` and the message edit, ensuring that the valid approval/denial event is still emitted.

RICARDO51167:
Follow-up commit `35c58533` addresses the review feedback.

- **Private-chat access control:** private chats are now denied by default. The channel configuration adds `allowed_private_user_ids` and an explicit `allow_public_private_chats` opt-in. Unauthorized users are rejected before session routing, album buffering, or attachment download; `/start` only returns their Telegram user ID so the channel owner can add it to the allowlist.
- **Approval callback reliability:** valid callbacks now emit `ChannelConfirmationResultEvent` before any Telegram UI work. `query.answer()` and message edits are handled independently as best-effort operations, so transient Telegram failures cannot leave a session parked. Callback state is stored in the shared message bus so it works with the latest connection-free outbound delivery model.
- **Flood-limit delivery:** removed the 30-second `RetryAfter` cutoff. Telegram’s requested retry delay is honored within the existing 300-second delivery budget, and later text chunks are not sent after an earlier chunk fails.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1705 @@
...
+                self.status.last_error = ""
+        except Exception:  # pylint: disable=broad-except
+            logger.exception(
+                "Telegram channel '%s' failed to process a message",
+                self._channel_id,
+            )
+
+    def _is_private_user_allowed(self, message: "Message", user: Any) -> bool:
+        """Return whether a sender may enter a private chat with the bot."""
+        if str(message.chat.type) != "private":
```
qbc2016: **Do not let Telegram group chats bypass the new private-user access control**

Every non-private chat is accepted unconditionally here. Unless group additions are disabled separately through BotFather, an arbitrary Telegram user can add this bot to a new group, mention or reply to it, and invoke the configured agent without appearing in `allowed_private_user_ids`.

Please add an explicit Telegram group access policy, such as an `allowed_group_chat_ids` allowlist with public group access as an opt-in, and enforce it before remembering the chat, downloading media, or emitting an event. If unrestricted group access is intentional, it should at least be exposed clearly as a security-sensitive configuration option rather than being enabled implicitly.

examples/agent_service/README.md
```diff
@@ -64,6 +64,90 @@ After that, you can set the API endpoint `http://localhost:8000` in the Web UI a
 
 <img src="https://gw.alicdn.com/imgextra/i2/O1CN01Phmg1G1brIVC8WXyU_!!6000000003518-2-tps-2938-1736.png" alt="Web UI Screenshot" width="100%">
 
+## Telegram Channel
```
qbc2016: **Remove the Telegram-specific changes from this example README**

This README is intended to document how to install and run the Agent Service example. It does not contain platform-specific setup or implementation guides for the other registered channels, so adding a detailed Telegram-only section here is inconsistent with the existing documentation structure and unnecessarily expands the scope of this example README.

Please remove the Telegram-specific section from this file. Channel documentation is maintained in the [AgentScope channel documentation](https://docs.agentscope.io/versions/2.0.8dev/en/deploy/channel) and its [documentation repository](https://github.com/agentscope-ai/docs). If you would like to add a Telegram setup and usage guide, please consider submitting a corresponding PR to that repository.

src/agentscope/app/channel/_clients.py
```diff
@@ -213,7 +218,18 @@ async def _run() -> None:
             """Feed the run's event stream to the channel."""
             try:
                 async with aclosing(events):
-                    await channel.send_response(target, events)
+                    await asyncio.wait_for(
+                        channel.send_response(target, events),
+                        timeout=RESPONSE_TIMEOUT_SECS,
```
qbc2016: **Do not put a fixed deadline around the entire channel reply stream**

This PR introduces a 300-second timeout around the complete `send_response()` stream in the shared `ChannelClients` layer. The timeout therefore applies to every channel, not only to Telegram API retries.

If an agent run takes more than five minutes, `wait_for()` cancels  send_response()`, closes the event stream, and drops all remaining output without persisting or re-queuing the delivery. A Telegram `RetryAfter` longer than the remaining timeout is also still lost for the same reason.

Please avoid applying a wall-clock timeout to the complete reply stream. Telegram-specific retry limits should remain inside the Telegram adapter, and deliveries that cannot complete within a local retry budget should be persisted or re-queued rather than silently cancelled.

qbc2016 (commented):
Please see the inline comments, and also strengthen the new tests to compare complete deterministic structures where possible.

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Addressed. Telegram group and supergroup chats are now denied by default, with `allowed_group_chat_ids` and an explicit `allow_public_group_chats` opt-in. Authorization is checked before remembering the chat, buffering/downloading media, or emitting events, and is also applied to approval callbacks.

examples/agent_service/README.md
RICARDO51167: Addressed. The Telegram-specific section has been removed from the Agent Service example README. Telegram remains listed as a supported channel, while platform-specific setup documentation is left to the official documentation repository.

src/agentscope/app/channel/_clients.py
RICARDO51167: The shared 300-second deadline has been removed. For Telegram flood limits, this implementation takes the full-wait approach: every RetryAfter delay is honored and the same operation is retried without a local exhaustion branch. Durable recovery across process restarts is not introduced or claimed by this PR; that would require a shared delivery outbox and ACK/progress model across all channels.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1781 @@
...
+            await application.start()
+            self.status.state = "connected"
+            self.status.last_error = ""
+            assert self._fatal_event is not None
+            await self._fatal_event.wait()
+        except InvalidToken as error:
+            raise _PermanentTelegramError(
+                "Telegram rejected the configured bot token.",
+            ) from error
+        except Conflict as error:
```
DavdGao, line 459: Treating `Conflict` as permanent kills the channel for good — the task parks in `while True: sleep(30)` and never reconnects. `getUpdates` only allows one poller, so on a rolling restart of the channel worker the new pod gets `Conflict` while the old one is still polling, parks forever, and once the old pod exits nobody polls at all. The channel stays dead until someone edits the record.

This is different from Feishu/DingTalk (platform picks one connection) and Discord (broadcasts), so it needs explicit handling here. You already have the bus via `bind_message_bus` — suggested to take a per-channel `message_bus.acquire_lock()` before `start_polling`, so exactly one replica polls and the others stand by and take over when the lease expires. At minimum retry `Conflict` with backoff instead of parking.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1781 @@
...
+            default=False,
+            title="Show tool process",
+            description="Show tool calls and results inline in the reply",
+        )
+        show_thinking: bool = Field(
+            default=False,
+            title="Show thinking",
+            description="Show model reasoning inline in the reply",
+        )
+        allow_public_private_chats: bool = Field(
```
DavdGao, line 227: `allow_public_private_chats` / `allowed_private_user_ids` / `allow_public_group_chats` / `allowed_group_chat_ids` plus the `/start` reply at L706 are a new access-control concept no other channel has, and they all default to False — a freshly created channel just silently ignores everything. Telegram bots being publicly reachable is a real problem, but it's not this PR's problem, and if we want it, it belongs in routing as a generic capability rather than one channel's Config.

Suggested to drop these from this PR and open a separate issue for channel-level access control.

tests/channel_clients_test.py
```diff
@@ -274,6 +307,93 @@ async def test_an_unbuildable_channel_delivers_nothing(self) -> None:
         ) as clients:
             await self._deliver(clients)
 
+    async def test_delivery_waits_for_the_complete_reply_stream(self) -> None:
```
DavdGao, line 310: This test and the `tests/channel_stream_test.py` change have nothing to do with Telegram, please split them out.

`test_delivery_waits_for_the_complete_reply_stream` also patches `agentscope.app.channel._clients.asyncio.wait_for` and asserts no timeout >= 60s is used, but `_clients.py` never calls `asyncio.wait_for` at all — the test asserts an implementation detail that doesn't exist.

pyproject.toml
```diff
@@ -81,6 +81,8 @@ storage-s3 = ["aioboto3"]
 channel = [
     "lark-oapi>=1.4.0",
     "discord.py>=2.3",
+    "python-telegram-bot[callback-data]>=22.8,<23.0",
```
DavdGao, line 84: Two things here:

- `[callback-data]` isn't used — arbitrary callback data is never enabled (the test at `telegram_channel_test.py:324` even asserts `callback_data_cache is None`), and it's a per-process LRU so it wouldn't work across processes anyway. Suggested to drop the extra.
- The `<23.0` upper bound is inconsistent with `lark-oapi>=1.4.0` / `discord.py>=2.3`, and would block picking up Bot API 10.1 rich messages / 10.2 ephemeral messages later. Suggested to use `python-telegram-bot>=22.8`.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1781 @@
...
+        self,
+        chat_id: str,
+        preview: _StreamPreview,
+        text: str,
+    ) -> None:
+        """Best-effort preview update that can never block final delivery."""
+        if preview.disabled or not text:
+            return
+        try:
+            chunks = self._formatted_chunks(text)
```
DavdGao, line 1113: The full markdown parse + HTML chunking runs before the throttle check, so every streamed event re-parses the whole reply even when the update is dropped 1ms later. Suggested to move the `now - preview.last_update < _STREAM_MIN_INTERVAL_SECS` check above `_formatted_chunks`.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1781 @@
...
+    from ...message_bus import MessageBus
+    from ....tool import ToolBase
+    from ....workspace import WorkspaceBase
+    from ._markdown import _TelegramTextChunk
+
+
+_POLL_TIMEOUT_SECS = 30
+_POLL_READ_TIMEOUT_SECS = 40
+_ALBUM_SETTLE_SECS = 0.8
+_STREAM_MIN_INTERVAL_SECS = 1.0
```
DavdGao, line 68: 1s is 3x over Telegram's group limit — the [FAQ](https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this) says a bot can't send more than 20 messages per minute to the same group, and edits count. You'll hit `RetryAfter` constantly, and `_update_stream_preview` doesn't go through `_retry_api`, so the first one just disables the whole preview.

Suggested to use a separate interval for the `edit` path (3s+) and keep 1s only for private-chat drafts.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1781 @@
...
+        self,
+        operation: Callable[[], Awaitable[_T]],
+    ) -> _T:
+        from telegram.error import BadRequest, NetworkError, RetryAfter
+
+        network_failures = 0
+        while True:
+            try:
+                return await operation()
+            except RetryAfter as error:
```
DavdGao, line 1754: The `RetryAfter` branch is an unbounded loop — `network_failures` is reset every time and there's no cap on total attempts or total wait, so a bot that keeps getting flood-limited leaves this task hanging forever. Suggested to add a total budget (attempts or elapsed seconds) and give up after it.

src/agentscope/app/channel/_base.py
```diff
@@ -389,6 +390,18 @@ async def aclose(self) -> None:
         loop keeps releasing its own. Default: nothing to do.
         """
 
+    def bind_message_bus(self, message_bus: "MessageBus") -> None:
```
DavdGao, line 393: Adding a hook to `ChannelBase` for one channel is a big deal — the module docstring says a channel only keeps a connection, normalises events and sends replies, and today all bus access lives in the gateway/dispatcher. The need is real though (callback_data is capped at 64 bytes and the card is sent from a different process than the one receiving the click), so I'm fine keeping it.

Just don't document it as Telegram-specific — describe it as shared state for channels that need cross-process correlation, and use it for the poller lock too.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1781 @@
...
+
+        for tool in request.tool_calls:
+            base = {
+                "tool_call_id": tool.id,
+                "chat_id": event.chat_id,
+                "agent_id": str(event.metadata.get("agent_id", "")),
+                "session_id": str(event.metadata.get("session_id", "")),
+            }
+            try:
+                allow_callback = await self._store_approval_callback(
```
DavdGao, line 1419: Storing two tokens per tool call means two registry writes, and clicking Allow only retires the allow token — the deny one lingers for 24h. Suggested to store one payload per tool call and put the decision in the callback data itself, e.g. `as:a:<token>` / `as:d:<token>`, so both buttons share one entry that gets deleted once.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1781 @@
...
+        return ""
+
+    # -- Outbound replies and approvals -------------------------------
+
+    async def _new_stream_preview(
+        self,
+        event: ChannelEvent,
+    ) -> _StreamPreview:
+        """Choose native private-chat drafts or editable messages."""
+        chat_type = str(event.metadata.get("chat_type", ""))
```
DavdGao, line 1053: `event.metadata["chat_type"]` is never set on this path — `send_response` is only ever called from `ChannelClients.deliver`, whose synthetic event carries just `session_id` and `agent_id`. So this branch is dead and every chat pays an extra `getChat` on its first reply. Suggested to drop the metadata lookup and rely on `chat_kind` alone, or have the delivery target carry the chat type.

src/agentscope/app/channel/_telegram/_channel.py
```diff
@@ -0,0 +1,1781 @@
+# -*- coding: utf-8 -*-
```
DavdGao, line 1: 1781 lines in one file, larger than Feishu (1362) and DingTalk (1226). The markdown part is already split out; suggested to move the inbound normalisation (`_select_media` / `_download_media` / album buffering / entity parsing / `_structured_text`, ~450 lines) into `_inbound.py` as well.

tests/telegram_channel_test.py
```diff
@@ -0,0 +1,1878 @@
...
+def _deterministic_event(event: ChannelEvent) -> dict[str, Any]:
+    """Dump an event without generated timestamps and block ids."""
+    payload = event.model_dump(exclude={"received_at"})
+    for block in payload["content"]:
+        block.pop("id", None)
+        block.pop("created_at", None)
+    return payload
+
+
+def _expected_text_event(
```
DavdGao, line 100: `_expected_text_event` builds the expected value through a helper, which means you can't see what the test actually expects without jumping around. Suggested to write the expected dict as a literal in each assertion — use `AnyString()` from `tests/utils.py` for the generated fields instead of excluding them.

docs/NEWS.md
```diff
@@ -3,7 +3,7 @@
 <!-- To update news in READMEs, modify this file and push to trigger the workflow. -->
 
 - **[2026-08] `FEAT`:** Console supported — test and debug agents in the terminal. [Example](https://github.com/agentscope-ai/agentscope/tree/main/examples/console) | [Docs](https://docs.agentscope.io/latest/en/building-blocks/console)
-- **[2026-08] `INTE`:** Feishu (Lark) and Discord channels supported. [Feishu](https://docs.agentscope.io/latest/en/deploy/channel/feishu) | [Discord](https://docs.agentscope.io/latest/en/deploy/channel/discord)
+- **[2026-08] `INTE`:** Feishu (Lark), Discord, and Telegram channels supported. [Feishu](https://docs.agentscope.io/latest/en/deploy/channel/feishu) | [Discord](https://docs.agentscope.io/latest/en/deploy/channel/discord)
```
DavdGao, line 6: Feishu and Discord both link to their docs page, Telegram doesn't. Also the trailing-newline changes in this file and `examples/agent_service/README.md` are unrelated noise, suggested to revert them.

DavdGao (commented):
Overall the platform side is solid — HTML parse mode instead of MarkdownV2, private-chat `sendMessageDraft` vs. group edit matching Telegram's "drafts are private-chat only" restriction, and the cross-process approval token are all the right calls.

Main things to fix are the `Conflict` handling (a rolling restart currently kills the channel permanently) and trimming the scope: the access-control config and the unrelated test changes should come out. Details inline.

Two things worth a follow-up issue rather than this PR, once PTB catches up (22.8 is still on Bot API 10.0): `sendRichMessage` (Bot API 10.1) would give native tables instead of the ASCII-in-`<pre>` rendering, and ephemeral messages (10.2) would make group approvals visible only to the requester — right now anyone in the group can click Allow.

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Addressed in ae94c3f4. Polling now runs under a per-channel shared MessageBus lease, so only one replica calls getUpdates. Conflict is treated as retryable and triggers exponential reconnect; focused tests cover lease serialization, cancellation release, and repeated Conflict recovery.

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Addressed in ae94c3f4. The Telegram-specific private/group access-control fields and /start response were removed. I opened #2621 to track platform-agnostic inbound access control in the routing layer.

tests/channel_clients_test.py
RICARDO51167: Addressed in ae94c3f4. The unrelated complete-stream timeout test and the channel_stream_test.py change are absent from the final diff. channel_clients_test.py now only covers the shared bus binding required by listener coordination and cross-process callback correlation.

pyproject.toml
RICARDO51167: Addressed in ae94c3f4: the dependency is now python-telegram-bot>=22.8, without the unused callback-data extra or an upper bound.

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Addressed in ae94c3f4. The throttle and retry-not-before checks now run before Markdown parsing; the regression test verifies _formatted_chunks is not called for a throttled update.

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Addressed in ae94c3f4. Private drafts retain a 1s interval, group message edits use 3.1s, and RetryAfter establishes a preview cooldown instead of permanently disabling the preview.

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Addressed in ae94c3f4. Each logical Telegram API operation now has a 300s monotonic deadline shared across RetryAfter and network retries, and raises a classified budget-exhausted result when the next delay cannot fit.

src/agentscope/app/channel/_base.py
RICARDO51167: Addressed in ae94c3f4. ChannelBase documents bind_message_bus as generic cross-process correlation/coordination, and Telegram uses the shared bus both for approval state and its polling lease.

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Addressed in ae94c3f4. Allow and Deny now share one registry payload and encode only the decision plus the common token as as:a:<token> / as:d:<token>. The payload is retired after submission.

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Addressed in ae94c3f4. _new_stream_preview now relies on chat_kind(event.chat_id); the unreachable event.metadata["chat_type"] lookup was removed.

src/agentscope/app/channel/_telegram/_channel.py
RICARDO51167: Addressed in ae94c3f4. Inbound filtering, album handling, entity parsing, media download, and structured-message normalization moved into _telegram/_inbound.py.

tests/telegram_channel_test.py
RICARDO51167: Addressed in ae94c3f4. Expected ChannelEvent values are now literal dictionaries, with AnyString() only for generated IDs and timestamps.

docs/NEWS.md
RICARDO51167: Addressed in aca3222b. NEWS and its generated README copies now link Telegram in English and Chinese, and the unrelated EOF-newline changes were removed. The corresponding setup guide is in agentscope-ai/docs#68.

RICARDO51167:
The latest review feedback is addressed in `ae94c3f4` and `aca3222b`, with `upstream/main` merged in `8810b5f7`.

Highlights:
- one shared listener lease with retryable `Conflict` recovery
- Telegram-specific access-control config removed; generic follow-up tracked in #2621
- unrelated stream/client regression tests removed from the PR diff
- `python-telegram-bot>=22.8` without the callback-data extra or upper bound
- pre-render throttling, a 3.1s group edit cadence, preview `RetryAfter` cooldown, and a bounded per-operation retry budget
- one shared approval payload per tool call
- inbound normalization split into `_inbound.py` and deterministic event assertions expanded
- Telegram documentation added in agentscope-ai/docs#68; Bot API follow-up tracked in #2622

Verification:
- focused channel suite: 92 passed, 18 subtests passed
- full pre-commit suite: passed
- GitHub Actions: Ubuntu, macOS, Windows, pre-commit, and PR-title checks passed

@DavdGao @qbc2016, all requested changes are pushed and the current CI matrix is green. Could you please take another look?

tests/telegram_channel_test.py
```diff
@@ -0,0 +1,1934 @@
...
+from agentscope.message import (
+    Base64Source,
+    DataBlock,
+    TextBlock,
+    ToolCallBlock,
+    URLSource,
+)
+from agentscope.message._block import ToolResultState
+from agentscope.permission import PermissionBehavior
+from tests.utils import AnyString
```
qbc2016, line 54: Please follow the existing repository convention:

```python
from utils import AnyString
```

qbc2016 (commented):
Please see the inline comments and use complete structural assertions in the new tests.

Some event and Markdown tests now compare complete structures, but many other deterministic results are still checked field by field or with partial `assertIn` assertions. This includes returned message blocks, structured location/contact text, Telegram API call arguments, approval callback payloads, and tool results.

Please prefer `assertEqual` against the complete dataclass, `model_dump()`, call list, or exact string wherever the result is deterministic, using matchers only for genuinely generated fields. This will ensure that unexpected additional or changed fields cannot pass unnoticed.

 Please also merge the latest `main` branch and fix the conflicts. Others LGTM.

RICARDO51167:
@qbc2016 Thanks for the follow-up review. The remaining assertions are strengthened in `2834b187`, on top of `7c90efae`.

- Test utilities follow the repository convention (`from utils import AnyString, AnyValue`).
- Deterministic results now use complete structures: configuration/capabilities, media selection and message blocks, location/contact text, Telegram API call lists (including file bytes and filenames), delivery results, approval keyboards and shared payloads, confirmation events, tool results/permissions, and Markdown chunks. Matchers are limited to generated IDs and timing values.
- The latest upstream `main` (`083cbd19`) was merged in `d2ab23d2`, including resolution of the README conflicts.

Validation on the pushed commit:
- Focused channel and lifecycle suites: **116 passed, 18 subtests passed**.
- Full `pre-commit run --all-files`: passed.
- Four previously undetected mutation probes are now rejected by assertions: wrong image/file chat ID, wrong draft chat ID, an extra approval-keyboard row, and missing bold formatting in long Markdown chunks.

Could you please take another look?

qbc2016 (approved):
LGTM.

PR #2275, opened 2026-08-09
fix(memory): isolate agentic memory retrieval tasks

### LINKED ISSUE #2274: [Bug]: AgenticMemoryMiddleware shared instances can clobber concurrent retrieval tasks
### Summary

`AgenticMemoryMiddleware` keeps the async retrieval task in a single instance-level `_retrieval_task`. If one middleware instance is shared by multiple agents/sessions, a later reply can overwrite the earlier task, and the earlier reply's `finally` block can cancel the later session's retrieval instead of its own.

### Reproduction

At `29b592358c2e983a0d10dd5227316b7a02d8c23a`, this can be reproduced with two concurrent `on_reply` calls on the same middleware instance:

1. Session A starts `on_reply` and creates retrieval task A.
2. Session B starts before A finishes and overwrites `_retrieval_task` with retrieval task B.
3. Session A exits its handler first.
4. A's `finally` cancels the current `_retrieval_task`, which is task B.

In a minimal hook-level reproduction, `cancelled_after_a` contains `sess-b`.

### Expected behavior

Each session/agent should only observe and clean up its own in-flight retrieval task. A reply finishing in session A should not cancel session B's retrieval.

### Actual behavior

The task slot is shared across the middleware instance, so concurrent replies can clobber each other. This can drop a relevant memory hint for the wrong session and can leave the earlier task orphaned.

### Suggested fix

Mirror the isolation pattern already used by `ReMeMiddleware`: keep retrieval tasks in a per-session/agent mapping, consume the matching key in `on_reasoning`, and cancel only the matching task in `on_reply` cleanup. A regression test should cover two concurrent replies sharing one `AgenticMemoryMiddleware` instance.

helloJamest:
## Summary

- isolate `AgenticMemoryMiddleware` async retrieval tasks by session/agent identity instead of one shared instance slot
- make `on_reasoning` consume only the matching retrieval task and make `on_reply` cleanup cancel only the matching task
- add regression tests for concurrent replies sharing one middleware instance across different sessions and different agents in the same session

## Root cause

`AgenticMemoryMiddleware` stored the in-flight retrieval task in a single `_retrieval_task` attribute. When the same middleware instance was shared across concurrent agents/sessions, a later reply could overwrite that slot, and an earlier reply's cleanup could cancel the later session's retrieval task.

The first fix keyed tasks by `session_id`, but that still aliased distinct agents that share the same session. The retrieval task map now keys by `(session_id, id(agent))`, which matches the in-process lifetime of the async task and keeps same-session agents isolated.

Closes #2274

## Test plan

- `PYTHONPATH=src;tests python -m unittest tests.middleware_filesystem_memory_test.AgenticMemoryMiddlewareTest.test_concurrent_replies_isolate_retrieval_tasks`
- `PYTHONPATH=src;tests python -m unittest tests.middleware_filesystem_memory_test.AgenticMemoryMiddlewareTest.test_concurrent_replies_isolate_same_session_agents`
- `PYTHONPATH=src;tests python -m unittest tests.middleware_filesystem_memory_test`
- `pre-commit run --files src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py tests/middleware_filesystem_memory_test.py`

### DIFF
--- a/src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py
+++ b/src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py
@@ -479,7 +479,7 @@ class AgenticMemoryMiddleware(MiddlewareBase):
         # In-flight asynchronous retrieval tasks started in ``on_reply`` and
         # consumed in ``on_reasoning``. Keyed per session/agent so a shared
         # middleware instance never lets one concurrent reply clobber another.
-        self._retrieval_tasks: dict[object, asyncio.Task] = {}
+        self._retrieval_tasks: dict[tuple[str | None, int], asyncio.Task] = {}
 
     @staticmethod
     def _truncate_if_needed(content: str, max_length: int) -> str:
@@ -683,10 +683,10 @@ class AgenticMemoryMiddleware(MiddlewareBase):
     # ========================================================================
 
     @staticmethod
-    def _retrieval_key_of(agent: "Agent") -> object:
+    def _retrieval_key_of(agent: "Agent") -> tuple[str | None, int]:
         """Return the key used to isolate one in-flight retrieval task."""
         session_id = getattr(getattr(agent, "state", None), "session_id", None)
-        return session_id if session_id is not None else id(agent)
+        return (session_id, id(agent))
 
     @staticmethod
     def _format_manifest(headers: list[_MemoryFileHeader]) -> str:
--- a/src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py
+++ b/src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py
@@ -476,11 +476,10 @@ class AgenticMemoryMiddleware(MiddlewareBase):
         self._parameters = parameters or self.Parameters()
         self._backend = backend or LocalBackend()
 
-        self._cached_input: str | None = None
-        # The in-flight asynchronous retrieval task started in ``on_reply``
-        # and consumed in ``on_reasoning``. Kept on the instance so the
-        # reasoning hook can poll for completion across iterations.
-        self._retrieval_task: asyncio.Task | None = None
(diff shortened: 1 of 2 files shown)


iamroylim (changes requested):
The new mapping still aliases distinct agents that share the same non-null session_id: _retrieval_key_of returns session_id instead of an agent-specific key. AgentState documents that one session can maintain independent state for each agent. At exact head 675b0b7, I reran the added suite (8 passed), then repeated the concurrent hook reproduction with agent-a and agent-b both using session_id="same-session". Starting B cancels A as stale, and when A exits its finally block it pops and cancels B; cancelled_after_a is ["agent-a", "agent-b"] and the task map is empty. Please key by agent identity as well (for example a tuple of session_id and id(agent), or another stable per-agent identifier) and add the same-session/two-agent regression. The current different-session test does not cover this case.

helloJamest:
Addressed in `78a2794`.

The retrieval task key now includes both `session_id` and the in-process agent identity: `(session_id, id(agent))`. That keeps the original different-session isolation and also prevents same-session agents from aliasing each other's in-flight retrieval tasks.

I added `test_concurrent_replies_isolate_same_session_agents`, which reproduces the reviewed case: with the previous `session_id`-only key, agent-b was cancelled when agent-a finished; with the new key it remains isolated until its own cleanup.

Validation:
- `PYTHONPATH=src;tests python -m unittest tests.middleware_filesystem_memory_test.AgenticMemoryMiddlewareTest.test_concurrent_replies_isolate_retrieval_tasks tests.middleware_filesystem_memory_test.AgenticMemoryMiddlewareTest.test_concurrent_replies_isolate_same_session_agents`
- `PYTHONPATH=src;tests python -m unittest tests.middleware_filesystem_memory_test`
- `pre-commit run --files src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py tests/middleware_filesystem_memory_test.py`

PR #2241, opened 2026-08-05
fix(utils): make DictMixin honor missing-attribute semantics

jesseedcp:
## AgentScope Version

2.0.5 (`main`, base commit `2c2f612`)

## Description

Fixes #2239

`DictMixin.__getattr__` delegated directly to `dict.__getitem__`, so a missing attribute raised `KeyError`. This violated Python's attribute protocol and broke consumers that probe optional attributes, including `hasattr()`, `getattr(..., default)`, and `copy.deepcopy()`.

This PR:

- translates a missing mapping key into `AttributeError` for attribute-style access;
- preserves the existing built-in `dict.__getitem__` dispatch for present keys;
- preserves `KeyError` for direct mapping access such as `response["missing"]`;
- adds public `ChatResponse` regression tests for missing attributes and independent nested deep copies.

Behavioral note: missing attribute access now raises the standard `AttributeError` instead of the previous non-standard `KeyError`. Direct mapping access is unchanged.

### Validation

- `python -m pytest tests/mixin_test.py tests/model_response_test.py -q -p no:cacheprovider` — 18 passed
- `pre-commit run --all-files` — all hooks passed
- Full-suite comparison in the same Windows environment:
  - patched branch: 1428 passed, 287 skipped, 92 failed
  - clean base `2c2f612`: 1426 passed, 287 skipped, 92 failed
  - the same 92 environment/optional-dependency failures occur on both revisions; this branch adds two passing regression tests and no new failures

## Checklist

- [x] An issue has been created for this PR
- [x] I have read the [CONTRIBUTING.md](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md)
- [x] Docstrings are in Google style
- [x] Related documentation has been updated (not applicable: this fixes existing behavior without adding or changing a documented public API)
- [x] Code is ready for review

PR #2281, opened 2026-08-11
fix(utils): honor Python attribute lookup semantics

### LINKED ISSUE #2239: [Bug]: DictMixin raises KeyError for missing attributes and breaks deepcopy
## Prerequisites

- [x] I searched the existing issues and discussions and could not find a duplicate.
- [x] This is a reproducible bug rather than a usage question.

## Background / Description

I encountered this while post-processing an AgentScope model response. I needed to preserve the original `ChatResponse` for logging before modifying its `content` and `metadata`, so I attempted to create an independent copy with `copy.deepcopy()`.

I expected the response to be copied normally. Instead, the operation failed with `KeyError('__deepcopy__')`.

The failure is reproducible with an empty `ChatResponse`, so it is not caused by a nested content block. Missing-attribute checks also fail: `hasattr(response, "missing")` and `getattr(response, "missing", "fallback")` raise `KeyError` instead of returning `False` and `"fallback"`.

`ChatResponse` inherits from `DictMixin`, whose `__getattr__` currently delegates directly to `dict.__getitem__`. A missing mapping key raises `KeyError`, while Python's attribute protocol expects `__getattr__` to raise `AttributeError` when an attribute is unavailable.

A focused fix would translate the `KeyError` raised during attribute lookup into `AttributeError`, while preserving `KeyError` for direct item access such as `response["missing"]`. I plan to submit a small PR with regression tests for `deepcopy()`, `hasattr()`, `getattr(..., default)`, and direct mapping access.

## Error Messages

```shell
AgentScope: 2.0.4.post1
  File "<stdin>", line 8, in <module>
  File "C:\Users\dd\AppData\Local\Programs\Python\Python312\Lib\copy.py", line 141, in deepcopy
    copier = getattr(x, "__deepcopy__", None)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: '__deepcopy__'
```

## Steps to Reproduce

```python
from copy import deepcopy

response = ChatResponse(content=[], is_last=True)

deepcopy(response)
hasattr(response, "missing")
getattr(response, "missing", "fallback")
```

Expected behavior:

```python
copied = deepcopy(response)
assert copied == response
assert copied is not response
assert hasattr(response, "missing") is False
assert getattr(response, "missing", "fallback") == "fallback"
```

Actual behavior:

```text
deepcopy(response)                         -> KeyError('__deepcopy__')
hasattr(response, "missing")               -> KeyError('missing')
getattr(response, "missing", "fallback")   -> KeyError('missing')
```

## Environment

- AgentScope Version: observed in installed `2.0.4.post1`; confirmed on current `main` (`2.0.5`) at commit `2c2f612`
- Python Version: `3.12.10`
- OS: Windows `10.0.26200.0`

uuzzrm:
## Summary

Fixes #2239.

`DictMixin` uses dictionary keys for attribute-style access, but a missing key currently raises `KeyError`. That breaks Python's attribute lookup contract: `hasattr()` and `getattr(..., default)` cannot handle the missing attribute, and `copy.deepcopy()` fails while probing for `__deepcopy__`.

This change translates missing mapped attributes into `AttributeError` while keeping direct mapping access unchanged, so `response["missing"]` still raises `KeyError`.

## Tests

- `uv run pre-commit run --files src/agentscope/_utils/_mixin.py tests/model_response_test.py`
- `uv run pytest tests/model_response_test.py -q` (19 passed)
- `uv run pytest tests -q` (1764 passed, 147 skipped; one pre-existing failure in `tests/mcp_sse_client_test.py::SseSchemaDefsPreservationTest::test_defs_preserved_and_titles_stripped`, also reproduced on `origin/main`)

## Notes

No public API names or serialized response fields change; this only restores standard Python behavior for missing attributes.

### DIFF
--- a/src/agentscope/_utils/_mixin.py
+++ b/src/agentscope/_utils/_mixin.py
@@ -1,9 +1,16 @@
 # -*- coding: utf-8 -*-
 """The mixin for agentscope."""
+from typing import Any
 
 
 class DictMixin(dict):
     """The dictionary mixin that allows attribute-style access."""
 
     __setattr__ = dict.__setitem__
-    __getattr__ = dict.__getitem__
+
+    def __getattr__(self, name: str) -> Any:
+        """Return a mapped value or signal a missing attribute."""
+        try:
+            return self[name]
+        except KeyError as exc:
+            raise AttributeError(name) from exc
--- a/tests/model_response_test.py
+++ b/tests/model_response_test.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """Unit tests for :class:`agentscope.model.ChatResponse` and its
 ``append_*`` helpers."""
+from copy import deepcopy
+from unittest import TestCase
 from unittest.async_case import IsolatedAsyncioTestCase
 
 from utils import AnyString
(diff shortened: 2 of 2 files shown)


iamroylim (approved):
Verified at exact head 61007437819a1848d159e5aa1b5d7255cbb4d756. Translating only missing mapping keys to AttributeError restores Python attribute lookup semantics while preserving direct KeyError behavior. I reran tests/model_response_test.py (19 passed) and the pre-commit file set; all checks passed.

uuzzrm:
@iamroylim Thanks for verifying the exact head and approving this. The current head `61007437` is ready from my side. Could a maintainer merge it when convenient?

DavdGao:
Fixed in #2241

Issue #1711, opened 2026-05-28
[Bug]:ToolOffloadMiddleware Prematurely Terminates Conversation When Tools Are Offloaded


ljl3937:
# Bug Report: ToolOffloadMiddleware Prematurely Terminates Conversation When Tools Are Offloaded

**Describe the bug**
When a tool execution exceeds the default timeout (10 seconds), `ToolOffloadMiddleware` offloads the tool to background and returns a synthetic `ToolResponse` with state `SUCCESS`. This causes the agent's ReAct loop to incorrectly assume the task is complete and exit prematurely, even though the background tool is still running. When the background tool finally completes, the retrigger mechanism fails to properly resume the conversation because:

1. The injected result is a `HintBlock` (plain text notification) rather than a proper `ToolResultBlock`
2. The agent has already exited the ReAct loop and session is no longer running
3. The conversation ends with incomplete/truncated responses

This is a critical bug in the `ToolOffloadMiddleware` design - the offloaded tool response should not be marked as `SUCCESS`, and the retrigger mechanism cannot reliably resume agent execution.

**To Reproduce**

1. Create an agent with a tool that takes longer than 10 seconds to execute (e.g., a file writing tool with large content, or a tool with sleep/delay)

2. Configure the agent via Web UI with default settings:
   ```python
   from agentscope.app import create_app
   from agentscope.app.storage import RedisStorage
   
   app = create_app(storage=RedisStorage())
   # ToolOffloadMiddleware is automatically added with default timeout_secs=10.0
   ```

3. Send a message that triggers the slow tool execution via Web UI chat

4. Observe the behavior:
   - After ~10 seconds, the tool is offloaded to background
   - Agent receives a synthetic `ToolResponse` with placeholder text: "Tool 'xxx' is taking longer than expected..."
   - Agent incorrectly assumes task is complete and generates a final response
   - Conversation ends prematurely
   - The actual background tool result arrives later but cannot be properly injected

**Expected behavior**
When a tool execution times out and is offloaded to background:

1. The agent should NOT mark the tool call as `SUCCESS` or exit the ReAct loop
2. The agent should either:
   - Wait for the background tool to complete (blocking execution)
   - Continue with other tool calls while the offloaded tool runs in background
   - Properly handle the injected `ToolResultBlock` when retriggered after background completion
3. The conversation should NOT end until all tools (including offloaded ones) have completed and their results have been properly processed
[...]

PR #2286, opened 2026-08-11
fix(service): preserve busy background wakeups

Riofd:
## Summary

- defer plain wake triggers while their target session still holds the run lock instead of dropping them
- check the inbox under the session lock before a wake-only run, so a deferred or duplicate wake cannot start an empty model turn on built-in transports
- add non-destructive queue length support for the in-memory and Redis message buses, with a backward-compatible fallback for custom transports

## Why

A background completion can enqueue its inbox entry and wake trigger after the live run has performed its final inbox drain but just before that run releases the session lock. The dispatcher previously dropped the wake because the session was still busy, leaving the inbox entry stranded with no later trigger.

Retrying the wake closes that timing window. Pairing the retry with an inbox check under the same distributed lock prevents a delayed or duplicate wake from invoking the model after another run has already consumed the inbox entry.

## Scope

Refs #1711 and #1868. This is a focused fix for wake delivery during normal session completion. It does not change offloaded-tool result semantics, and it does not implement Redis consumer groups or crash recovery for in-flight wake triggers. Custom message-bus transports that do not implement non-destructive queue inspection preserve delivery but cannot suppress duplicate empty wakes.

## Validation

- `pytest tests`: 1832 passed, 89 skipped
- focused wake-up and message-bus suites: 102 passed

### DIFF
--- a/src/agentscope/app/_bus_ops.py
+++ b/src/agentscope/app/_bus_ops.py
@@ -90,9 +90,9 @@ async def enqueue_run_trigger(
 
     ``kind`` selects how the dispatcher handles the entry:
 
-    - ``wake`` — idle-session wake-up.  The dispatcher skips the entry
-      when the session is already running (the live run drains the inbox
-      itself).  ``inputs`` must be ``None``.
+    - ``wake`` — idle-session wake-up.  The dispatcher re-queues the entry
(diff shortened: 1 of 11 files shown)


Riofd:
This issue was solved in latest 2.0.6. Close it.

PR #2289, opened 2026-08-11
fix(app): purge worker workspace state with leader session

### LINKED ISSUE #2288: [Bug]: Deleting a team leader session leaves worker workspace and MCP state behind
### Prerequisites

### Background / Description

Deleting a team leader session cascades through storage and removes the team's
worker sessions, but `SessionService` only purges the leader's workspace scope.

Before storage mutation, the service collects worker session IDs for run
cancellation and message-bus cleanup:

```python
worker_sids = await self._team_worker_session_ids(...)
all_sids = [session_id, *worker_sids]
```

This is sufficient for cancellation and bus cleanup because those operations
only require a session ID. However, workspace cleanup additionally requires the
session owner, agent ID, and persisted workspace ID.

After `storage.delete_session()` dissolves the team, the worker records no
longer exist. The final workspace cleanup therefore only calls:

```python
await workspace.purge_session(
    agent_id=agent_id,
    session_id=session_id,
)
```

where `agent_id` and `session_id` still refer to the leader.

This leaves worker-scoped state behind:

- stateful MCP instances and subprocesses;
- persisted `.mcp` declarations;
- session offload files;
- `_mcp_last_used` capacity records.

This also affects shared workspaces. `WorkspaceBase.purge_session()` scopes MCP
state by `(agent_id, session_id)`, so purging the leader does not implicitly
purge worker scopes.

Expected behavior:

- every session removed by the storage cascade has its workspace scope purged;
- created workers are purged along with the leader;
- invited agents only lose their borrowed team session scope;
- an invited agent's ordinary sessions and agent-level state survive;
- one failed workspace purge does not prevent cleanup of later targets.

### Error Messages

```shell
Observed deterministic cleanup calls:


{
  "bus_purged": [
    "leader-session",
    "worker-session"
  ],
  "workspace_purged": [
    ["leader-agent", "leader-session"]
  ]
}


Expected `workspace_purged` to also contain:


["worker-agent", "worker-session"]
```

### Steps to Reproduce

1. Create a leader agent and session.
2. Create a team from that session.
3. Add a created worker. The worker receives its own agent/session scope while
   sharing the leader workspace.
4. Register worker-scoped MCP state or offload session data.
5. Delete the leader session through
   `DELETE /sessions/{leader_session_id}`.
6. Observe that storage removes the team and worker records.
7. Observe that message-bus state is purged for both leader and worker sessions.
8. Inspect calls to `Workspace.purge_session()` or inspect the remaining
   workspace state.

A minimal service-level regression assertion is:

```python
assert workspace.purge_session.await_args_list == [
    call(
        agent_id="leader-agent",
        session_id="leader-session",
    ),
    call(
        agent_id="worker-agent",
        session_id="worker-session",
    ),
]
```

On the affected implementation, only the first call is made.

### Environment

- AgentScope Version: 2.0.6
- Python Version: 3.11.7
- OS: Windows 10 (10.0.26200)

redSun64:
### Background

Deleting a team leader session triggers a storage cascade that removes the team's worker sessions. Before this change, `SessionService.delete_session()` collected the worker session IDs so that it could cancel their runs and purge their message-bus state, but workspace cleanup was only performed for the leader session.

This difference matters because workspace state is scoped by both `agent_id` and `session_id`. Purging the leader scope does not remove state belonging to a worker, even when the leader and workers use the same physical workspace. The remaining worker state can include MCP clients and subprocesses, persisted `.mcp` declarations, offload files, and `_mcp_last_used` capacity records.

The worker session records are no longer available after the storage cascade, so the missing agent and workspace identities cannot be reconstructed safely at cleanup time.

### What changed

- Added the private `_SessionDeleteTarget` value object to retain the `user_id`, `agent_id`, `session_id`, and `workspace_id` needed for cleanup.
- Replaced the worker-session-ID-only lookup with `_session_delete_targets()`, which snapshots the leader and every session that the storage layer will cascade-delete before storage is mutated.
- Continued to use the collected session IDs for run cancellation and message-bus cleanup.
- After a successful storage deletion, resolved the persisted workspace for each captured target and called `purge_session()` with that target's exact `(agent_id, session_id)` scope.
- Kept workspace cleanup best-effort and isolated per target: a failure while purging one scope is logged but does not undo the committed storage deletion or prevent later scopes from being purged.
- Added focused regression tests for ordinary sessions, created workers, invited workers, shared workspaces, and per-target failure isolation.

### Why this approach

The fix captures complete deletion targets before calling `storage.delete_session()` because this is the last point at which every worker's session record and persisted workspace binding are reliably available. Keeping only session IDs is insufficient for workspace cleanup, while attempting to query the records after deletion is too late.

The implementation deliberately mirrors the storage cascade instead of changing storage APIs or moving workspace concerns into the storage layer. This preserves the existing separation of responsibilities: storage owns durable records, while `SessionService` coordinates storage, message-bus, run, and workspace lifecycle cleanup.

Team members are obtained from the role-aware team roster. This is important for invited members because an invited agent may also have unrelated ordinary sessions. Using the exact borrowed team session ensures that deleting the leader cleans only the session participating in the team; it does not delete or purge the invited agent's other sessions or agent-wide state.

The code still purges every `(agent_id, session_id)` independently when multiple sessions share one workspace. A shared workspace is only the physical container; its MCP and offload state remains session-scoped, so one purge call per deleted session is required.

Finally, cleanup remains best-effort because the durable deletion may already have committed before a workspace backend fails or becomes unavailable. Each target has its own exception boundary so one failure does not leave all later worker scopes unprocessed.

### Scope and compatibility

This is a focused, single-purpose bug fix, as requested by `CONTRIBUTING.md`. It changes two files: the session-deletion orchestration and its regression tests.

- No public API is changed.
- No dependency is added.
- No optional dependency import is introduced.
- No migration or deprecation is required.
- Ordinary non-team session deletion preserves its existing behavior.
- Invited agents and their unrelated ordinary sessions remain intact.

### Documentation

No companion documentation PR is required. The change corrects internal resource-lifecycle behavior and does not introduce a module, public API, configuration option, tutorial change, or other user-facing interface.

### Testing

Full pre-commit validation:

```powershell
python -m pre_commit run --all-files --show-diff-on-failure
```

All hooks passed, including AST/YAML/TOML/JSON checks, trailing whitespace, trailing commas, mypy, black, flake8, pylint, and Pyroma.

Focused unit and integration coverage:

```powershell
python -m pytest -q `
  tests/service_session_delete_test.py `
  tests/storage_redis_test.py::TestTeamCascade `
  tests/service_team_tools_test.py `
  tests/workspace_local_test.py::TestLocalWorkspaceMCPScoping::test_purge_session_drops_scope_and_offload
```

Result:

```text
49 passed
```

Fixes #2288

## Checklist

- [x] An issue has been created for this PR.
- [x] I have read the [CONTRIBUTING.md](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md).
- [x] Related documentation is not required because no public or user-facing interface changed.


DavdGao (commented):
Thanks for pointing out this bug. Some suggestions:

1. **Reuse `SessionService.delete_team()` instead of re-deriving the cascade.** It's already the role-aware version of this exact walk (`_session.py:460-469`). Something like:

```python
# in delete_session, after _purge_subagent_hitl, before cancel/storage delete
session = await self._storage.get_session(user_id, agent_id, session_id)
if session and session.team_id:
    team = await self._storage.get_team(user_id, session.team_id)
    if team is not None and team.session_id == session_id:
        await self.delete_team(user_id, session.team_id)
```

Then `_team_worker_session_ids` and `_SessionDeleteTarget` can both go away.

2. **This also fixes the other half of the leak.** Storage deletes `role == "created"` members via `delete_agent` (`_redis_storage.py:1743-1751`), so their skill partition needs `purge_agent()` too — the current patch only restores `purge_session()`. Delegating gets it for free via `SessionService.delete_agent()` (`_session.py:557`).

3. **Keep the call after `_purge_subagent_hitl`** — it resolves the session's role from the team record (`_session.py:681`).

4. **Add a cycle guard in `delete_team`:** skip members whose `session_id == team.session_id`. Shouldn't happen, but a corrupt roster listing the leader as its own member would recurse.

redSun64:
> Thanks for pointing out this bug. Some suggestions:
> 
> 1. **Reuse `SessionService.delete_team()` instead of re-deriving the cascade.** It's already the role-aware version of this exact walk (`_session.py:460-469`). Something like:
> 
> ```python
> # in delete_session, after _purge_subagent_hitl, before cancel/storage delete
> session = await self._storage.get_session(user_id, agent_id, session_id)
> if session and session.team_id:
>     team = await self._storage.get_team(user_id, session.team_id)
>     if team is not None and team.session_id == session_id:
>         await self.delete_team(user_id, session.team_id)
> ```
> 
> Then `_team_worker_session_ids` and `_SessionDeleteTarget` can both go away.
> 
> 2. **This also fixes the other half of the leak.** Storage deletes `role == "created"` members via `delete_agent` (`_redis_storage.py:1743-1751`), so their skill partition needs `purge_agent()` too — the current patch only restores `purge_session()`. Delegating gets it for free via `SessionService.delete_agent()` (`_session.py:557`).
> 3. **Keep the call after `_purge_subagent_hitl`** — it resolves the session's role from the team record (`_session.py:681`).
> 4. **Add a cycle guard in `delete_team`:** skip members whose `session_id == team.session_id`. Shouldn't happen, but a corrupt roster listing the leader as its own member would recurse.

Thanks for the detailed suggestions. I’ve updated the implementation accordingly:

- Reused `SessionService.delete_team()` in `delete_session()` after `_purge_subagent_hitl()`.
- Removed the manually derived worker-session cascade.
- Created members now go through `delete_agent()`, ensuring both session state and agent-level skill partitions are purged.
- Invited members only have their borrowed team session removed.
- Added cycle guards for corrupt rosters that list the leader as a member.
- Added regression tests covering role-aware cleanup, HITL ordering, and the cycle guards across the service, Redis, and SQL paths.

All related tests and pre-commit checks pass.

PR #2298, opened 2026-08-12
feat(tts): add MiniMax speech synthesis

octo-patch:
Reason: Add native MiniMax T2A support to AgentScope's TTS framework.

This change:

- registers MiniMax credentials and TTS model discovery;
- supports streaming and non-streaming T2A responses across configurable regional endpoints;
- exposes voice, audio, language, and endpoint parameters;
- adds model cards for the supported speech models and focused unit coverage.

Checks:

- `.venv312/bin/python -m pytest tests/tts_*_test.py -q` (77 passed, 5 skipped)
- `.venv312/bin/python -m black --check --line-length 79 src/agentscope/tts/_minimax/_model.py tests/tts_minimax_test.py`
- `.venv312/bin/python -m flake8 --extend-ignore=E203 src/agentscope/credential/__init__.py src/agentscope/credential/_factory.py src/agentscope/credential/_minimax.py src/agentscope/tts/__init__.py src/agentscope/tts/_minimax/__init__.py src/agentscope/tts/_minimax/_model.py tests/tts_minimax_test.py`
- `.venv312/bin/python -m mypy --disallow-untyped-defs --disallow-incomplete-defs --ignore-missing-imports --follow-imports=skip --explicit-package-bases src/agentscope/credential/_minimax.py src/agentscope/tts/_minimax/_model.py tests/tts_minimax_test.py`

### DIFF
--- a/src/agentscope/credential/_minimax.py
+++ b/src/agentscope/credential/_minimax.py
@@ -10,7 +10,7 @@ if TYPE_CHECKING:
     from ..tts import TTSModelBase
 
 
-class MiniMaxCredential(CredentialBase):
+class MiniMaxCredential(CredentialBase):  # pylint: disable=abstract-method
     """The credential for MiniMax APIs."""
 
     model_config = ConfigDict(
--- a/src/agentscope/credential/_minimax.py
+++ b/src/agentscope/credential/_minimax.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+"""The MiniMax credential."""
+from typing import Literal, Type, TYPE_CHECKING
+
+from pydantic import ConfigDict, Field, SecretStr
+
+from ._base import CredentialBase
+
+if TYPE_CHECKING:
+    from ..tts import TTSModelBase
+
+
+class MiniMaxCredential(CredentialBase):
+    """The credential for MiniMax APIs."""
+
+    model_config = ConfigDict(
+        title="MiniMax API",
+    )
+
+    type: Literal["minimax_credential"] = "minimax_credential"
+    """The credential type."""
+
+    api_key: SecretStr = Field(
+        description="The MiniMax API key.",
+        title="API Key",
+    )
+    """The API key."""
+
+    @classmethod
+    def get_tts_model_classes(cls) -> list[Type["TTSModelBase"]]:
+        """Return the MiniMax TTS model classes."""
+        from ..tts import MiniMaxTTSModel
+
+        return [MiniMaxTTSModel]
--- a/src/agentscope/credential/_minimax.py
+++ b/src/agentscope/credential/_minimax.py
@@ -7,7 +7,11 @@ from pydantic import ConfigDict, Field, SecretStr
 from ._base import CredentialBase
 
 if TYPE_CHECKING:
+    from ..model import ModelCard
     from ..tts import TTSModelBase
+
+
+_MINIMAX_GLOBAL_BASE_URL = "https://api.minimax.io"
 
 
 class MiniMaxCredential(CredentialBase):  # pylint: disable=abstract-method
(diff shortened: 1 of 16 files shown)


octo-patch:
I fixed the pre-commit pylint error for the TTS-only credential and pushed commit 64edb41. I ran the repository hooks against the changed file and the MiniMax TTS unit tests (6 passed).

src/agentscope/tts/_minimax/_model.py
```diff
@@ -0,0 +1,295 @@
...
+            description="The number of audio channels.",
+        )
+
+        language_boost: str | None = Field(
+            default="auto",
+            title="Language Boost",
+            description="The language or dialect optimization mode.",
+        )
+
+        api_url: str = Field(
```
qbc2016: **Security:** `api_url` is exposed as a per-model parameter. In a shared-credential setup, a user with read access to another user's credential can point this value to an arbitrary endpoint. `get_tts_model()` then resolves the owner's raw credential, and the client sends that API key in the `Authorization` header. This allows credential exfiltration and also creates an SSRF primitive.

Could you move the endpoint configuration to `MiniMaxCredential` as an owner-controlled `base_url`, and keep `/v1/t2a_v2` as the TTS-specific path in this model? Using `base_url` also avoids a naming conflict if a MiniMax chat model is added later. If arbitrary compatible endpoints are unnecessary, restricting this to allowlisted regional origins would be even safer.

src/agentscope/tts/_minimax/_model.py
```diff
@@ -0,0 +1,295 @@
...
+            "audio_setting": {
+                "sample_rate": self.parameters.sample_rate,
+                "bitrate": self.parameters.bitrate,
+                "format": self.parameters.response_format,
+                "channel": self.parameters.channel,
+            },
+            **kwargs,
+            "model": self.model,
+            "text": text,
+            "stream": self.stream,
```
qbc2016, line 171: **Correctness:** MiniMax includes the complete synthesized audio in the final streaming chunk by default. This generator yields the preceding incremental chunks and then yields that aggregate chunk as another increment, so consumers that concatenate the chunks will duplicate the earlier audio.

Please set `stream_options.exclude_aggregated_audio` to `true` for streaming requests, or handle the final `status == 2` payload as an aggregate rather than an incremental chunk. The streaming test should also model the provider behavior by returning `first + second` in the final chunk.

src/agentscope/tts/_minimax/_model.py
```diff
@@ -0,0 +1,295 @@
...
+            description="The output volume multiplier.",
+        )
+
+        pitch: int = Field(
+            default=0,
+            title="Pitch",
+            description="The voice pitch adjustment.",
+        )
+
+        response_format: Literal["mp3", "wav", "flac", "pcm"] = Field(
```
qbc2016: **Correctness:** MiniMax streaming synthesis currently supports only MP3, while this model always uses streaming in the application path. Exposing `response_format` therefore allows users to select formats that will fail at runtime.

Please remove this parameter and always send `"format": "mp3"` in `audio_setting`. The response media type can consequently be fixed to `audio/mpeg`.

src/agentscope/credential/_minimax.py
```diff
@@ -0,0 +1,34 @@
...
+
+from pydantic import ConfigDict, Field, SecretStr
+
+from ._base import CredentialBase
+
+if TYPE_CHECKING:
+    from ..tts import TTSModelBase
+
+
+class MiniMaxCredential(CredentialBase):  # pylint: disable=abstract-method
```
qbc2016, line 17: **Compatibility:** MiniMax currently supports only TTS, but the model discovery route directly calls `credential_cls.get_chat_model_class().list_models()`. This invokes the inherited `get_chat_model_class()` implementation and raises
`NotImplementedError`, causing `/model/?provider=minimax_credential` to return a server error.

Could you override `list_models()` in `MiniMaxCredential` to return an empty list and update the discovery route to call `credential_cls.list_models()` instead? `CredentialBase.list_models()` already delegates to the chat model class for chat-capable credentials, while this override would allow a TTS-only credential to return no chat models without changing `ChatModelBase`.

qbc2016 (commented):

octo-patch:
Thanks for the review. I moved the allowlisted regional base URL into `MiniMaxCredential`, fixed TTS-only discovery, forced MP3 streaming with aggregate-audio protection, and updated the tests and model cards. I ran the focused MiniMax tests (6 passed), Black, Flake8, and Python compilation.

PR #2300, opened 2026-08-13
refactor(app): make the message bus queue at-least-once, and move its users onto it

DavdGao:
## Problem

`queue_drain` is the bus's only way to consume a queue, and it is **at-most-once**: `XRANGE` then `XDEL`, so an entry is gone the instant it is read.

```python
entries = await bus.queue_drain(key, max_count=n)   # entry deleted here
...                                                  # crash / exception → lost
```

Three consequences, all of which show up in production:

| | |
|---|---|
| **Loss** | A consumer that dies — or merely raises — between reading and finishing loses the entry. Nothing can recover it. |
| **Duplication** | `XRANGE` and `XDEL` are two commands, so two processes can read the same entry before either deletes it. |
| **Failure == success** | There is no way to say "I could not handle this, put it back". |

The docstring assumes a *single consumer per key*, but the architecture violates it: `WakeupDispatcher` and `IndexTaskConsumer` are one-per-process and every process drains the same key. Of the queues in use today, only the session inbox is genuinely single-consumer — and single-consumer only prevents duplication, not loss.

This is the shared root cause behind #1868 (wake-ups executing N times across instances) and #2227 (resume triggers duplicated or permanently lost).

## What changed

The queue had two states — *in queue* and *gone*. This adds the missing third one:

```
                claim                      ack
  available ──────────────▶ claimed ──────────────▶ deleted
      ▲                        │
      ├──── release ───────────┤   failure path: available again at once
      └──── (idle takeover) ───┘   crash path: next claim picks it up
```

```python
queue_claim(key, *, consumer, max_count, min_idle_secs) -> [(id, payload)]
queue_ack(key, *, consumer, entry_ids) -> None
queue_release(key, *, consumer, entry_ids) -> None
```

Two deliberate choices:

- **Recovery lives inside `queue_claim`**, which takes over idle entries before topping up with fresh ones. No periodic sweep to remember to run, and a crashed consumer's work is picked up by whoever claims next.
- **No lease heartbeat.** A heartbeat would only guard against a claim outliving `min_idle_secs`, and it fails silently exactly when it is needed — the event loop is busiest when the work is slowest. Instead the contract is *claim → one durable write → ack*, with the real work after the ack, so a claim never lives long enough to be taken over.

## Atomicity

| Operation | Redis | Atomic |
|---|---|---|
| `queue_push` | `XADD` | ✅ single command |
| `queue_claim` | `XAUTOCLAIM` then `XREADGROUP` | ✅ each single command; worst case between them is a smaller batch |
| `queue_ack` | `XACK` + `XDEL` in one MULTI/EXEC | ✅ never both acknowledged and still pending |
| `queue_release` | `XCLAIM … JUSTID IDLE` | ✅ single command |

`queue_release` uses `XCLAIM`'s `IDLE` option to stamp the entry as long-idle, so the next claim takes it over. That keeps the **original id and arrival order** — unlike `XACK` + re-`XADD`, which would renumber, reorder, and lose the entry if the process died between the two commands.

The in-memory bus mirrors the same semantics with an available list plus a claimed map, and both backends run the same test class.

## `queue_drain` is deprecated, not rewritten

It stays on its own `XRANGE`/`XDEL` path rather than being layered over `queue_claim`. Sharing a key between the two protocols is unsafe: `XREADGROUP` does not remove entries from the stream, so an older process's `XRANGE` would still see — and `XDEL` — entries a newer process is holding in the group's PEL. A rolling deploy would hit exactly that.

Keeping the paths separate means **this PR changes no existing behaviour**: every current caller keeps the semantics it has today, and queues migrate one at a time onto fresh keys.

## Not in this PR

- **Consumer migration.** At-least-once means redelivery is normal, so each consumer needs its own idempotency story first — the session inbox can dedupe on `HintBlock.id` against the context it already persisted, `wake` triggers need an empty-work guard, and `resume` needs a real idempotency key (#2227).
- **`XTRIM MINID` housekeeping** for entries left behind when a process dies between `XACK` and `XDEL`. Harmless (they are unreachable, not redelivered), but they should be swept eventually.

## Requirements

`XAUTOCLAIM` needs **Redis 7.0+** on the server. `storage-redis` only pins the client library; the server requirement should be documented at deploy time.

## Tests

14 new cases per backend (`TestWorkQueue`, in both `service_message_bus_test.py` and `in_memory_message_bus_test.py`): disjoint concurrent claims, claimed entries hidden until idle, takeover preserving id and payload, ack removing an entry for good, release skipping the idle wait, release preserving arrival order, idempotent ack, and delete dropping in-flight entries.

`pre-commit run --all-files` clean; 1808 passed. The 5 `TestPerAgentMCPAPI` failures also fail on `main` and are unrelated.


DavdGao:
The fix addresses the wrong layer — it leaves the message bus bugs intact while piling on complexity.

PR #2308, opened 2026-08-13
fix(mcp): allow reconnecting stateful clients

### LINKED ISSUE #2307: [Bug]: MCPClient.connect() after close() reuses an exhausted context manager, so reconnect always crashes
### Prerequisites

### Background / Description

`MCPClient.connect()` and `close()` both document `connect() -> close() -> connect()` on the same instance as the intended recovery path after a transient MCP server failure (their own error text: connect's `RuntimeError` says "Call close() before reconnecting", close's says "Call connect() first"). That path is broken for both transports.

`close()` never resets `self._client` after `self._stack.aclose()` exhausts it. `connect()` only rebuilds `self._client` when `self._client is None and self.mcp_config.type == "http_mcp"`, so:

- stdio: `self._client` is built once in `_initialize_client()` at construction time and is never rebuilt, so it's always non-None and the guard never fires.
- http: `close()` never sets `self._client` back to `None`, so the same guard silently no-ops there too.

Either way, the second `connect()` re-enters an already-exhausted `@asynccontextmanager`, which is single-use by contextlib's own contract, and raises.

### Error Messages

```
AttributeError: '_AsyncGeneratorContextManager' object has no attribute 'args'
```

### Steps to Reproduce

1. Code (`echo_server.py`, a minimal stdio MCP server using the real `mcp` SDK):

```python
import asyncio
from mcp.server.lowlevel import Server
from mcp.server.stdio import stdio_server
import mcp.types as types

server = Server("echo")

@server.list_tools()
async def list_tools():
    return [types.Tool(name="echo", description="echo", inputSchema={"type": "object"})]

@server.call_tool()
async def call_tool(name, arguments):
    return [types.TextContent(type="text", text=arguments.get("text", ""))]

async def main():
    async with stdio_server() as (read_stream, write_stream):
        await server.run(read_stream, write_stream, server.create_initialization_options())

if __name__ == "__main__":
    asyncio.run(main())
```

```python
import asyncio, sys
from agentscope.mcp import MCPClient, StdioMCPConfig

async def main():
    client = MCPClient(
        name="echo",
        is_stateful=True,
        mcp_config=StdioMCPConfig(command=sys.executable, args=["echo_server.py"]),
    )
    await client.connect()
    await client.close()
    await client.connect()  # raises

asyncio.run(main())
```

2. Run: `python repro.py`
3. See error above on the second `connect()`

Ran this against a real stdio subprocess (no mocks) in a clean `python:3.11-slim` container, `pip install mcp<2.0.0 pydantic httpx`, against current `main` (`31e0ec6`).

### Environment

- AgentScope Version: 2.0.6
- Python Version: 3.11 (reproduced in a clean container; not version-specific, the bug is in the context-manager lifecycle)
- OS: Linux (container)

### Additional context

Disclosure: I used an AI coding assistant to help find and verify this.

Two-line fix shape, mirroring what `connect()` already does for `http_mcp`: reset `self._client = None` in `close()`'s `finally` block, and have `connect()` rebuild `self._client` for stdio too (re-run the `_initialize_client()` construction) instead of only doing it for `http_mcp`. Happy to send a PR for this if useful.

Solaris-star:
## AgentScope Version

2.0.6 (current `main` at `31e0ec61`)

## Description

Fixes #2307.

`MCPClient` transport context managers are single-use. After `connect()` followed by `close()`, the stateful client retained the exhausted transport and a second `connect()` reused it. This caused reconnects to fail for both stdio and HTTP configurations.

This change:

- clears the transport reference during `close()`;
- recreates the transport only when the next `connect()` needs one;
- adds focused lifecycle regression tests for stdio and HTTP clients.

No public API shape or dependency changes are included.

## Verification

- `PYTHONPATH=src pytest -q tests/mcp_client_reconnect_test.py` — 2 passed
- `PYTHONPATH=src pytest -q tests/mcp_streamable_http_client_test.py tests/mcp_sse_client_test.py` — 5 passed
- `pre-commit run --files src/agentscope/mcp/_mcp_client.py tests/mcp_client_reconnect_test.py` — all hooks passed
- `git diff --check` — passed

## Checklist

### DIFF
--- a/src/agentscope/mcp/_mcp_client.py
+++ b/src/agentscope/mcp/_mcp_client.py
@@ -248,6 +248,7 @@ class MCPClient(BaseModel):
         except Exception:
             await self._stack.aclose()
             self._stack = None
+            self._client = None
             raise
 
     async def close(self, ignore_errors: bool = True) -> None:
--- a/src/agentscope/mcp/_mcp_client.py
+++ b/src/agentscope/mcp/_mcp_client.py
@@ -164,15 +164,7 @@ class MCPClient(BaseModel):
         self._initialize_client()
 
     def _initialize_client(self) -> None:
-        """Pre-build the stdio client context manager.
-
-        Only the stdio transport is materialised at construction time —
-        ``StdioServerParameters`` carry process-launch details that are
-        cheap to bind upfront. HTTP transports are built lazily inside
-        :meth:`connect` (or per-call inside :meth:`_get_client_gen` for
-        stateless mode), since each ``streamable_http_client`` /
-        ``sse_client`` is a one-shot context manager.
-        """
+        """Pre-build the stdio client context manager."""
         if self.mcp_config.type == "stdio_mcp":
             config = self.mcp_config
             self._client = stdio_client(
@@ -233,10 +225,15 @@ class MCPClient(BaseModel):
                 "Call close() before reconnecting.",
             )
 
-        # Create HTTP client if needed
-        if self._client is None and self.mcp_config.type == "http_mcp":
-            self._client = self._create_http_client()
+        # Transports are one-shot context managers. Recreate them before every
+        # connection so connect() -> close() -> connect() starts a fresh one.
+        if self._client is None:
+            if self.mcp_config.type == "http_mcp":
+                self._client = self._create_http_client()
+            else:
+                self._initialize_client()
 
+        assert self._client is not None
         self._stack = AsyncExitStack()
 
         try:
@@ -288,6 +285,7 @@ class MCPClient(BaseModel):
                 str(e),
             )
         finally:
+            self._client = None
             self._stack = None
             self._session = None
             self._is_connected = False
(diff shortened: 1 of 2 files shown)


src/agentscope/mcp/_mcp_client.py
DavdGao, line 250: We should also reset the client here `self._client = None`

DavdGao (commented):
Plz see inline comments, others lgtm

src/agentscope/mcp/_mcp_client.py
Solaris-star: Done — added `self._client = None` in the connect() except block, plus a regression test (`test_failed_connect_can_be_retried`) that verifies a failed first connect still allows a successful retry with a fresh transport. All 3 reconnect tests + 5 existing MCP tests pass.

Solaris-star:
Addressed the inline review — `self._client = None` is now set in the `connect()` except block so that a failed connect discards the one-shot transport and a retry recreates it.

**Changes since initial push:**
- `src/agentscope/mcp/_mcp_client.py`: +1 line — `self._client = None` in the except block
- `tests/mcp_client_reconnect_test.py`: added `test_failed_connect_can_be_retried` — fails a first connect with a `_FailOnceSession` that raises on first `initialize()`, then asserts the retry succeeds with a fresh transport

**Test results:**
```
3 passed in 0.14s (reconnect tests)
5 passed in 76.80s (existing MCP tests)
```

**Commit:** `b341ce76` (Solaris-star <820622658@qq.com>)

DavdGao (approved):
lgtm

PR #2311, opened 2026-08-14
fix(mcp): detect SSE transport by URL path so query strings work

### LINKED ISSUE #2309: [Bug]:   SSE MCP servers whose URL carries a query string (e.g. `?key=...`) are misdetected   as `streamable_http` transport, causing `McpError: Session terminated`
### Prerequisites

### Background / Description

In agentscope 2.0.6, MCPClient auto-detects the MCP transport (SSE vs streamable HTTP) purely from the URL suffix in
  agentscope/mcp/_mcp_client.py::_create_http_client:

  def _create_http_client(self):
      config = self.mcp_config
      if config.url.endswith("/sse") or config.url.endswith("/messages/"):
          return sse_client(url=config.url, headers=config.headers, timeout=config.timeout)
      # ... otherwise
      return streamable_http_client(url=config.url, http_client=http_client)

  HttpMCPConfig exposes no transport field, so detection relies entirely on this endswith check. The check ignores the query string. Real-world SSE
  servers that pass their API key in the query string — e.g. the official Amap (高德地图) MCP at https://mcp.amap.com/sse?key=<API_KEY> — have a URL
  that ends with /sse?key=..., not /sse. So:

  - endswith("/sse") → False
  - endswith("/messages/") → False

  … and the client wrongly falls through to streamable_http_client. A streamable-HTTP client then tries to speak streamable-HTTP to an SSE endpoint; the
  protocols don't match, the initialize handshake never completes, and the session is torn down. This is silent and easy to misdiagnose as a
  network/key problem, because any SSE URL without a query string (e.g. http://host:port/sse) connects fine.

  This affects any SSE MCP server whose URL includes a query string — a very common pattern for third-party/key-based SSE MCPs, not only Amap.

### Error Messages

```shell
Calling MCPClient.list_tools() directly raises:

  mcp.shared.exceptions.McpError: Session terminated

  Relevant traceback (mcp 1.26.0):

    File ".../mcp/client/session.py", line 171, in initialize
      result = await self.send_request(...)
    File ".../mcp/shared/session.py", line 306, in send_request
      raise McpError(response_or_error.error)
  mcp.shared.exceptions.McpError: Session terminated

  When used inside a Toolkit (the typical path), the exception is caught and logged as a warning, so the MCP is simply dropped from the tool group:

  WARNING | _toolkit:_get_available_tools:527 - Skipping MCP '<name>' in group 'basic':
  listing its tools failed with Session terminated
```

### Steps to Reproduce

  1. Install: pip install agentscope==2.0.6 mcp==1.26.0 httpx==0.28.1
  2. Run the following (any SSE MCP whose URL has a query string reproduces it; Amap is used as a public example):

  import asyncio
  from agentscope.mcp import MCPClient, HttpMCPConfig

  async def main():
      # SSE transport, but the API key sits in the query string
      client = MCPClient(
          name="amap",
          is_stateful=False,
          mcp_config=HttpMCPConfig(
              url="https://mcp.amap.com/sse?key=YOUR_AMAP_KEY",
              timeout=30.0,
          ),
      )
      tools = await client.list_tools()        # raises McpError: Session terminated
      print([t.name for t in tools])

  asyncio.run(main())

  3. Expected: tools are listed (the SSE endpoint is reachable and serves tools/list).
  Actual: mcp.shared.exceptions.McpError: Session terminated.
  4. Control experiment proving the root cause is the query string: the same SSE server reached with a query-less URL behaves correctly, while forcing
  streamable_http_client against it fails identically:

  # (A) FAILS — what agentscope 2.0.6 wrongly does (streamable against an SSE server)
  async with streamable_http_client(url="https://mcp.amap.com/sse?key=YOUR_AMAP_KEY") as s:
      async with ClientSession(s[0], s[1]) as sess:
          await sess.initialize()   # -> McpError: Session terminated

  # (B) WORKS — what agentscope should have selected
  async with sse_client(url="https://mcp.amap.com/sse?key=YOUR_AMAP_KEY", timeout=30.0) as (r, w):
      async with ClientSession(r, w) as sess:
          await sess.initialize()
          res = await sess.list_tools()   # -> 15 tools returned

### Environment

───────────┬──────────────────────────────────────────────────────────────┐
  │ Component  │                           Version                            │
  ├────────────┼──────────────────────────────────────────────────────────────┤
  │ agentscope │ 2.0.6 (latest release on PyPI as of report)                  │
  ├────────────┼──────────────────────────────────────────────────────────────┤
  │ mcp        │ 1.26.0                                                       │
  ├────────────┼──────────────────────────────────────────────────────────────┤
  │ httpx      │ 0.28.1                                                       │
  ├────────────┼──────────────────────────────────────────────────────────────┤
  │ Python     │ 3.12                                                         │
  ├────────────┼──────────────────────────────────────────────────────────────┤
  │ OS         │ macOS (platform-independent — pure protocol-detection logic) │
  └────────────┴──────────────────────────────────────────────────────────────┘

  Root Cause

  agentscope/mcp/_mcp_client.py, MCPClient._create_http_client (lines ~189–213 in 2.0.6): the transport is chosen by config.url.endswith("/sse"), which
  does not strip the query string, so SSE URLs of the form .../sse?key=... are misclassified as streamable HTTP.

1753936488:
## Problem

SSE MCP servers whose URL carries a query string fail to connect.
`MCPClient.list_tools()` / `get_tool()` raise:

```
mcp.shared.exceptions.McpError: Session terminated
```

and when the client is registered in a `Toolkit`:

```
WARNING - Skipping MCP '<name>' in group 'basic': listing its tools failed with Session terminated
```

This is a very common shape for key-based SSE MCP servers — e.g. the official Amap (高德地图) MCP at `https://mcp.amap.com/sse?key=API_KEY`.

## Root cause

`_create_http_client` detects the transport with `url.endswith('/sse')`. An SSE URL with a query string ends with `?key=...`, not `/sse`, so the check is `False` and the client wrongly falls through to `streamable_http_client`, which cannot handshake against an SSE server.

## Fix

Inspect only the URL **path** via `urlsplit` (which strips the query and fragment) instead of the full URL. Behaviour is unchanged for existing URLs without a query/fragment.

```python
path = urlsplit(config.url).path
if path.endswith("/sse") or path.endswith("/messages/"):
    return sse_client(...)
```

## Test

Added `test_stateless_client_with_query_string_in_url` in `tests/mcp_sse_client_test.py`. It connects to the existing local FastMCP SSE server through `http://127.0.0.1:<port>/sse?key=API_KEY` — fails with `Session terminated` before the fix, passes after.

The transport-selection logic was also verified directly for:

- `.../sse`, `.../sse?key=...`, `.../sse#frag`, `.../mcp_server/sse?k=v`, `.../messages/?s=1` -> `sse`
- `.../mcp`, `.../mcp?session_id=abc` -> `streamable_http`

### DIFF
--- a/src/agentscope/mcp/_mcp_client.py
+++ b/src/agentscope/mcp/_mcp_client.py
@@ -3,6 +3,7 @@
 import re
 from contextlib import AsyncExitStack, _AsyncGeneratorContextManager
 from typing import Any, TYPE_CHECKING
+from urllib.parse import urlsplit
 
 import httpx
 import mcp.types
@@ -192,8 +193,14 @@ class MCPClient(BaseModel):
         """Create an HTTP MCP client (SSE or streamable HTTP)."""
         config = self.mcp_config
 
-        # Determine transport from URL
-        if config.url.endswith("/sse") or config.url.endswith("/messages/"):
+        # Determine transport from the URL *path* only. Inspecting the full
+        # URL with endswith would misdetect SSE endpoints whose URL carries
+        # a query string (e.g. https://mcp.amap.com/sse?key=API_KEY): such
+        # URLs no longer end with '/sse' and would fall through to the
+        # streamable HTTP transport, which then fails to handshake against
+        # an SSE server with 'Session terminated'.
+        path = urlsplit(config.url).path
+        if path.endswith("/sse") or path.endswith("/messages/"):
             return sse_client(
                 url=config.url,
                 headers=config.headers,
--- a/tests/mcp_sse_client_test.py
+++ b/tests/mcp_sse_client_test.py
@@ -279,6 +279,36 @@ class SseMCPClientTest(IsolatedAsyncioTestCase):
         await stateful_client.close()
         self.assertFalse(stateful_client.is_connected)
 
+    async def test_stateless_client_with_query_string_in_url(self) -> None:
+        """An SSE URL carrying a query string must still be detected as SSE.
+
+        Regression test: ``_create_http_client`` used to detect the transport
+        with ``url.endswith('/sse')``, which fails for URLs that carry a query
+        string such as ``http://host:port/sse?key=API_KEY`` (a very common
+        shape for key-based SSE MCP servers, e.g. the Amap MCP at
+        https://mcp.amap.com/sse?key=API_KEY). Such URLs were misdetected as
+        streamable HTTP, and the client then failed to handshake against the
+        SSE server with ``McpError: Session terminated``.
+
+        After the fix, only the URL path is inspected, so the query string no
+        longer affects transport detection.
+        """
+        client = MCPClient(
+            name="test_sse_query_client",
+            is_stateful=False,
+            mcp_config=HttpMCPConfig(
+                type="http_mcp",
+                url=f"http://127.0.0.1:{self.port}/sse?key=API_KEY",
+            ),
+        )
+
+        mcp_tool_1 = await client.get_tool("tool_1")
+        res: ToolChunk = await mcp_tool_1(arg1="123", arg2=[1, 2, 3])
+        self.assertEqual(
+            res.content[0].text,
+            "arg1: 123, arg2: [1, 2, 3]",
+        )
+
 
 class SseSchemaDefsPreservationTest(IsolatedAsyncioTestCase):
     """End-to-end tests for $defs preservation in MCP tool schemas.


1753936488:
fix bug report #2309 https://github.com/agentscope-ai/agentscope/issues/2309

DavdGao (approved):
lgtm

PR #2318, opened 2026-08-14
fix(agent): keep prompt cache tokens in usage

### LINKED ISSUE #2305: [Bug]: Prompt cache tokens (cache_input_tokens, cache_creation_input_tokens) are dropped in ModelCallEndEvent and _save_to_context
## Background

AgentScope's `ChatUsage` dataclass correctly captures prompt cache tokens (`cache_input_tokens`, `cache_creation_input_tokens`) from provider APIs. The OpenAI-compatible model adapter (`_openai_chat/_model.py`) correctly parses `prompt_tokens_details.cached_tokens` from both DeepSeek and GLM (Zhipu AI) APIs.

However, these cache tokens are **silently dropped** in two places in `_agent.py`, making it impossible for middleware (e.g., `on_reply` hook via `ModelCallEndEvent`) to record prompt cache hit rates.

### 1. `ModelCallEndEvent` construction (`_agent.py` ~line 1410)

```python
yield ModelCallEndEvent(
    reply_id=self.state.reply_id,
    input_tokens=completed_response.usage.input_tokens
    if completed_response.usage
    else 0,
    output_tokens=completed_response.usage.output_tokens
    if completed_response.usage
    else 0,
    finished_reason=completed_response.finished_reason,
)
```

`completed_response.usage` is a `ChatUsage` that has `cache_input_tokens` and `cache_creation_input_tokens`, but `ModelCallEndEvent` only reads `input_tokens` and `output_tokens`. The cache fields are not passed to the event.

### 2. `_save_to_context` usage merging (`_agent.py` ~line 2912)

```python
if tail.usage is None:
    tail.usage = msg_usage
else:
    tail.usage.input_tokens += msg_usage.input_tokens
    tail.usage.output_tokens += msg_usage.output_tokens
```

When merging usage into the message context, only `input_tokens` and `output_tokens` are accumulated. `cache_input_tokens` and `cache_creation_input_tokens` are not merged.

### 3. `ModelCallEndEvent` class definition (`event/_event.py`)

The `ModelCallEndEvent` class itself has no `cache_input_tokens` or `cache_creation_input_tokens` fields:

```python
class ModelCallEndEvent(EventBase):
    type: Literal[EventType.MODEL_CALL_END] = EventType.MODEL_CALL_END
    reply_id: str
    input_tokens: int
    output_tokens: int
    finished_reason: FinishedReason = Field(default=FinishedReason.COMPLETED)
```

## Impact

- Token usage middleware using `on_reply` hook cannot record cache hit rates
- `on_model_call` hook receives an `async_generator` for streaming models, so it cannot drain the generator to read the final `ChatResponse.usage` without breaking streaming
- Applications relying on AgentScope's usage numbers cannot see actual cache utilization, making it impossible to assess cost savings or optimize for cache efficiency

## Verification

Direct API calls to GLM-5.2 confirm prompt cache is working server-side:

```
Call 1: input=6016  cached=0     ← first call, no cache
Call 2: input=6016  cached=6016  ← 100% cache hit (same prefix)
Call 3: input=6023  cached=5992  ← 99.5% cache hit (growing conversation)
```

But `TokenUsageMiddleware` (using `on_reply` hook) always records `cache_read_tokens=0` because the cache data is dropped before the event reaches the middleware.

## Proposed fix

1. Add `cache_input_tokens: int = 0` and `cache_creation_input_tokens: int = 0` fields to `ModelCallEndEvent`
2. Pass these fields when constructing the event in `_agent.py`
3. Merge cache fields in `_save_to_context`

## Environment

- AgentScope 2.0.5 / 2.0.6 (checked latest `main` branch — issue persists)
- Models: GLM-5.2 (Zhipu AI), DeepSeek, OpenAI-compatible providers
- Python 3.14

yingz23:
## AgentScope Version

2.0.6 (`import agentscope; print(agentscope.__version__)`)

## Description

Fixes #2305. `cached_tokens` is already parsed into `ChatUsage.cache_input_tokens`
by the model adapters (DashScope, DeepSeek, OpenAI-compatible, etc.), but it was
dropped before reaching `ModelCallEndEvent` and the message `'usage'`, so
middleware couldn't see prompt-cache hits.

Changes:
- Add `cache_input_tokens` / `cache_creation_input_tokens` to `ModelCallEndEvent` and `Usage`
- Pass them when building the event; construct/accumulate them in `_save_to_context`
- Add regression test `test_usage_cache_tokens_are_not_dropped`

Both new fields default to `0`, so it's backward compatible.
Test: `tests/agent_basic_test.py::AgentBasicTest::test_usage_cache_tokens_are_not_dropped`.

Happy to open a matching docs PR if you think the new fields should be documented — see https://docs.agentscope.io/versions/2.0.7dev/en/building-blocks/message-and-event#reconstruct-messages-from-events

## Checklist

- [x]  I have read the `https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md`
- [ ]  Related documentation has been updated (e.g. links, examples, etc.) in `https://github.com/agentscope-ai/docs`


iamroylim (changes requested):
Blocking event-replay gap at head 8dd62e9b475b8f938abb685e811e9872a5740834:

Msg.append_event(ModelCallEndEvent(...)) still constructs Usage from only input_tokens and output_tokens. A non-zero cache-token event therefore replays as zeros.

Minimal reproducer: create AssistantMsg(id="reply-1"), append a ModelCallEndEvent with input_tokens=100, output_tokens=20, cache_input_tokens=60, and cache_creation_input_tokens=40, then inspect msg.usage. Actual: cache_input_tokens=0 and cache_creation_input_tokens=0.

The new event_to_message_test.py expectations use only default-zero cache values, so the regression is not exercised. Please pass and accumulate both cache fields in Msg.append_event, and make at least one replay test use non-zero values across two MODEL_CALL_END events.

Verification: uv run pytest tests/agent_basic_test.py::AgentBasicTest::test_usage_cache_tokens_are_not_dropped tests/event_to_message_test.py -q passes (4 tests), while the non-zero replay reproducer above fails.

yingz23:
> Blocking event-replay gap at head [8dd62e9](https://github.com/agentscope-ai/agentscope/commit/8dd62e9b475b8f938abb685e811e9872a5740834):
> 
> Msg.append_event(ModelCallEndEvent(...)) still constructs Usage from only input_tokens and output_tokens. A non-zero cache-token event therefore replays as zeros.
> 
> Minimal reproducer: create AssistantMsg(id="reply-1"), append a ModelCallEndEvent with input_tokens=100, output_tokens=20, cache_input_tokens=60, and cache_creation_input_tokens=40, then inspect msg.usage. Actual: cache_input_tokens=0 and cache_creation_input_tokens=0.
> 
> The new event_to_message_test.py expectations use only default-zero cache values, so the regression is not exercised. Please pass and accumulate both cache fields in Msg.append_event, and make at least one replay test use non-zero values across two MODEL_CALL_END events.
> 
> Verification: uv run pytest tests/agent_basic_test.py::AgentBasicTest::test_usage_cache_tokens_are_not_dropped tests/event_to_message_test.py -q passes (4 tests), while the non-zero replay reproducer above fails.

Good catch — thanks! Fixed `Msg.append_event` to pass and accumulate both cache fields, and updated `event_to_message_test.py` to exercise a non-zero replay across two `MODEL_CALL_END` events.

Verified with `pytest tests/agent_basic_test.py::AgentBasicTest::test_usage_cache_tokens_are_not_dropped tests/event_to_message_test.py -q` → 4 passed.

iamroylim (approved):
Verified the event-replay fix at b01d70f310dcc70e41d6525008c17180dd1faef3. `Msg.append_event` now preserves and accumulates both non-zero cache token fields across MODEL_CALL_END events, and the regression test covers the exact two-event replay path that previously failed. Local verification: 4 focused tests passed; changed-file pre-commit passed.

iamroylim:
The current head now fails the refreshed full test matrix because the event serialization contract grew, but the HITL expected-event fixtures did not.

Reproduction on b01d70f310dcc70e41d6525008c17180dd1faef3 (macOS, Python 3.11):

    uv run --python 3.11 --extra dev pytest -q tests/hitl_external_execution_test.py::AgentExternalExecutionTest::test_single_external_execution

Result: 1 failed. MODEL_CALL_END now serializes cache_input_tokens=0 and cache_creation_input_tokens=0, while the expected dictionaries in hitl_external_execution_test.py (and the analogous HITL mixed/user-confirmation fixtures) still omit both keys. The remote macOS run reports the same mechanism across 11 tests; the cache-token regression test itself passes.

Please update those expected MODEL_CALL_END payloads (ideally through their shared fixture/helper if there is one) to include both default-zero fields. This looks like a fixture-contract update, not a rollback of the cache-token fields.

src/agentscope/agent/_agent.py
```diff
@@ -1453,14 +1412,15 @@ async def _reasoning_impl(
...
             reply_id=self.state.reply_id,
-            input_tokens=completed_response.usage.input_tokens
-            if completed_response.usage
-            else 0,
-            output_tokens=completed_response.usage.output_tokens
-            if completed_response.usage
-            else 0,
+            input_tokens=usage.input_tokens if usage else 0,
+            output_tokens=usage.output_tokens if usage else 0,
+            cache_input_tokens=usage.cache_input_tokens if usage else 0,
```
qbc2016: **Blocking:** Please normalize the optional cache-token values before passing them into Pydantic models.

`ChatUsage` is a dataclass and therefore does not enforce its `int` annotations at runtime. Several adapters copy optional SDK fields with `getattr(..., 0)`, which can still return `None` when the attribute exists but is null. For example, OpenAI defines `PromptTokensDetails.cached_tokens` as `Optional[int]`.

Consequently, a response with `usage.cache_input_tokens is None` currently causes `ModelCallEndEvent` to raise a `ValidationError` and abort an otherwise successful reply.

Please handle both levels of absence explicitly:

```python
cache_input_tokens=(
    (usage.cache_input_tokens or 0)
    if usage is not None
    else 0
),
cache_creation_input_tokens=(
    (usage.cache_creation_input_tokens or 0)
    if usage is not None
    else 0
),
```
The same normalization is also required when constructing Usage at lines 3010–3011. Please add a regression test where both cache fields are None.

src/agentscope/agent/_agent.py
```diff
@@ -3047,6 +3007,8 @@ def _save_to_context(
             Usage(
                 input_tokens=usage.input_tokens,
                 output_tokens=usage.output_tokens,
+                cache_input_tokens=usage.cache_input_tokens,
```
qbc2016: **Blocking:** The cache-token values also need to be normalized here.

Even if the `ModelCallEndEvent` construction is fixed, `ChatUsage.cache_input_tokens` and `cache_creation_input_tokens` can still be `None`. Passing either value into the Pydantic `Usage` model will raise a `ValidationError` in `_save_to_context`.

Please coerce both values to zero:

```python
cache_input_tokens=usage.cache_input_tokens or 0,
cache_creation_input_tokens=(
    usage.cache_creation_input_tokens or 0
),
```

Please also cover this path with a `ChatUsage` containing explicit `None` cache fields.

src/agentscope/agent/_agent.py
qbc2016, line 1633: **Blocking:** The cache-token values are still dropped from the final message returned by `Agent.reply()`.

`_save_to_context` now correctly stores the cache counts in `last_ctx.usage`, but `final_usage` reconstructs `Usage` using only `input_tokens` and `output_tokens`. Because the new cache fields default to zero, the returned message silently reports incorrect values.

A minimal reproduction currently produces:

```python
reply.usage.model_dump()
# cache_input_tokens=0, cache_creation_input_tokens=0

agent.state.context[-1].usage.model_dump()
# cache_input_tokens=60, cache_creation_input_tokens=40
```

Please copy both fields when constructing `final_usage`:

```python
cache_input_tokens=last_ctx.usage.cache_input_tokens,
cache_creation_input_tokens=(
    last_ctx.usage.cache_creation_input_tokens
),
```

Please also extend the regression test to assert the complete `usage` structure returned by `await agent.reply(...)`, rather than checking only the event stream and saved context.

qbc2016 (commented):
Please see the inline comments. Please also merge the latest `main` branch and fix the unit tests.

src/agentscope/agent/_agent.py
yingz23: fixed:
```
            cache_input_tokens=(usage.cache_input_tokens or 0) if usage else 0,
            cache_creation_input_tokens=(
                (usage.cache_creation_input_tokens or 0) if usage else 0
            ),
```

src/agentscope/agent/_agent.py
yingz23: fixed

```
                cache_input_tokens=usage.cache_input_tokens or 0,
                cache_creation_input_tokens=(
                    usage.cache_creation_input_tokens or 0
                ),
```

src/agentscope/agent/_agent.py
yingz23: fixed:
```
                    cache_input_tokens=(
                        last_ctx.usage.cache_input_tokens or 0
                    ),
                    cache_creation_input_tokens=(
                        last_ctx.usage.cache_creation_input_tokens or 0
                    ),
```

yingz23:
Thanks a lot for all the detailed reviews!

At first I thought those 11 HITL test failures were pre-existing environment issues. Then I realized they came from the new serialization contract — fixed all HITL fixtures to add the two zero-default cache fields.

Also normalized both cache fields at `ModelCallEndEvent` and `Usage` construction with `(x or 0)`, added a regression test that explicitly sets both `ChatUsage` cache fields to `None`.

Full test matrix (minus the container/e2e tests that need extra local setup) all passes now:

```
PYTHONDONTWRITEBYTECODE=1 python -m pytest tests/ --ignore-glob='tests/backend_*.py' --ignore-glob='tests/workspace_*.py' --ignore='tests/test_e2e_api.py' --ignore='tests/test_e2e_docker_mcp.py' --ignore='tests/test_e2e_e2b_mcp.py' -q 2>&1 | tail -6

...
1632 passed, 3 skipped, 223 warnings, 375 subtests passed in 143.32s (0:02:23)
```

tests/agent_basic_test.py
```diff
@@ -659,6 +663,153 @@ async def test_non_streaming_reasoning(self) -> None:
         context_dicts = [msg.model_dump() for msg in self.agent.state.context]
         self.assertListEqual(context_dicts, expected_context_after_reply)
 
+    async def test_usage_cache_tokens_are_not_dropped(self) -> None:
```
qbc2016, line 666: Please follow the existing whole-structure assertion pattern used throughout this test file.

Instead of asserting individual fields, serialize the complete object with `model_dump()` and keep dynamic fields such as `id` and `created_at` in the expected structure using `AnyString()`. The existing `_get_event_base(reply_id)` and `_get_msg_base()` helpers already provide this pattern.

For example, the event assertion can be written as:

```python
self.assertEqual(
    end_event.model_dump(),
    {
        **self._get_event_base(self.agent.state.reply_id),
        "type": "MODEL_CALL_END",
        "input_tokens": 100,
        "output_tokens": 20,
        "cache_input_tokens": 60,
        "cache_creation_input_tokens": 40,
        "finished_reason": "completed",
    },
)
```

The usage assertion can similarly compare the complete serialized structure:

```python
self.assertEqual(
    usage.model_dump(),
    {
        "input_tokens": 100,
        "output_tokens": 20,
        "cache_input_tokens": 60,
        "cache_creation_input_tokens": 40,
    },
)
```

Please apply the same approach to the `None` normalization and `Agent.reply()` return-value tests. For complete message assertions, reuse `_get_msg_base()` and represent dynamic content-block IDs and timestamps with `AnyString()`.

qbc2016 (commented):

tests/agent_basic_test.py
yingz23: Refactored the tests to use the whole-structure assertion pattern with _get_event_base/_get_msg_base and AnyString.

Appreciate the careful review!

qbc2016 (approved):
LGTM.

yingz23:
> Happy to open a matching docs PR if you think the new fields should be documented — see https://docs.agentscope.io/versions/2.0.7dev/en/building-blocks/message-and-event#reconstruct-messages-from-events

Happy to open a matching docs PR if you think the new fields should be documented — see https://docs.agentscope.io/versions/2.0.7dev/en/building-blocks/message-and-event#reconstruct-messages-from-events

PR #2323, opened 2026-08-14
fix(model): propagate max-token termination reasons

### LINKED ISSUE #2321: [Bug]: Agent输出中断的问题
### Prerequisites

### Background / Description

agent对话超过一定的轮次，或者输出超过一定的长度，Agent回复会突然中止，没有任何的提示

### Error Messages

```shell
无
```

### Steps to Reproduce

无

### Environment

- AgentScope Version:2.05
- Python Version:3.11
- OS:Windows

RICARDO51167:
## AgentScope Version

2.0.6

### Background

When an upstream model stops because it reaches its output token limit, AgentScope currently loses the provider’s termination reason and reports the response as `completed`.

This makes truncated responses indistinguishable from normally completed responses. In a ReAct loop, it may also allow another reasoning iteration or attempt to execute an incomplete tool call.

This is separate from the O(n²) streaming accumulation issue already addressed by #2158.

### Changes

- Added `MAX_TOKENS = "max_tokens"` to model-level and reply-level finished reasons.
- Normalized explicit token-limit signals from supported providers:
  - OpenAI Chat, DeepSeek, Moonshot, DashScope and Ollama
  - OpenAI Responses
  - Anthropic
  - Gemini
  - xAI
- Preserved the original provider finish reason in `metadata["provider_finished_reason"]`.
- Ensured streaming adapters emit a carrier chunk when the final provider frame contains only a finish reason or usage.
- Preserved the finished reason, usage and metadata during stream accumulation.
- Updated the ReAct loop to stop immediately after a max-token termination:
  - Keep partial text, thinking content and usage.
  - Do not start another reasoning iteration.
  - Do not execute truncated tool calls.
  - Close incomplete tool calls with a synthetic error result so tool calls and results remain paired in the context.
  - Emit a system hint explaining that the output token limit was reached.
  - Propagate `max_tokens` through `ModelCallEndEvent`, `ReplyEndEvent` and the final `AssistantMsg`.
- Added an always-visible console warning for token-limit truncation.
- Kept existing behavior unchanged for normal provider finish reasons such as `stop`, `end_turn`, `tool_calls` and `tool_use`.

Existing consumers do not need to migrate unless they exhaustively match finished-reason enum values. Such consumers should handle the new `max_tokens` wire value.

Fixes #2321

### How to test

- Run all static checks:

  `pre-commit run --all-files`

- Run the model, Agent, event and console tests covering this change.

Test results:

- `pre-commit run --all-files`: passed
- Relevant model, Agent, event and console tests: 217 passed
- Remaining non-environment-dependent test suite: 1753 passed, 133 skipped
- Local-service tests requiring unrestricted socket binding could not complete in the restricted test environment; the affected S3, MCP, Milvus and Bubblewrap tests do not overlap with the changed code paths.

Coverage added for:

- Streaming and non-streaming token-limit mapping across all affected model adapters.
- Provider finish-reason metadata and unchanged normal completion behavior.
- Finish-only streaming carrier frames, usage and metadata accumulation.
- Partial text preservation after truncation.
- Safe termination of truncated tool calls without executing the tool.
- Event-to-message conversion and console warning output.

## Checklist

- [x] I have read the [[CONTRIBUTING.md](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md)](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md)
- [x] Related documentation has been updated — inline API documentation was updated; no separate documentation PR is needed because this change introduces no new configuration or usage workflow

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -792,8 +792,24 @@ class Agent:
         self,
         finished_reason: FinishedReason,
         metadata: dict[str, Any],
+        model_response_start: int,
     ) -> tuple[list[AgentEvent], ReplyEndEvent | None, Msg | None]:
-        """Build the reply outcome for a terminal model response."""
+        """Build the reply outcome for a terminal model response.
+
+        Args:
+            finished_reason (`FinishedReason`):
+                The normalized model termination reason.
+            metadata (`dict[str, Any]`):
+                Metadata propagated from the model response.
+            model_response_start (`int`):
+                The index where the current model response begins in the
+                accumulated reply context.
+
+        Returns:
+            `tuple[list[AgentEvent], ReplyEndEvent | None, Msg | None]`:
+                Synthetic terminal events, the reply-end event, and the
+                terminal message when one is required.
+        """
         if finished_reason == FinishedReason.COMPLETED:
             return [], None, None
 
@@ -815,7 +831,7 @@ class Agent:
                 id=self.state.reply_id,
                 name=self.name,
                 content=(
-                    deepcopy(last_ctx.content)
+                    deepcopy(last_ctx.content[model_response_start:])
                     if last_ctx is not None
                     else _MAX_TOKENS_HINT
                 ),
@@ -994,6 +1010,15 @@ class Agent:
                         async for evt in self._inject_runtime_state():
                             yield evt
 
+                        last_ctx_before_reasoning = self._get_last_msg()
+                        model_response_start = (
+                            len(last_ctx_before_reasoning.content)
+                            if last_ctx_before_reasoning is not None
+                            and last_ctx_before_reasoning.id
+                            == self.state.reply_id
+                            else 0
+                        )
+
                         # Perform reasoning
                         model_finished_reason = FinishedReason.COMPLETED
                         model_finished_metadata: dict[str, Any] = {}
@@ -1023,6 +1048,7 @@ class Agent:
                         ) = await self._handle_model_termination(
                             model_finished_reason,
                             model_finished_metadata,
+                            model_response_start,
                         )
                         for termination_event in termination_events:
                             yield termination_event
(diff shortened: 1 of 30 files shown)


src/agentscope/agent/_agent.py
```diff
@@ -754,23 +758,72 @@ async def _close_unfinished_tool_calls(
...
+                    result_state=ToolResultState.ERROR,
+                )
+            ]
+
+            last_ctx = self._get_last_msg()
+            terminal_msg = AssistantMsg(
+                id=self.state.reply_id,
+                name=self.name,
+                content=(
+                    deepcopy(last_ctx.content)
```
iluv7: Could we avoid copying `last_ctx.content` here? It contains the full ReAct trace accumulated during the current reply, so a `MAX_TOKENS` response may unexpectedly include previous thinking, tool calls, and tool results, while normal completion returns only `completed_response.content`.

Since the loop is already terminated and earlier blocks are available through SSE and `state.context`, it may be safer to return only the current partial response plus synthetic results for its unfinished tool calls.

iluv7 (commented):
👀

src/agentscope/agent/_agent.py
RICARDO51167: Thanks for catching this. Fixed in `85ee2a8d`.

The `MAX_TOKENS` path now records the start of the current model response before reasoning and only returns content added from that point onward. The final `AssistantMsg` therefore contains only:

- the current partial response;
- the max-token hint;
- synthetic error results for unfinished tool calls.

Earlier ReAct thinking, completed tool calls, and tool results remain available in `state.context` and the event stream, but are no longer duplicated into the final message.

I also added a multi-round regression test covering a successful tool round followed by a token-truncated response.

Validation results:

- `pytest tests/agent_basic_test.py`: 12 passed
- relevant model/agent/event/console tests: 218 passed, 6 subtests passed
- `pre-commit run --all-files`: passed

PR #2325, opened 2026-08-15
fix(app): interrupt assembling chat run via local registry check (#2320)

### LINKED ISSUE #2320: [Bug]: interrupt during _run_impl assembly phase cancels nothing — the run completes and a full "abandoned" reply is persisted
### Prerequisites

### What I was trying to do

Call `ChatService.interrupt(user_id, session_id, agent_id)` to stop an
in-progress chat run **while the run is still in its assembly phase** — i.e. the
window between `chat_service.run(...)` spawning `_run_impl` and the run acquiring
the session lock. This window is noticeable on the **first message of a session**,
where `workspace_manager.get_workspace(...)` / `get_tts_model` / middleware / KB /
toolkit / model assembly (steps 1–6 of `_run_impl`) can take on the order of
seconds when materializing a cold workspace.

### Expected behavior

The run should be cancelled, exactly as interrupting during the locked generation
phase (step 7) does: the agent's `CancelledError` cleanup runs, the client receives
a `ReplyEndEvent(INTERRUPTED)` terminal event, and **no full reply is persisted**.
During the window the 409 guard (run already in `ChatRunRegistry`) should keep
working so a re-send is correctly rejected until the cancelled run releases its slot.

### What actually happens

`interrupt` returns normally (no exception), but the run is **not** stopped. It
proceeds to step 7, generates the complete reply, and the `finally` block's
**shielded** persistence (`_chat.py:963-985`) writes the **full reply** to storage.
The transcript gains an assistant message the user never wanted (a "ghost/abandoned"
reply). The user's interrupt intent, instead of cancelling the run, gets enqueued as
a `UserInterruptEvent` resume trigger that only runs *after* the abandoned run
finishes and releases the lock — by which time the session is idle and the trigger
no-ops.

### Root cause

`ChatService.interrupt` (`src/agentscope/app/_service/_chat.py:388`) uses
`message_bus.is_locked(session_lock)` as the **sole** "is a run active" judge
(lines 429-432):

```python
# _chat.py:429-445
if await self._message_bus.is_locked(
    MessageBusKeys.session_lock(session_id),
):
    await self._message_bus.publish(               # cancel path
        MessageBusKeys.session_interrupt_channel(),
        {"session_id": session_id},
    )
    return

await enqueue_run_trigger(                         # resume-trigger path
    self._message_bus,
    user_id=user_id, session_id=session_id, agent_id=agent_id,
    kind=MessageBusKeys.WAKEUP_KIND_RESUME,
    inputs=UserInterruptEvent(reply_id=session.state.reply_id),
)
```

But `_run_impl` (`_chat.py:503`) has an assembly phase (steps 1-6, lines 524-780)
that runs **before** `acquire_lock` (step 7, line 798):

```python
# _chat.py:524-   (assembly, steps 1-6)
try:
    agent_record = await self._access.resolve_agent(...)   # 536
    session_record = await self._storage.get_session(...)  # 551
    workspace = await self._workspace_manager.get_workspace(...)  # 556 — slow on cold build
    ...middlewares / TTS / KB / toolkit / model / assemble...
except Exception as e:                                     # 781 — catches only Exception
    async with self._message_bus.acquire_lock(...):        # 786 — only the failure path
        await self._report_failure(...)
    return
# step 7:
async with self._message_bus.acquire_lock(                 # 798 — lock acquired here
    lock_key, ttl_secs=MessageBusKeys.SESSION_RUN_TTL_SECS,
):
    ... actual generation ...
```

During assembly `is_locked` is `False`, so `interrupt` takes the **resume-trigger**
branch and never publishes on `session_interrupt_channel`. The assembly task is
therefore never cancelled.

### Key point: the cancel path *can* cancel an in-assembly task — the gap is branch selection, not capability

`CancelDispatcher._interrupt_session`
(`src/agentscope/app/_manager/_cancel_dispatcher.py:227-246`) keys off the
**registry** (the in-process asyncio task handle), **not** `is_locked`:

```python
# _cancel_dispatcher.py:227-246
def _interrupt_session(self, session_id: str) -> None:
    task = self._registry.get(session_id)
    if task is not None and not task.done():
        logger.info("CancelDispatcher: interrupting local chat run for session %s", session_id)
        task.cancel()
```

The assembly task is already in the registry (it occupies the 409 slot), so
publishing on the interrupt channel *would* cancel it. And an assembly-phase
`CancelledError` **does not persist** — `reply_msg` is not yet bound, and
`CancelledError` is a `BaseException` not caught by the `except Exception` at line
781, so it bubbles out with no `finally` upsert (`reply_msg is None`). The
capability is present; `interrupt` simply chooses not to invoke it when
`is_locked` is `False`.

### Why the full reply survives

```python
# _chat.py:963-985 (finally, inside the step-7 lock)
async def _persist() -> None:
    if reply_msg is not None:
        await self._storage.upsert_message(user_id, session_id, reply_msg)  # full reply
    await self._storage.update_session_state(...)
    await self._message_bus.log_trim(events_key)
persist_task = asyncio.create_task(_persist())
try:
    await asyncio.shield(persist_task)        # shielded — survives outer cancel
except asyncio.CancelledError:
    await persist_task                        # await persistence, then propagate
```

Even a cancellation that arrives in step 7 is shielded so the full `reply_msg`
lands before the lock is released. For the assembly-phase case the run isn't
cancelled at all, so the complete reply is persisted as a matter of course.

### Suggested fix directions (least → most invasive)

1. **Judge by "active run", not only `is_locked`** (recommended, targeted). Before
   falling to the resume-trigger branch, check whether a run is active: in-process,
   `ChatRunRegistry.get(session_id)` (task not done) → treat as active and publish on
   `session_interrupt_channel` (same as the `is_locked=True` branch). For multi-replica
   deployments the registry is per-process, so a distributed "run alive" marker is
   needed (set at spawn, **before** `acquire_lock`; cleared at task done) and the judge
   becomes `is_locked OR run_alive`. This realizes the intent of "lock acquired at
   spawn" via a lightweight marker instead of actually holding the lock early. The
   assembly task is already in the registry, so the cancel reaches it, and an
   assembly-phase cancel persists nothing (see above).
2. **Synthesize an INTERRUPTED reply on assembly-phase cancel** (complements #1).
   Add `except asyncio.CancelledError:` to the assembly try block (parallel to the
   `except Exception` at line 781; `CancelledError` is not caught by `Exception`).
   Mirror `_report_failure` (`_chat.py:315-386`) but emit
   `finished_reason=ReplyFinishedReason.INTERRUPTED` (no `error` field) with
   `AssistantMsg(content=[])`, publish + upsert best-effort, then `raise` to honour
   asyncio semantics. Gives the client a clean terminal event and an empty (not full)
   assistant message — the same shape as an early mid-reply cancel. `_report_failure`
   should not be parameterized for this (it carries `ERROR` + `error` semantics); use a
   separate sibling. Only meaningful once #1 lets the cancel reach the assembly task.
3. **Move `acquire_lock` to `_run_impl` entry** (thorough, most invasive). Acquire the
   lock before the assembly try so `is_locked` is `True` throughout assembly and the
   existing judge works unchanged. Caveats: the lock is non-reentrant (Redis `SET NX` /
   in-memory `asyncio.Lock`), so the nested `acquire_lock` at lines 786-790 (the
   assembly failure path) must be removed (call `_report_failure` directly — lock already
   held); confirm no assembly step takes `session_lock` (lines 786-790 are the only
   re-entry point). Lock is held longer (cold `get_workspace` build under lock); rely on
   heartbeat (TTL/2) renewal, and let the build emit heartbeats or raise
   `SESSION_RUN_TTL_SECS` if cold builds can exceed it. Direction #2 is still needed for
   the terminal event.

### Boundaries / notes

- The **409 busy guard** ("already has an active chat run") is normal concurrency
  protection (in-process registry + Redis distributed lock, two layers) and is **not**
  part of this bug; it continues to work during the affected window.
- **`404 ⇔ no run`**: when the session does not exist, `trigger_chat_run` itself 404s
  before spawning, so a 404 cannot coexist with in-flight generation. (A 404 observed
  here belongs to a different path — session ownership / id resolution — and is unrelated
  to the persisted-reply issue.)
- A **downstream platform cannot cleanly fix this**: if it merely "publishes cancel after
  an `is_locked` timeout", it **loses the user message** — the Case A user-input upsert
  (`_chat.py:823-834`) happens inside the step-7 lock, *after* assembly, so cancelling
  the assembly task never reaches it. The root-cause fix belongs upstream (interrupt
  judge / lock timing).

### Error Messages

```shell
**None.** This is a silent logic/race bug: `interrupt()` returns normally, no exception
is raised and nothing is logged as an error. The run simply isn't cancelled, runs to
completion, and the full reply is persisted by the shielded `finally`. The only
observable artifact is an unexpected assistant message in the session transcript and
the absence of a `REPLY_END(INTERRUPTED)` terminal event on the event stream.
```

### Steps to Reproduce

The trigger is timing-based: `interrupt` must land during the assembly phase (before
`acquire_lock` at `_chat.py:798`). To make it deterministic, slow down
`workspace_manager.get_workspace` so the assembly phase lasts long enough to interrupt
into. Sketch (adapt the fakes to your test harness; the essential point is a slow
`get_workspace` so the interrupt lands before the lock appears):

1. Code:

```python
import asyncio
import agentscope  # 2.0.6
from agentscope.app.message_bus import InMemoryMessageBus
from agentscope.app.message_bus._keys import MessageBusKeys
# ...your ChatService construction: storage / workspace_manager /
#    resource_access / ChatRunRegistry / WakeupDispatcher / CancelDispatcher...

class SlowWorkspaceManager:
    """Fake that makes the assembly phase deterministic-slow."""
    async def get_workspace(self, user_id, agent_id, session_id, *a, **kw):
        await asyncio.sleep(5)          # hold the run in assembly (pre-lock)
        return await self._real.get_workspace(user_id, agent_id, session_id, *a, **kw)

# --- wire SlowWorkspaceManager into ChatService ---

# 1. spawn the run (first user message → cold/slow assembly)
run_task = asyncio.create_task(
    chat_service.run(user_id, agent_id, session_id, user_msg)
)

# 2. while the run is still assembling (is_locked is False), interrupt
await asyncio.sleep(1)                  # ensure we're inside the 5s assembly window
assert not await message_bus.is_locked(MessageBusKeys.session_lock(session_id))
await chat_service.interrupt(user_id, session_id, agent_id)   # returns normally

# 3. let the un-cancelled run finish
await run_task

# 4. observe: an unwanted FULL assistant reply was persisted
messages = await storage.list_messages(user_id, agent_id, session_id)
assistant_replies = [m for m in messages if m.name == agent_id]
assert assistant_replies and assistant_replies[-1].content   # ← BUG: full abandoned reply

# and the event log has NO REPLY_END(INTERRUPTED) — the interrupt never took effect
```

2. Run: `python repro_assembly_interrupt.py`
3. See: `interrupt()` returns, the run completes, a full assistant message is
   persisted, no `REPLY_END(INTERRUPTED)` is emitted.

**Control / expected (interrupt during step-7 locked generation)**: if you move the
`interrupt` call to *after* `is_locked` becomes `True` (i.e. past `acquire_lock` at
line 798), the cancel path fires, the agent's `CancelledError` cleanup runs, an
`INTERRUPTED` terminal event is emitted, and the full reply is not persisted — this is
the behaviour the assembly-phase case should match.

### Environment

- AgentScope Version: 2.0.6
- Python Version: 3.11.15
- OS: Linux 6.2.0-39-generic x86_64

YuhaoLin2005:
## Why

Fixes #2320. During the **assembly phase** of `ChatService._run_impl` (steps 1–6: workspace materialisation, TTS, middleware, KB, toolkit, model) — which on a session's **first message** can take seconds on a cold workspace — the distributed session lock is **not yet acquired** (step 7). `ChatService.interrupt` uses `message_bus.is_locked(session_lock)` as its **sole** "is a run active" judge, so an interrupt landing in this window takes the **resume-trigger** branch instead of the **cancel** branch:

- the run is never cancelled → it completes and the shielded `finally` persists the **full reply** → a "ghost/abandoned" assistant message the user never wanted;
- the `UserInterruptEvent` resume trigger only fires *after* the run finishes and releases the lock, by which time the session is idle and the trigger no-ops.

The cancel machinery is already in place: `CancelDispatcher._interrupt_session` looks the run up in `ChatRunRegistry` (the in-process asyncio task handle), and an assembly-phase `CancelledError` is a `BaseException` that neither `_run_impl`'s `except Exception` nor `run`'s `except Exception` swallows — so it bubbles out with `reply_msg` unbound and **nothing persisted**. The gap is purely **branch selection**.

## Scope of this PR (read this first)

**This PR fully fixes the assembly-window interrupt for single-process deployments** (embedded / one-worker mode, where the interrupt and the run share a process). It does **not** yet fix the same window in **multi-replica** deployments: `ChatRunRegistry` is per-process, so if the run is assembling on worker A and the interrupt lands on worker B, B sees `is_locked == False` *and* an empty local registry and still falls through to the resume-trigger branch — the ghost reply persists, unchanged. Closing that needs a distributed run-alive marker (set at spawn, before `acquire_lock`), which is a follow-up. The issue itself acknowledges this.

## Fix (targeted — the issue's recommended direction #1)

`ChatService.interrupt` now treats a run as active when **either** the distributed lock is held **or** a non-finished chat-run task is live in this process's `ChatRunRegistry`:

```python
if await self._message_bus.is_locked(
    MessageBusKeys.session_lock(session_id),
) or self._has_live_local_run(session_id):
    ...  # publish on session_interrupt_channel (cancel path)
```

`_has_live_local_run` is a small helper reading the already-injected per-process registry (no new state, no new dependency):

```python
if self._chat_run_registry is None:
    return False
task = self._chat_run_registry.get(session_id)
return task is not None and not task.done()
```

The registry is **injected via the constructor** (`ChatRunRegistry | None = None`) and wired at the single production construction site in `_lifespan.py`. `None` preserves the previous behaviour for test / alternate wiring — zero API break. (Note: this means a future construction site that forgets to wire the registry silently keeps the pre-fix behaviour; the `interrupt` docstring flags this.)

The assembly task is already registered at spawn (it occupies the 409 "busy" slot), so publishing on the interrupt channel reaches it immediately, and an in-assembly cancel persists nothing.

## Behaviour change in the fixed window (honest disclosure)

For the assembly window this PR *changes* what the user sees — it does not merely make the old path work:

- **Before**: the run completes → the user's input `Msg` **and** a full ghost assistant reply are both persisted.
- **After** (single-process): the run is cancelled before any reply exists → **nothing is persisted**, and no `ReplyStartEvent` / `ReplyEndEvent` is emitted. Because the input `Msg` is only persisted after assembly (inside the step-7 lock), **the user's message is dropped from the transcript** rather than answered-with-a-ghost.

Synthesising an `INTERRUPTED` terminal event (and persisting an empty assistant message) for assembly-phase cancels is the issue's **direction #2** and is left as a follow-up — direction #1 is its prerequisite. This trade-off (ghost reply → silent cancel, no terminal event) is exactly what the issue's direction #1 implies; we flag it rather than bury it.

## Tests

`tests/service_interrupt_dispatch_test.py` — 5 tests:

- **assembly window** (no lock + live local task) → cancel path (the regression; fails without the fix);
- **lock held** (mid-reply / HITL-parked) → cancel path (unchanged);
- **idle** (no lock + empty registry) → resume trigger (unchanged);
- **registry not wired** (`None`) → resume trigger (legacy fallback preserved);
- **end-to-end**: real `InMemoryMessageBus` + real `CancelDispatcher` + real registry — `ChatService.interrupt` on a task blocked *before* acquiring any lock, asserting the task is actually cancelled. Guards the fix's premise ("the cancel machinery already works for an assembly task") rather than asserting it.

Locally verified: the new file plus the existing `service_agent_interrupt_test.py`, `service_cancel_dispatcher_test.py`, `service_wakeup_dispatcher_test.py` and `service_chat_middleware_factory_test.py` all pass (31 tests, 0 failures).

## Impact on existing behaviour

The lock-held path, the resume-trigger path, and the `None`-registry fallback are byte-for-byte unchanged. The only production behaviour that changes is the single-process assembly window — the target of the fix. Multi-replica deployments are unchanged (and remain a follow-up).

### DIFF
--- a/src/agentscope/app/_lifespan.py
+++ b/src/agentscope/app/_lifespan.py
@@ -143,6 +143,7 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
             scheduler_manager=scheduler,
             background_task_manager=bg_manager,
             message_bus=message_bus,
+            chat_run_registry=chat_run_registry,
             resource_access_service=resource_access_service,
             knowledge_base_manager=knowledge_base_manager,
             extra_agent_middlewares=app.state.extra_agent_middlewares,
--- a/src/agentscope/app/_service/_chat.py
+++ b/src/agentscope/app/_service/_chat.py
@@ -29,7 +29,11 @@ from ..message_bus import MessageBus, MessageBusKeys
 from ..rag.knowledge_base_manager import KnowledgeBaseManagerBase
 from ..channel import ChatKind
 from ..storage import StorageBase, AgentRecord, SessionRecord, SessionSource
-from .._manager import BackgroundTaskManager, SchedulerManager
+from .._manager import (
+    BackgroundTaskManager,
+    ChatRunRegistry,
+    SchedulerManager,
+)
 from ..workspace_manager import WorkspaceManagerBase
 from ..middleware import (
     InboxMiddleware,
@@ -101,6 +105,7 @@ class ChatService:
         custom_agent_cls: type[Agent] | None = None,
         extra_projectors: list[EventProjector] | None = None,
         channel_dispatcher: "ChannelLifecycleDispatcher | None" = None,
+        chat_run_registry: ChatRunRegistry | None = None,
     ) -> None:
         """Initialize chat service.
 
@@ -167,12 +172,20 @@ optional):
                 The node's channel dispatcher, forwarded to
                 :func:`get_toolkit` so a channel-originated session's
                 agent gets that channel's platform tools.
+            chat_run_registry (`ChatRunRegistry | None`, optional):
+                Per-process registry of in-flight chat-run asyncio
+                tasks. Consulted by :meth:`interrupt` to detect a run
+                that is live locally but has not yet acquired the
+                distributed session lock (during agent assembly).
+                ``None`` (tests / legacy callers) disables the registry
+                check, preserving the previous behaviour.
         """
         self._storage = storage
         self._workspace_manager = workspace_manager
         self._scheduler_manager = scheduler_manager
         self._background_task_manager = background_task_manager
         self._message_bus = message_bus
+        self._chat_run_registry = chat_run_registry
         self._access = resource_access_service
         self._knowledge_base_manager = knowledge_base_manager
         self._extra_agent_middlewares = extra_agent_middlewares
@@ -398,11 +411,27 @@ optional):
 
         Two paths, chosen by session liveness:
 
-        - **Running** (lock held): publish on the interrupt channel so
-          the local :class:`~agentscope.app._manager.CancelDispatcher`
-          cancels its chat-run task; the agent's ``CancelledError``
-          cleanup runs (fake tool results for pending calls, fallback
-          message, ``ReplyEndEvent(INTERRUPTED)``).
+        - **Running** (lock held, or a run is live in this process):
+          publish on the interrupt channel so the local
+          :class:`~agentscope.app._manager.CancelDispatcher` cancels
+          its chat-run task. Once the reply has started, the agent's
+          ``CancelledError`` cleanup runs (fake tool results for
+          pending calls, fallback message, ``ReplyEndEvent(INTERRUPTED)``).
+          The in-process registry check closes the agent-assembly
+          window — before :meth:`_run_impl` acquires the session lock
+          (step 7) — during which the run is already cancellable but
+          ``is_locked`` is not yet ``True``.
+        - **Assembly-window consequences** (same-process): the run is
+          cancelled before any reply exists, so nothing is persisted
+          and no ``ReplyStartEvent`` / ``ReplyEndEvent`` is emitted —
+          the input ``Msg`` is only persisted after assembly (inside
+          the step-7 lock), so it is dropped from the transcript.
+          Synthesising an ``INTERRUPTED`` terminal event for that case
+          is a follow-up (issue direction #2). The registry is
+          per-process: in multi-replica deployments an interrupt
+          landing on a non-hosting worker during assembly still falls
+          through to the resume path (a distributed run-alive marker
+          is the follow-up).
         - **Not running**: enqueue a ``resume`` trigger carrying a
           :class:`UserInterruptEvent`. If the session is parked on
           HITL, the agent short-circuits into the same cleanup path;
@@ -431,7 +460,7 @@ optional):
 
         if await self._message_bus.is_locked(
             MessageBusKeys.session_lock(session_id),
-        ):
+        ) or self._has_live_local_run(session_id):
             await self._message_bus.publish(
                 MessageBusKeys.session_interrupt_channel(),
                 {"session_id": session_id},
@@ -446,6 +475,38 @@ optional):
             kind=MessageBusKeys.WAKEUP_KIND_RESUME,
             inputs=UserInterruptEvent(reply_id=session.state.reply_id),
         )
+
+    def _has_live_local_run(self, session_id: str) -> bool:
+        """Whether a chat-run task for ``session_id`` is live in this
+        process.
+
+        The distributed session lock is acquired only after agent
+        assembly (step 7 of :meth:`_run_impl`); during assembly the run
+        is already live locally but ``is_locked`` still reports
+        ``False``. Consulting the per-process registry closes that
+        window so an interrupt lands on the cancel path instead of being
+        enqueued as a ``resume`` trigger the assembling run cannot
+        process.
+
+        Scope: the registry is per-process, so this only detects runs
+        hosted in *this* process. A run assembling on another worker is
+        invisible here; multi-replica deployments need a distributed
+        run-alive marker (issue follow-up).
+
+        Args:
+            session_id (`str`):
+                The session whose local run to look up.
+
+        Returns:
+            `bool`:
+                ``True`` when a non-finished chat-run task is
+                registered for the session; ``False`` when the service
+                was built without a registry or the task is finished.
+        """
+        if self._chat_run_registry is None:
+            return False
+        task = self._chat_run_registry.get(session_id)
+        return task is not None and not task.done()
 
     @staticmethod
     def _skip_parked_wakeup(
--- a/tests/service_interrupt_dispatch_test.py
+++ b/tests/service_interrupt_dispatch_test.py
@@ -1,0 +1,244 @@
+# -*- coding: utf-8 -*-
+# pylint: disable=missing-class-docstring,missing-function-docstring
+"""Unit tests for :meth:`ChatService.interrupt` liveness dispatch.
+
+Regression test for the agent-assembly interrupt race (issue #2320): a
+chat run is spawned — and registered in ``ChatRunRegistry`` — *before*
+:meth:`ChatService._run_impl` acquires the distributed session lock, so
+``is_locked`` alone cannot tell "a reply is in flight here" during that
+window. The fix makes :meth:`ChatService.interrupt` also consult the
+per-process registry.
+
+Covers the four dispatch outcomes:
+
+- assembly window — live local task, lock not held → interrupt channel
+  (cancel path; previously misrouted to a resume trigger)
+- lock held (mid-reply / HITL-parked)             → interrupt channel
+  (cancel path; unchanged)
+- idle — no local task, lock not held             → resume trigger
+  (unchanged)
+- registry not wired (``None``), lock not held    → resume trigger
+  (legacy fallback preserved)
+"""
+import asyncio
+import unittest
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, patch
+
+from agentscope.app._manager import (
+    BackgroundTaskManager,
+    CancelDispatcher,
+    ChatRunRegistry,
+)
+from agentscope.app._service import ChatService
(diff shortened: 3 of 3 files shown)


iamroylim (approved):
Reviewed at `5722eefca31cd3c5064b4e0eab22c8667672f413`. I traced the liveness decision through the registry spawn, lifespan wiring, interrupt channel, and cancel dispatcher: using the live local task only for the pre-lock assembly window preserves the existing lock-held and idle paths.

I also ran the new dispatch test plus the existing cancel-dispatcher, wakeup-dispatcher, and agent-interrupt suites locally: **25 passed**. The stated boundary is accurate: cross-replica assembly interrupts still need a distributed run-alive marker, and a terminal `INTERRUPTED` event before reply creation remains follow-up scope. No blocking issue found in this patch.

YuhaoLin2005:
Thanks @iamroylim — really appreciate the careful trace through the registry spawn, lifespan wiring, interrupt channel, and cancel dispatcher, plus the independent run of the interrupt-related suites. The confirmation that the lock-held and idle paths stay untouched is exactly the reassurance this kind of race fix needs.

Agreed the cross-replica run-alive marker and a terminal `INTERRUPTED`-before-reply event are follow-ups, not scope creep for this patch.

I'll leave the PR as-is unless you'd like anything adjusted.

Issue #2042, opened 2026-07-10
feat(service): support sharing agent with MCPs and skills


DavdGao:
**Background**

As a follow-up to #1844, while we already support sharing credentials, agents, and knowledge bases, the current agent sharing mechanism is incomplete. 

At the moment, only the basic agent configuration is shared — i.e., `name`, `system prompt`, and related configs. However, a well-tuned agent typically also includes MCPs and skills (even memory), which are equally important and should be carried over during sharing. Without them, a shared agent loses much of its value.

**The Challenge**

MCPs, skills, and even memory files are all persisted within the agent's workspace(sandbox filesystem). This raises a fundamental question: should sharing an agent also involve sharing its workspace?

The current answer is **no**. The isolation model in our workspace manager is hierarchical: `user_id` is the top-level boundary, with `agent_id` and `session_id` nested within. This means when a user receives a shared agent and creates a new session, a **new workspace is created under their own account** — which is the correct behavior, as sharing a workspace across users would risk data interference.

However, this creates a gap: **the original agent's MCPs and skills need to be carried over into the new user's workspace**, and there is currently no mechanism to do so.

**Proposed Solution (Under Discussion)**

One candidate solution is to allow an agent to declare a set of **default MCPs and skills** as part of its configuration. When a brand-new workspace is created for that agent (regardless of isolation level), these defaults would be automatically applied as the initial setup.

However, this approach has known limitations:
1. MCP configs may contain sensitive information (e.g. API keys) and therefore cannot be naively copied across users
2. Skills present an additional challenge — we currently lack a dedicated file service for skills (though we do have blob abstraction and S3 integration for the RAG pipeline, it is not designed to serve skills)

March-77:
[...]

PR #2326, opened 2026-08-15
fix(app): support cross-owner agent invites

### LINKED ISSUE #2319: [Bug]: AgentInvite lists cross-owner shared agents but rejects them as no longer invitable
### Prerequisites

## Background

Agent sharing (#1844) and the host-resolvable invite pool (#1999) allow an agent owned by one user to be visible to another user through a `ResourceAccessPolicy`.

When an owner creates an invitable agent and a policy grants a viewer READ access:

1. `get_toolkit` correctly calls `ResourceAccessService.list_resource(viewer_id, AGENT)`.
2. The shared agent is included in the viewer leader's `AgentInvite` target enum.
3. Calling `AgentInvite` fails with `agent ... is no longer invitable`.

This is distinct from #2042: the failure occurs before MCP/skill/workspace inheritance is relevant.

## Expected behavior

The viewer should be able to invite the shared agent while preserving split ownership:

- the existing `AgentRecord` remains owned by the resource owner;
- the team and borrowed `SessionRecord` belong to the viewer/caller;
- `TeamMember.owner_id` records the shared agent definition owner;
- access is re-evaluated at invocation time so revocation fails closed.

## Actual behavior

The shared agent appears in `AgentInvite` but cannot be invited.

## Root cause

The toolkit discovers own + shared records through `ResourceAccessService.list_resource`, but `AgentInvite.__call__` re-fetches with:

```python
fresh = await self._storage.get_agent(self._user_id, invited.id)
```

For a viewer calling an owner-owned agent, this becomes `get_agent(viewer_id, shared_agent_id)` and returns `None`.

The fresh lookup should use the resource access service (for example, `resolve_agent(viewer_id, agent_id)`) so access is checked again and the real owner record is returned.

There is a second ownership mismatch when the member is saved:

```python
TeamMember(owner_id=self._user_id, ...)
```

For a shared agent this should use the resolved agent's owner, while the borrowed session should continue to use the viewer/caller as `user_id`.

Related TeamSay and team/session deletion paths should also be covered by cross-owner regression tests so definition lookup uses the agent owner while borrowed-session cleanup uses the team/viewer owner.

The same raw lookup and member owner assignment are still present on current `main` at the time of filing.

### Error Messages

```shell
AgentInvite: agent 'Shared Specialist' is no longer invitable.
```

### Steps to Reproduce

A minimal regression test can be added to `tests/service_team_tools_test.py` using its existing `_AgentInviteTestBase` fixture:

```python
class TestAgentInviteCrossOwner(_AgentInviteTestBase):
    async def test_invites_shared_agent(self) -> None:
        from agentscope.app.storage._model._agent import InviteConfig

        shared_owner = "platform-owner"
        shared_agent = AgentRecord(
            user_id=shared_owner,
            source="user",
            data=AgentData(
                name="Shared Specialist",
                system_prompt="I am a shared specialist.",
                context_config=ContextConfig(),
                react_config=ReActConfig(),
                invite_config=InviteConfig(
                    invitable=True,
                    invite_description="Shared expert.",
                ),
            ),
        )
        await self.storage.upsert_agent(shared_owner, shared_agent)

        tool = AgentInvite(
            storage=self.storage,
            message_bus=self.bus,
            workspace_manager=self.workspace_manager,
            user_id=self.user_id,  # viewer / team owner
            session_id=self.leader_session.id,
            agent_id=self.leader_agent.id,
            invitable_pool=[shared_agent],  # supplied by list_resource()
        )

        target = f"Shared Specialist@{shared_agent.id[:8]}"
        chunk = await tool(target=target, prompt="Please handle this.")

        self.assertNotEqual(chunk.state, ToolResultState.ERROR)

        leader_session = await self.storage.get_session(
            self.user_id,
            self.leader_agent.id,
            self.leader_session.id,
        )
        team = await self.storage.get_team(
            self.user_id,
            leader_session.team_id,
        )
        member = team.data.members[0]
        self.assertEqual(member.owner_id, shared_owner)

        borrowed = await self.storage.get_session(
            self.user_id,
            shared_agent.id,
            member.session_id,
        )
        self.assertIsNotNone(borrowed)
```

Run:

```bash
pytest -q tests/service_team_tools_test.py -k AgentInviteCrossOwner
```

On current code, the assertion fails because `chunk.state` is `error` with the message shown above.

### Environment

- AgentScope Version: 2.0.6
- Python Version: 3.13.5
- OS: macOS 15.3.2
- Reproduced with the existing in-memory team-tool test fixture
- Current main inspected on 2026-08-14 and still contains the same lookup/owner logic

logicwu0:
## Summary

- re-resolve invite targets through `ResourceAccessService` so cross-owner access is checked again at invocation time
- preserve the shared agent definition owner while keeping the borrowed session in the team owner namespace
- clean up cross-owner borrowed sessions correctly in the service, Redis, and SQL storage paths
- add regression coverage for successful invitations, revoked access, workspace isolation, routing ownership, and team deletion

## Tests

- `pre-commit run --files <changed files>`
- `pytest tests/service_team_tools_test.py tests/storage_redis_test.py -q`
- `pytest tests/storage_sql_test.py -q`
- `pytest tests/service_toolkit_test.py tests/resource_access_policy_test.py -q`

Closes #2319

### DIFF
--- a/src/agentscope/app/_tool/_agent_invite.py
+++ b/src/agentscope/app/_tool/_agent_invite.py
@@ -34,6 +34,7 @@ from ..._utils._common import _generate_id
 
 if TYPE_CHECKING:
     from ..message_bus import MessageBus
+    from .._service._access import ResourceAccessService
     from ..storage import AgentRecord, StorageBase
     from ..workspace_manager import WorkspaceManagerBase
 
@@ -168,6 +169,7 @@ class AgentInvite(_TeamToolBase):
         session_id: str,
         agent_id: str,
         invitable_pool: list["AgentRecord"],
+        resource_access_service: "ResourceAccessService | None" = None,
     ) -> None:
         """Bind the base dependencies plus the invitable pool snapshot.
 
@@ -188,6 +190,10 @@ class AgentInvite(_TeamToolBase):
             invitable_pool (`list[AgentRecord]`):
                 Snapshot of currently-invitable agents. Must be
                 non-empty — the caller skips construction otherwise.
+            resource_access_service (`ResourceAccessService | None`):
+                Resolves the selected agent against the caller's current
+                access grants. ``None`` preserves direct construction for
+                owner-only integrations.
         """
         super().__init__(
             storage,
@@ -201,6 +207,7 @@ class AgentInvite(_TeamToolBase):
         self._pool_by_id: dict[str, "AgentRecord"] = {
             a.id: a for a in invitable_pool
         }
+        self._resource_access_service = resource_access_service
 
         # Build enum + a human-readable per-target rundown for the LLM.
         enum_values = [
@@ -266,12 +273,10 @@ class AgentInvite(_TeamToolBase):
 
             team = await self._require_leader_team("invite members")
 
-            # Re-fetch fresh — the snapshot could be stale if the user
-            # just toggled the invite off.
-            fresh = await self._storage.get_agent(
-                self._user_id,
-                invited.id,
-            )
+            # Re-fetch fresh — both the invite settings and a cross-owner
+            # access grant may have changed since the toolkit snapshot was
+            # assembled.
+            fresh = await self._resolve_fresh_agent(invited)
             if (
                 fresh is None
                 or not fresh.data.invite_config.invitable
@@ -401,7 +406,7 @@ class AgentInvite(_TeamToolBase):
             team.data.members = [
                 *existing_members,
                 TeamMember(
-                    owner_id=self._user_id,
+                    owner_id=invited.user_id,
                     agent_id=invited.id,
                     session_id=borrowed.id,
                     role="invited",
@@ -453,6 +458,30 @@ class AgentInvite(_TeamToolBase):
                 content=[TextBlock(text=f"AgentInvite failed: {e}")],
                 state=ToolResultState.ERROR,
             )
+
+    async def _resolve_fresh_agent(
+        self,
+        invited: "AgentRecord",
+    ) -> "AgentRecord | None":
+        """Resolve a snapshot entry without losing its owner namespace."""
+        if self._resource_access_service is None:
+            return await self._storage.get_agent(invited.user_id, invited.id)
+
+        # Keep the web-framework exception dependency local to the app-only
+        # access-service path. A revoked share is a normal stale-snapshot
+        # outcome for this tool, not an unhandled tool failure.
+        from fastapi import HTTPException
+
+        try:
+            fresh = await self._resource_access_service.resolve_agent(
+                self._user_id,
+                invited.id,
+            )
+            return fresh if fresh.user_id == invited.user_id else None
+        except HTTPException as exc:
+            if exc.status_code == 404:
+                return None
+            raise
 
 
 def _resolve_target(
--- a/src/agentscope/app/_tool/_agent_invite.py
+++ b/src/agentscope/app/_tool/_agent_invite.py
@@ -5,10 +5,11 @@ Unlike :class:`AgentCreate`, which spawns a brand-new worker
 (``source='team'``) from a :class:`SubAgentTemplate`, this tool
 **borrows** a pre-existing user-owned agent by minting a fresh
 team-scoped :class:`SessionRecord` on top of the *existing*
-:class:`AgentRecord`.  The borrowed agent keeps its system prompt,
-context/react configs, workspace, MCP, skills, and model choice; only
-a new session is created so it can hold a parallel conversation for
-the team.
+:class:`AgentRecord`. The borrowed agent keeps its definition, including
+its system prompt and context/react configs. An agent owned by the leader
+may reuse its existing session configuration; a cross-owner agent instead
+runs in a fresh leader-owned workspace with the leader's chat model and
+without the owner's MCPs, skills, cache, or session permissions.
 
 When the team is dissolved or the leader is deleted, only the borrowed
 session is cleaned up — the underlying :class:`AgentRecord` survives
@@ -189,7 +190,9 @@ class AgentInvite(_TeamToolBase):
                 The calling agent id.
             invitable_pool (`list[AgentRecord]`):
                 Snapshot of currently-invitable agents. Must be
-                non-empty — the caller skips construction otherwise.
+                non-empty — the caller skips construction otherwise. When
+                no access service is supplied, the constructor is responsible
+                for ensuring every snapshot entry is safe for the caller.
             resource_access_service (`ResourceAccessService | None`):
                 Resolves the selected agent against the caller's current
                 access grants. ``None`` preserves direct construction for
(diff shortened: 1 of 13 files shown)


src/agentscope/app/storage/_model/_team.py
```diff
@@ -22,13 +22,9 @@ class TeamMember(BaseModel):
...
     owner_id: str = Field(
         description=(
-            "Owner of the member's agent. Always equals the team owner "
-            "in today's user-only invite pool, but is stored explicitly "
-            "so a future admin-share layer (agents borrowed across users) "
-            "can slot in without a schema migration. Distinct name from "
-            ":attr:`TeamRecord.user_id` on purpose — the surrounding "
-            "team already carries the team owner in context, so calling "
-            "this field ``user_id`` too would be ambiguous."
+            "Owner of the member's agent definition. For a cross-owner "
```
DavdGao, line 25: Widening this to the definition owner means every reader that feeds it into `get_agent` now crosses into another user's namespace. `_router/_session.py:98` does exactly that with no policy check — I confirmed that after the grant is revoked, team detail still returns the shared agent's full `AgentData` including `system_prompt`. Chat itself fails closed via `resolve_agent`, but this read path doesn't, so it should go through `ResourceAccessService` too.

Same pattern at `_team_say.py:237` and `_agent_create.py:415`, though those only surface the agent name into a routing directory.

src/agentscope/app/storage/_redis_storage.py
```diff
@@ -1745,7 +1746,7 @@ async def delete_team(self, user_id: str, team_id: str) -> bool:
                 await self.delete_agent(member.owner_id, member.agent_id)
             else:  # invited
                 await self.delete_session(
-                    member.owner_id,
+                    user_id,
```
DavdGao, line 1749: This covers team deletion, but not the reverse direction: `delete_agent` scrubs team back-references only within `list_teams(user_id)`, so when the *owner* deletes a shared agent the borrower's session and its `TeamMember` both survive. `TeamSay` then silently drops the member (`get_agent` → `None` → `continue`) and opening the borrowed session 404s.

Partly pre-existing — the sessions router already lets a viewer create a session on a shared agent — but this PR is what puts such a session into a team roster. Probably worth a note or a follow-up issue rather than fixing here.

src/agentscope/app/_tool/_agent_invite.py
```diff
@@ -289,12 +296,10 @@ async def __call__(
...
-            # Re-fetch fresh — the snapshot could be stale if the user
-            # just toggled the invite off.
-            fresh = await self._storage.get_agent(
-                self._user_id,
-                invited.id,
-            )
+            # Re-fetch fresh — both the invite settings and a cross-owner
+            # access grant may have changed since the toolkit snapshot was
+            # assembled.
+            fresh = await self._resolve_fresh_agent(invited)
```
DavdGao, line 282: Once this can return a cross-owner record, the comment at :353 stops being accurate: it says the invited agent's own primary session is preferred because it already has MCP / skills / cache set up, but :360 looks up `list_sessions(self._user_id, ...)`, which for a cross-owner invite never finds the owner's session and always takes the fresh-workspace branch.

The behaviour is right and the new test locks the isolation in — the comment just needs to say so. Worth being explicit too that a borrowed cross-owner agent therefore runs with the leader's chat model and none of the owner's MCP / skills (#2042).

src/agentscope/app/_tool/_agent_invite.py
```diff
@@ -478,6 +483,30 @@ async def __call__(
                 state=ToolResultState.ERROR,
             )
 
+    async def _resolve_fresh_agent(
+        self,
+        invited: "AgentRecord",
+    ) -> "AgentRecord | None":
+        """Resolve a snapshot entry without losing its owner namespace."""
+        if self._resource_access_service is None:
+            return await self._storage.get_agent(invited.user_id, invited.id)
```
DavdGao, line 475: This moved from `get_agent(self._user_id, ...)` to `get_agent(invited.user_id, ...)`, which drops the implicit "the pool may only hold the caller's agents" guard — pool correctness is now entirely the constructor's responsibility. Fine for the direct-construction path, but worth a line in the docstring.

src/agentscope/app/_tool/_agent_invite.py
```diff
@@ -478,6 +483,30 @@ async def __call__(
...
+
+        # Keep the web-framework exception dependency local to the app-only
+        # access-service path. A revoked share is a normal stale-snapshot
+        # outcome for this tool, not an unhandled tool failure.
+        from fastapi import HTTPException
```
DavdGao: This doesn't need to be function-local. The import that has to stay deferred is `.._service._access` (circular via `_toolkit` → `.._tool`); `fastapi` isn't circular, and the whole `app` package already hard-depends on the `service` extra.

src/agentscope/app/_tool/_agent_invite.py
```diff
@@ -478,6 +483,30 @@ async def __call__(
...
+        try:
+            fresh = await self._resource_access_service.resolve_agent(
+                self._user_id,
+                invited.id,
+            )
+            return fresh if fresh.user_id == invited.user_id else None
+        except HTTPException as exc:
+            if exc.status_code == 404:
+                return None
+            raise
```
DavdGao: Unreachable — `resolve_agent` only ever raises 404, via `_not_found`. A non-raising resolve variant on `ResourceAccessService` (returning `None`) would collapse `_resolve_fresh_agent` to a couple of lines and keep the web-framework exception out of the tool layer entirely.

DavdGao (commented):
Checked this out locally — the related tests pass and `pre-commit` is clean on the changed files. The core fix and the `owner_id` widening look right, and existing records need no migration since `owner_id == team owner == agent owner` for them. Left notes inline; the first one (team detail no longer re-checking the policy) is the only thing I'd call a real regression, the rest are follow-ups or cleanups.

src/agentscope/app/storage/_model/_team.py
logicwu0: Fixed the sensitive team-detail path. `_build_team_detail` now resolves each member through `ResourceAccessService.try_resolve_agent` and verifies the resolved owner before returning agent data. I also added a regression test covering grant revocation. I left the name-only routing lookups unchanged in this focused fix.

src/agentscope/app/storage/_redis_storage.py
logicwu0: Agreed. I kept the reverse owner-deletion cascade out of this PR as suggested, so this change remains focused on cross-owner invitation and the sensitive read-path regression.

src/agentscope/app/_tool/_agent_invite.py
logicwu0: Updated the comments to describe the cross-owner behavior explicitly: a borrowed agent uses a fresh leader-owned workspace and the leader chat model, without reusing the owner session, MCP configuration, skills, cache, or permissions.

src/agentscope/app/_tool/_agent_invite.py
logicwu0: Updated the constructor documentation to state that, without a resource access service, pool correctness is the caller’s responsibility.

src/agentscope/app/_tool/_agent_invite.py
logicwu0: Removed the function-local FastAPI import together with the web-framework exception handling from the tool layer.

src/agentscope/app/_tool/_agent_invite.py
logicwu0: Added the non-raising `ResourceAccessService.try_resolve_agent` variant and simplified `_resolve_fresh_agent` to use it directly.

DavdGao (approved):

PR #2332, opened 2026-08-17
fix(rag): isolate concurrent middleware inputs

### LINKED ISSUE #2331: [Bug]: shared RAGMiddleware mixes static queries across concurrent agents
### Prerequisites

- [x] I searched existing issues and pull requests for RAGMiddleware concurrency and cached-input isolation.
- [x] This is a reproducible bug, not a usage question.

### Background / Description

`RAGMiddleware` in static mode stores the current reply query in one instance-level `_cached_inputs` slot. When one middleware instance is shared by two agents and their replies overlap, the second `on_reply` overwrites that slot before the first agent reaches `on_reasoning`.

A deterministic interleaving is:

1. Agent A enters `on_reply` with `query-a` and pauses before reasoning.
2. Agent B enters `on_reply` with `query-b` and pauses before reasoning.
3. Agent A resumes. Its `on_reasoning` searches for `query-b`, so session A receives retrieval context for session B's query.
4. Agent A exits and its `finally` clears the shared slot.
5. Agent B resumes and skips retrieval entirely because the slot is now `None`.

This is the same shared-middleware isolation class as #2274, but it affects the synchronous static RAG query scratchpad rather than AgenticMemory's background task.

### Expected behavior

Each concurrent reply should search with its own input. Finishing one reply must not clear another reply's pending query.

### Actual behavior

The deterministic hook-level reproduction records one embedding call:

```text
[['user: query-b']]
```

Expected:

```text
[['user: query-a'], ['user: query-b']]
```

The first call is cross-session data contamination; the missing second call is a silent retrieval omission.

### Proposed fix

Store the reply-local scratchpad in a `ContextVar` and set/reset it with a token around the downstream `on_reply` chain. That keeps values isolated between asyncio tasks, preserves nested hook behavior, and avoids a mutable per-session registry.

### Verification

Add a deterministic regression test that starts two `on_reply` calls on one middleware instance, barriers both before `on_reasoning`, then resumes A followed by B and asserts both exact embedding queries.

Reproduced on current `main` at `3a4e2ae` with Python 3.12.

fszcd:
## AgentScope Version

2.0.6 (current `main` at `3a4e2ae`)

## Description

Fixes #2331.

`RAGMiddleware` static mode kept reply inputs in one instance field. When a middleware instance was shared across concurrent agents, the later reply overwrote that field: the first agent searched with the second agent's query, and the first reply's cleanup then caused the second agent to skip retrieval.

This change stores the reply-local input blocks in a `ContextVar` and restores the prior value with its token when the reply chain exits. This isolates concurrent asyncio tasks and preserves nested middleware calls without maintaining a mutable session registry.

The regression test deterministically barriers two replies before reasoning, resumes them in sequence, and verifies that both exact queries reach the embedding model.

### Validation

```bash
uv run pytest tests/middleware_rag_test.py tests/middleware_test.py -q
# 41 passed

uv run pre-commit run --files src/agentscope/middleware/_rag.py tests/middleware_rag_test.py
# all hooks passed
```

## Checklist

- [x] Related documentation has been updated (not needed: no public API or documented behavior changes)

### DIFF
--- a/src/agentscope/middleware/_rag.py
+++ b/src/agentscope/middleware/_rag.py
@@ -31,6 +31,7 @@ service, to the knowledge-base manager) that constructs the
 """
 import asyncio
 import json
+from contextvars import ContextVar
 from copy import deepcopy
 from typing import (
     TYPE_CHECKING,
@@ -589,7 +590,9 @@ class RAGMiddleware(MiddlewareBase):
         # original reply inputs (which are no longer in
         # ``agent.state.context`` by the time ``on_reasoning`` runs in
         # a tool-call loop).  Cleared in ``on_reply``'s finally.
-        self._cached_inputs: list[TextBlock | DataBlock] | None = None
+        self._cached_inputs: ContextVar[
+            list[TextBlock | DataBlock] | None
+        ] = ContextVar("rag_middleware_cached_inputs", default=None)
 
     # ------------------------------------------------------------------
     # Agentic mode — expose the search tool
@@ -655,6 +658,7 @@ class RAGMiddleware(MiddlewareBase):
         ):
             msgs = inputs
 
+        cached_inputs: list[TextBlock | DataBlock] | None = None
         if msgs:
             # Deepcopy because we are about to mutate the first text block of
             # each message to prepend the speaker name — never touch the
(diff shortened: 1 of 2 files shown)


fszcd:
Closing this for now because it has been inactive for over a month and is based on an older tree. If the change is still useful later, it can be revisited against the current branch. Thank you.

PR #2334, opened 2026-08-17
fix: prevent double end signal in stream_printing_messages and add NotImplementedError in UserAgent.observe

Ricardo-M-L:
This PR addresses: prevent double end signal in stream_printing_messages and add NotImplementedError in UserAgent.observe


iamroylim:
The current diff against `main` includes 26 commits, 38 files, and 4,727 additions, including Tablestore, formatter, session, and documentation changes unrelated to the two fixes in this PR. That makes both review and CI attribution unsafe. Please rebase or cherry-pick the intended head commit (`5e3e6a7531f4a6ee1728fb83bca84478c98764d3`) onto current `main`, so the PR only contains the focused `_functional.py` / `_user_agent.py` changes plus targeted tests.

PR #2345, opened 2026-08-18
fix(formatter): include media type in DashScope audio data URLs

BlessedWithLuck1105:
## Summary

`DashScopeChatFormatter` builds base64 audio data URLs as `data:;base64,<data>`, which omits the media type. Image and video sources already include it (`data:<media_type>;base64,...`), so audio is the only inconsistent case. DashScope's compatible API expects the media type in data URLs for audio payloads.

This fixes both the `Base64Source` and the local `file://` URL paths.

## Tests

- Updated the three existing assertions in `tests/formatter_dashscope_test.py` to expect the media type in the data URL.
- `python -m pytest tests/formatter_dashscope_test.py -q` → 15 passed

## Notes

- No breaking changes.

### DIFF
--- a/src/agentscope/formatter/_dashscope_formatter.py
+++ b/src/agentscope/formatter/_dashscope_formatter.py
@@ -186,7 +186,7 @@ class _DashScopeFormatterBase(FormatterBase, ABC):
             return {
                 "type": "input_audio",
                 "input_audio": {
-                    "data": f"data:;base64,{source.data}",
+                    "data": f"data:{source.media_type};base64,{source.data}",
                     "format": fmt,
                 },
             }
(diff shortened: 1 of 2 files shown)


qbc2016:
Please provide a real DashScope API verification for `qwen3.5-omni-plus`, covering both WAV and MP3 base64 inputs. The [official Qwen-Omni example](https://help.aliyun.com/en/model-studio/qwen-omni#1c7a1764cfjbs) currently uses `data:;base64,...`. Updating snapshot assertions alone does not establish that the new provider wire format is required or regression-free.

BlessedWithLuck1105:
Verified against the official Alibaba Cloud docs (https://help.aliyun.com/zh/model-studio/qwen-omni): the OpenAI-compatible example for qwen3.5-omni-plus passes base64 audio as `"data": f"data:;base64,{base64_audio}"` — the `data:;base64,<data>` form without a media type is the documented DashScope wire format, and the original code was correct. My change would have broken that compatibility, so I'm closing this PR.

PR #2346, opened 2026-08-18
fix(model): only send enable_thinking when thinking is enabled

BlessedWithLuck1105:
## Summary

`DashScopeChatModel.Parameters.thinking_enable` is a plain `bool` defaulting to `False`, so the guard `if self.parameters.thinking_enable is not None:` is always true. Every request therefore sends `extra_body["enable_thinking"] = false`, even when thinking was never requested. Some DashScope endpoints reject the explicit `enable_thinking: false` flag (HTTP 400), so the default path can fail where the flag is simply omitted.

The condition now sends `enable_thinking` only when thinking is enabled, matching the truthiness check already used in `_parse_stream_response` for the same field.

## Tests

- Added `test_default_thinking_not_sent` (no `enable_thinking` in `extra_body` with defaults) and `test_thinking_enabled_sent` (`enable_thinking: true` when enabled).
- `python -m pytest tests/model_dashscope_test.py -q` → 16 passed

## Notes

- No breaking changes.

### DIFF
--- a/src/agentscope/formatter/_dashscope_formatter.py
+++ b/src/agentscope/formatter/_dashscope_formatter.py
@@ -186,7 +186,7 @@ class _DashScopeFormatterBase(FormatterBase, ABC):
             return {
                 "type": "input_audio",
                 "input_audio": {
-                    "data": f"data:;base64,{source.data}",
+                    "data": f"data:{source.media_type};base64,{source.data}",
                     "format": fmt,
                 },
             }
@@ -200,7 +200,7 @@ class _DashScopeFormatterBase(FormatterBase, ABC):
                 return {
                     "type": "input_audio",
                     "input_audio": {
-                        "data": f"data:;base64,{encoded}",
+                        "data": f"data:{source.media_type};base64,{encoded}",
                         "format": fmt,
                     },
                 }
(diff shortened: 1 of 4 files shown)


qbc2016:
Thanks for identifying this issue and adding the related tests. I agree that we should avoid sending unsupported parameters to DashScope models, but I am not sure the current change preserves the intended semantics of `thinking_enable`.

Since `thinking_enable` is a plain boolean that defaults to `False`, `False` currently means that AgentScope should disable thinking. Changing the condition to a truthiness check conflates “disabled” with “not specified” and prevents AgentScope from sending `enable_thinking=False`. For models whose server-side default is thinking-enabled, users would therefore no longer be able to disable thinking through this parameter.

For thinking-only models that cannot disable thinking, AgentScope should represent thinking as enabled so that its internal state remains consistent with the model’s actual behavior.

Could you please share the exact DashScope endpoint/model and the 400 response showing this issue? In particular, does it reject the `enable_thinking` parameter itself, or only `enable_thinking=False` for certain models?

If the parameter is accepted, the existing request construction may not need to change. If only certain models do not support it, perhaps we could gate the parameter based on model capability instead of omitting `False` for every DashScope model.

BlessedWithLuck1105:
Verified: per the official docs, `enable_thinking` defaults vary per model (e.g. qwen3.6-max-preview defaults to true, qwen3.7-max to false). Sending `enable_thinking: false` is therefore required to express `thinking_enable=False` on models whose server-side default is thinking-enabled, and omitting it would silently leave thinking on. The truthiness check conflates "disabled" with "unspecified", so the original request construction stands. I'm closing this PR.

PR #2351, opened 2026-08-18
fix(formatter): keep images accumulated before a tool call

BlessedWithLuck1105:
## Summary

In `OllamaChatFormatter.format`, the `ToolCallBlock` branch reset the accumulated `images` list without flushing it, so an assistant message containing image `DataBlock`s followed by a `ToolCallBlock` silently lost the images. Every sibling flush site in the same method preserves images: the `HintBlock` branch, the `ToolResultBlock` branch, and the end-of-message flush.

The branch now flushes pending text + images into their own message before emitting the `tool_calls` message (whose content stays empty in that case); when no images are present, the previous behavior — text carried on the `tool_calls` message — is unchanged.

## Tests

- Added `test_chat_formatter_image_before_tool_call_kept`: an assistant message with text + image + tool call formats to two messages, the first carrying the text and the images list.
- `python -m pytest tests/formatter_ollama_test.py -q` → 11 passed

## Notes

- No breaking changes.

### DIFF
--- a/src/agentscope/formatter/_ollama_formatter.py
+++ b/src/agentscope/formatter/_ollama_formatter.py
@@ -186,32 +186,41 @@ class OllamaChatFormatter(_OllamaFormatterBase):
                     pass
 
                 elif isinstance(block, ToolCallBlock):
-                    messages.append(
-                        {
-                            "role": msg.role,
-                            "content": "\n".join(content_parts)
-                            if content_parts
-                            else "",
-                            "tool_calls": [
-                                {
-                                    "function": {
-                                        "name": block.name,
-                                        # Ollama SDK expects a dict, not a
-                                        # JSON string. Use the repair helper
-                                        # so a truncated input (from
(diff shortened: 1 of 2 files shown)


qbc2016:
When `images` is non-empty, this emits two consecutive assistant messages. Ollama's server-side `collate` merges consecutive same-role messages by appending only `Content` to the first message, so the second message's `ToolCalls` are discarded before template rendering. Please keep `content`, `images`, and `tool_calls` on the same message and update the test to expect one assistant message.

BlessedWithLuck1105:
Good catch on the server-side `collate` merging consecutive same-role messages — a separate images message would indeed be collapsed away. Updated the branch to keep `content`, `images`, and `tool_calls` on a single assistant message (no behavior change when no images are present), and the test now expects exactly one message. Pushed as `58076f0`; all pre-commit hooks and formatter tests pass.

qbc2016 (approved):
LGTM.

PR #2382, opened 2026-08-20
fix(middleware): isolate static RAG inputs across concurrent replies

### LINKED ISSUE #2331: [Bug]: shared RAGMiddleware mixes static queries across concurrent agents

weichu2002:
## Summary

Fixes #2331.

### Root cause

In static mode, `RAGMiddleware` stored reply inputs in one mutable instance attribute, `_cached_inputs`. A shared middleware instance therefore allowed an overlapping reply to overwrite another reply's retrieval query. When the first reply completed, its cleanup also cleared the second reply's pending query.

### Fix

`_cached_inputs` is now an instance-owned `ContextVar`. `on_reply` sets the reply-local value and retains the returned Token; its `finally` block calls `reset(token)`.

`ContextVar` values are isolated per asyncio task, so concurrent replies on the same middleware instance read their own retrieval inputs. Token reset is essential: it restores the prior value for nested calls and ensures exception or cancellation cleanup does not erase an outer reply's context. This preserves the existing API and serial behavior.

### Regression coverage

Added a deterministic concurrency regression test using `asyncio.Event` (no timing sleeps):

1. Reply A captures `query-a` and pauses.
2. Reply B captures `query-b` and pauses on the same middleware instance.
3. A resumes and must search `query-a`.
4. After A exits and cleans up, B resumes and must still search `query-b`.

The test asserts both exact embedding inputs, covering both cross-request contamination and cleanup interference.

### Verification

Run successfully:

- `python -m py_compile src\\agentscope\\middleware\\_rag.py tests\\middleware_rag_concurrency_test.py`
- `git diff --check`

Attempted but blocked by the local environment:

- `python -m pytest tests/middleware_rag_concurrency_test.py -q` — pytest was not installed.
- `python -m unittest tests/middleware_rag_concurrency_test.py -v` — failed before collection because `docstring_parser` was missing.
- Created an E-drive-only virtual environment and ran `pip install -e '.[dev]'`; installation made no progress and was stopped, so the targeted RAG suite, pre-commit, type checks, and full test suite could not be run in this environment.

### DIFF
--- a/src/agentscope/middleware/_rag.py
+++ b/src/agentscope/middleware/_rag.py
@@ -31,6 +31,7 @@ service, to the knowledge-base manager) that constructs the
 """
 import asyncio
 import json
+from contextvars import ContextVar
 from copy import deepcopy
 from typing import (
     TYPE_CHECKING,
@@ -588,8 +589,11 @@ class RAGMiddleware(MiddlewareBase):
         # consumed by ``on_reasoning`` so the auto-search can use the
         # original reply inputs (which are no longer in
         # ``agent.state.context`` by the time ``on_reasoning`` runs in
-        # a tool-call loop).  Cleared in ``on_reply``'s finally.
-        self._cached_inputs: list[TextBlock | DataBlock] | None = None
+        # a tool-call loop).  It must be task-local because middleware
+        # instances may be shared by concurrent agents.
+        self._cached_inputs: ContextVar[
+            list[TextBlock | DataBlock] | None
+        ] = ContextVar("rag_middleware_cached_inputs", default=None)
 
     # ------------------------------------------------------------------
     # Agentic mode — expose the search tool
@@ -655,12 +659,13 @@ class RAGMiddleware(MiddlewareBase):
         ):
             msgs = inputs
 
+        blocks: list[TextBlock | DataBlock] | None = None
         if msgs:
             # Deepcopy because we are about to mutate the first text block of
             # each message to prepend the speaker name — never touch the
             # caller's message objects.
             msgs = deepcopy(msgs)
-            blocks: list[TextBlock | DataBlock] = []
+            blocks = []
             for msg in msgs:
                 if not msg.content:
                     continue
@@ -671,13 +676,13 @@ class RAGMiddleware(MiddlewareBase):
                     blocks.append(TextBlock(text=speaker))
                 blocks.extend(msg.content)
 
-            self._cached_inputs = blocks
+        token = self._cached_inputs.set(blocks)
 
         try:
             async for evt in next_handler(**input_kwargs):
                 yield evt
         finally:
-            self._cached_inputs = None
+            self._cached_inputs.reset(token)
 
     async def on_reasoning(
         self,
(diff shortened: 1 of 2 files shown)


src/agentscope/middleware/_rag.py
```diff
@@ -671,13 +676,13 @@ async def on_reply(
                     blocks.append(TextBlock(text=speaker))
                 blocks.extend(msg.content)
 
-            self._cached_inputs = blocks
+        token = self._cached_inputs.set(blocks)
```
AmirF194, line 679: Couldn't get the actual test suite running here (disk-constrained sandbox: `pip install -e .` finished but pytest then hung with no output), so I built a standalone repro of just this pattern instead, no dependencies: a ContextVar set/reset around an async-generator on_reply/on_reasoning chain, two replies driven by separate `asyncio.create_task`, same shape as `middleware_rag_concurrency_test.py`.

Pre-fix (plain instance attribute) reproduces #2331 exactly: task A's search picks up B's query, and B comes back empty once A's cleanup clears the shared slot. Post-fix (the `.set()`/`.get()` at 679/723) isolates the two correctly in both orderings I tried.

I also checked whether this codebase ever drives concurrent `on_reply` calls from a single Task instead of separate ones, since that would defeat ContextVar isolation. Every concurrency pattern I found (`_dispatcher.py`, the workspace managers) uses `asyncio.gather`/`create_task`, so the isolation should hold for how replies actually get scheduled here.

PR #2393, opened 2026-08-21
feat(sop): standard operating procedures — fixed milestones, each verified

### LINKED ISSUE #2401: feat(sop): support sdk-level SOP creation, execution, and verification
**Background**

- The SDK-level implementation focuses on the core logic of **SOP creation, execution, and verification**, without any dependency on external storage or frontend interfaces. It also includes first-class support for **Human-in-the-Loop (HITL)** interactions, where user approval/external execution may be required at each SOP step.
- The SDK-level implementation will serve as the **foundational layer** for the service-level SOP implementation.

**Changes**

- A new `agentscope.sop` module in agentscope library.
- Basic abstractions for the SOP functionality.
- Documentations.
- Unittests.

DavdGao:
## What

A SOP is **a fixed skeleton with free flesh**: which milestones there are, in what order, and what each must prove is written by a person and can be read without running anything; how any of it gets done belongs to the agents.

```python
sop = SOP(name="客户投诉处理", steps=[
    SOPStep("核实事实",   "...", executor=support, verifier=auditor),
    SOPStep("拟补偿方案", "...", executor=policy,  verifier=SupervisorApproval()),
    SOPStep("写回复客户", "...", executor=writer,  verifier=safety),
])
engine = SOPEngine(sop)
async for event in engine.reply_stream(UserMsg(name="user", content=complaint)):
    ...
```

Two tests draw the boundary. *Is this a SOP?* — if you cannot say how many steps it has and who does each without running it, no. *Should this be a step?* — if nobody checks anything at that point, no.

## What the engine does — and deliberately does not

It walks the steps in order, routes a resumption event to whichever step parked, and spends each step's `max_attempts`. A refusal goes back to the executor verbatim, with which attempt it is; exhaustion fails the run rather than falling through.

It sees **verdicts, never verifiers**. A verifier is anything satisfying `AgentLike` — an `Agent` with tools, or a twenty-line object that asks a person and stops. `SOPStep` (executor + verifier) is the usual shape but a convenience, not a law: subclass `SOPStepBase` and a step can be anything that fills in its run state on time.

## Definition and run are separate things

A definition is code — it holds live agents — and is never written to. A run is plain data (`SOPRunState`), owned by the engine, handed to each step one record at a time. One definition can drive any number of engines; a run outlives the process it started in and comes back through `SOPEngine(sop, state)`.

`PipelineProtocol` gains `.state` to say so for every pipeline. `AgentLike` extends it with the typed-task half a step needs of its executor; a SOP satisfies the base protocol but not `AgentLike`, since an outside schema has no place in a procedure whose steps each carry their own.

## Relationship to `GoalPipeline`

`GoalPipeline` is the one-step case: a goal handed in at run time, verified until it passes. A SOP is what you write when the milestones are known before the run is. They stay separate modules for now; the state `GoalPipeline` would need to be resumable is two fields, and is a follow-up.

## Example

`examples/sop/` handles a customer complaint by the book: establish the facts (judged by an auditor agent that reads the same records), propose redress (signed off by a person, without holding anything open), write the reply (checked against what was approved). Say `n` at the approval and give a reason — it goes back to the policy agent verbatim.

## Not in this PR

SOP-level events, forwarding `UserInterruptEvent` to the parked step, an `impossible` verdict, token/time budgets, and the service layer. Each is scoped separately.

## Testing

`tests/sop_engine_test.py` — 8 cases against scripted executors: ordering and handover, refusal → critique → retry, budget exhaustion, park in the executor, park in the verifier without redoing the work, verifier sees what the step was given, export/import across a fresh engine, mismatched state refused. `tests/pipeline_goal_test.py` unchanged and passing. `pre-commit run` clean.


DavdGao:
Pushed `94302866d` — stripped the service-only concepts that had leaked into the SDK model.

Each of these needed a scheduler, a workspace manager, channels or a message bus underneath, and none of those exist at this layer. Keeping them here meant the model described a service it could not reach:

| Removed | Actually lives in |
|---|---|
| `TriggerKind` (`schedule`, `event`) | `app/_manager/_scheduler/`, the channel / bus layer |
| `WorkspacePolicy`, `SOPRun.workspace_key` | `workspace_id`, `WorkspaceManager`, `IsolationPolicy` — all `app/` |
| `ChannelKind`, `NotifyTarget`, `SOPStep.notify`, the `Announce` action | `app/channel/_feishu`, `_discord` |
| `SOP.allow_agent_chat` | team tools / message bus |
| `SOP.fixed_sessions`, `SOP.persistent_workspace_key` | sessions and workspaces are both service-side |

A service that has those wraps this definition rather than pushing its own fields down into it.

Same reasoning killed the session vocabulary. `ExecutorMode` is now `reuse_state` / `reset_state` / `new_agent` — describing what happens to the agent's own state, which this layer does have, instead of naming a service-side session it does not. `StepRun` likewise records one opaque `executor_ref` in place of an agent id plus a session id.

What is left is the whole definition:

```
SOP        id, name, description, inputs, steps, created_at
SOPStep    id, title, instruction, blocked_by, executor, acceptance, max_attempts
```

DavdGao:
Pushed `6d4890035`. A step now holds its agent rather than a reference to one.

At this layer authoring a SOP is coding — you already have the agent in hand, so the whole indirection was service-shaped thinking:

```python
SOPStep(
    title="Write the code",
    agent=Agent(..., offloader=workspace),
    acceptance=Acceptance(criteria="Changes stay inside src/"),
)
```

`Executor`, `AgentSpec` and `ExecutorMode` are gone. The three modes turned out not to be policy at all — they were descriptions of how you construct:

| Was a mode | Is now just construction |
|---|---|
| `reuse_state` | run the same SOP twice with the same agent |
| `reset_state` | build a fresh agent per run |
| `new_agent` | same, and give each step its own |

Sharing one agent between steps likewise shares its context and workspace; giving each its own does not. None of that needs a field.

`AgentSpec` was the clearest tell: name, system prompt and model all as *strings*. That only exists so something can build an agent from a stored record — which is a service's job.

So a definition is no longer serialisable, and deliberately. A service that stores SOPs keeps its own records and builds one of these before running it, exactly the way `AgentData` becomes a live `Agent` in `_chat.py` today. `SOPRun` stays plain data, and that is the half worth persisting.

This costs `_core.py` nothing — it never read `executor`, only `id`, `blocked_by`, `title` and `max_attempts`. Same ten tests, still green.

```
SOP      name, steps, id, description, inputs, created_at
SOPStep  title, agent, acceptance, id, instruction, blocked_by, max_attempts
```

PR #2397, opened 2026-08-21
fix(model): use conservative fallback token estimate

### LINKED ISSUE #2385: [Bug]: Context compaction fails and crashes with 400 error due to token underestimation in count_tokens
### Prerequisites

### Background / Description

In multi-turn agent sessions involving non-ASCII text (e.g., Chinese), numerical data (financial records, SQL), or tool outputs (JSON), AgentScope's context compaction and length guardrails fail completely. This causes requests to exceed the backend model's context window and crash with **HTTP 400 (`context_length_exceeded` / `BadRequestError`)**.

### Root Cause
`ChatModelBase.count_tokens` uses a fixed heuristic (`len(utf8_bytes) / 4`). While this works for standard English prose, it systematically underestimates tokens by **2x to 3x** on non-ASCII text, numbers, and structured tool outputs.

Because of this underestimation:
1. The session is mistakenly estimated to be well below the `0.8 * context_window` compaction threshold, bypassing compaction.
2. The full context is sent to the LLM backend (e.g., vLLM / OpenAI-compatible endpoint), which rejects it with HTTP 400.
3. The internal `drop-oldest` retry mechanism inside memory compression also relies on `count_tokens`, so it fails to detect the overflow, terminating the entire agent task.

### Benchmark Data (Tested against Qwen on vLLM)
We compared `ChatModelBase.count_tokens` against the ground-truth `usage.input_tokens` returned by the model backend:

| Workload Scenario | Chars | UTF-8 Bytes | AgentScope Estimate (`bytes/4`) | Actual Usage (`input_tokens`) | Underestimation Ratio |
| :--- | :---: | :---: | :---: | :---: | :---: |
| Pure English query | 95 | 95 | 24 | 66 | **+175.0% (2.75x)** |
| Pure Chinese query | 44 | 124 | 31 | 74 | **+138.7% (2.39x)** |
| Financial & SQL Data | 377 | 473 | 118 | 315 | **+166.9% (2.67x)** |
| Tool Result (JSON) | 344 | 344 | 86 | 266 | **+209.3% (3.09x)** |
| Multi-turn Mix Context | 531 | 587 | 147 | 271 | **+84.4% (1.84x)** |

*(Note: Both streaming final chunk and non-streaming responses reliably return `ChatUsage(input_tokens, output_tokens)`).*

### Error Messages

```shell
openai.BadRequestError: Error code: 400 - {'error': {'message': "This model's maximum context length is 65536 tokens. However, your messages resulted in 65537 tokens. Please reduce the length of the messages.", 'type': 'invalid_request_error', 'param': 'messages', 'code': 400}}

  File "agentscope/agent/_react_agent.py", line 150, in reply
    response = await self._model(messages)
  File "agentscope/model/_openai_model.py", line 120, in __call__
    return await self._async_client.chat.completions.create(...)
openai.BadRequestError: Error code: 400 - {'message': 'This model maximum context length is 65536 tokens...'}
```

### Steps to Reproduce

1. Code:

```python
import asyncio
from agentscope.model import OpenAIChatModel, OpenAICredential
from agentscope.message import Msg, TextBlock

async def main():
    model = OpenAIChatModel(
        model_name="Qwen3.8-27B",
        credential=OpenAICredential(api_key="your_key", base_url="http://localhost:8000/v1"),
        stream=False,
    )
    
    # Financial SQL query with dense numbers
    text = (
        "SELECT order_id, sum(amount * (1 - discount)) as net_total "
        "FROM orders WHERE amount > 10000.50 AND customer_id = 88392019482 "
        "GROUP BY order_id HAVING net_total > 500000.00;"
    )
    msg = Msg(name="user", role="user", content=[TextBlock(type="text", text=text)])
    
    # 1. AgentScope Estimate
    estimated = await model.count_tokens([msg], None)
    # 2. Actual LLM Provider Usage
    resp = await model([msg])
    actual = resp.usage.input_tokens
    
    print(f"AgentScope count_tokens: {estimated}")
    print(f"Model Actual input_tokens: {actual}")
    print(f"Underestimated ratio: {actual / estimated:.2f}x")

if __name__ == "__main__":
    asyncio.run(main())
```
2. Run: python reproduce.py
3. See output: count_tokens reports ~118 tokens while actual LLM usage is ~315 tokens (underestimated by ~2.67x). In long multi-turn sessions, this error compounds until physical context limit is exceeded and the backend returns 400.

### Environment

```text
- OS: Linux (Ubuntu 22.04 / Docker)
- Python Version: 3.11.15
- AgentScope Version: 2.0.4 / 2.0.6
- Backend Model: Qwen3.8-27B (hosted via vLLM / OpenAI-compatible endpoint)

Oxygen56:
Fixes #2385

## AgentScope Version

2.0.6

## Description

`ChatModelBase.count_tokens` previously estimated all text as one token per
four UTF-8 bytes. That fallback can undercount token-dense numeric and
structured input, wide non-ASCII text, and hidden provider template overhead,
allowing context compression to be skipped before the provider rejects the
request.

This change adds two conservative fallback layers while leaving exact
provider-specific counters authoritative:

- class-aware estimates preserve the existing behavior for ordinary ASCII
  prose, while assigning local weight to dense numbers, structured punctuation,
  wide/full-width characters, and decodable textual data blocks;
- after a successful model call, the fallback records the provider's real
  `usage.input_tokens` for that agent session and estimates later turns as the
  larger of the heuristic or the latest usage plus the incremental input
  surface.

Calibration is isolated per agent session, works for streaming and
non-streaming responses, is bounded for shared model instances, and is cleared
after context replacement so an old large context cannot force repeated early
compression. There are no public API, dependency, or serialized-data changes.

## Verification

- The exact five payloads supplied in #2385 are covered. Their fallback
  estimates are 33/41/149/151/143 tokens versus pure-tokenizer counts of
  26/24/149/134/136, so none undercount and none exceed 2x.
- Old-vs-new provider-boundary regressions cover SQL, numeric, CJK, JSON, and
  mixed content: the legacy estimate sends oversized input and receives
  `context_length_exceeded`; the new fallback compresses first.
- Provider-usage regressions cover fixed template overhead, multi-turn deltas,
  session isolation, streaming final usage, authoritative provider counters,
  bounded session retention, and clearing after context compression.
- A fallback-driven overflow regression covers compression-request failure,
  drop-oldest retry, and successful provider delivery without the evicted
  context.
- 97 focused tests plus 15 subtests passed locally; formatting, type, lint, and
  package checks pass.

## Checklist

- [x] Related documentation has been updated (not applicable; no public API or
  setup behavior changed)

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -5,6 +5,7 @@ import collections
 import inspect
 import re
 import warnings
+from contextlib import contextmanager
 
 from asyncio import Queue
 from copy import deepcopy
@@ -17,6 +18,7 @@ from typing import (
     List,
     TYPE_CHECKING,
     Type,
+    Iterator,
 )
 
 import jsonschema
@@ -336,12 +338,13 @@ class Agent:
                 A final reply message.
         """
         final_msg: Msg | None = None
-        async for evt_or_msg in self._reply(
-            inputs=inputs,
-            structured_schema=structured_schema,
-        ):
-            if isinstance(evt_or_msg, Msg):
-                final_msg = evt_or_msg
+        with self._model_calibration_scope():
+            async for evt_or_msg in self._reply(
+                inputs=inputs,
+                structured_schema=structured_schema,
+            ):
+                if isinstance(evt_or_msg, Msg):
+                    final_msg = evt_or_msg
         if final_msg is None:
             raise RuntimeError("Agent did not produce a final message.")
         return final_msg
@@ -432,7 +435,10 @@ class Agent:
 
         # Count the current tokens
         kwargs = await self._prepare_model_input()
-        estimated_tokens = await self.model.count_tokens(**kwargs)
+        estimated_tokens = await self._count_model_tokens(
+            kwargs["messages"],
+            kwargs.get("tools"),
+        )
 
         # Skip if no compression is needed
         threshold = cfg.trigger_ratio * self.model.context_size
@@ -534,7 +540,7 @@ class Agent:
             },
         ]
         context_overflow = False
-        estimated_compression_tokens = await self.model.count_tokens(
+        estimated_compression_tokens = await self._count_model_tokens(
             messages,
             compression_tool_schema,
         )
@@ -550,7 +556,7 @@ class Agent:
         # Compress the messages
         res = None
         try:
-            res = await self.model.generate_structured_output(
+            res = await self._generate_structured_output(
                 messages=messages,
                 structured_model=cfg.summary_schema,
             )
@@ -575,7 +581,7 @@ class Agent:
                         ]
                     )
                     estimated_compression_tokens = (
-                        await self.model.count_tokens(
+                        await self._count_model_tokens(
                             messages,
                             compression_tool_schema,
                         )
@@ -589,7 +595,7 @@ class Agent:
                         break
 
                 try:
-                    res = await self.model.generate_structured_output(
+                    res = await self._generate_structured_output(
                         messages=messages,
                         structured_model=cfg.summary_schema,
                     )
@@ -656,6 +662,7 @@ class Agent:
             # Update the context and summary
             self.state.summary = new_summary
             self.state.context = msgs_to_reserve
+            self._clear_model_usage_anchor()
 
             logger.info(
                 "[AGENT %s]: The context compression finished.",
@@ -1388,7 +1395,10 @@ class Agent:
         if self.state.cur_iter == 0:
             # Count the current tokens
             kwargs = await self._prepare_model_input()
-            input_tokens = await self.model.count_tokens(**kwargs)
+            input_tokens = await self._count_model_tokens(
+                kwargs["messages"],
+                kwargs.get("tools"),
+            )
 
             trigger_tokens = int(
                 self.context_config.trigger_ratio * self.model.context_size,
@@ -1704,7 +1714,9 @@ class Agent:
             # Given event, required but not match
             extra_ids = set(
                 _.tool_call.id for _ in event.confirm_results
-            ) - set(awaiting_confirmations)
+            ) - set(
+                awaiting_confirmations,
+            )
             if extra_ids:
                 raise ValueError(
                     f"Received UserConfirmResultEvent with tool call ids "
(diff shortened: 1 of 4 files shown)


weike-zhang:
Thanks for this fix. One concern with a uniform `(bytes * 4 + 4) // 5` (~0.8 tokens/byte) is that it over-counts ASCII/English — the dominant content — by roughly 3×. Example: 135 bytes of English prose measures ~31 tokens with `cl100k_base` (~0.23 tokens/byte), but this formula returns ~108, so compaction would trigger ~3× early and waste a large share of the context window.

A CJK-aware estimate avoids that regression: count CJK/full-width characters as 1 token each (they occupy 3 UTF-8 bytes but tokenize at ~1.2 tokens each, vs the 0.75 that `bytes/4` implies) and keep `bytes/4` for the remaining bytes. That fixes the non-ASCII undercount from #2385 without changing ASCII behavior. Happy to share the calibrated numbers if useful.

qiboyu301-crypto:
@Oxygen56 Thanks for addressing #2385. I have one framework-level concern about applying the calibration through ChatModelBase.count_tokens().
In this PR, the calibrated result is used not only for full-request threshold checks, but also for AgentScope’s internal partial-input measurements, including _split_context_for_compression, boundary-block checks, _split_tool_result_for_compression, and drop-oldest retries.
Those calls may pass only a suffix or a single tool result, while the calibration anchor was created from a previous full request. For example, if the anchor is (anchor_estimate=100K, usage_tokens=120K), counting a 5K suffix returns:
max(5K, 120K + max(0, 5K - 100K)) = 120K
With reserve_ratio=0.1, AgentScope may then treat the suffix as exceeding the reserved budget and move the newest context into msgs_to_compress. This could cause over-compression and potentially reintroduce loss of the latest user message.
More generally, this changes count_tokens(messages, tools) from a deterministic estimate of the supplied input into a session-history-dependent lower-bound function. That may affect callers that rely on additive or monotonic estimates for arbitrary slices.
Could calibration be restricted to full-request measurements, while internal suffix/boundary/tool-result calculations use a pure additive heuristic? Alternatively, the anchor could be applied only when the current input is demonstrably compatible with the anchored request surface.
It would be useful to add regression tests covering:
- suffix and boundary token counts after a full-request anchor;
- single ToolResultBlock counting;
- drop-oldest retry behavior;
- reserve_ratio=0.1;
- preservation of the newest user message when a previous anchor exists.

qiboyu301-crypto:
We have encountered a similar issue in a production-oriented AgentScope integration and would be happy to help investigate this further. If the concern is confirmed, we can help prepare the regression tests or a follow-up patch that separates full-request calibration from partial-input counting.

PR #2406, opened 2026-08-24
fix(gemini): exclude tool-execution tokens from output_tokens

### LINKED ISSUE #2405: [Bug]: GeminiChatModel reports inflated output_tokens when a response includes tool-execution usage
### Prerequisites

### Background / Description

When talking to a Gemini model through `GeminiChatModel`, the `output_tokens` value on
the returned `ChatResponse.usage` can be noticeably larger than the number of tokens
Gemini actually generated. This shows up whenever the underlying response's
`usage_metadata` carries a nonzero `tool_use_prompt_token_count` (the token cost of
results from Gemini's own server-side tools, such as built-in Google Search grounding
or code execution, fed back into the model). I expected `output_tokens` to track only
the tokens the model produced; instead it also picks up that tool-execution cost.

Anything downstream that uses `output_tokens` for cost or budget accounting, including
agentscope's own `ReplyBudgetControlMiddleware`, ends up counting more output tokens
than the model actually generated for these calls.

### Error Messages

None. No exception is raised, the reported usage numbers are just wrong.

### Steps to Reproduce

This reproduces the discrepancy directly against an unmodified checkout, using a
`GenerateContentResponse` shaped the way the `google-genai` SDK documents (the four
usage fields sum to `total_token_count`), so no live Gemini API call is needed.

```python
from datetime import datetime
from google.genai import types as gt
from agentscope.model._gemini._model import GeminiChatModel

usage_metadata = gt.GenerateContentResponseUsageMetadata(
    prompt_token_count=500,
    candidates_token_count=120,      # tokens the model actually generated
    tool_use_prompt_token_count=300, # tool-execution results fed back in as input
    thoughts_token_count=0,
    total_token_count=920,
)
response = gt.GenerateContentResponse(
    candidates=[gt.Candidate(content=gt.Content(parts=[gt.Part(text="hello")], role="model"))],
    usage_metadata=usage_metadata,
)

model = GeminiChatModel.__new__(GeminiChatModel)
chat_response = model._parse_completion_response(datetime.now(), response)
print(chat_response.usage.output_tokens)  # 420
```

Run: `python repro.py`

Observed: `output_tokens` is `420`.
Expected: something close to `120`, the model's own `candidates_token_count`, since
that is the only field the SDK documents as "the total number of tokens in the
generated candidates."

### Environment

- AgentScope Version: HEAD `da00849f` (unreleased, checked out directly)
- google-genai: 2.19.0
- Python Version: 3.11
- OS: Linux (Docker, `python:3.11-slim`)

iluv7:
## AgentScope Version

2.0.6

## Description

Gemini reports prompt, candidate, server-side tool-use, and thinking tokens separately in `usage_metadata`. `GeminiChatModel` previously derived `output_tokens` as `total_token_count - prompt_token_count`, which incorrectly classified `tool_use_prompt_token_count` as model output. These tokens come from server-side tool results (for example, Google Search grounding or code execution) that are fed back to the model as input.

This caused incorrect input/output usage reporting and distorted weighted cost or budget accounting such as `ReplyBudgetControlMiddleware`.

Fixes #2405

**Changes:**

- Classify tool-use prompt tokens as input: `input_tokens = prompt_token_count + tool_use_prompt_token_count`
- Count only model-generated candidate and thinking tokens as output: `output_tokens = candidates_token_count + thoughts_token_count`
- When `candidates_token_count` is unavailable, fall back to `total_token_count - input_tokens`, so tool-use tokens are still excluded from output
- Add regression coverage for non-streaming, streaming, thinking-token, and missing-candidate-count paths in `tests/model_gemini_test.py`

For example, given `prompt=500`, `candidates=120`, `tool_use=300`, `thoughts=10`, and `total=930`, the model now reports:

```text
input_tokens = 800
output_tokens = 130
```

The two values preserve the API-reported total while correctly classifying tool-use results as input rather than output.

**How to test:**

```bash
pytest tests/model_gemini_test.py
```

## Checklist

- [x] An issue has been created for this PR (#2405)
- [x] Related documentation has been updated (N/A — no doc changes needed)

### DIFF
--- a/tests/model_gemini_test.py
+++ b/tests/model_gemini_test.py
@@ -6,6 +6,7 @@ Tests cover both non-streaming and streaming modes.
 Gemini uses google.genai client with async iterator streaming.
 """
 import json
+from types import SimpleNamespace
 from typing import Any
 import unittest
 from unittest import IsolatedAsyncioTestCase
@@ -68,9 +69,14 @@ def _mock_completion(
     resp.candidates = [MagicMock()]
     resp.candidates[0].content = MagicMock()
     resp.candidates[0].content.parts = parts
-    resp.usage_metadata = MagicMock()
-    resp.usage_metadata.prompt_token_count = 10
-    resp.usage_metadata.candidates_token_count = 5
+    resp.usage_metadata = SimpleNamespace(
+        prompt_token_count=10,
+        candidates_token_count=5,
+        tool_use_prompt_token_count=0,
+        thoughts_token_count=0,
+        total_token_count=15,
+        cached_content_token_count=0,
+    )
     return resp
 
 
@@ -84,9 +90,14 @@ def _make_stream_chunk(
     chunk.candidates = [MagicMock()]
     chunk.candidates[0].content = MagicMock()
     chunk.candidates[0].content.parts = parts
-    chunk.usage_metadata = MagicMock()
-    chunk.usage_metadata.prompt_token_count = 10
-    chunk.usage_metadata.candidates_token_count = 5
+    chunk.usage_metadata = SimpleNamespace(
+        prompt_token_count=10,
+        candidates_token_count=5,
+        tool_use_prompt_token_count=0,
+        thoughts_token_count=0,
+        total_token_count=15,
+        cached_content_token_count=0,
+    )
     return chunk
 
 
@@ -243,6 +254,32 @@ class TestGeminiNonStream(IsolatedAsyncioTestCase):
             ),
         )
 
+    async def test_usage_excludes_tool_use_tokens(self) -> None:
+        """Usage counts only model-generated tokens, not tool-use tokens."""
+        parts = [_make_part(text="hello")]
+        resp = _mock_completion(parts)
+        # Mirror the SDK's documented invariant: total = prompt +
+        # candidates + tool_use_prompt + thoughts.
+        resp.usage_metadata = SimpleNamespace(
+            prompt_token_count=500,
+            candidates_token_count=120,
+            tool_use_prompt_token_count=300,
+            thoughts_token_count=0,
+            total_token_count=920,
+            cached_content_token_count=50,
+        )
+        self.mock_client.aio.models.generate_content = AsyncMock(
+            return_value=resp,
+        )
+
+        result = await self.model([])
+
+        self.assertEqual(result.usage.input_tokens, 500)
+        # Not total - prompt (420), which wrongly includes the 300
+        # tool-execution tokens.
+        self.assertEqual(result.usage.output_tokens, 120)
+        self.assertEqual(result.usage.cache_input_tokens, 50)
+
 
 # ---------------------------------------------------------------------------
 # Streaming tests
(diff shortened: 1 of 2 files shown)


DavdGao:
@iluv7 Is this pr ready for review?

iluv7:
> @iluv7 Is this pr ready for review?

Yep.

tests/model_gemini_test.py
```diff
@@ -243,6 +254,53 @@ async def test_thinking_response(self) -> None:
...
+            total_token_count=930,
+            cached_content_token_count=50,
+        )
+        self.mock_client.aio.models.generate_content = AsyncMock(
+            return_value=resp,
+        )
+
+        result = await self.model([])
+
+        self.assertEqual(result.usage.input_tokens, 800)
```
qbc2016: Please assert the complete usage structure instead of checking selected fields individually. This verifies all normalized fields and would catch unexpected values such as `cache_input_tokens=None`, without introducing a direct dependency on `ChatUsage` in this provider test.

For example:

```python
self.assertEqual(
    dict(result.usage),
    {
        "input_tokens": 800,
        "output_tokens": 130,
        "time": result.usage.time,
        "cache_creation_input_tokens": 0,
        "cache_input_tokens": 50,
        "type": "chat",
        "metadata": None,
    },
)
```

The separate `input_tokens + output_tokens == 930` assertion is redundant
because both values are covered by the complete structure assertion.

tests/model_gemini_test.py
```diff
@@ -243,6 +254,53 @@ async def test_thinking_response(self) -> None:
...
+            total_token_count=810,
+            cached_content_token_count=0,
+        )
+        self.mock_client.aio.models.generate_content = AsyncMock(
+            return_value=resp,
+        )
+
+        result = await self.model([])
+
+        self.assertEqual(result.usage.input_tokens, 800)
```
qbc2016: Please assert the complete usage structure in this fallback test as well:

```
self.assertEqual(
    dict(result.usage),
    {
        "input_tokens": 800,
        "output_tokens": 10,
        "time": result.usage.time,
        "cache_creation_input_tokens": 0,
        "cache_input_tokens": 0,
        "type": "chat",
        "metadata": None,
    },
)
```

This ensures that the fallback path preserves all expected usage defaults.

tests/model_gemini_test.py
```diff
@@ -309,6 +367,31 @@ async def test_stream_text(self) -> None:
...
+        self.mock_client.aio.models.generate_content_stream = AsyncMock(
+            return_value=_MockAsyncStream([chunk]),
+        )
+
+        gen = await self.model([])
+        responses = [r async for r in gen]
+
+        # The delta and the accumulated final response both carry the
+        # usage of the last chunk.
+        self.assertEqual(responses[0].usage.input_tokens, 800)
```
qbc2016: Please assert the complete usage structure for the streamed delta, then compare
the complete final usage against it:

```python
self.assertEqual(
    dict(responses[0].usage),
    {
        "input_tokens": 800,
        "output_tokens": 130,
        "time": responses[0].usage.time,
        "cache_creation_input_tokens": 0,
        "cache_input_tokens": 50,
        "type": "chat",
        "metadata": None,
    },
)
self.assertEqual(responses[-1].usage, responses[0].usage)
```

This covers every usage field while also verifying that the accumulated final response carries the same usage snapshot.

qbc2016 (commented):

tests/model_gemini_test.py
iluv7: Updated to assert the complete normalized usage structure and removed the redundant token-sum assertion. Addressed in 181b9131.

tests/model_gemini_test.py
iluv7: Updated the fallback test to assert the complete usage structure, including all default fields. Addressed in 181b9131.

tests/model_gemini_test.py
iluv7: Updated the streaming test to assert the complete delta usage and verify that the final response carries the same usage snapshot. Addressed in 181b9131.

qbc2016 (approved):
LGTM.

iluv7:
Hi, @qbc2016 
While reviewing usage accounting across the other providers, I noticed a similar issue in XAIChatModel: output_tokens currently includes only completion_tokens, while the xAI SDK reports reasoning_tokens separately. This causes reasoning tokens to be omitted in both streaming and non-streaming paths. I’ll submit a PR to fix it later.

iluv7:
> Hi, @qbc2016 While reviewing usage accounting across the other providers, I noticed a similar issue in XAIChatModel: output_tokens currently includes only completion_tokens, while the xAI SDK reports reasoning_tokens separately. This causes reasoning tokens to be omitted in both streaming and non-streaming paths. I’ll submit a PR to fix it later.

Here: https://github.com/agentscope-ai/agentscope/pull/2461

PR #2417, opened 2026-08-25
fix(web_ui): correct thinking/reply elapsed time across UTC offsets

lwylab:
## Summary

Fixes incorrect thinking/reply elapsed badges (e.g. flashing **8h** while streaming in UTC+8, then snapping to **1s** after THINKING_BLOCK_END) when the Agent Service runs in UTC (Docker) and the browser is in a non-UTC timezone.

## Root cause

1. **Backend**: naive datetime.now().isoformat() timestamps without timezone (EventBase.created_at, some Msg defaults).
2. **Web UI**: while streaming, elapsed time mixed Date.now() (browser local) with 
ew Date(naive ISO) (parsed as local), producing a UTC-offset flash.

## Changes

- Python: default timestamp factory now emits **UTC ISO 8601 with Z suffix**; EventBase and Msg use _generate_timestamp.
- Web UI (examples/web_ui): add useElapsedSeconds hook with parseApiTimestamp(); use it in ThinkingBlockView and message footer badges.
- Tests: 	ests/timestamp_factory_test.py.

## Test plan

- [x] pytest tests/timestamp_factory_test.py
- [ ] Manual: Docker/UTC backend + UTC+8 browser, verify thinking header shows ~1s while streaming (not ~8h)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/web_ui/frontend/src/components/chat/ASMessageBubble.tsx
+++ b/examples/web_ui/frontend/src/components/chat/ASMessageBubble.tsx
@@ -22,6 +22,7 @@ import {
 import * as mime from 'mime-types';
 import { useEffect, useRef, useState } from 'react';
 
+import { useElapsedSeconds } from '@/hooks/useElapsedSeconds';
 import { renderToolCall } from './tool-renderers';
 import { countDiffStats, DiffStats, getResultDiff } from './tool-renderers/_shared';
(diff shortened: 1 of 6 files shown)


lwylab:
Friendly bump — this PR has been open for ~3 weeks with no review yet.

Would a maintainer have a chance to take a look when convenient?
cc @DavdGao @qbc2016

PR #2426, opened 2026-08-26
fix(model): preserve encrypted reasoning items in OpenAI Responses history

### LINKED ISSUE #2420: [Bug]: OpenAI Responses multi-turn fails with 400 "Referenced reasoning item ... not found or has expired" on stateless upstreams — formatter drops encrypted_content
### Prerequisites

### Background / Description

Multi-turn conversations with a Responses-API reasoning model (e.g. `muse-spark-1.2-contributor` via OpenCode Zen/Go, `https://opencode.ai/zen/go/v1/responses`) fail on the 2nd turn with:

`400 invalid_request_error: Referenced reasoning item 'rs_...' was not found or has expired.`

The first turn works; every subsequent turn in the same session fails, making the session unusable.

Root cause: `OpenAIResponseFormatter` replays a `reasoning` item in multi-turn history as `{"type":"reasoning","id":...,"summary":[...],"content":[]}`. At parse time (`_parse_completion_response` / `_parse_stream_response` in `agentscope/model/_openai_response/_model.py`) only `reasoning_item_id` + summary text are kept; the **complete raw item is discarded**, in particular `encrypted_content`. Stateless upstreams (OpenCode Zen/Go, and Meta's own Responses API) validate a replayed reasoning item against its full content; a bare id + empty content fails their validation.

### Error Messages

```
openai.BadRequestError: Error code: 400 - {'model': 'muse-spark-1.2-contributor',
'error': {'type': 'invalid_request_error',
'message': "Error from provider (Console Go): Upstream request failed: [invalid_request_error]
Referenced reasoning item 'rs_6a8d...:rs_01a0...' was not found or has expired."}}
```

### Steps to Reproduce

1. Configure an OpenAI-compatible Responses provider pointing at a stateless gateway, e.g.:
   ```python
   from agentscope.model import OpenAIResponseModel
   model = OpenAIResponseModel(
       credential=OpenAICredential(id="t", api_key="...", base_url="https://opencode.ai/zen/go/v1"),
       model="muse-spark-1.2-contributor",
       parameters=OpenAIResponseModel.Parameters(max_tokens=800, thinking_enable=True, reasoning_effort="low"),
       stream=True,
   )
   ```
2. Turn 1: `await model([Msg("user", "hi")])` → OK (response contains a reasoning item with `encrypted_content`).
3. Turn 2: replay the assistant message (including the reasoning item) plus a new user message → 400 `Referenced reasoning item ... not found or has expired`.

Direct API evidence (curl, no SDK):
- Replaying the **complete** reasoning item (including `encrypted_content`) → multi-turn works (3 consecutive turns OK).
- Replaying only `{"type":"reasoning","id":...,"summary":[],"content":[]}` (current formatter behavior) → 400.
- `previous_response_id` is not supported by this upstream (`referenced response not found or expired` — the gateway is stateless).
- The upstream does not offer plaintext summaries: `include: ["reasoning.summary"]` is rejected (only `reasoning.encrypted_content` is available).

### Environment

- AgentScope Version: 2.0.4.post1
- Python Version: 3.11
- OS: Ubuntu 24.04

### Suggested fix (verified locally)

1. In `_parse_completion_response` / `_parse_stream_response`, attach the full raw item dict (`item.model_dump()`) to the `ThinkingBlock` (e.g. extra field `reasoning_item_raw`; `ThinkingBlock` already allows extra fields).
2. In `_openai_response_formatter.py`, when `reasoning_item_raw` is present, append it verbatim to the request input instead of reconstructing `{"type":"reasoning","id":...,"summary":...,"content":[]}`. Fall back to current behavior when raw data is unavailable.

Patch applied locally (agentscope 2.0.4.post1 + QwenPaw 2.1.0) and verified: 1st turn preserves `encrypted_content` (2791 chars), 2nd turn replays it and no longer returns 400; conversation continues normally.

Related: #2071 (preserve OpenAI Responses reasoning history replay) — this issue is about the raw item payload being dropped, which the current code still does.

---

qbc2016:
## AgentScope Version

2.0.7

## Description

Preserve complete OpenAI Responses reasoning items, including `encrypted_content`, across streaming and non-streaming message round-trips. The formatter now replays validated raw reasoning items and falls back to the existing ID-based format for backward compatibility.

## Checklist


DavdGao:
The direction is right — losing `encrypted_content` breaks reasoning continuity under `store=false`. Two things before this lands:

**1. The feature is unreachable by default.** `OpenAIResponseModel` never sets `include=["reasoning.encrypted_content"]`, and OpenAI only returns `encrypted_content` when it is explicitly requested. With the default config `_dump_reasoning_item` therefore always captures an item without it, and nothing changes. Users have to know to push `include` + `store=False` through `generate_kwargs`. Please expose this as an option on the model, or at least document it.

**2. Streaming keeps only the last reasoning item.** In `_parse_stream_response` every reasoning item in `resp.output` is appended under the same `block_id=thinking_id`. `append_thinking` matches the existing block and `setattr`s over it, so a response carrying several reasoning items (common when a reasoning model makes multiple tool calls in one turn) retains only the last item's `reasoning_item_raw` — the earlier encrypted payloads are dropped. Pre-existing for `reasoning_item_id`, but it defeats the purpose of this PR. `_parse_completion_response` is fine, it builds one block per item.

**Minor:** `_is_valid_reasoning_item_raw` is ~25 lines with a single call site, validating data we produced ourselves in `_dump_reasoning_item`; an id match plus a dict check would cover the hand-constructed case. Same for the `getattr(item, "model_dump", None)` guard and the `reasoning_metadata` dict assembly that exists only to avoid storing a `None`. Also, the `typing` import was reflowed to multi-line without adding a symbol — pure diff noise.

qbc2016:
Thanks for pointing this out. According to the current [OpenAI documentation](https://developers.openai.com/api/docs/guides/reasoning#preserve-reasoning-without-stored-responses), reasoning items include `encrypted_content` by default when `store=false` or when ZDR is enabled. The legacy `include=["reasoning.encrypted_content"]` option is still accepted for compatibility but is no longer required.

The provider reported in #2420 also returns `encrypted_content` without an explicit `include`. For legacy Responses-compatible providers that still require it, users can already pass both `store` and `include` through `generate_kwargs`.

Therefore, I have not added another model option for this behavior.

DavdGao (approved):

PR #2432, opened 2026-08-26
fix(agent): emit final-only streaming text

### LINKED ISSUE #2430: [Bug]: reply_stream() emits no text events for a final-only streaming ChatResponse
### Background / Description

`Agent.reply_stream()` normally exposes model text through `TextBlockStartEvent`, `TextBlockDeltaEvent`, and `TextBlockEndEvent`.

`ChatResponse.is_last=True` is documented to mean that `content` is the complete response. A streaming `ChatModelBase` implementation may therefore legally produce a single final `ChatResponse` containing all text.

In that final-only case, the text is persisted into `agent.state.context`, but it is never streamed to the default `reply_stream()` caller.

### Actual Behavior

For a streaming model that yields only:

```python
ChatResponse(
    content=[TextBlock(text="hello world")],
    is_last=True,
)
```

Observed events include:

```text
ModelCallEndEvent
ReplyEndEvent
```

No text events are emitted:

```text
TextBlockStartEvent
TextBlockDeltaEvent
TextBlockEndEvent
```

Yet `agent.state.context` contains `"hello world"`.

### Expected Behavior

For a final-only response containing one `TextBlock("hello world")`, `reply_stream()` should expose equivalent text events so callers do not lose the model output.

### Steps to Reproduce

This uses a deterministic local fake model. It does not require an API key, network access, a model provider, or GPU.

```python
import asyncio
from typing import Any

from pydantic import BaseModel

from agentscope.credential import CredentialBase
from agentscope.message import Msg, TextBlock, UserMsg
from agentscope.model import ChatModelBase, ChatResponse
from agentscope.tool import ToolChoice


class DummyCredential(CredentialBase):
    @classmethod
    def get_chat_model_class(cls):
        return FakeModel


class FakeModel(ChatModelBase):
    class Parameters(BaseModel):
        pass

    def __init__(self):
        super().__init__(
            credential=DummyCredential(),
            model="fake",
            parameters=self.Parameters(),
            stream=True,
            max_retries=0,
        )
        self.formatter = OpenAIChatFormatter()

    async def _call_api(
        self,
        model_name: str,
        messages: list[Msg],
        tools: list[dict] | None = None,
        tool_choice: ToolChoice | None = None,
        **kwargs: Any,
    ):
        async def gen():
            yield ChatResponse(
                content=[TextBlock(text="hello world", id="t1")],
                is_last=True,
                id="final",
            )

        return gen()


async def main():
    agent = Agent(
        name="assistant",
        system_prompt="You are helpful.",
        model=FakeModel(),
    )

    events = []
    async for event in agent.reply_stream(UserMsg(name="user", content="hi")):
        events.append(type(event).__name__)

    print(events)
    print([b.text for b in agent.state.context[-1].content if b.type == "text"])


asyncio.run(main())
```

Output on current `main`:

```text
['ReplyStartEvent', 'HintBlockEvent', 'ModelCallStartEvent', 'ModelCallEndEvent', 'ReplyEndEvent']
['hello world']
```

### Root Cause Analysis

`Agent._reasoning_impl()` stores an `is_last` `ChatResponse` as the completed response but does not pass that final chunk through `_convert_chat_response_to_event()`.

This assumes that equivalent content was already emitted by earlier partial chunks. That assumption is false for a valid final-only streaming response.

### Relevant Code Paths

- `src/agentscope/agent/_agent.py`
- `src/agentscope/model/_model_response.py`

### Environment

- AgentScope main: `142c341`
- Python version: `3.12.13`
- OS: WSL2 / Ubuntu 22.04.4 LTS

### Additional Context

The `ChatResponse.is_last` contract documents that its `content` is the complete response when true.

A fix must avoid duplicating text for providers that emit partial deltas and then return a final aggregate response.

I'd be happy to submit a focused fix with a regression test if this behavior is considered unintended.

Ricky-7-Yan:
## AgentScope Version

2.0.7 (`main` at `142c3411`)

## Description

Fixes #2430.

A streaming `ChatModelBase` may legally return all text only in its final `ChatResponse`. `Agent._reasoning_impl()` previously stored that text in context but skipped event conversion for the final chunk, so `reply_stream()` callers received no text events.

This change:

- tracks whether the stream emitted any non-final content;
- emits block events for final-only text, thinking, and data content;
- leaves aggregate final chunks untouched after normal deltas, avoiding duplicate output;
- retains the existing final-only tool-call acting contract.

## Testing

- `python -m pytest tests/agent_basic_test.py tests/agent_injection_test.py tests/agent_interrupt_test.py tests/agent_structured_output_test.py -q` — 43 passed
- `pre-commit run --files src/agentscope/agent/_agent.py tests/agent_basic_test.py` — passed
- Full `pytest tests -q` was also attempted in a partial Windows dev environment: 1765 passed, 319 skipped, and 82 optional/integration tests failed because the complete `dev` extra could not be installed after the Windows `ripgrep` source build stalled. None of the Agent test modules failed.

No public API was added or changed, so no companion documentation update is required.

## Checklist

- [x] Related documentation has been updated (not applicable: no public API or documentation contract change)

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -1539,6 +1539,7 @@ class Agent:
             "data": [],
         }
         completed_response: ChatResponse | None = None
+        has_streamed_content = False
 
         # Check if res is an async generator (streaming response)
         if inspect.isasyncgen(res):
@@ -1546,8 +1547,31 @@ class Agent:
                 # Save the last chunk with completed response
                 if chunk.is_last:
                     completed_response = chunk
+                    if not has_streamed_content:
+                        # Surface final-only text without replaying aggregate
+                        # tool, thinking, or data content as duplicate deltas.
+                        text_blocks = [
+                            block
+                            for block in chunk.content
+                            if isinstance(block, TextBlock)
+                        ]
+                        if text_blocks:
+                            event_stream = (
+                                self._convert_chat_response_to_event(
+                                    block_ids,
+                                    ChatResponse(
+                                        content=text_blocks,
+                                        is_last=False,
+                                    ),
+                                )
+                            )
+                            async for evt in event_stream:
+                                yield evt
 
                 else:
+                    has_streamed_content = has_streamed_content or bool(
+                        chunk.content,
+                    )
                     # Convert the chunk into events
                     async for evt in self._convert_chat_response_to_event(
                         block_ids,
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -1548,19 +1548,21 @@ class Agent:
                 if chunk.is_last:
                     completed_response = chunk
                     if not has_streamed_content:
-                        # Surface final-only text without replaying aggregate
-                        # tool, thinking, or data content as duplicate deltas.
-                        text_blocks = [
+                        # Surface final-only user-visible content without
+                        # replaying an aggregate response after streamed
+                        # chunks. Tool calls keep their existing acting-path
+                        # event contract.
+                        event_blocks = [
                             block
                             for block in chunk.content
-                            if isinstance(block, TextBlock)
+                            if not isinstance(block, ToolCallBlock)
                         ]
-                        if text_blocks:
+                        if event_blocks:
                             event_stream = (
                                 self._convert_chat_response_to_event(
                                     block_ids,
                                     ChatResponse(
-                                        content=text_blocks,
+                                        content=event_blocks,
                                         is_last=False,
                                     ),
                                 )
@@ -1569,9 +1571,7 @@ class Agent:
                                 yield evt
 
                 else:
-                    has_streamed_content = has_streamed_content or bool(
-                        chunk.content,
-                    )
+                    has_streamed_content = True
                     # Convert the chunk into events
                     async for evt in self._convert_chat_response_to_event(
                         block_ids,
--- a/tests/agent_basic_test.py
+++ b/tests/agent_basic_test.py
@@ -18,6 +18,8 @@ from agentscope.permission import (
     PermissionContext,
 )
 from agentscope.message import (
+    Base64Source,
+    DataBlock,
     TextBlock,
     ThinkingBlock,
     ToolCallBlock,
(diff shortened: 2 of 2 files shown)


src/agentscope/agent/_agent.py
```diff
@@ -1539,15 +1539,39 @@ async def _reasoning_impl(
...
         # Check if res is an async generator (streaming response)
         if inspect.isasyncgen(res):
             async for chunk in res:
                 # Save the last chunk with completed response
                 if chunk.is_last:
                     completed_response = chunk
+                    if not has_streamed_content:
+                        # Surface final-only text without replaying aggregate
+                        # tool, thinking, or data content as duplicate deltas.
+                        text_blocks = [
```
DavdGao: Why are the thinking and data blocks being filtered out here?

src/agentscope/agent/_agent.py
```diff
@@ -1539,15 +1539,39 @@ async def _reasoning_impl(
...
+                                        content=text_blocks,
+                                        is_last=False,
+                                    ),
+                                )
+                            )
+                            async for evt in event_stream:
+                                yield evt
 
                 else:
+                    has_streamed_content = has_streamed_content or bool(
```
DavdGao: Suggested to be `has_streamed_content = Ture`

DavdGao (changes requested):
Plz see inline comments

src/agentscope/agent/_agent.py
Ricky-7-Yan: Agreed. I removed the text-only filter in 287509e. Final-only responses now convert thinking and data blocks as well; only tool-call blocks retain the existing acting-path event contract. The regression test now covers final-only thinking, text, and base64 data events.

src/agentscope/agent/_agent.py
Ricky-7-Yan: Updated in 287509e: every non-final chunk now sets has_streamed_content = True, independent of whether that chunk has content.

DavdGao:
Thanks for the PR. Not taking this one — reason in https://github.com/agentscope-ai/agentscope/issues/2430#issuecomment-5724944727

PR #2437, opened 2026-08-27
fix(gemini): sanitize nullable type arrays

### LINKED ISSUE #2436: [Bug]: Gemini tool schema sanitizer leaves nullable JSON Schema type arrays unsupported
### Describe the bug

`_sanitize_schema_for_gemini` handles the common `anyOf: [X, {"type": "null"}]` shape, but it does not sanitize the equivalent JSON Schema type-array form, for example:

```json
{"type": ["string", "null"]}
```

That shape is produced by real tool-schema pipelines such as Pydantic's `model_json_schema(union_format="primitive_type_array")` and OpenAPI 3.1 nullable schemas. MCP tools also expose `inputSchema` as JSON Schema, so this can flow into Gemini function declarations through AgentScope's tool formatting path.

### Current behavior

For a tool parameter schema like:

```json
{
  "type": "object",
  "properties": {
    "name": {
      "type": ["string", "null"],
      "description": "Optional name"
    }
  }
}
```

AgentScope currently keeps the `type` array unchanged. In Gemini function declarations, `type` is a single enum value rather than a repeated field, so this can be rejected before the model call runs.

There is a second related edge case: `anyOf` null branches are only removed when the branch is exactly `{"type": "null"}`. A valid annotated null schema such as `{"type": "null", "description": "No value"}` is not recognized as null and is recursively rewritten to `{"type": "object"}`.

### Expected behavior

The Gemini sanitizer should:

- inline `type: ["string", "null"]` to `type: "string"`
- convert multi-type nullable arrays such as `type: ["string", "integer", "null"]` into non-null `anyOf` alternatives
- treat annotated null schemas inside `anyOf` as null branches

### References

- JSON Schema allows `type` to be an array and includes `null` as a type: https://json-schema.org/understanding-json-schema/reference/type
- Pydantic supports `union_format="primitive_type_array"`: https://docs.pydantic.dev/latest/api/base_model/#pydantic.main.BaseModel.model_json_schema
- MCP tool `inputSchema` is JSON Schema: https://modelcontextprotocol.io/specification/2025-06-18/server/tools

helloJamest:
## Summary

- sanitize nullable JSON Schema `type` arrays before passing tool schemas to Gemini function declarations
- inline single non-null nullable arrays such as `type: ["string", "null"]`
- convert multi-type nullable arrays to non-null `anyOf` alternatives only when no existing `anyOf` constraints would be overwritten
- fail explicitly for multi-type nullable arrays combined with existing `anyOf`, preserving the tool contract instead of silently weakening it
- treat annotated `null` schemas in `anyOf` as null branches instead of rewriting them to `object`

Closes #2436

## Tests

- `PYTHONPATH=src python -m pytest tests/model_gemini_test.py -q`
- `pre-commit run --config .pre-commit-config.yaml --files src/agentscope/model/_gemini/_model.py tests/model_gemini_test.py`

### DIFF
--- a/src/agentscope/model/_gemini/_model.py
+++ b/src/agentscope/model/_gemini/_model.py
@@ -22,6 +22,14 @@ else:
     GenerateContentResponse = Any
 
 
+def _is_null_schema(schema: Any) -> bool:
+    """Return whether a schema represents only JSON null."""
+    return isinstance(schema, dict) and schema.get("type") in (
+        "null",
+        ["null"],
+    )
+
+
 def _sanitize_schema_for_gemini(schema: Any) -> Any:
     """Sanitize a JSON schema to be compatible with the Gemini API.
 
@@ -32,6 +40,10 @@ def _sanitize_schema_for_gemini(schema: Any) -> Any:
     - ``additionalProperties``: removed entirely.
     - ``const``: converted to an equivalent single-value ``enum``,
       since Gemini's ``Schema`` model does not support ``const``.
+    - ``type`` arrays containing ``"null"``: simplified to the single
+      non-null type or converted to ``anyOf`` for multiple non-null
+      types, matching nullable JSON Schema emitted by Pydantic and
+      OpenAPI 3.1.
     - ``anyOf`` containing a ``{"type": "null"}`` entry: simplified to
       the single non-null type. If there is exactly one non-null
       alternative it is inlined directly; otherwise the ``anyOf`` is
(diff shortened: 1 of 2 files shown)


src/agentscope/model/_gemini/_model.py
```diff
@@ -63,8 +75,17 @@ def _sanitize_schema_for_gemini(schema: Any) -> Any:
...
-    if schema.get("type") == "null":
+    if _is_null_schema(schema):
         schema["type"] = "object"
+    elif isinstance(schema.get("type"), list):
+        non_null_types = [v for v in schema["type"] if v != "null"]
+        if len(non_null_types) == 1:
+            schema["type"] = non_null_types[0]
+        elif non_null_types:
+            schema.pop("type")
+            schema["anyOf"] = [{"type": type_} for type_ in non_null_types]
```
qbc2016, line 92: **Blocking: preserve an existing `anyOf`**

JSON Schema allows `type` and `anyOf` to appear together, in which case their constraints are conjunctive. This assignment unconditionally replaces an existing `anyOf`. For example, `{"type": ["string", "integer", "null"], "anyOf": [{"enum": ["auto"]}, {"enum": [0]}, {"type": "null"}]}` is currently restricted to `"auto"`, `0`, or `null`, but this sanitizer turns it into any string or integer and silently drops the original constraints.

That is worse than the current explicit validation failure because the model can now emit arguments outside the tool contract. Could we either compose the generated type alternatives with the existing `anyOf`, or fail explicitly when the two constructs cannot be represented safely in Gemini's supported schema subset? Please also add a regression test for this combination.

qbc2016 (commented):

src/agentscope/model/_gemini/_model.py
helloJamest: Addressed in `93632cb`.

The multi-type nullable `type` array conversion now checks for an existing `anyOf` first and raises `ValueError` instead of overwriting conjunctive constraints. I also added a regression test using the enum example shape so this cannot silently weaken the tool contract again.

Validation:
- `PYTHONPATH=src python -m pytest tests/model_gemini_test.py -q` (`30 passed`)
- `pre-commit run --config .pre-commit-config.yaml --files src/agentscope/model/_gemini/_model.py tests/model_gemini_test.py`

qbc2016 (approved):
LGTM.

Issue #2423, opened 2026-08-26
[Bug]: windows 启动  NotImplementedError   asyncio.create_subprocess_exec(...)


john00010:
### Prerequisites

- [x] I have searched the existing [issues](https://github.com/agentscope-ai/agentscope/issues) and [discussions](https://github.com/agentscope-ai/agentscope/discussions), and this is not a duplicate.
- [x] This is a bug, not a usage question. (For questions, please use [Discussions](https://github.com/agentscope-ai/agentscope/discussions/new?category=general) instead.)

### Background / Description

前端提示 Cannot reach the server. Check the server address and your network.
后端错误：
  INFO:     127.0.0.1:8349 - "GET /sessions/?agent_id=9401304258f247d1b3d279908246eec8 HTTP/1.1" 200 OK
  INFO:     127.0.0.1:8835 - "GET /sessions/c9f8b56b4ebc4867b1226a5844691dac/stream?agent_id=9401304258f247d1b3d279908246eec8 HTTP/1.1" 200 OK
  INFO:     127.0.0.1:9869 - "GET /workspace/skill?agent_id=9401304258f247d1b3d279908246eec8&session_id=c9f8b56b4ebc4867b1226a5844691dac HTTP/1.1" 500 Internal Server Error
  ERROR:    Exception in ASGI application
  Traceback (most recent call last):
    File "D:\www\a_bot\agentscope\.venv\Lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 416, in run_asgi
      result = await app(  # type: ignore[func-returns-value]
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "D:\www\a_bot\agentscope\.venv\Lib\site-packages\uvicorn\middleware\proxy_headers.py", line 63, in __call__
      return await self.app(scope, receive, send)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "D:\www\a_bot\agentscope\.venv\Lib\site-packages\fastapi\applications.py", line 1163, in __call__
      await super().__call__(scope, receive, send)
    File "D:\www\a_bot\agentscope\.venv\Lib\site-packages\starlette\applications.py", line 96, in __call__
      await self.middleware_stack(scope, receive, send)
    File "D:\www\a_bot\agentscope\.venv\Lib\site-packages\starlette\middleware\errors.py", line 186, in __call__
      raise exc
[...]

PR #2444, opened 2026-08-27
fix(app): use Proactor event loop on Windows for LocalBackend subprocesses

lwylab:
## Summary

On Windows, Agent Service workspace routes (`/workspace/skill`, MCP listing, shell tools, etc.) can raise `NotImplementedError` from `asyncio.create_subprocess_exec`.

**Root cause (documented behavior):**
- Per [Python asyncio platform notes](https://docs.python.org/3/library/asyncio-platforms.html#windows), `SelectorEventLoop` does **not** support subprocesses on Windows; only `ProactorEventLoop` does.
- `WindowsProactorEventLoopPolicy` has been the default since Python 3.8, but **uvicorn `reload=True`** installs `WindowsSelectorEventLoopPolicy` (see [encode/uvicorn#529](https://github.com/encode/uvicorn/issues/529)), which breaks `LocalBackend.exec_shell`.

**Fix:**
- Add `ensure_windows_proactor_event_loop_policy()` and call it from `create_app()` and when importing the local backend module.
- Convert the opaque `NotImplementedError` into a clear `RuntimeError` with actionable guidance.
- Disable uvicorn reload on Windows in `examples/agent_service/main.py`, and document the limitation in the example README with a link to the Python docs.
- Add a Windows regression test for `LocalBackend.exec_shell`.

## Test plan

- [x] `python -m unittest tests.asyncio_windows_policy_test -v` on Windows
- [x] Start `examples/agent_service/main.py` without reload and confirm `/workspace/skill` returns 200 (no `NotImplementedError`)
- [ ] CI unit tests on Linux remain green (no Windows-only regressions)

## References

- https://docs.python.org/3/library/asyncio-platforms.html#windows
- https://github.com/encode/uvicorn/issues/529

### DIFF
--- a/examples/agent_service/README.md
+++ b/examples/agent_service/README.md
@@ -62,11 +62,12 @@ pnpm dev
 
 After that, you can set the API endpoint `http://localhost:8000` in the Web UI and start experiencing the agent service.
 
-> **Windows note:** keep `reload=False` when running `python main.py` on
-> Windows. Uvicorn's reloader can select an asyncio event loop that does
-> not support subprocesses, which breaks local workspace routes (skills,
-> MCP listing, etc.). AgentScope now sets a subprocess-capable loop
-> policy during `create_app()`, but the reloader may still win the race.
+> **Windows note:** On Windows, Python's
+> [`SelectorEventLoop` does not support subprocesses](https://docs.python.org/3/library/asyncio-platforms.html#windows);
+> only `ProactorEventLoop` does. Uvicorn's `reload=True` forces the
+> selector loop, which breaks AgentScope's `LocalWorkspace` (skills, MCP
+> listing, shell tools, etc.). This example disables reload on Windows;
+> AgentScope also sets `WindowsProactorEventLoopPolicy` in `create_app()`.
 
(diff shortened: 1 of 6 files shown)


lwylab:
Friendly bump — this PR has been open for ~3 weeks with no review yet.

Would a maintainer have a chance to take a look when convenient?
cc @DavdGao @qbc2016

DavdGao:
The reload part is already on main — `examples/agent_service/main.py` has `reload=sys.platform != "win32"` with the same reasoning, and #2423 was closed as completed back in August.

The policy helper also can't do what it claims. uvicorn builds its loop with `asyncio.Runner(loop_factory=...)` (`uvicorn/_compat.py`), and `Runner` with an explicit factory never consults the event loop policy — `asyncio_loop_factory` returns `SelectorEventLoop` on win32 whenever `use_subprocess` (`reload or workers > 1`) is true, no matter what policy we set. The other way to end up on a Selector loop, an embedding app or Jupyter, has already created the loop by the time agentscope is imported, so setting the policy then does nothing either. Calling it at `_backend.py` import time would also mean importing agentscope mutates the process-global policy, which a library shouldn't do.

Closing this. The one piece worth having is the clearer error in `exec_shell` — happy to take that as a small separate PR.

PR #2446, opened 2026-08-27
Fix confirmation card overlapping chat messages

john00010:
## Summary

The confirmation card was absolutely positioned above the composer, which caused it to overlap and obscure the latest chat messages.

This change keeps the confirmation card in the normal layout flow and limits its height to 40vh with vertical scrolling. The conversation remains visible while large confirmation content can still be reviewed.

## Testing

- `pnpm build` (from `examples/web_ui/frontend`)


DavdGao:
@john00010 Actually, the confirmation card is intentionally designed to overlay the text input component. I'm not sure whether it would be better to push the messages above the card instead.

DavdGao:
Closing for now, and plz feel free to open new one if needed

PR #2459, opened 2026-08-28
fix(middleware): isolate concurrent retrieval tasks per session

### LINKED ISSUE #2274: [Bug]: AgenticMemoryMiddleware shared instances can clobber concurrent retrieval tasks

hata33:
Fixes #2274

## AgentScope Version

2.0.7.post1 (main @ e90f1c75)

## Description

**Problem.** `AgenticMemoryMiddleware` kept its in-flight retrieval task in a single instance-level `_retrieval_task` slot. When one middleware instance is shared by multiple agents/sessions, a later reply overwrites the earlier session's task, and the earlier reply's `on_reply` `finally` block then cancels the **later** session's retrieval instead of its own (see the hook-level reproduction in #2274).

**Changes** (mirrors the per-session isolation pattern already used by `ReMeMiddleware`):
- Replace the single `_retrieval_task` slot with `_retrieval_tasks: dict[str, Task]`, keyed by `agent.state.session_id` read live at hook time.
- `on_reply` discards any stale task left for its own session before starting a new one, and its `finally` consumes/cancels **only** this session's task; tasks of other in-flight sessions are untouched.
- `on_reasoning` consumes the matching session's task only.
- `_cached_input` (the other piece of per-reply state previously kept on the instance) is now a local variable in `on_reply`.

**Scope.** This isolates replies from *different* sessions sharing one middleware instance. Two concurrent replies within the *same* session (e.g. agents restored from one chat session, `app/_service/_chat.py:982`) can still clobber each other's task; per-reply keying would be needed for that and is left as future work.

No public API, configuration, or default behavior changed.

**Tests.** Regression test `test_concurrent_replies_isolate_retrieval_per_session`: two agents sharing one middleware instance, retrieval stubbed to block on an event so the interleaving is deterministic. Asserts behavior only: session A finishing its reply cancels only A's own task, session B's retrieval is not cancelled, and B's `on_reasoning` afterwards injects B's own hint.

## Checklist

- [x]  Related documentation has been updated (e.g. links, examples, etc.) in [documentation repository](https://github.com/agentscope-ai/docs)

Local checks: `pytest` on the middleware suites — all pass; `black --line-length=79`, `flake8 --extend-ignore=E203`, `mypy`, `pylint` clean on the changed files.

### DIFF
--- a/src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py
+++ b/src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 import asyncio
 import re
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, AsyncGenerator, Callable
+from typing import TYPE_CHECKING, Any, AsyncGenerator, Callable
 
 from pydantic import BaseModel, Field
 
@@ -476,11 +476,12 @@ class AgenticMemoryMiddleware(MiddlewareBase):
         self._parameters = parameters or self.Parameters()
         self._backend = backend or LocalBackend()
 
-        self._cached_input: str | None = None
-        # The in-flight asynchronous retrieval task started in ``on_reply``
-        # and consumed in ``on_reasoning``. Kept on the instance so the
-        # reasoning hook can poll for completion across iterations.
-        self._retrieval_task: asyncio.Task | None = None
+        # In-flight background retrieval per session (started in ``on_reply``,
+        # consumed/injected in ``on_reasoning``, cleaned up in ``on_reply``'s
+        # finally). Keyed by ``session_id`` so one middleware shared across
+        # agents keeps each session's retrieval isolated — a concurrent reply
+        # in another session never clobbers this one's task.
+        self._retrieval_tasks: dict[Any, asyncio.Task] = {}
 
     @staticmethod
     def _truncate_if_needed(content: str, max_length: int) -> str:
@@ -565,14 +566,25 @@ class AgenticMemoryMiddleware(MiddlewareBase):
 
         return f"{current_prompt}\n\n{content}"
 
+    @staticmethod
+    def _session_id_of(agent: "Agent") -> str | None:
+        """Read the ``session_id`` live from the agent.
+
+        The retrieval task is keyed by ``session_id``, read from
+        ``agent.state.session_id`` at hook time and threaded through per
+        call — **never** stored on the middleware — so a single instance
+        shared across agents keeps each session's retrieval isolated.
+        """
+        return getattr(getattr(agent, "state", None), "session_id", None)
+
     async def on_reply(
         self,
         agent: "Agent",
         input_kwargs: dict,
         next_handler: Callable[..., AsyncGenerator],
     ) -> AsyncGenerator:
-        """Cache the user input and kick off an asynchronous retrieval task
-        that runs concurrently with the agent reply.
+        """Kick off an asynchronous retrieval task that runs concurrently
+        with the agent reply.
 
         Args:
             agent (`Agent`):
@@ -586,6 +598,7 @@ class AgenticMemoryMiddleware(MiddlewareBase):
             `Any`:
                 Items yielded by ``next_handler``.
         """
+        session_id = self._session_id_of(agent)
 
         if self._parameters.retrieval_async:
             inputs = input_kwargs.get("inputs")
@@ -598,8 +611,9 @@ class AgenticMemoryMiddleware(MiddlewareBase):
             elif isinstance(inputs, Msg):
                 msgs = [inputs]
 
+            cached_input = None
             if msgs is not None:
-                self._cached_input = "\n".join(
+                cached_input = "\n".join(
                     [
                         f"{_.name}: " + _.get_text_content()
                         for _ in msgs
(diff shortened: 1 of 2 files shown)


src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py
```diff
@@ -476,11 +476,12 @@ def __init__(
...
-        # The in-flight asynchronous retrieval task started in ``on_reply``
-        # and consumed in ``on_reasoning``. Kept on the instance so the
-        # reasoning hook can poll for completion across iterations.
-        self._retrieval_task: asyncio.Task | None = None
+        # In-flight background retrieval per session (started in ``on_reply``,
+        # consumed/injected in ``on_reasoning``, cleaned up in ``on_reply``'s
+        # finally). Keyed by ``session_id`` so one middleware shared across
+        # agents keeps each session's retrieval isolated — a concurrent reply
+        # in another session never clobbers this one's task.
+        self._retrieval_tasks: dict[Any, asyncio.Task] = {}
```
DavdGao: `dict[Any, asyncio.Task]` should be `dict[str, asyncio.Task]` — `session_id` is always a `str`. Then the `Any` import isn't needed. Also the 5-line comment here can be one line, e.g. "In-flight retrieval per session, so a shared instance never clobbers another session's task."

DavdGao (commented):
The root cause and the fix are right — I reproduced it locally: with the old code, session A's `finally` cancels B's task and B ends up with no hint. The per-session dict matches what `ReMeMiddleware` already does, so the direction is fine.

Two things before merging:
- `_session_id_of` is unnecessary — one line, used twice, and its defensive `getattr` chain can't ever trigger. Inline it.
- The new test asserts on `middleware._retrieval_tasks` internals, so on the old code it fails with `AttributeError` instead of on the actual behavior. The behavioral assertions alone are already enough (I checked).

Also worth stating the scope in the description: this isolates different sessions, but two concurrent replies within the *same* session still clobber each other.

src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py
```diff
@@ -565,14 +566,25 @@ async def on_system_prompt(
 
         return f"{current_prompt}\n\n{content}"
 
+    @staticmethod
+    def _session_id_of(agent: "Agent") -> str | None:
```
DavdGao: This helper is a one-liner used twice, suggest dropping it and writing `session_id = agent.state.session_id` at both call sites. The nested `getattr` fallback can't trigger either: `Agent.__init__` always does `self.state = state or AgentState()` (`agent/_agent.py:169`), and `AgentState.session_id` is a `str` with a default factory (`state/_state.py:181`), so it's never missing or `None`.

(Same applies to `_reme/_middleware.py:297` — either clean both up or leave a note that you're following the existing pattern.)

src/agentscope/middleware/_longterm_memory/_agentic_memory/_middleware.py
```diff
@@ -598,39 +611,45 @@ async def on_reply(
...
                         f"{_.name}: " + _.get_text_content()
                         for _ in msgs
                         if _.get_text_content() is not None
                     ],
                 )
 
+            # Discard any stale task left for this session (a previous turn
+            # that never reached its finally is unexpected, but never leak
+            # one).
+            stale = self._retrieval_tasks.pop(session_id, None)
```
DavdGao, line 613: Keying by `session_id` doesn't isolate two concurrent replies within one session: this `pop`+`cancel` kills the earlier reply's live task, and then the earlier reply's `finally` pops and cancels *this* one's. `app/_service/_chat.py:982` assigns the same `session_id` to agents restored from one chat session, so it's reachable. Not asking you to fix it here, but please note the scope in the PR description.

tests/middleware_filesystem_memory_test.py
```diff
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
+# pylint: disable=protected-access,unused-argument
```
DavdGao: Disabling `protected-access,unused-argument` for the whole file to accommodate one test is too broad — it relaxes the checks on every other test here. Most of it goes away if the internals-based assertions below are dropped; for the stub you can use `patch.object(middleware, "_retrieve_relevant_files", ...)` and name the unused arg `_query`.

tests/middleware_filesystem_memory_test.py
```diff
@@ -774,3 +776,111 @@ async def test_agent_skips_retrieval_when_async_retrieval_disabled(
...
+        release = asyncio.Event()
+        cancelled = {session_a: False, session_b: False}
+        hints = {
+            session_a: "hint for session a",
+            session_b: "hint for session b",
+        }
+        session_of = {id(agent_a): session_a, id(agent_b): session_b}
+
+        async def fake_retrieve(agent: Agent, query: str) -> str:
+            """Block on ``release`` so the task lifetime is deterministic.
```
DavdGao: Full Google-style Args/Returns on a test stub is a bit heavy, one line is enough here.

tests/middleware_filesystem_memory_test.py
```diff
@@ -774,3 +776,111 @@ async def test_agent_skips_retrieval_when_async_retrieval_disabled(
...
+        await started[session_a].wait()
+        reply_b = asyncio.create_task(drive_on_reply(agent_b))
+        await started[session_b].wait()
+
+        # Session A finishes its reply while session B's retrieval is still
+        # in flight: A's cleanup must cancel only A's own task.
+        gate[session_a].set()
+        await reply_a
+        task_b = middleware._retrieval_tasks.get(session_b)
+        b_task_alive_after_a_exit = task_b is not None and not task_b.done()
```
DavdGao: `b_task_alive_after_a_exit` / `a_entry_left_after_a_exit` / `tasks_left_after_both_replies` all read `middleware._retrieval_tasks`, which is why the test fails with `AttributeError` on the pre-fix code rather than on the wrong behavior — it's locking the implementation, not the bug. Suggest keeping only `a_retrieval_cancelled`, `b_retrieval_cancelled` and `b_hints`; I checked that those alone already fail before the fix.

hata33:
Thanks for the thorough review — all points addressed in 32f88b61:

- **Inlined `_session_id_of`** into `session_id = agent.state.session_id` at both call sites; type narrowed to `dict[str, asyncio.Task]` and the `Any` import dropped; the init comment reduced to two lines.
- **Test cleanup**: removed the file-level `pylint: disable=protected-access,unused-argument`; the stub is now installed via `patch.object` with the unused arg named `_query`; stub docstrings reduced to one line; dropped the three internals-based assertions (`b_task_alive_after_a_exit` / `a_entry_left_after_a_exit` / `tasks_left_after_both_replies`) and kept only the behavioral ones.
- **Scope note added** to the description: this isolates different sessions; two concurrent replies within the same session (reachable via `app/_service/_chat.py:982`) can still clobber each other, left as future work.

On the `_reme/_middleware.py:297` note: I left ReMe's `_session_id_of` untouched here to keep this PR focused — happy to clean it up in a separate follow-up PR if you'd like.

PR #2463, opened 2026-08-29
fix(gemini): preserve tool call thought signatures

### LINKED ISSUE #2462: [Bug]: Gemini 3 tool calls fail with 400 — `function_call` loses `thought_signature` on history round-trip
### Prerequisites

### Background / Description

When using the native `GeminiChatModel` with a Gemini 3 model
(`gemini-3-flash-preview`) and function calling, the *first* tool-call round
succeeds, but the *next* request (which re-sends the assistant message
containing the previous `function_call`) is rejected by the API:

**Root cause**

- `agentscope/model/_gemini/_model.py` (lines ~376 and ~433) parses the
  response and stores `part.thought_signature` **base64-encoded into
  `ToolCallBlock.id`**.
- `agentscope/formatter/_gemini_formatter.py` (line ~204) re-serializes
  history but only writes `function_call: {id, name, args}` — it never
  restores the `thought_signature` field.

So the signature is effectively dropped on the round-trip.

### Error Messages

```shell
google.genai.errors.ClientError: 400 Bad Request. {
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts.
                This is required for tools to work correctly, and missing
                thought_signature may lead to degraded model performance.
                Additional data, function call `get_weather`, position 3.
                Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures
                for more details.",
    "status": "INVALID_ARGUMENT"
  }
}

Gemini 2.5 models tolerate the missing `thought_signature` (it is optional);
**Gemini 3 enforces it** — hence the 400.
```

### Steps to Reproduce

import base64
from agentscope.formatter import GeminiChatFormatter
from agentscope.message import Msg, ToolCallBlock

call_id = base64.b64encode(bytes(range(32))).decode("utf-8")
msg = Msg(name="assistant", role="assistant", content=[
    ToolCallBlock(id=call_id, name="get_weather", input='{"city": "Beijing"}'),
])
body = asyncio.run(GeminiChatFormatter().format([msg]))
# -> {"function_call": {"id": "...", "name": "get_weather", "args": {...}}}
#    ^^^ no "thought_signature" key

### Environment

- agentscope 2.0.4.post1 (bug still present in 2.0.7.post1, verified against
  the released wheel)
- google-genai 2.19.0
- Python 3.12, macOS

Oxygen56:
Fixes #2462

## AgentScope Version

2.0.7.post1 (main @ e90f1c7)

## Description

Gemini 3 requires the opaque `thought_signature` from a function-call response to be returned with that same history part on the next request. The model parser retained the signature in the tool-call ID, but the message-formatting path dropped the signature metadata, causing the next tool-call round to fail with HTTP 400.

This change preserves the base64-encoded signature as provider-specific `ToolCallBlock` metadata and restores it as raw bytes when formatting Gemini history. Both streaming and non-streaming responses use the same round trip, and streamed provider metadata is retained even when it arrives on a later delta.

When Gemini provides a function-call ID, that ID is preserved independently from the signature and reused for the matching function response. When Gemini omits the ID, AgentScope generates one while keeping the signature as separate metadata. Existing persisted blocks without separate signature metadata are not heuristically migrated because an arbitrary ID may also look like Base64.

## Verification

- Reproduced the missing `thought_signature` on the original base revision.
- Verified mocked Gemini responses through parsing, message serialization/restoration, history formatting, function-response ID mapping, generated IDs, and later-delta signature accumulation for streaming and non-streaming paths.
- `.venv/bin/python -m pytest tests/model_gemini_test.py tests/model_base_test.py tests/formatter_gemini_test.py tests/model_response_test.py tests/message_test.py -q` — 83 passed.
- `.venv/bin/pre-commit run --files src/agentscope/message/_block.py src/agentscope/model/_gemini/_model.py src/agentscope/model/_utils.py src/agentscope/formatter/_gemini_formatter.py tests/model_gemini_test.py tests/model_base_test.py` — all hooks passed.

## Checklist

- [x] Related documentation has been updated (not applicable: no public API or setup behavior changed)

### DIFF
--- a/src/agentscope/model/_gemini/_model.py
+++ b/src/agentscope/model/_gemini/_model.py
@@ -394,10 +394,12 @@ class GeminiChatModel(ChatModelBase):
 
                     # Tool call
                     if part.function_call:
+                        thought_signature: str | None = None
                         if part.thought_signature:
-                            call_id = base64.b64encode(
+                            thought_signature = base64.b64encode(
                                 part.thought_signature,
                             ).decode("utf-8")
+                            call_id = thought_signature
                         else:
                             call_id = part.function_call.id or _generate_id()
 
@@ -408,6 +410,7 @@ class GeminiChatModel(ChatModelBase):
                                 part.function_call.args or {},
                                 ensure_ascii=False,
                             ),
+                            thought_signature=thought_signature,
                         )
 
             usage = self._extract_usage(chunk.usage_metadata, start_datetime)
@@ -451,19 +454,22 @@ class GeminiChatModel(ChatModelBase):
 
                 if part.function_call:
                     keyword_args = part.function_call.args or {}
+                    thought_signature: str | None = None
                     if part.thought_signature:
-                        call_id = base64.b64encode(
+                        thought_signature = base64.b64encode(
                             part.thought_signature,
                         ).decode("utf-8")
+                        call_id = thought_signature
                     else:
                         call_id = part.function_call.id or _generate_id()
-                    content_blocks.append(
-                        ToolCallBlock(
-                            id=call_id,
-                            name=part.function_call.name,
-                            input=json.dumps(keyword_args, ensure_ascii=False),
-                        ),
+                    tool_call = ToolCallBlock(
+                        id=call_id,
+                        name=part.function_call.name,
+                        input=json.dumps(keyword_args, ensure_ascii=False),
                     )
+                    if thought_signature:
+                        tool_call.thought_signature = thought_signature
+                    content_blocks.append(tool_call)
 
         usage = self._extract_usage(response.usage_metadata, start_datetime)
 
--- a/src/agentscope/model/_gemini/_model.py
+++ b/src/agentscope/model/_gemini/_model.py
@@ -399,9 +399,11 @@ class GeminiChatModel(ChatModelBase):
                             thought_signature = base64.b64encode(
                                 part.thought_signature,
                             ).decode("utf-8")
-                            call_id = thought_signature
-                        else:
-                            call_id = part.function_call.id or _generate_id()
+                        call_id = (
+                            part.function_call.id
+                            or thought_signature
+                            or _generate_id()
+                        )
 
                         delta_res.append_tool_call(
                             block_id=call_id,
@@ -459,9 +461,11 @@ class GeminiChatModel(ChatModelBase):
                         thought_signature = base64.b64encode(
                             part.thought_signature,
                         ).decode("utf-8")
-                        call_id = thought_signature
-                    else:
-                        call_id = part.function_call.id or _generate_id()
+                    call_id = (
+                        part.function_call.id
+                        or thought_signature
+                        or _generate_id()
+                    )
                     tool_call = ToolCallBlock(
                         id=call_id,
                         name=part.function_call.name,
--- a/src/agentscope/model/_gemini/_model.py
+++ b/src/agentscope/model/_gemini/_model.py
@@ -399,11 +399,7 @@ class GeminiChatModel(ChatModelBase):
                             thought_signature = base64.b64encode(
                                 part.thought_signature,
                             ).decode("utf-8")
-                        call_id = (
-                            part.function_call.id
-                            or thought_signature
-                            or _generate_id()
-                        )
+                        call_id = part.function_call.id or _generate_id()
 
                         delta_res.append_tool_call(
                             block_id=call_id,
@@ -461,19 +457,22 @@ class GeminiChatModel(ChatModelBase):
                         thought_signature = base64.b64encode(
                             part.thought_signature,
                         ).decode("utf-8")
-                    call_id = (
-                        part.function_call.id
-                        or thought_signature
-                        or _generate_id()
+                    call_id = part.function_call.id or _generate_id()
+                    content_blocks.append(
+                        ToolCallBlock(
+                            id=call_id,
+                            name=part.function_call.name,
+                            input=json.dumps(
+                                keyword_args,
+                                ensure_ascii=False,
+                            ),
+                            **(
+                                {"thought_signature": thought_signature}
+                                if thought_signature
+                                else {}
+                            ),
+                        ),
                     )
-                    tool_call = ToolCallBlock(
-                        id=call_id,
-                        name=part.function_call.name,
-                        input=json.dumps(keyword_args, ensure_ascii=False),
-                    )
-                    if thought_signature:
-                        tool_call.thought_signature = thought_signature
-                    content_blocks.append(tool_call)
 
         usage = self._extract_usage(response.usage_metadata, start_datetime)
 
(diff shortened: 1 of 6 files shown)


src/agentscope/model/_gemini/_model.py
```diff
@@ -394,10 +394,12 @@ async def _parse_stream_response(
 
                     # Tool call
                     if part.function_call:
+                        thought_signature: str | None = None
                         if part.thought_signature:
-                            call_id = base64.b64encode(
+                            thought_signature = base64.b64encode(
                                 part.thought_signature,
                             ).decode("utf-8")
+                            call_id = thought_signature
```
iluv7: I took another look at the previous implementation. My understanding is that this did not surface with Gemini 2.5 because thought signatures were not required to be round-tripped, whereas Gemini 3 validates them and therefore exposes the issue.

The underlying ID problem seems to have existed for a while: when a thought signature is present, we Base64-encode it and use it as `call_id`, silently discarding `part.function_call.id`. Using the signature as the call ID happens to work because Gemini currently treats the ID as a correlation key, and the same value is used for both the function call and its response.

Now that `thought_signature` is preserved as a separate field, would it be better to keep the model-generated call ID as well?

```python
if part.thought_signature:
    thought_signature = base64.b64encode(
        part.thought_signature,
    ).decode("utf-8")

call_id = part.function_call.id or _generate_id()
```

Of course, we should also consider backward compatibility with existing persisted sessions or downstream code that may rely on the previous signature-as-ID behavior. What do you think?

src/agentscope/model/_gemini/_model.py
```diff
@@ -451,19 +454,22 @@ def _parse_completion_response(
 
                 if part.function_call:
                     keyword_args = part.function_call.args or {}
+                    thought_signature: str | None = None
                     if part.thought_signature:
-                        call_id = base64.b64encode(
+                        thought_signature = base64.b64encode(
                             part.thought_signature,
                         ).decode("utf-8")
+                        call_id = thought_signature
```
iluv7: The same applies here in the non-streaming path.

iluv7 (commented):
I left a small comment.

src/agentscope/model/_gemini/_model.py
Oxygen56: Thanks for catching this. I updated both parser paths so the model-provided function_call.id is preserved independently from thought_signature, and the matching function response keeps that ID. When Gemini omits an ID, the existing signature-as-ID fallback remains to avoid changing that behavior. I added streaming and non-streaming coverage for both cases. Existing persisted IDs are not guessed as signatures because a valid ID may also look like Base64.

src/agentscope/model/_gemini/_model.py
Oxygen56: Updated the non-streaming path in the same way. Its round-trip test now verifies the original function-call ID, the separate thought signature, and the matching function-response ID.

src/agentscope/model/_gemini/_model.py
```diff
@@ -394,12 +394,16 @@ async def _parse_stream_response(
...
                         if part.thought_signature:
-                            call_id = base64.b64encode(
+                            thought_signature = base64.b64encode(
                                 part.thought_signature,
                             ).decode("utf-8")
-                        else:
-                            call_id = part.function_call.id or _generate_id()
+                        call_id = (
+                            part.function_call.id
+                            or thought_signature
```
DavdGao: Since we store the thought_signature within the tool call block directly, plz remove line 404 here.

src/agentscope/model/_gemini/_model.py
```diff
@@ -408,6 +412,7 @@ async def _parse_stream_response(
                                 part.function_call.args or {},
                                 ensure_ascii=False,
                             ),
+                            thought_signature=thought_signature,
```
DavdGao, line 411: This only survives when the signature happens to arrive on the first delta of the block. `_AccToolCallBlock.append` (model/_utils.py:82) doesn't carry over model_extra, unlike `_AccThinkingBlock.append` right above it, so a signature on any later delta is silently dropped. Suggest copying those three lines into `_AccToolCallBlock.append`. The new streaming test only feeds a single chunk, so it doesn't catch this.

src/agentscope/model/_gemini/_model.py
```diff
@@ -451,19 +456,24 @@ def _parse_completion_response(
...
+                        part.function_call.id
+                        or thought_signature
+                        or _generate_id()
+                    )
+                    tool_call = ToolCallBlock(
+                        id=call_id,
+                        name=part.function_call.name,
+                        input=json.dumps(keyword_args, ensure_ascii=False),
                     )
+                    if thought_signature:
```
DavdGao: The streaming path passes the signature through `append_tool_call(..., thought_signature=...)`, but here the block is built first and then mutated. Suggest constructing it in one go so both paths read the same.

tests/model_gemini_test.py
```diff
@@ -549,6 +640,86 @@ async def test_stream_tool_calls_without_id(self) -> None:
...
+        ]
+        self.mock_client.aio.models.generate_content_stream = AsyncMock(
+            return_value=_MockAsyncStream(chunks),
+        )
+
+        gen = await self.model([])
+        responses = [response async for response in gen]
+        final_block = responses[-1].content[0]
+
+        self.assertEqual(final_block.id, signature_b64)
```
DavdGao: The other tool call tests in this file compare the whole `result.content` against a literal list of blocks (line 180), these new ones assert field by field which is harder to follow. Suggest keeping the existing style. Also the two id-less tests (line 311 and 717) only assert the block and never the formatted output, which is the part the issue is actually about.

DavdGao (commented):
Plz see inline comments, others lgtm

src/agentscope/model/_gemini/_model.py
Oxygen56: Removed the signature-as-ID fallback in both streaming and non-streaming paths. An id-less function call now gets a generated ID while the signature remains separate provider metadata, with deterministic coverage for both paths.

src/agentscope/model/_gemini/_model.py
Oxygen56: Updated `_AccToolCallBlock.append` to carry non-None provider extras from every delta. I also added a multi-delta regression where `thought_signature` arrives on the later delta.

src/agentscope/model/_gemini/_model.py
Oxygen56: Updated the non-streaming path to construct `ToolCallBlock` with `thought_signature` in one step.

tests/model_gemini_test.py
Oxygen56: Reworked the new tests to compare the complete `ToolCallBlock` content. Both id-less streaming and non-streaming cases now format the parsed history and verify the generated ID and `thought_signature` in the outgoing Gemini parts.

PR #2471, opened 2026-08-31
fix(tool): split read-only bash commands on every separator

### LINKED ISSUE #2470: [Bug]: `&` and newline separators bypass the Bash read-only auto-allow, executing unapproved commands in DEFAULT permission mode
## Prerequisites

## Background / Description

`Bash.check_permissions()` auto-ALLOWs a command in **every** permission mode — including
`DEFAULT` — when `BashCommandParser.is_read_only_command()` returns `True`
(`src/agentscope/tool/_builtin/_bash.py:270`). The safety checks that follow it (dangerous
command patterns, dangerous paths, dangerous removal paths) are never reached, because the
read-only branch returns first.

`is_read_only_command()` only treats a command as compound when one of `&&`, `||`, `;` or
`|` appears literally in the string:

```python
# src/agentscope/tool/_builtin/_bash_parser.py
if any(op in cmd for op in ["&&", "||", ";", "|"]):
    ...  # split on the AST, require every subcommand to be read-only
# Single command - check directly
return self._is_single_command_read_only(cmd)
```

Bash has more command separators than those four. `&` (background) and a plain newline also
terminate a command, and neither is in the list. For `ls & rm -rf /tmp/pwn` or
`ls\nrm -rf /tmp/pwn` the classifier therefore skips the compound path and judges the whole
string by its read-only *head*, so it reports read-only and the tool auto-allows a command
that deletes files.

The same string is then handed to the shell unchanged, which runs both parts.

`check_injection_risk()` does not catch these either: `&` and newline produce plain
`command` nodes in the tree-sitter AST, not any of the `DANGEROUS_NODE_TYPES`
(`command_substitution`, `subshell`, control flow, ...). So the well-known
`ls $(rm -rf /)` vector is blocked, while `ls & rm -rf /` is not.

Affected separators, all classified read-only today:

| command | `is_read_only_command` | `check_permissions` |
|---|---|---|
| `ls; rm -rf /tmp/pwn` | `False` | ASK (correct) |
| `ls && rm -rf /tmp/pwn` | `False` | ASK (correct) |
| `ls $(rm -rf /tmp/pwn)` | `False` | ASK (correct) |
| `ls & rm -rf /tmp/pwn` | **`True`** | **ALLOW** |
| `ls -la&rm -rf /tmp/pwn` | **`True`** | **ALLOW** |
| `ls\nrm -rf /tmp/pwn` | **`True`** | **ALLOW** |
| `ls\r\nrm -rf /tmp/pwn` | **`True`** | **ALLOW** |
| `ls\n\nrm -rf /tmp/pwn` | **`True`** | **ALLOW** |
| `ls &\nrm -rf /tmp/pwn` | **`True`** | **ALLOW** |

This matters because the read-only fast path is exactly the path that exists to run without
asking the user. A model that emits a multi-line bash block — which is entirely ordinary
output — gets its trailing commands executed with no confirmation, and a prompt-injected one
can reach it deliberately. This is the same class of gap as #2003 (`find -delete` classified
read-only), just through the separator list rather than the predicate list.

## Expected Behavior

A command is read-only only when **every** subcommand it will actually run is read-only,
whatever separator joins them.

## Error Messages

```text
permission mode : PermissionMode.DEFAULT
command         : 'ls /tmp/demo & rm -f /tmp/demo/important.txt'
decision        : PermissionBehavior.ALLOW — Permission granted for read-only command
read_only       : True
victim exists   : True
victim exists   : False  <-- after execution
```

## Steps to Reproduce

Save and run against current `main`. It only touches a private temporary directory.

```python
import asyncio
import os
import tempfile

from agentscope.permission import PermissionContext
from agentscope.tool._builtin._bash import Bash


async def main() -> None:
    for sep in ["&", "\n"]:
        with tempfile.TemporaryDirectory() as d:
            victim = os.path.join(d, "important.txt")
            with open(victim, "w", encoding="utf-8") as f:
                f.write("payroll")

            command = f"ls {d}{sep}rm -f {victim}"
            tool = Bash()
            context = PermissionContext()

            decision = await tool.check_permissions({"command": command}, context)
            print("mode     :", context.mode)
            print("command  :", repr(command))
            print("decision :", decision.behavior, "-", decision.message)
            print("read_only:", await tool.check_read_only({"command": command}))
            print("exists   :", os.path.exists(victim))

            async for _ in await tool(command=command):
                pass

            print("exists   :", os.path.exists(victim), "<-- after execution")
            print()


asyncio.run(main())
```

Both separators print `PermissionBehavior.ALLOW` with `read_only: True`, and the file is
gone afterwards without any confirmation being requested.

## Environment

- AgentScope Version: 2.0.7.post1 (also reproduces on `main`)
- Python Version: 3.11.11
- OS: macOS 15 (the classifier is platform-independent)

tancheng33:
## AgentScope Version

2.0.7.post1 / current `main`

## Description

Fixes #2470.

`BashCommandParser.is_read_only_command()` decided whether a command was compound by
scanning the raw string for `&&`, `||`, `;` and `|`. Bash separates commands with more than
those four, so `ls & rm -rf /tmp/pwn` and `ls\nrm -rf /tmp/pwn` took the single-command path
and were judged by their read-only head alone. `Bash.check_permissions()` auto-ALLOWs
whatever that classifier calls read-only, in every permission mode including `DEFAULT`, and
returns before the dangerous-command / dangerous-path checks — so the trailing command ran
with no confirmation.

The AST split already handles these separators: `split_compound_command()` collects every
`command` node, and tree-sitter parses `ls & rm -rf x` and `ls\nrm -rf x` as two `command`
nodes. The textual pre-check was the only thing keeping them off that path. This PR drops
the pre-check and always splits on the AST, keeping the existing behaviour for parse
failures (conservative `False`) and for redirections (`>` still returns early).

**Changes**

- `src/agentscope/tool/_builtin/_bash_parser.py`: `is_read_only_command()` always splits
  subcommands from the parsed AST and requires every one of them to be read-only.
- `tests/permission_bash_parser_test.py`: parser-level cases for `&`, `\n`, `\r\n`, blank
  lines and `&\n`, plus cases asserting that read-only sequences joined by those separators
  stay read-only (`ls &`, `ls\ncat file.txt`, `git status\ngit log`).
- `tests/builtin_bash_test.py`: tool-level cases asserting `check_permissions()` no longer
  returns `ALLOW` and `check_read_only()` returns `False` for those commands.

**Testing**

- `pytest tests` — 2043 passed, 147 skipped, 400 subtests passed.
- `pre-commit run --files <changed files>` — all hooks pass.
- Before the fix the new tests fail; after it, `ls & rm -rf /tmp/pwn` and
  `ls\nrm -rf /tmp/pwn` return the bypass-immune ASK that `ls; rm -rf /tmp/pwn` already got.
- Checked for regressions across ordinary read-only commands (`ls -la`, `git log --oneline
  -5`, `cat a.txt | head -5`, `git diff && git status`, `cd /tmp; ls`, `echo 'a && b'`,
  `NODE_ENV=prod ls`, heredocs, multi-line read-only sequences): all still classified
  read-only.

## Checklist

- [ ] Related documentation has been updated (no doc change needed — internal behaviour fix)

### DIFF
--- a/src/agentscope/tool/_builtin/_bash_parser.py
+++ b/src/agentscope/tool/_builtin/_bash_parser.py
@@ -132,6 +132,27 @@ READ_ONLY_COMMANDS = {
     "pip show",
 }
 
+# AST node types that only group other statements. The split traverses
+# into them and collects the statements they contain.
+COMPOUND_NODE_TYPES = {
+    "program",
+    "list",
+    "pipeline",
+    "command_list",
+    "redirected_statement",
+    "negated_command",
+}
+
+# AST node types that carry no statement of their own, so nothing is
+# collected from them (a redirect target is not a command, and the read-only
+# check rejects output redirections separately).
+NON_STATEMENT_NODE_TYPES = {
+    "comment",
+    "file_redirect",
+    "heredoc_redirect",
+    "herestring_redirect",
+}
+
 FIND_MUTATING_PREDICATES = {
     "-delete",
     "-exec",
@@ -156,9 +177,10 @@ class BashCommandParser:
         """Check if a command is read-only (safe to auto-allow).
 
         For compound commands, ALL subcommands must be read-only for the
-        entire command to be considered read-only. Subcommands are split
-        from the parsed AST, so every bash command separator counts — not
-        only ``&&``, ``||``, ``;`` and ``|`` but also ``&`` and newlines.
+        entire command to be considered read-only. Subcommands come from
+        :meth:`split_compound_command`, which splits on the parsed AST, so
+        the separator used does not matter and a statement it cannot
+        classify makes the whole command non-read-only.
 
         Commands with output redirections (>, >>) are NOT considered read-only.
 
@@ -178,23 +200,19 @@ class BashCommandParser:
         if ">" in cmd:
             return False
 
-        # Split into subcommands and check each one. The split always goes
-        # through the AST: a textual scan for "&&"/"||"/";"/"|" misses the
-        # remaining separators, and "ls & rm -rf /" or "ls\nrm -rf /" would
-        # then be judged by their read-only head alone.
+        # Always split on the AST: a textual scan for the separators misses
+        # "&" and newlines.
         try:
             tree = self.parser.parse(bytes(cmd, "utf8"))
-            root = tree.root_node
-            subcommands = self.split_compound_command(root, cmd)
+            subcommands = self.split_compound_command(tree.root_node, cmd)
         except Exception:
             # If parsing fails, be conservative
             return False
 
-        # All subcommands must be read-only
-        for subcmd in subcommands:
-            if not self._is_single_command_read_only(subcmd.strip()):
-                return False
-        return True
+        return all(
+            self._is_single_command_read_only(subcmd.strip())
+            for subcmd in subcommands
+        )
 
     def _is_single_command_read_only(self, cmd: str) -> bool:
         """Check if a single (non-compound) command is read-only.
@@ -526,7 +544,12 @@ class BashCommandParser:
     def split_compound_command(self, root: Node, command: str) -> List[str]:
         """Split compound commands using tree-sitter for precise parsing.
 
-        Recognizes: &&, ||, ;, |
+        Every statement the shell would run is returned, whatever separator
+        joins them — ``&&``, ``||``, ``;``, ``|``, ``&`` and newlines alike.
+        Statements that are not plain commands (``export``/``declare``,
+        bare assignments, ``unset``, ...) are returned as-is rather than
+        dropped, so callers judging the split cannot mistake them for
+        nothing having happened.
 
         Args:
             root (`Node`):
@@ -541,20 +564,26 @@ class BashCommandParser:
         subcommands = []
 
         def extract_commands(node: Node) -> None:
-            """Recursively extract commands from AST."""
+            """Recursively extract statements from AST."""
             if node.type == "command":
                 # Extract command text
                 cmd_text = command[node.start_byte : node.end_byte]
                 subcommands.append(cmd_text)
-            elif node.type in ["list", "pipeline", "command_list"]:
-                # Recursively process compound structures
+            elif node.type in COMPOUND_NODE_TYPES:
+                # Recursively process compound structures, skipping the
+                # separator tokens between the statements they group
                 for child in node.children:
-                    if child.type not in ["&&", "||", ";", "|", "|&"]:
+                    if child.is_named:
                         extract_commands(child)
-            else:
-                # Continue traversing
-                for child in node.children:
-                    extract_commands(child)
+            elif node.type in NON_STATEMENT_NODE_TYPES:
+                # Nothing executable here
+                return
+            elif node.is_named:
+                # An unrecognized statement — ``export PATH=/evil``, a bare
+                # assignment, a subshell, ... Return its text instead of
+                # dropping it, so a caller that requires every subcommand to
+                # be read-only fails closed on what we cannot classify.
+                subcommands.append(command[node.start_byte : node.end_byte])
 
         extract_commands(root)
         return subcommands if subcommands else [command]
--- a/src/agentscope/tool/_builtin/_bash_parser.py
+++ b/src/agentscope/tool/_builtin/_bash_parser.py
@@ -155,8 +155,10 @@ class BashCommandParser:
     def is_read_only_command(self, command: str) -> bool:
         """Check if a command is read-only (safe to auto-allow).
 
-        For compound commands (&&, ||, ;, |), ALL subcommands must be
-        read-only for the entire command to be considered read-only.
+        For compound commands, ALL subcommands must be read-only for the
+        entire command to be considered read-only. Subcommands are split
+        from the parsed AST, so every bash command separator counts — not
+        only ``&&``, ``||``, ``;`` and ``|`` but also ``&`` and newlines.
 
         Commands with output redirections (>, >>) are NOT considered read-only.
 
@@ -176,25 +178,23 @@ class BashCommandParser:
         if ">" in cmd:
             return False
 
-        # Check if it's a compound command
-        if any(op in cmd for op in ["&&", "||", ";", "|"]):
-            # Split into subcommands and check each one
-            try:
-                tree = self.parser.parse(bytes(cmd, "utf8"))
-                root = tree.root_node
-                subcommands = self.split_compound_command(root, cmd)
+        # Split into subcommands and check each one. The split always goes
+        # through the AST: a textual scan for "&&"/"||"/";"/"|" misses the
+        # remaining separators, and "ls & rm -rf /" or "ls\nrm -rf /" would
+        # then be judged by their read-only head alone.
+        try:
+            tree = self.parser.parse(bytes(cmd, "utf8"))
+            root = tree.root_node
+            subcommands = self.split_compound_command(root, cmd)
+        except Exception:
+            # If parsing fails, be conservative
+            return False
 
-                # All subcommands must be read-only
-                for subcmd in subcommands:
-                    if not self._is_single_command_read_only(subcmd.strip()):
-                        return False
-                return True
-            except Exception:
-                # If parsing fails, be conservative
+        # All subcommands must be read-only
+        for subcmd in subcommands:
+            if not self._is_single_command_read_only(subcmd.strip()):
                 return False
-
-        # Single command - check directly
-        return self._is_single_command_read_only(cmd)
+        return True
 
     def _is_single_command_read_only(self, cmd: str) -> bool:
         """Check if a single (non-compound) command is read-only.
--- a/src/agentscope/tool/_builtin/_bash_parser.py
+++ b/src/agentscope/tool/_builtin/_bash_parser.py
@@ -217,14 +217,22 @@ class BashCommandParser:
     def _is_single_command_read_only(self, cmd: str) -> bool:
         """Check if a single (non-compound) command is read-only.
 
+        A leading variable assignment makes the command non-read-only,
+        whatever follows it: ``PATH=/tmp/evil ls`` runs ``ls`` from a
+        directory the caller chose.
+
         Args:
             cmd (`str`):
-                A single command string (no &&, ||, ;, |)
+                A single statement from :meth:`split_compound_command`,
+                which is not always a plain command
 
         Returns:
             `bool`:
                 True if the command is read-only, False otherwise
         """
+        if self._has_leading_assignment(cmd):
+            return False
+
         # Check exact match in read-only commands
         if cmd in READ_ONLY_COMMANDS:
             return True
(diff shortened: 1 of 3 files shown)


src/agentscope/tool/_builtin/_bash_parser.py
```diff
@@ -155,8 +155,10 @@ def __init__(self) -> None:
     def is_read_only_command(self, command: str) -> bool:
         """Check if a command is read-only (safe to auto-allow).
 
-        For compound commands (&&, ||, ;, |), ALL subcommands must be
-        read-only for the entire command to be considered read-only.
+        For compound commands, ALL subcommands must be read-only for the
+        entire command to be considered read-only. Subcommands are split
+        from the parsed AST, so every bash command separator counts — not
+        only ``&&``, ``||``, ``;`` and ``|`` but also ``&`` and newlines.
```
DavdGao: This claims every separator counts now, but `export` / `unset` / bare assignments still fall out of the split because they aren't `command` nodes. Suggest not claiming full coverage until that's handled.

src/agentscope/tool/_builtin/_bash_parser.py
```diff
@@ -176,25 +178,23 @@ def is_read_only_command(self, command: str) -> bool:
...
+            # If parsing fails, be conservative
+            return False
 
-        # Single command - check directly
-        return self._is_single_command_read_only(cmd)
+        # All subcommands must be read-only
+        for subcmd in subcommands:
+            if not self._is_single_command_read_only(subcmd.strip()):
+                return False
+        return True
```
DavdGao: The 4-line comment repeats what the docstring above already says, one line is enough.

And the loop can just be:

```python
return all(self._is_single_command_read_only(s.strip()) for s in subcommands)
```

DavdGao (commented):
The direction is right, but the fix only covers half of the problem.

`split_compound_command()` only collects `command` nodes, so statements that aren't `command` nodes are silently dropped from the split:

```
"ls\nexport PATH=/tmp/evil"  ->  ['ls']
"ls\nPATH=/tmp/evil"         ->  ['ls']
"ls\nunset PATH"             ->  ['ls']
```

They parse as `declaration_command` / `variable_assignment`. So `ls\nPATH=/tmp/evil\nls` splits into `['ls', 'ls']`, is judged read-only and gets auto-ALLOWed in DEFAULT mode, and the second `ls` runs from the poisoned PATH. I verified this against real bash. `check_injection_risk()` doesn't catch it either, since `variable_assignment` is not in `DANGEROUS_NODE_TYPES`.

Not a regression, it returned True before this PR too. But now that everything goes through the AST split, the node types `split_compound_command()` recognizes are the only thing left guarding the auto-ALLOW, so its blind spots matter more than they did before.

Suggest making `extract_commands()` fail closed: treat a statement type it doesn't recognize as non-read-only instead of dropping it, so these degrade to ASK rather than ALLOW. If you'd rather keep this PR minimal that's fine, but then please soften the new docstring and open a follow-up issue, because "every bash command separator counts" reads like the whole class is closed now.

One more, outside the diff: `split_compound_command()`'s docstring (line 529) still says `Recognizes: &&, ||, ;, |`, which is exactly the claim this PR disproves, and that function is the only path now. Worth fixing here, that stale line is part of how #2470 happened.

src/agentscope/tool/_builtin/_bash_parser.py
tancheng33: Confirmed, and thanks — I reproduced all three against `main`:

```
"ls\nexport PATH=/tmp/evil"   -> ['ls']   declaration_command
"ls\nPATH=/tmp/evil"          -> ['ls']   variable_assignment
"ls\nunset PATH"              -> ['ls']   unset_command
```

(`unset` is its own `unset_command` node rather than a `declaration_command`, but it was
dropped the same way.)

I went with the fail-closed option rather than softening the docstring, since you're right
that the recognized node types are now the only thing guarding the auto-ALLOW.

**What changed in `split_compound_command()`**

`extract_commands()` no longer has a catch-all "keep traversing" branch that can walk past a
statement without collecting it. Node types are now explicitly one of three kinds:

- `COMPOUND_NODE_TYPES` — `program`, `list`, `pipeline`, `command_list`,
  `redirected_statement`, `negated_command`: pure grouping, traverse into their named
  children (which also drops the separator tokens, since those are anonymous).
- `NON_STATEMENT_NODE_TYPES` — `comment`, `file_redirect`, `heredoc_redirect`,
  `herestring_redirect`: nothing executable, skip.
- anything else that is a named node — return its source text as a subcommand.

So `export PATH=/evil`, a bare assignment, `unset`, `declare`, `readonly`, and any node type
tree-sitter grows in the future all land in the third bucket, get handed to
`_is_single_command_read_only()`, and fail there. The split for your example is now
`['ls', 'PATH=/tmp/evil', 'ls']`.

**Outcome**

```
ls\nPATH=/tmp/evil\nls   read_only=False   ASK
ls\nexport PATH=/x\nls   read_only=False   ASK
ls\nunset PATH           read_only=False   ASK
PATH=/x && ls            read_only=False   ASK
```

The ASK holds even for a user who has allow-listed `ls` (`PermissionRule("Bash", "ls:*",
ALLOW)`): the extra statement changes the extracted prefixes, so rule matching does not
cover it either.

**Regressions checked** — these all stay read-only: `ls &`, `ls\ncat file.txt`,
`git status\ngit log`, `ls # comment`, `ls\n# comment\ncat f`, `cat < input.txt`,
`cat <<EOF\nhello\nEOF`, `! ls`, `NODE_ENV=prod ls`, `cat a.txt | head -5`,
`git diff && git status`, `cd /tmp; ls`, `echo 'a && b'`. The heredoc and redirect cases are
why `file_redirect` / `heredoc_redirect` are in the skip set — without it, `cat < in.txt`
would have started asking.

Also done:
- `split_compound_command()`'s `Recognizes: &&, ||, ;, |` line is gone; it now states that
  every separator is covered and that unclassifiable statements are returned rather than
  dropped.
- `is_read_only_command()`'s docstring no longer claims the separator class is closed; it
  points at the split and says an unclassifiable statement makes the command non-read-only.
- Comment trimmed to one line, loop replaced with `all(...)`.

Tests added: the parser-level vectors, an explicit assertion on the split output for
`ls\nPATH=/tmp/evil\nls`, the read-only cases above as regression guards, and the tool-level
`check_permissions()` assertions for the PATH-poisoning form.

`pytest tests` — 2045 passed, 147 skipped, 415 subtests passed. `pre-commit` clean.

tancheng33:
Pushed 09a34ae — took the fail-closed route on `extract_commands()` rather than softening the docstring, and dropped the stale `Recognizes: &&, ||, ;, |` line you flagged. Full response in the [review thread](https://github.com/agentscope-ai/agentscope/pull/2471#discussion_r3892454074).

src/agentscope/tool/_builtin/_bash_parser.py
```diff
@@ -541,20 +564,26 @@ def split_compound_command(self, root: Node, command: str) -> List[str]:
...
-                    extract_commands(child)
+            elif node.type in NON_STATEMENT_NODE_TYPES:
+                # Nothing executable here
+                return
+            elif node.is_named:
+                # An unrecognized statement — ``export PATH=/evil``, a bare
+                # assignment, a subshell, ... Return its text instead of
+                # dropping it, so a caller that requires every subcommand to
+                # be read-only fails closed on what we cannot classify.
+                subcommands.append(command[node.start_byte : node.end_byte])
```
DavdGao, line 589: The `command` branch is redundant: `command` is a named node and isn't in `NON_STATEMENT_NODE_TYPES`, so this last branch already appends the same text. Checked on ~2900 generated statement/separator combinations, the two versions give identical splits. Suggested to be:

```python
def extract_commands(node: Node) -> None:
    """Recursively extract statements from AST."""
    if node.type in COMPOUND_NODE_TYPES:
        for child in node.children:
            if child.is_named:
                extract_commands(child)
    elif node.is_named and node.type not in NON_STATEMENT_NODE_TYPES:
        subcommands.append(command[node.start_byte : node.end_byte])
```

Also the 4-line comment here is the same thing I asked to trim last round, one line is enough.

tests/permission_bash_parser_test.py
```diff
@@ -282,6 +282,74 @@ async def test_mutating_find_commands_are_not_read_only(self) -> None:
...
+                self.assertFalse(
+                    self.parser.is_read_only_command(cmd),
+                    f"Expected '{cmd}' to be non-read-only",
+                )
+
+    async def test_split_keeps_non_command_statements(self) -> None:
+        """Test the split returns statements that are not plain commands."""
+        command = "ls\nPATH=/tmp/evil\nls"
+        tree = self.parser.parser.parse(bytes(command, "utf8"))
+        self.assertEqual(
```
DavdGao: `assertListEqual` here rather than `assertEqual`.

DavdGao (commented):
Verified the separator fix does what it says — `ls & rm -rf /usr` and `ls\nrm -rf /usr` go from ALLOW to a bypass-immune ASK, and the read-only cases (heredocs, `cat < f`, `! ls`, pipelines, comments) all still pass. Two things left.

First, outside the diff, `_is_single_command_read_only()` line 244. The PATH poisoning is only half fixed. The split handles `ls\nPATH=/tmp/evil\nls`, but the inline form is one `command` node so it never reaches the split, and that loop skips any leading `VAR=value` and judges `ls`:

```
PATH=/tmp/evil ls           -> read_only=True -> ALLOW
LD_PRELOAD=/tmp/evil.so ls  -> read_only=True -> ALLOW
BASH_ENV=/tmp/evil ls       -> read_only=True -> ALLOW
```

That's the more natural way to write the attack than `PATH=/x && ls`, so the class isn't closed. Suggested to treat a leading assignment as non-read-only there too — or, if you'd rather keep this PR minimal, leave it and open a follow-up issue, but then the docstring and your comment shouldn't say the PATH case is handled.

Second, `extract_commands` has one branch too many, see inline.

Note the CI hasn't actually run on this PR, only `notify`. I'll approve the workflows.

tancheng33:
Both fixed in c5be4e2, and you were right that the class was not closed.

`PATH=/tmp/evil ls` is a single `command` node so it never reaches the split, and the loop at line 244 skipped every leading `VAR=value` token before looking up the base command, so it judged `ls`. Reproduced all three.

`_is_single_command_read_only()` now checks the parsed command for a `variable_assignment` child and returns False when it has one, so the inline form ends up where `PATH=/tmp/evil && ls` already was:

```
PATH=/tmp/evil ls               False   ASK
LD_PRELOAD=/tmp/evil.so ls      False   ASK
BASH_ENV=/tmp/evil ls           False   ASK
IFS=. NODE_ENV=prod git status  False   ASK
```

The token loop that skipped assignments is gone with it, since nothing reaches it any more. One side effect of removing it: a quoted `"a=b" ls`, which bash does not treat as an assignment either, used to be read-only because that loop skipped any token containing `=`. It asks now, which is what bash trying to run a command named `a=b` deserves.

One behaviour change worth calling out, because I listed it as a regression check in the PR description: `NODE_ENV=prod ls` was read-only and now asks. I did not key this off `SAFE_ENV_VARS`, since `PATH`, `SHELL`, `EDITOR` and `PAGER` are all in that set and `_extract_command_prefix()` uses it too, so fixing it there would have grown this past #2470. Say the word if you would rather keep the harmless prefixes auto-allowed and handle the env-var set separately.

A value that merely contains `=` is unaffected: `echo a=b` and `grep 'k=v' file.txt` stay read-only, both added as guards. Also rechecked and unchanged: `ls -la`, `cat a.txt | head -5`, `git diff && git status`, `cd /tmp; ls`, `cat < in.txt`, heredocs, `! ls`, `ls &`, `ls # comment`, `ls\ncat file.txt`.

Two things I checked while I was in there, neither of which needs a change:

- The wrapper spelling is already covered. `env`, `command`, `nice`, `timeout`, `nohup`, `stdbuf` and `setsid` are in neither `SAFE_COMMANDS` nor `READ_ONLY_COMMANDS`, so `env PATH=/tmp/evil ls` was already non-read-only.
- `is_read_only_command()` does return True for `echo $(rm -rf /tmp/x)`, but `check_injection_risk()` runs ahead of it in both entry points, so `check_permissions()` gives the bypass-immune ASK at step 0 and `check_read_only()` returns False before the classifier is consulted.

Also updated the `cmd` arg description on `_is_single_command_read_only()`, which still said "A single command string (no &&, ||, ;, |)". Since the last round the split hands it bare assignments too, and that is the kind of stale line we agreed is worth not leaving around.

Tests: parser-level cases for the inline vectors, the two `=`-in-a-value guards, and the tool-level `check_permissions()` list extended with the three.

`pytest tests` gives 2046 passed, 147 skipped, 425 subtests passed. pre-commit clean. Thanks for approving the workflows.

src/agentscope/tool/_builtin/_bash_parser.py
tancheng33: Took your version. Confirmed the same way: 2304 statement/separator combinations give identical splits between the two, and reading it, `command` is named and outside `NON_STATEMENT_NODE_TYPES`, so the branch order never mattered either. Comment is one line now, sorry for leaving it in last round.

tests/permission_bash_parser_test.py
tancheng33: Done.

PR #2474, opened 2026-08-31
fix(agent): repair interrupted tool call arguments

### LINKED ISSUE #2469: [Bug]: interrupting a streaming tool call persists truncated JSON and breaks subsequent turns
### Prerequisites

### Background / Description

When a model stream is interrupted while it is still generating tool-call arguments, AgentScope can persist the partially accumulated `ToolCallBlock.input` into the conversation history.

If the interruption occurs before the tool-call arguments form valid JSON, the malformed input is still retained in the session history. AgentScope later marks the tool call as `finished` and appends an interrupted `ToolResultBlock`, but the original malformed `ToolCallBlock.input` remains unchanged.

On the next user turn, `OpenAIChatFormatter` replays this malformed value directly as:

```text
tool_calls[].function.arguments
```

For OpenAI-compatible backends that validate historical tool calls, such as vLLM with the `qwen3_coder` tool-call parser, the whole request is then rejected with HTTP 400.

Because the malformed tool call remains in persistent session history, every subsequent turn in the same session fails. Creating a new session works normally.

This is not a context-length issue. In the observed session, only about 34k of a 128k context window was used.

### Observed persisted state

The interrupted tool call was stored approximately as:

```json
{
  "type": "tool_call",
  "id": "chatcmpl-tool-...",
  "name": "Write",
  "input": "{\"file_path\": \"/workspace/scratch/verify2896.py\", \"content\": \"Verify the standard DP ... print(f\"MISMATCH: s1={s1} s2={'",
  "state": "finished"
}
```

The corresponding result was:

```text
<system-reminder>The tool call has been interrupted by the user.</system-reminder>
```

The `input` field is clearly incomplete JSON: the `content` string and outer object are never closed.

### Suspected root cause

The current interruption path appears to be:

1. `src/agentscope/model/_base.py`
   - catches `asyncio.CancelledError`
   - keeps the accumulated partial streaming response
   - returns it as an interrupted `ChatResponse`

2. `src/agentscope/model/_utils.py`
   - `_StreamAccumulator` concatenates partial `ToolCallBlock.input` fragments
   - no JSON-validity check is performed before `build()`

3. `src/agentscope/agent/_agent.py`
   - `_reasoning_impl()` saves the completed/interrupted response content into context
   - interruption cleanup later marks the tool call as `finished`
   - an interrupted tool result is appended
   - the malformed `ToolCallBlock.input` itself is not removed or sanitized

4. `src/agentscope/formatter/_openai_formatter.py`
   - historical `ToolCallBlock.input` is replayed unchanged as `function.arguments`

So the lifecycle becomes:

```text
streaming tool arguments
        ↓
user interrupts
        ↓
partial ToolCallBlock is preserved
        ↓
malformed JSON is saved to context
        ↓
tool call is marked finished
        ↓
next turn replays malformed arguments
        ↓
OpenAI-compatible backend returns HTTP 400
```

### Expected behavior

After an interruption, the session history should remain valid and the user should be able to continue the same conversation.

An incomplete tool call should not later be replayed as a normal historical tool call containing malformed JSON.

Possible implementations could discard, sanitize, or otherwise make interrupted incomplete tool-call arguments safe for history replay.

### Related work

This seems related to, but not fully covered by:

- #1806 / #1995 — graceful agent interruption and context/tool-state consistency
- #2012 — repair truncated tool-call input in Anthropic/Gemini/Ollama formatters
- #2075 — Web UI handling of incomplete JSON during tool-call streaming/interruption

In particular, #2012 already notes that truncated `ToolCallBlock.input` can be produced by streaming interruption, but `OpenAIChatFormatter` still forwards the raw string. That is safe for the OpenAI SDK itself, but not necessarily for strict OpenAI-compatible backends that parse historical tool-call arguments.

### Error Messages

```shell
openai.BadRequestError: Error code: 400 - {
  'error': {
    'message': 'Unterminated string starting at: line 1 column 62 (char 61)',
    'type': 'BadRequestError',
    'param': None,
    'code': 400
  }
}

Model Qwen3.8-27B exhausted all 1 attempt(s)

Reply failed; reported to the client as invalid_request.
```

### Steps to Reproduce

1. Start an AgentScope application with:

   - streaming enabled
   - an OpenAI-compatible model backend
   - a tool whose arguments may contain a long string, e.g. `Write(file_path, content)`
   - persistent session storage

2. Ask the agent to generate and write a sufficiently long source file so that the `content` argument is emitted over multiple streaming chunks.

3. While the model is still streaming the `Write` tool-call arguments, click the Web UI **Stop / Interrupt** button.

4. Inspect the persisted conversation context.

   An interrupted tool call can be found whose `input` contains incomplete JSON, for example:

   ```json
   {
     "name": "Write",
     "input": "{\"file_path\":\"/workspace/test.py\",\"content\":\"...unfinished",
     "state": "finished"
   }
   ```

   An interrupted tool result is also present.

5. Send any new user message in the same session.

6. Observe that AgentScope replays the malformed historical tool-call arguments to the OpenAI-compatible backend.

7. With vLLM using the `qwen3_coder` tool-call parser, the request fails with:

   ```text
   Unterminated string starting at ...
   ```

   and HTTP 400 is returned.

8. Send another message in the same session.

   The same error occurs again because the malformed historical tool call is still present.

9. Create a new session and send a message.

   The request succeeds because the malformed historical tool call is no longer part of the context.

### Environment

- AgentScope Version: 2.0.7
- Python Version: 3.11.15
- OS: Linux / Docker
- Model: Qwen3.8-27B
- Model Backend: vLLM 0.27.1
- API: OpenAI-compatible Chat Completions
- Context Size: 128k
- Tool Call Parser: qwen3_coder
- Streaming: enabled
- Session Storage: Redis

iluv7:
## AgentScope Version

`2.0.7.post1`

## Description

Fixes https://github.com/agentscope-ai/agentscope/issues/2469.

When a streaming model call is interrupted while generating tool-call arguments, the partial JSON could be persisted and replayed on later turns. Strict OpenAI-compatible backends such as vLLM then rejected the request with HTTP 400.

## What Changed

- Adds a helper that validates and repairs tool-call arguments as a JSON object string.
- Repairs partial tool-call arguments during interruption cleanup.
- Adds defensive repair in `OpenAIChatFormatter` for existing sessions containing malformed tool-call arguments.
- Adds regression tests for interrupted streams and malformed historical tool calls.

### Testing

```bash
PYTHONPATH=src .venv/bin/pytest -q \
  tests/agent_interrupt_test.py \
  tests/formatter_openai_chat_test.py \
  tests/model_base_test.py \
  tests/formatter_anthropic_test.py \
  tests/formatter_gemini_test.py \
  tests/formatter_ollama_test.py
```

Result: `80 passed`.

`black`, `flake8`, `mypy`, and `pylint` also pass for the changed files.

## Checklist

- [x] An issue has been created for this PR (#2469)
- [ ] I have read the [CONTRIBUTING.md](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md)
- [x] Related documentation has been updated (not applicable for this bug fix)

### DIFF
--- a/src/agentscope/model/_base.py
+++ b/src/agentscope/model/_base.py
@@ -34,6 +34,28 @@ _TOOL_CHOICE_LITERAL_MODES = {"auto", "none", "required"}
 _MULTIMODAL_DATA_BLOCK_TOKEN_ESTIMATE = 2000
 
 
+def _repair_tool_call_arguments(response: ChatResponse) -> ChatResponse:
+    """Repair incomplete tool arguments in a final model response."""
+    for block in response.content:
+        if not isinstance(block, ToolCallBlock):
+            continue
+
+        try:
+            parsed = json.loads(block.input)
+            if isinstance(parsed, dict):
+                continue
+        except json.JSONDecodeError:
+            pass
+
+        try:
+            parsed = _json_loads_with_repair(block.input)
+        except ToolJSONDecodeError:
+            continue
+        block.input = json.dumps(parsed, ensure_ascii=False)
+
+    return response
+
+
 class ChatModelBase:
     """The base class for chat models."""
 
@@ -255,7 +277,7 @@ class ChatModelBase:
         # Consume the model calling result
         # =====================================================================
         if isinstance(res, ChatResponse):
-            return res
+            return _repair_tool_call_arguments(res)
 
         async def _stream() -> AsyncGenerator[ChatResponse, None]:
             """The wrapper around model calling."""
@@ -279,13 +301,14 @@ class ChatModelBase:
                             continue
                     else:
                         yield_acc_res = False
+                        _repair_tool_call_arguments(chunk)
                     yield chunk
             except asyncio.CancelledError:
                 acc_res.finished_reason = FinishedReason.INTERRUPTED
                 yield_acc_res = True
 
             if yield_acc_res:
-                yield acc_res.build()
+                yield _repair_tool_call_arguments(acc_res.build())
 
         return _stream()
 
--- a/tests/agent_interrupt_test.py
+++ b/tests/agent_interrupt_test.py
@@ -12,6 +12,7 @@
   loop.
 """
 import asyncio
+import json
 from typing import Any, AsyncGenerator
 from unittest.async_case import IsolatedAsyncioTestCase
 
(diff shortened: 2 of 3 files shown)


iluv7:
Once the approach #2516 is agreed upon and the follow-up PR is opened, I will remove the OpenAI-specific formatter changes from this PR, as that logic will be moved to and handled by the new PR.

iluv7:
Updated this PR based on the #2516 above:
- Moved tool-call argument repair to the response side of ChatModelBase.__call__(), covering non-streaming responses, provider-supplied final chunks, and accumulated responses after interruption.
- Only rewrites the input when repair succeeds; unrecoverable input remains unchanged so Toolkit can return ToolJSONDecodeError.
- Removed the OpenAI formatter safeguard and the repair in _close_unfinished_tool_calls.
- Added regression tests for interruption, output truncation, and unrecoverable inputs.

Existing sessions containing malformed history remain out of scope. All related tests and pre-commit checks pass.

Issue #2481, opened 2026-09-01
feat(rag): support RAGFlow in a new `RAGFlowKnowledgeBase` class rather than VDB


DavdGao:
**Background**

Through recent investigation into RAGFlow, we found that it provides a complete, end-to-end RAG pipeline — including document parsing, chunking, indexing, and retrieval — as a fully managed solution. This makes it fundamentally different from a traditional vector database (VDB).

Forcing RAGFlow into our existing VDB abstraction would be a poor fit: it would either require bypassing RAGFlow's core strengths (e.g., its built-in document processing and indexing pipeline) or lead to an awkward, leaky abstraction. To fully leverage what RAGFlow offers, it should instead be integrated at the Knowledge layer, sitting alongside the existing Knowledge class rather than beneath it.

**Changes**

Add RAGFlowKnowledge class under the rag module
Implement the same interface as the existing Knowledge class, ensuring consistency and interoperability across different knowledge backends
Encapsulate RAGFlow's document parsing, indexing, and retrieval capabilities within the new class

MannXo:
I'll take this if no one's working on it yet.

DavdGao:
@MannXo Thanks for your interest! Since RAGFlow offers a more complete, end-to-end pipeline compared to a traditional VDB, there may well be additional functionalities worth exposing beyond the current Knowledge class interface. If you identify any such capabilities during your exploration, feel free to raise them here — we're happy to discuss and align on the interface design before diving into implementation.

MannXo:
Thanks, that offer to align first is very welcome, because the exploration turned up more than I expected. Findings below, ending with the decisions I think are yours to make.

## Where RAGFlow fits cleanly

The agent-facing surface is smaller than it looks. `RAGMiddleware` only ever touches `name`, `description` and `search()` on a knowledge base, and `search()` already takes text rather than a vector. That maps almost exactly onto `POST /api/v1/retrieval`, so the retrieval half needs no new concepts.

The document lifecycle also lines up better than I expected. `KnowledgeDocumentStatus` is already `pending / parsing / chunking / indexing / ready / error`, and RAGFlow reports `run`, `progress`, `progress_msg`, `chunk_count` and `token_count` per document. Its server-side parse maps onto states we already model rather than needing new ones.

## Where it does not

[...]

PR #2485, opened 2026-09-01
feat(rag): add RAGFlowKnowledge knowledge-layer backend

ConcaSun:
Support RAGFlow as a managed, end-to-end RAG pipeline at the knowledge layer, sitting alongside KnowledgeBase rather than underneath VectorStoreBase. RAGFlow owns document parsing, chunking, indexing, and retrieval on the server side, so RAGFlowKnowledge accepts raw document bytes and delegates the heavy lifting to RAGFlow.

Mirrors the KnowledgeBase interface (search, insert_document, delete_document, list_documents, list_chunks) for consistency across knowledge backends. The ragflow-sdk dependency is lazily imported and ships behind a new vdb-ragflow extra.

Refs #2481

## PR Title Format

## AgentScope Version

## Description

[Please describe the background, purpose, changes made, and how to test this PR]

## Checklist

### DIFF
--- a/examples/rag/README.md
+++ b/examples/rag/README.md
@@ -6,6 +6,7 @@ Two library-mode walk-throughs of `agentscope.rag` — no FastAPI service, no ma
 | --- | --- |
 | [`index_and_search.py`](./index_and_search.py) | The minimal pipeline: parse → chunk → embed → insert, then `KnowledgeBase.search`. Start here. |
 | [`integrate_with_agent.py`](./integrate_with_agent.py) | Attaches the same `KnowledgeBase` to an `Agent` via `RAGMiddleware`, in both `static` (auto-inject) and `agentic` (tool-driven) modes. |
+| [`ragflow_knowledge.py`](./ragflow_knowledge.py) | Uses `RAGFlowKnowledge` — a knowledge-layer integration that delegates parsing/chunking/indexing/retrieval to a RAGFlow server instead of running the pipeline locally. |
 
 Both examples use an in-memory Qdrant store (`location=":memory:"`) and the DashScope `text-embedding-v4` model, so no external services are required. The sections below show how to swap in Milvus Lite, MongoDB, or Elasticsearch instead; those backends need additional setup.
 
@@ -225,6 +226,58 @@ store = ElasticsearchStore(
 - For service mode, construct
   `CollectionPerKbManager(storage=storage, vector_store=store)` and pass it
   to `create_app(knowledge_base_manager=...)`.
+
+### RAGFlow (managed pipeline)
+
+[RAGFlow](https://ragflow.io/) is a managed, end-to-end RAG pipeline: it
+owns document parsing, chunking, indexing, and retrieval on the server
+side.  Instead of being forced underneath `VectorStoreBase`, it is exposed
+as `RAGFlowKnowledge` — a knowledge-layer handle with the same interface
+as `KnowledgeBase` (`search`, `insert_document`, `delete_document`,
+`list_documents`).
+
+Install the optional extra:
+
+```bash
+uv pip install "agentscope[vdb-ragflow]"
+# Or from source (repo root)
+uv pip install -e ".[vdb-ragflow]"
+```
+
+**Prerequisites**
+
+- A reachable RAGFlow instance (e.g. via Docker Compose) with a dataset
+  already created in the RAGFlow console.  RAGFlow's dataset is the
+  knowledge base — note its id and set its embedding model / parsing
+  strategy in the console (RAGFlow, not AgentScope, runs those steps).
+- An API key and base URL available in the environment (do not hard-code
+  credentials).
+
+```bash
+export RAGFLOW_API_KEY="ragflow-xxxxx"
+export RAGFLOW_BASE_URL="http://localhost:9380"
+export RAGFLOW_DATASET_ID="kb-xxxxx"
+```
+
+Then run the example — no parser, chunker, embedding model, or vector
+store is needed on your side:
+
+```bash
+python examples/rag/ragflow_knowledge.py
+```
+
+**Notes**
+
+- `insert_document` uploads raw document bytes; RAGFlow parses, chunks,
+  and indexes them server-side.  This deliberately differs from
+  `KnowledgeBase.insert_document`, which takes pre-embedded `Chunk`
+  objects because AgentScope runs the pipeline locally.
+- Retrieval tuning (`top_k`, `similarity_threshold`,
+  `vector_similarity_weight`, optional rerank) is configured on
+  `RAGFlowConfig`.
+- RAGFlow scores like a hybrid of vector + keyword similarity rather than
+  a plain cosine distance, so `score_threshold` behaves a little
+  differently from the pure vector backends.
 
 ### Choosing a vector backend
 
--- a/examples/rag/README.md
+++ b/examples/rag/README.md
@@ -1,6 +1,6 @@
 # RAG Examples
 
-Two library-mode walk-throughs of `agentscope.rag` — no FastAPI service, no manager, no message bus. Each script wires the building blocks (parser, chunker, embedding model, vector store, `KnowledgeBase` handle) by hand so the data flow is visible end-to-end.
+Library-mode walk-throughs of `agentscope.rag` — no FastAPI service, no manager, no message bus. The first two scripts wire the building blocks (parser, chunker, embedding model, vector store, `KnowledgeBase` handle) by hand so the data flow is visible end-to-end; the third delegates the whole pipeline to a RAGFlow server via `RAGFlowKnowledge`.
 
 | Script | What it shows |
 | --- | --- |
@@ -232,11 +232,13 @@ store = ElasticsearchStore(
 [RAGFlow](https://ragflow.io/) is a managed, end-to-end RAG pipeline: it
 owns document parsing, chunking, indexing, and retrieval on the server
 side.  Instead of being forced underneath `VectorStoreBase`, it is exposed
-as `RAGFlowKnowledge` — a knowledge-layer handle with the same interface
-as `KnowledgeBase` (`search`, `insert_document`, `delete_document`,
-`list_documents`).
+as `RAGFlowKnowledge` — a knowledge-layer handle that exposes the same
+operations as `KnowledgeBase` (`search`, `insert_document`,
+`delete_document`, `list_documents`, `list_chunks`).
 
-Install the optional extra:
+Install the optional extra (kept opt-in — it is **not** part of
+`agentscope[full]`, because `ragflow-sdk` does not yet support
+Python 3.13+):
 
 ```bash
 uv pip install "agentscope[vdb-ragflow]"
@@ -272,6 +274,14 @@ python examples/rag/ragflow_knowledge.py
   and indexes them server-side.  This deliberately differs from
   `KnowledgeBase.insert_document`, which takes pre-embedded `Chunk`
   objects because AgentScope runs the pipeline locally.
+- RAGFlow indexing is **asynchronous** — `insert_document` returns once
+  the upload is accepted, but a document becomes searchable only after
+  RAGFlow finishes parsing it.  The example polls the document's parse
+  status before searching (see `wait_until_indexed`).
+- Because the two handles expose *operations with the same names but
+  different signatures*, a caller cannot transparently swap a
+  `RAGFlowKnowledge` in for a `KnowledgeBase` without adapting the
+  `insert_document` / `search` calls.
 - Retrieval tuning (`top_k`, `similarity_threshold`,
   `vector_similarity_weight`, optional rerank) is configured on
   `RAGFlowConfig`.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,6 +110,11 @@ vdb-milvus = [
 ]
 vdb-mongodb = ["pymongo>=4.7"]
 vdb-elasticsearch = ["elasticsearch[async]>=8.12"]
+# RAGFlow is a managed end-to-end RAG pipeline (parsing, chunking,
+# indexing, retrieval).  It is integrated at the knowledge layer as
+# `RAGFlowKnowledge` rather than underneath VectorStoreBase, so it ships
+# as its own extra.
+vdb-ragflow = ["ragflow-sdk>=0.17.0"]
 # ------------ RAG ------------
 # Document parsers; pick a vector store via the vdb-* extras.
 rag = [
@@ -148,6 +153,7 @@ k8s = ["agentscope[workspace-k8s]"]
 milvuslite = ["agentscope[vdb-milvus]"]
 mongodb = ["agentscope[vdb-mongodb]"]
 elasticsearch = ["agentscope[vdb-elasticsearch]"]
+ragflow = ["agentscope[vdb-ragflow]"]
 mem0 = ["agentscope[memory-mem0]"]
 reme = ["agentscope[memory-reme]"]
 # ------------ Full ------------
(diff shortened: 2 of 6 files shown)


DavdGao (commented):
It's suggsted to add a new abstrction class for KnowledgeBase, e.g. `KnowledgeBaseBase`, so that we can align all its sub classes, i.e. `RAGFlowKnowledgeBase` and `KnowledgeBase`.

Others plz see inline comments

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+                score = self._extract_score(chunk)
+                threshold = (
+                    self._config.similarity_threshold
+                    if score_threshold is None
+                    else score_threshold
+                )
+                if threshold is not None and score < threshold:
+                    continue
+                document_id = (
+                    getattr(chunk, "dataset_id", None)
```
DavdGao: This is the dataset id, not the document id — `ragflow_sdk.Chunk` has `document_id`. All hits end up with the same id and the dedup key degenerates to `(dataset_id, rank)`, dropping hits from different documents. Suggested to use `chunk.document_id` and dedup on `chunk.id`.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+            value = getattr(chunk, attr, None)
+            if value is not None:
+                try:
+                    return float(value)
+                except (TypeError, ValueError):
+                    break
+        return 0.0
+
+    @staticmethod
+    def _to_agentscope_chunk(chunk: Any, index: int) -> Chunk:
```
DavdGao: `chunk_index` should be the 0-based index within the document, not the retrieval rank, and `total_chunks=0` is wrong. RAGFlow doesn't return either, suggested to keep `chunk.id` in metadata and dedup on that. Also this and `_extract_score` are single-use, suggested to inline them.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+
+        merged = sorted(
+            best.values(),
+            key=lambda result: result.score,
+            reverse=True,
+        )
+        return merged[:top_k] if top_k > 0 else merged
+
+    @staticmethod
+    def _extract_score(chunk: Any) -> float:
```
DavdGao: `ragflow_sdk.Chunk` always initializes `similarity`, the fallback never triggers. Suggested to use `chunk.similarity` directly.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+                    source=getattr(document, "name", "") or "",
+                    chunk_count=int(getattr(document, "chunk_count", 0) or 0),
+                    metadata={
+                        "parse_progress": getattr(
+                            document,
+                            "progress",
+                            None,
+                        ),
+                        "run": getattr(document, "run", None),
+                        "size": getattr(document, "size", None),
```
DavdGao: All these fields are always initialized in `ragflow_sdk.Document.__init__`, so the `getattr` fallbacks and the `if not document_id` guard are dead code. Suggested to access them directly, same for the ones in `search` / `list_chunks`.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+            chunk_index=index,
+            total_chunks=0,
+            metadata=metadata,
+        )
+
+    # ------------------------------------------------------------------
+    # Document management
+    # ------------------------------------------------------------------
+
+    async def insert_document(
```
DavdGao: `upload_documents` doesn't parse or index — `dataset.async_parse_documents([doc.id])` is needed, otherwise the document never becomes searchable (the example uploads and searches right away). Suggested to also note in the docstring that indexing is async here, unlike `KnowledgeBase.insert_document`.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+            dataset.upload_documents,
+            [{"display_name": filename, "blob": blob}],
+        )
+        if not documents:
+            raise RuntimeError(
+                f"RAGFlow did not return a document after uploading "
+                f"{filename!r} to dataset {self._config.dataset_id!r}.",
+            )
+        document = documents[0]
+        return getattr(document, "id", None) or document_id or _generate_id()
```
DavdGao: `document_id` is never used, and the `_generate_id()` fallback returns an id that doesn't exist on the server, which breaks `delete_document`. Suggested to drop the argument and return `document.id`.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+            `list[DocumentSummary]`:
+                One summary per document in the dataset, in server-defined
+                order.
+        """
+        dataset = await self._get_dataset()
+        documents = await asyncio.to_thread(
+            dataset.list_documents,
+            id=None,
+            page=1,
+            page_size=30,
```
DavdGao, line 454: This caps `list_documents` at 30 documents while `KnowledgeBase` returns all. Suggested to page until a short page comes back.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+            limit (`int`, defaults to ``30``):
+                Maximum number of chunks to return.
+
+        Returns:
+            `list[Chunk]`:
+                At most ``limit`` chunks.
+        """
+        document = await self._get_document(document_id)
+        # RAGFlow pages from 1 and numbers chunks from 1.
+        page = offset // limit + 1 if limit > 0 else 1
```
DavdGao, line 501: `offset=5, limit=30` maps to page 1 and returns chunks 0-29 instead of 5-34; `base_index` has the same issue. Suggested to fetch `page_size=offset + limit` and slice `[offset:]`. The docstring also claims a 1→0 rebase that the code doesn't do.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+                f"RAGFlow document {document_id!r} not found in dataset "
+                f"{self._config.dataset_id!r}.",
+            )
+        return documents[0]
+
+    # ------------------------------------------------------------------
+    # Search
+    # ------------------------------------------------------------------
+
+    async def search(
```
DavdGao, line 254: `KnowledgeBase.search` takes `list[str | TextBlock | DataBlock]`. Suggested to accept `DataBlock` too and drop those entries before calling RAGFlow, like `KnowledgeBase` does for text-only models.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+                )
+                for text in query_texts
+            ),
+        )
+
+        best: dict[tuple[str, int], VectorSearchResult] = {}
+        for query_results in results_per_query:
+            for index, chunk in enumerate(query_results):
+                score = self._extract_score(chunk)
+                threshold = (
```
DavdGao, line 340: `score_threshold=None` means no filtering in `KnowledgeBase`, and the server already applied `similarity_threshold` in `retrieve`. Suggested to filter only when `score_threshold is not None`.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+        if self._client is None:
+            from ragflow_sdk import RAGFlow
+
+            self._client = RAGFlow(
+                api_key=self._config.api_key,
+                base_url=self._config.base_url,
+            )
+        return self._client
+
+    async def _get_dataset(self) -> "Dataset":
```
DavdGao, line 204: Every operation does a `list_datasets` round-trip. Suggested to cache the resolved dataset, like `KnowledgeBase.ensure_collection` does.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,556 @@
...
+    # Read-only accessors
+    # ------------------------------------------------------------------
+
+    @property
+    def config(self) -> RAGFlowConfig:
+        """The bound :class:`RAGFlowConfig`."""
+        return self._config
+
+    @property
+    def api_key(self) -> str:
```
DavdGao, line 167: Nits: `api_key` / `base_url` / `dataset_id` just forward to `config`, suggested to remove; `get_client` -> `_get_client`; L351 `if top_k > 0 else merged` can just be a slice; L46 the SDK exports `DataSet`, not `Dataset`.

pyproject.toml
```diff
@@ -165,6 +171,7 @@ full = [
     "agentscope[vdb-milvus]",
     "agentscope[vdb-mongodb]",
     "agentscope[vdb-elasticsearch]",
+    "agentscope[vdb-ragflow]",
```
DavdGao: `ragflow-sdk` is `Requires-Python: >=3.10,<3.13` while agentscope is `>=3.11`, so this breaks `pip install "agentscope[full]"` (and `[dev]`) on 3.13+. Suggested to keep it opt-in only.

tests/rag_ragflow_test.py
```diff
@@ -0,0 +1,282 @@
...
+    doc.id = document_id
+    doc.name = name
+    doc.chunk_count = chunk_count
+    doc.progress = progress
+    doc.run = run
+    doc.size = 1024
+    return doc
+
+
+@unittest.skipUnless(
```
DavdGao: The client is fully mocked, nothing here needs the SDK — and once `vdb-ragflow` leaves `full`, these tests would never run in CI. Suggested to drop the skip.

tests/rag_ragflow_test.py
```diff
@@ -0,0 +1,282 @@
...
+                1,
+                doc_name="sick.txt",
+                similarity=0.7,
+            ),
+        ]
+        self.mock_client.retrieve.return_value = hits
+
+        results = await self.kb.search(["What is PTO?"], top_k=5)
+
+        self.assertEqual(len(results), 2)
```
DavdGao: Suggested to assert the whole list with `assertListEqual` against a literal expected structure, instead of field by field.

tests/rag_ragflow_test.py
```diff
@@ -0,0 +1,282 @@
...
+        self.assertIsInstance(results[0], VectorSearchResult)
+        # Ordered by descending similarity.
+        self.assertAlmostEqual(results[0].score, 0.9)
+        self.assertAlmostEqual(results[1].score, 0.7)
+        self.assertEqual(results[0].document_id, "kb-1")
```
DavdGao: This locks in the `dataset_id`-as-`document_id` bug.

examples/rag/README.md
```diff
@@ -226,6 +227,58 @@ store = ElasticsearchStore(
...
   to `create_app(knowledge_base_manager=...)`.
 
+### RAGFlow (managed pipeline)
+
+[RAGFlow](https://ragflow.io/) is a managed, end-to-end RAG pipeline: it
+owns document parsing, chunking, indexing, and retrieval on the server
+side.  Instead of being forced underneath `VectorStoreBase`, it is exposed
+as `RAGFlowKnowledge` — a knowledge-layer handle with the same interface
+as `KnowledgeBase` (`search`, `insert_document`, `delete_document`,
+`list_documents`).
```
DavdGao: "Same interface as `KnowledgeBase`" isn't accurate — `insert_document` takes different arguments and `search` accepts fewer types. Suggested to state what's actually shared once the abstraction is in.

src/agentscope/rag/_ragflow.py
ConcaSun: ok, I was using the dataset id as the document id. I've fixed ``search`` to read ``chunk.document_id`` (the real source-document id) and to dedup on the stable RAGFlow chunk id, so hits from different documents no longer collapse to the same key. Thanks for catching this.

src/agentscope/rag/_ragflow.py
ConcaSun: ok, since RAGFlow reports neither a per-document chunk index nor a total, I now keep the RAGFlow chunk id in ``metadata["ragflow_chunk_id"]`` and dedup on that, and I set ``chunk_index`` as a running 0-based ordinal within the document rather than a retrieval rank. I also inlined the single-use helpers. total_chunks stays 0, and the method docstring notes why

src/agentscope/rag/_ragflow.py
ConcaSun: yes, ``chunk`` always initialises similarity, so I removed ``_extract_score`` and ``read chunk.similarity`` directly, dropping the never-triggered fallback

src/agentscope/rag/_ragflow.py
ConcaSun: Understand, now I've removed the ``getattr`` fallbacks and the empty-id guard in ``list_documents``, and applied the same cleanup to ``search`` and ``list_chunks``

src/agentscope/rag/_ragflow.py
ConcaSun: I now call ``dataset.async_parse_documents([doc.id])`` inside ``insert_document`` after the upload so the document becomes searchable, and I noted in the docstring that RAGFlow indexing is asynchronous (unlike ``KnowledgeBase.insert_document``). The example also polls the document's parse status before searching.

pyproject.toml
ConcaSun: ok, I've changed it to keep it opt-in only, as an explicit ``agentscope[vdb-ragflow]``

tests/rag_ragflow_test.py
ConcaSun: ok, I've dropped the module-wide ``skipUnless`` so those tests always run in CI. I kept a ``skipUnless`` only on ``test_client_is_lazily_created``, which is the one test that genuinely constructs the real client to exercise the lazy import

tests/rag_ragflow_test.py
ConcaSun: I've rewritten the search and document tests to assert the whole result list against an explicit expected structure instead of checking fields one by one.

tests/rag_ragflow_test.py
ConcaSun: ok, The tests now assert real per-document ids coming from ``chunk.document_id`` rather than the dataset id

examples/rag/README.md
ConcaSun: ok, The README and module docstring now state that the handles expose same-named operations but with different signatures (``insert_document`` takes bytes + filename, ``search`` accepts only text queries), so a transparent drop-in swap isn't implied

ConcaSun:
@DavdGao Thanks a lot for the thorough review, the feedback really improved this PR. I've addressed all the points and pushed a new commit 0de522c , I also replied under each inline comment — when you have a chance, could you take another look? Think you.

wangtaotaotao95:
The issue claim was released and reassigned to me, so I continued this work in #2722. I cherry-picked both of your commits intact to preserve authorship and credit, then adapted the implementation to the maintainer-requested KnowledgeBaseBase abstraction and the Python 3.11-compatible async HTTP approach discussed on #2481. I have not closed this PR; if you are resuming it, please say so and I am happy to coordinate.

ConcaSun:
@wangtaotaotao95 I saw your PR. I appreciate that you preserved the commits and authorship. I’m happy for you to continue and finish this on top of my work. Please go ahead with #2722; I’m not planning to resume it myself right now. If you need anything from me or want to coordinate, just let me know.

DavdGao:
Thanks a lot for the contribution! This PR figured out how a RAGFlow knowledge base should look in AgentScope, and that it belongs to the knowledge layer rather than the vector store layer, which set the direction for the RAGFlow support. #2722 continues on top of your commits with the `KnowledgeBaseBase` abstraction, so closing this one in favor of it.

PR #2486, opened 2026-09-01
fix(rag): normalize distance scores so higher means more relevant

### LINKED ISSUE #2482: [Bug]: distance-based vector metrics are ranked in reverse by KnowledgeBase
## Bug description

The RAG pipeline assumes that a larger `VectorSearchResult.score` always means a more relevant result, but Qdrant and Milvus expose distance-based metrics where a smaller value is better.

This causes `KnowledgeBase.search` to reverse the ranking for:

- Qdrant `Euclid` and `Manhattan`
- Milvus Lite `L2`

The issue is in the shared post-processing path, not in the backend's nearest-neighbor search itself.

## Evidence in the current code

- `KnowledgeBase.search` unconditionally filters with `result.score < score_threshold`, keeps the largest score during deduplication, and sorts scores in descending order:
  [`src/agentscope/rag/_knowledge.py`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/rag/_knowledge.py#L251-L272)
- `VectorSearchResult.score` documents that cosine/dot-product scores are higher-is-better, while L2 distance is lower-is-better:
  [`src/agentscope/rag/_vdb/_vector_store.py`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/rag/_vdb/_vector_store.py#L53-L64)
- Qdrant allows `Euclid` and `Manhattan` metrics and returns the backend `point.score` unchanged:
  [`src/agentscope/rag/_vdb/_qdrant.py`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/rag/_vdb/_qdrant.py#L58-L65)
- Milvus Lite allows `L2` and returns the raw `distance` value:
  [`src/agentscope/rag/_vdb/_milvus_lite.py`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/rag/_vdb/_milvus_lite.py#L45-L49)
  [`src/agentscope/rag/_vdb/_milvus_lite.py`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/rag/_vdb/_milvus_lite.py#L517-L533)

## Minimal reproduction

Using an in-memory Qdrant store configured with `distance="Euclid"`, I inserted three vectors and searched with the nearest vector as the query.

The backend returned the expected distance order:

```
nearest = 0.0
near    = 1.0
far     = 10.0
```

However, passing the same store through `KnowledgeBase.search(top_k=3)` returned:

```
far, near, nearest
```

With `score_threshold=0.5`, the result with distance `0.0` was filtered out, while the farther results with distances `1.0` and `10.0` were retained.

This can be reproduced with the public `QdrantStore(distance="Euclid")` configuration and does not require a race, a failed request, or an unusual deployment.

## Impact

This affects:

- top-k ranking
- score-threshold filtering
- selecting the best result when deduplicating multiple query results
- result ordering when multiple knowledge bases are merged by the RAG middleware

The default cosine configuration is not affected, but the other publicly supported metrics do not preserve their intended semantics through `KnowledgeBase`.

## Proposed direction

The vector-store abstraction should expose one consistent score contract to its consumers. One possible direction is to normalize distance-based backend scores at the vector-store boundary so that all returned scores use “higher means more similar”; another is to carry metric direction through the abstraction and make filtering, deduplication, and merging metric-aware.

I would be happy to follow the maintainers' preferred design. In either case, the fix should include regression coverage for Qdrant Euclid and, if practical, Milvus Lite L2, including ranking and threshold behavior.

Please review this proposal when you have a moment. I would like to take this fix myself. If the scope and direction look right, please explicitly confirm that I may implement it. I will wait for maintainer approval before changing code, and then prepare a focused, test-first PR with the regression tests and the required formatting and test results. I will not start implementation or open a PR before that approval.

lyh7c00:
## AgentScope Version

Working against `2.0.7.post1` (`import agentscope; print(agentscope.__version__)`).

## Description

`KnowledgeBase.search` assumed that a higher `VectorSearchResult.score` always means a more relevant result. That assumption is wrong for distance-based metrics, where a **lower** score is more relevant (Qdrant `Euclid` / `Manhattan`, Milvus Lite `L2`), so those stores' results were ranked in reverse — the nearest match ended up last.

This PR normalizes the score inside the affected vector stores so `VectorSearchResult.score` always means "higher is more relevant", which keeps the shared `KnowledgeBase.search` post-processing unchanged:

- `QdrantStore`: negates the score for `Euclid` / `Manhattan`.
- `MilvusLiteStore`: negates the score for `L2` in `_extract_score`.

`MongoDB` and `Elasticsearch` are left untouched — their scores are already higher-is-better for every metric.

Fixes #2482

## Changes

- `src/agentscope/rag/_vdb/_qdrant.py`: negate the distance score for `Euclid` / `Manhattan`.
- `src/agentscope/rag/_vdb/_milvus_lite.py`: negate the L2 distance score in `_extract_score`.
- `tests/rag_knowledge_search_test.py`: add tests that a distance metric yields higher-is-better scores and that `KnowledgeBase.search` ranks the nearest match first.

## Testing

- New tests (Qdrant `Euclid` score normalization + `KnowledgeBase.search` nearest-first) pass.
- Existing `rag_vdb_qdrant_test` and `rag_vdb_mongodb_test` still pass.

## Checklist


DavdGao (commented):
Thanks for catching this. But the metric direction is internal to the vector store — callers of `search` shouldn't have to know it. Suggest normalizing the score inside each store so `VectorSearchResult.score` always means "higher is more relevant", instead of exposing `score_higher_is_better` and branching on it. `MilvusLiteStore._extract_score` is already this hook, and `_knowledge.py` then needs no change.

src/agentscope/rag/_vdb/_mongodb.py
```diff
@@ -105,6 +105,12 @@ def __init__(
         self._client_kwargs = client_kwargs or {}
         self._client: "AsyncMongoClient | None" = None
 
+    @property
+    def score_higher_is_better(self) -> bool:
+        """MongoDB ``euclidean`` is a distance metric (lower-is-better);
+        ``cosine`` and ``dotProduct`` are similarity metrics."""
+        return self._distance in ("cosine", "dotProduct")
```
DavdGao: MongoDB's `vectorSearchScore` is normalized to 0-1 and higher-is-better for all three similarities, `euclidean` included. Marking it lower-is-better reverses a ranking that currently works — suggest dropping this and leaving `_mongodb.py` untouched.

lyh7c00:
Thanks for the review @DavdGao! Reworked to normalize in the stores instead: Qdrant now negates `Euclid`/`Manhattan` scores and Milvus negates `L2` in `_extract_score`, so `score` is always "higher is more relevant" and `KnowledgeBase.search` stays as is. Left `MongoDB` untouched per your note. Updated the tests too.

DavdGao (approved):

PR #2488, opened 2026-09-01
fix(workspace): recover expired OpenSandbox sandboxes

### LINKED ISSUE #2483: [Bug]: Lack of a lifecycle recovery mechanism after sandbox expiration.
### Prerequisites

### Background / Description

- `OpenSandboxWorkspaceManager.get_workspace()` first checks the in-memory cache;
- If there is a cache hit, it returns the existing `OpenSandboxWorkspace` directly without checking if the Sandbox is still active;
- `OpenSandboxWorkspace.initialize()` is executed only upon a cache miss;
- During initialization, it queries for the same `workspace.id`:
- If found in `Running` state: reconnect; 
- If found in `Paused` state: call `resume()`;
- If not found: create a new Sandbox.

- `/D:/a_bot/agentscope/src/agentscope/app/workspace_manager/_opensandbox_workspace_manager.py:247` Returns immediately on cache hit;
- `/D:/a_bot/agentscope/src/agentscope/workspace/_opensandbox/_opensandbox_workspace.py:159` Searches for an existing Sandbox during initialization;
- `/D:/a_bot/agentscope/src/agentscope/workspace/_opensandbox/_opensandbox_workspace.py:276` Connects to or resumes the Sandbox if it is `Running` or `Paused`. 

Consequently, the current behavior is:
Sandbox TTL expires and it is destroyed
↓
The workspace remains in the cache
↓
The next message continues using the old connection
↓
`httpx.ConnectError`

Automatic recreation or resumption occurs only if one of the following happens:

- The AgentScope process restarts;
- The workspace is cleared from the manager's idle cache;
- The workspace is manually closed and re-initialized;
- Code is explicitly added to handle reconnection/recreation after a connection failure.

Image

### Error Messages

```shell
The conversation will fail in 5 minutes.
```

### Steps to Reproduce

OPEN_SANDBOX_REQUEST_TIMEOUT=600
OPEN_SANDBOX_TIMEOUT_SECONDS=300

### Environment

- AgentScope Version:
- Python Version:
- OS:windows

RICARDO51167:
## AgentScope Version

2.0.7.post1

## Description

Fixes #2483.

OpenSandbox workspaces could remain cached after their remote sandbox had expired. Subsequent requests would reuse the stale local workspace and fail instead of recovering automatically.

This PR:

- renews the sandbox lease while a workspace is active;
- checks the remote sandbox state before reusing a cached workspace;
- reuses and renews running sandboxes;
- reattaches paused sandboxes;
- recreates expired, deleted, or terminal sandboxes under the same workspace ID;
- avoids duplicate sandbox creation during concurrent recovery;
- cleans up partially initialized workspaces on failure;
- adds unit tests covering lease renewal, expiration recovery, paused sandboxes, transient errors, cleanup, and concurrent requests.

### DIFF
--- a/src/agentscope/app/workspace_manager/_opensandbox_workspace_manager.py
+++ b/src/agentscope/app/workspace_manager/_opensandbox_workspace_manager.py
@@ -23,8 +23,8 @@ Differences from the Docker manager:
   ``workspace_id``.
 * Idle workspaces are evicted by a background sweeper task started in
   :meth:`__aenter__` and cancelled in :meth:`__aexit__`.
-* Cached workspaces are checked against the OpenSandbox control plane before
-  reuse. Paused sandboxes are reattached; expired sandboxes are recreated.
+* Cached workspaces renew their sandbox lease on access. A renewal 404
+  evicts the stale connection so the sandbox can be recreated.
 * ``close_all`` fans calls out with :func:`asyncio.gather` because
   ``sandbox.pause()`` is a remote round-trip per sandbox.
 """
@@ -104,7 +104,8 @@ class OpenSandboxWorkspaceManager(WorkspaceManagerBase):
                 command timeout.
             timeout_seconds (`int`, defaults to `DEFAULT_TIMEOUT`):
                 Sandbox expiration lease and create/connect/resume timeout.
-                Open workspaces renew the finite lease in the background.
+                Defaults to 1800 seconds and renews on cache access, not
+                in the background. Set it longer than the longest turn.
             gateway_port (`int`, defaults to `DEFAULT_GATEWAY_PORT`):
                 TCP port the in-sandbox gateway listens on.
             env (`dict[str, str] | None`, optional):
@@ -199,64 +200,6 @@ class OpenSandboxWorkspaceManager(WorkspaceManagerBase):
         await ws.initialize()
         return ws
 
-    async def _reuse_cached(
-        self,
-        workspace_id: str,
-    ) -> OpenSandboxWorkspace | None:
-        """Return a live cached workspace or evict its stale connection.
-
-        Remote lifecycle calls run outside :attr:`_lock` so one slow control
-        plane does not block unrelated workspace ids. Cache identity is
-        rechecked under the lock before a result is reused or evicted.
-
-        Args:
-            workspace_id (`str`):
-                Cache key to inspect.
-
-        Returns:
-            `OpenSandboxWorkspace | None`:
-                The renewed cached workspace, or ``None`` when the caller
-                must build a replacement.
-        """
-        while True:
-            async with self._lock:
-                cached = self._cache.get(workspace_id)
-            if cached is None:
-                return None
-            ws, _ = cached
-            # pylint: disable-next=protected-access
-            lifecycle = await ws._refresh_remote_lifecycle()
-
-            async with self._lock:
-                current = self._cache.get(workspace_id)
-                if current is None or current[0] is not ws:
-                    continue
-                if lifecycle == "running":
-                    self._cache[workspace_id] = (ws, time.monotonic())
-                    return ws
-                self._cache.pop(workspace_id)
-
-            if lifecycle == "reattach":
-                logger.info(
-                    "OpenSandbox workspace %r is paused; reattaching "
-                    "sandbox %r",
-                    workspace_id,
-                    ws.sandbox_id,
-                )
-            else:
-                logger.warning(
-                    "OpenSandbox workspace %r lost sandbox %r; recreating "
-                    "it. Ephemeral workspace data may have been lost.",
-                    workspace_id,
-                    ws.sandbox_id,
-                )
-            # Do not pause a sandbox that may already be paused or gone. Drop
-            # the SDK transport locally, then let normal initialization resume
-            # the matching paused sandbox or create a fresh one when absent.
-            # pylint: disable-next=protected-access
-            await ws._discard_local_connection()
-            return None
-
     async def get_workspace(
         self,
         user_id: str,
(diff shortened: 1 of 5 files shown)


RICARDO51167:
Additional context for reviewers:

While investigating #2483, I found that similar cache-versus-remote-lifecycle mismatches may also affect E2B, Daytona, Kubernetes, and remote Docker. I documented the detailed findings in this issue comment: [link to the issue comment].

This PR is intentionally scoped to OpenSandbox only. The other backends have different lifecycle semantics and should be handled separately:

- E2B may require timeout renewal and stale-sandbox recovery.
- Daytona may require refreshing the remote state and starting or resuming the sandbox.
- Kubernetes and remote Docker may need reconciliation when cached remote resources are externally removed or stopped.

src/agentscope/workspace/_opensandbox/_opensandbox_workspace.py
```diff
@@ -138,6 +143,9 @@ def __init__(
 
         self._sandbox: Sandbox | None = None
         self._backend: OpenSandboxBackend | None = None
+        self._gateway: GatewayClient | None
```
DavdGao: Bare annotation with no assignment, and the base class already declares and initializes `_gateway` (`_sandboxed_base.py:61,142`). Dead line -- remove it along with the `GatewayClient` import above.

src/agentscope/app/workspace_manager/_opensandbox_workspace_manager.py
```diff
@@ -242,13 +303,11 @@ async def get_workspace(
                 agent_id=agent_id,
                 session_id="",
             )
+        assert workspace_id is not None
```
DavdGao: `assign_workspace_id` returns `str`, so this can never fire. Drop it.

tests/workspace_manager_opensandbox_test.py
```diff
@@ -0,0 +1,377 @@
...
+        if self.renew_error is not None:
+            raise self.renew_error
+        self.renewed.append(timeout)
+
+    async def close(self) -> None:
+        """Record local SDK transport cleanup."""
+        self.closed = True
+
+
+class TestOpenSandboxWorkspaceLease(IsolatedAsyncioTestCase):
```
DavdGao: `TestOpenSandboxWorkspaceLease` is workspace-level, it fits better in the existing `tests/workspace_opensandbox_test.py`.

DavdGao (commented):
Root cause is right and `renew()` is the right API. But the lease itself is untouched -- still 300s, which is shorter than a single reply, and that is what forces the background task on top of it.

Suggest bumping `DEFAULT_TIMEOUT` to 1800 and just calling `renew()` once in `get_workspace` on a cache hit. A lease longer than any single reply means a turn can't expire mid-flight, and a 404 from `renew()` already tells us the sandbox is gone, so it doubles as the liveness check -- no `get_info()` needed. Then `_renew_loop`, `_refresh_remote_lifecycle`, `_discard_local_connection` and `_reuse_cached` can all go.

I'd rather not add a background task here if we can avoid it. It brings cancellation, retry and give-up logic that all has to be right, and it keeps renewing for as long as the workspace sits in the cache (ttl=3600) rather than for as long as it is actually used.

Two things about the current shape I'd like to avoid either way:

- The pre-check can't deliver what it promises -- `get_info()` says running, the sandbox can still die a second later and the request fails anyway. Meanwhile it costs two extra round-trips per call, and `get_workspace` is on the hot path (`_service/_workspace.py:285` backs every workspace file API).
- Raising on `pending`/`pausing`/unknown states and re-raising non-404 errors means a control-plane hiccup now fails every request, while today the bug only bites when the sandbox actually expires.

RICARDO51167:
Thanks for the detailed review. I've simplified the implementation along the suggested lines:

- `DEFAULT_TIMEOUT` is now 1800 seconds.
- Cache hits make a single `renew()` call, with no additional `get_info()` pre-check.
- A renewal 404 triggers cleanup and recreation; other renewal errors are logged without rejecting the cached workspace.
- The background renewal loop and the additional lifecycle/reuse/discard helpers have been removed.
- The redundant `_gateway` annotation/import and workspace ID assertion have been removed, and workspace-level tests have been moved to the existing workspace test file.

I kept a single on-demand renewal when reattaching an existing sandbox, since the SDK's connect/resume timeouts only bound readiness checks rather than extending the lease. Initialization cleanup and concurrent-recovery coverage are also retained.

src/agentscope/app/workspace_manager/_opensandbox_workspace_manager.py
```diff
@@ -236,36 +241,37 @@ async def get_workspace(
...
+        # two replacement sandboxes for the same workspace id.
         async with self._lock:
-            cached = self._cache.get(workspace_id)
+            cached = self._cache.get(resolved_id)
             if cached is not None:
                 ws, _ = cached
-                self._cache[workspace_id] = (ws, time.monotonic())
-                return ws
+                # pylint: disable-next=protected-access
+                if await ws._renew_once():
```
jianpingpei, line 281: **[P2] Avoid manager-wide blocking during cache-hit renewal**

Could we address the cross-workspace blocking here before merging? `_renew_once()` is awaited while holding the manager-wide `self._lock`, so one slow renewal now blocks even cache hits for unrelated workspace IDs. `close`, `close_all`, and the sweeper also need this lock.

I reproduced this locally against a real OpenSandbox/Docker service, comparing this PR (`250dd521`) with its base (`180b409b`). With two already-initialized, cached workspaces, I injected a 1-second delay before workspace A's SDK `renew()` call (the real HTTP renewal still ran), then requested workspace B 50 ms later:

- PR: B waited **0.996 seconds**.
- Base: B returned in **less than 1 ms**.

Please keep same-workspace recovery deduplicated without holding the global cache lock across the remote renewal, e.g. with per-workspace serialization and a short cache critical section. A focused regression test should block A's renewal and verify that cached B can still be acquired before A is released, while retaining the existing same-ID recovery test.

RICARDO51167: Thanks for the clear reproduction. Fixed in `4860cf4e` by using per-workspace locking and keeping remote renewals outside the manager-wide lock. Added the suggested cross-workspace regression test while retaining same-ID recovery coverage.

jianpingpei: Thanks for the fix. I rechecked `4860cf4e`: remote renewal now runs outside the manager-wide lock, while same-workspace recovery remains serialized. The close and sweep paths also coordinate through the per-workspace lock, with idle timestamps rechecked before eviction.

Local validation: 20 tests passed, 11 subtests passed, and 1 live-service test was skipped because `OPENSANDBOX_DOMAIN` was not set. I also ran the new cross-workspace regression test against the previous head (`250dd521`): it fails there and passes with this fix. I did not rerun the live Docker E2E in this round.

This addresses my blocking concern. This thread can be marked as resolved; my account does not have permission to resolve it directly.

PR #2489, opened 2026-09-01
fix(app): offload parser execution in IndexWorker

### LINKED ISSUE #2176: IndexWorker blocks the event loop: lifespan never wires up parser_executor (CPU parsers run in the loop thread)

Luious-LYH:
Fixes #2176

## What does this PR do?

- Offload parser execution with `asyncio.to_thread` when no
  `parser_executor` is provided.
- Preserve the explicit `parser_executor` path for callers that opt
  into process-based parsing.
- Apply the default offloading behavior at `IndexWorker._parse`, so it
  covers both embedded and dedicated workers.

## Why?

`ParserBase` requires parser instances to be stateless or thread-safe,
but does not require them to be pickleable.

Using an implicit process pool would introduce a serialization
requirement for otherwise valid custom parsers. For example, a parser
holding a `threading.Lock` cannot be sent through a
`ProcessPoolExecutor`.

Using `asyncio.to_thread` by default keeps synchronous parser work off
the worker event loop without imposing that pickleability requirement.
Callers that explicitly need multi-process parsing can still provide a
`parser_executor`.

## Tests

- Added coverage for a non-pickleable parser.
- Added coverage that blocking parser work does not block the event loop.
- Added coverage for parser exception propagation.
- Added coverage for the explicit executor path.
- Added a deadlock safety watchdog to keep the responsiveness regression test from hanging if parsing regresses to inline execution.
- Related targeted tests pass locally.

### DIFF
--- a/src/agentscope/app/_lifespan.py
+++ b/src/agentscope/app/_lifespan.py
@@ -2,6 +2,7 @@
 """The lifespan of the agent service."""
 import socket
 import uuid
+from concurrent.futures import ProcessPoolExecutor
 from contextlib import AsyncExitStack, asynccontextmanager
 from typing import TYPE_CHECKING, Any, AsyncIterator
 
@@ -216,6 +217,10 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
             # re-enqueues from storage.
             if enable_index_worker:
                 node_id = f"{socket.gethostname()}:{uuid.uuid4().hex[:8]}"
+                # PDF and Office parsers are CPU-bound. Keep their work off
+                # the API event loop and share one pool across documents.
+                parser_executor = ProcessPoolExecutor(max_workers=4)
+                stack.callback(parser_executor.shutdown, wait=False)
                 worker = IndexWorker(
                     storage=storage,
                     blob_store=blob_store,
@@ -223,6 +228,7 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
                     parsers=app.state.knowledge_parsers,
                     chunkers=app.state.knowledge_chunkers,
                     node_id=node_id,
+                    parser_executor=parser_executor,
                 )
                 await stack.enter_async_context(
                     IndexTaskConsumer(
--- a/src/agentscope/app/_lifespan.py
+++ b/src/agentscope/app/_lifespan.py
@@ -2,7 +2,6 @@
 """The lifespan of the agent service."""
 import socket
 import uuid
-from concurrent.futures import ProcessPoolExecutor
 from contextlib import AsyncExitStack, asynccontextmanager
 from typing import TYPE_CHECKING, Any, AsyncIterator
 
@@ -217,10 +216,6 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
             # re-enqueues from storage.
             if enable_index_worker:
                 node_id = f"{socket.gethostname()}:{uuid.uuid4().hex[:8]}"
-                # PDF and Office parsers are CPU-bound. Keep their work off
-                # the API event loop and share one pool across documents.
-                parser_executor = ProcessPoolExecutor(max_workers=4)
-                stack.callback(parser_executor.shutdown, wait=False)
                 worker = IndexWorker(
                     storage=storage,
                     blob_store=blob_store,
@@ -228,7 +223,6 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
                     parsers=app.state.knowledge_parsers,
                     chunkers=app.state.knowledge_chunkers,
                     node_id=node_id,
-                    parser_executor=parser_executor,
                 )
                 await stack.enter_async_context(
                     IndexTaskConsumer(
--- a/src/agentscope/app/_service/_index_worker.py
+++ b/src/agentscope/app/_service/_index_worker.py
@@ -169,12 +169,11 @@ class IndexWorker:
                 renews periodically so long-running parses do not
                 trip the sweeper.
             parser_executor (`ProcessPoolExecutor | None`, optional):
-                Process pool used to off-load CPU-intensive parses
-                (PDF, Office).  ``None`` runs parses in the event-loop
-                thread, which is fine for plain text but unsafe for
-                third-party byte-oriented parsers.  Injected so a
-                single pool can be shared across the app (built in
-                lifespan).
+                Optional process pool used to off-load CPU-intensive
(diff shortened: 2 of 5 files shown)


DavdGao (changes requested):
Thanks for the fix. However, a process pool requires every parser to be picklable, while `ParserBase` only promises stateless/thread-safe — a custom parser holding a client or a lock works today and would start failing with `cannot pickle '_thread.lock' object`. It also only covers embedded mode; `run_worker()` still blocks the same way, and there the lease heartbeat can't renew either, so long parses get swept and re-queued.

Suggested to fix it in `IndexWorker._parse` instead: when no executor is given, run the parser through `asyncio.to_thread` rather than awaiting it inline. Measured on a pure-Python CPU loop — inline the loop gets 0 ticks for the whole parse, through a thread it ticks every 10ms, same parse time. No pickling, no new contract, and it covers both modes in one place. `parser_executor` stays for people who really want multi-core.

Luious-LYH:
Thanks for the detailed explanation — that makes sense. Moving the fallback into IndexWorker._parse is cleaner and avoids introducing a pickling requirement for parsers, while also covering both embedded and standalone worker modes. I'll update the implementation to use asyncio.to_thread when no parser_executor is provided, keep the explicit executor path intact, and adjust the tests accordingly.

Luious-LYH:
Updated based on your suggestion. I moved the default parser offloading into `IndexWorker._parse`, using `asyncio.to_thread` when no `parser_executor` is provided while preserving the explicit executor path.

This now covers both embedded and dedicated workers without introducing a pickleability requirement for custom parsers.

I also added regression coverage for non-pickleable parsers, event-loop responsiveness during blocking parser work, exception propagation, and the explicit executor path. The related targeted tests pass locally.

Thanks for the guidance — ready for re-review.

PR #2499, opened 2026-09-04
fix(mcp): clean up cancelled MCP connections

### LINKED ISSUE #2458: [Bug]: cancelling MCPClient.connect() leaks a partially initialized connection
### Prerequisites

### Background / Description

`MCPClient.connect()` can be cancelled while a stateful transport and session
are being initialized.

If cancellation occurs after the transport has entered but before
`session.initialize()` completes, the cancellation propagates without closing
the partially entered `AsyncExitStack`.

The resulting client reports `is_connected == False`, but still retains the
transport, session, and connection stack. Its public `close()` method rejects
the cleanup because the client is not marked as connected, and reconnecting
the same instance can reuse the consumed one-shot transport.

### Expected Behavior

Cancellation during `connect()` should propagate after all partially
initialized connection resources have been closed, leaving the `MCPClient`
reusable for a subsequent `connect()`.

### Error Messages

```text
cancellation propagated
is_connected: False
transport exits: 0
session exits: 0
close: MCP 'cancelled_connect' is not connected. Call connect() first.
retry: transport was reused
```

### Steps to Reproduce

Save and run the following script against the current `main` branch. It uses
only deterministic fake contexts; no MCP server or network access is needed.

```python
import asyncio

from agentscope.mcp import MCPClient, StdioMCPConfig


async def main() -> None:
    initialize_started = asyncio.Event()
    never_finish = asyncio.Event()

    class OneShotTransport:
        def __init__(self) -> None:
            self.enter_count = 0
            self.exit_count = 0

        async def __aenter__(self):
            self.enter_count += 1
            if self.enter_count > 1:
                raise RuntimeError("transport was reused")
            return object(), object()

        async def __aexit__(self, *_):
            self.exit_count += 1

    class BlockingSession:
        instance = None

        def __init__(self, *_):
            self.exit_count = 0
            BlockingSession.instance = self

        async def __aenter__(self):
            return self

        async def __aexit__(self, *_):
            self.exit_count += 1

        async def initialize(self):
            initialize_started.set()
            await never_finish.wait()

    transport = OneShotTransport()
    with patch(
        "agentscope.mcp._mcp_client.stdio_client",
        return_value=transport,
    ), patch(
        "agentscope.mcp._mcp_client.ClientSession",
        BlockingSession,
    ):
        client = MCPClient(
            name="cancelled_connect",
            is_stateful=True,
            mcp_config=StdioMCPConfig(command="unused"),
        )

        task = asyncio.create_task(client.connect())
        await initialize_started.wait()
        task.cancel()
        try:
            await task
        except asyncio.CancelledError:
            print("cancellation propagated")

        session = BlockingSession.instance
        leaked_stack = client._stack
        print("is_connected:", client.is_connected)
        print("transport exits:", transport.exit_count)
        print("session exits:", session.exit_count)

        try:
            await client.close()
        except RuntimeError as error:
            print("close:", error)

        try:
            await client.connect()
        except RuntimeError as error:
            print("retry:", error)

        # Clean up the fake contexts retained by the current implementation.
        await leaked_stack.aclose()


asyncio.run(main())
```

The same lifecycle failure was also verified with the real MCP stdio
transport and a local child process. After cancellation, the subprocess
remained alive until the retained stack was manually closed.

### Relevant Code Paths

- `src/agentscope/mcp/_mcp_client.py`
- `tests/mcp_client_reconnect_test.py`

### Environment

- AgentScope Version: 2.0.7.post1 (`main` at `e90f1c7592896cc95f6e5ee506194f533378247d`)
- Python Version: 3.12.13
- OS: WSL2/Linux

I'd be happy to submit a focused fix with a deterministic regression test.

014-code:
## Summary

- Clean up transport and session contexts when `MCPClient.connect()` is cancelled during initialization.
- Reset the client lifecycle state so the same client can reconnect with a fresh one-shot transport.
- Add a deterministic regression test covering cancellation, cleanup, and retry.

## Tests

- `.venv\Scripts\python.exe -m pytest tests\mcp_client_reconnect_test.py -q`
- `.venv\Scripts\python.exe -m pytest tests\mcp_client_reconnect_test.py::MCPClientReconnectTest::test_cancelled_connect_closes_partial_connection -vv -s`
- `.venv\Scripts\python.exe -m compileall -q src tests`
- `git diff --check`

Fixes #2458

### DIFF
--- a/src/agentscope/mcp/_mcp_client.py
+++ b/src/agentscope/mcp/_mcp_client.py
@@ -241,21 +241,28 @@ class MCPClient(BaseModel):
                 self._initialize_client()
 
         assert self._client is not None
-        self._stack = AsyncExitStack()
+        stack = AsyncExitStack()
+        self._stack = stack
 
         try:
-            context = await self._stack.enter_async_context(self._client)
+            context = await stack.enter_async_context(self._client)
             read_stream, write_stream = context[0], context[1]
             self._session = ClientSession(read_stream, write_stream)
-            await self._stack.enter_async_context(self._session)
+            await stack.enter_async_context(self._session)
             await self._session.initialize()
 
             self._is_connected = True
             logger.info("MCP connected: %s", self.name)
-        except Exception:
-            await self._stack.aclose()
-            self._stack = None
-            self._client = None
+        except BaseException:
+            # asyncio.CancelledError inherits BaseException, so a cancelled
+            # initialization must close every context entered so far.
+            try:
+                await stack.aclose()
+            finally:
+                self._client = None
+                self._stack = None
+                self._session = None
+                self._is_connected = False
             raise
 
     async def close(self, ignore_errors: bool = True) -> None:
--- a/src/agentscope/mcp/_mcp_client.py
+++ b/src/agentscope/mcp/_mcp_client.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """Unified MCP client implementation for AgentScope."""
+import asyncio
 import re
 from contextlib import AsyncExitStack, _AsyncGeneratorContextManager
 from typing import Any, TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


DavdGao (approved):
LGMT

PR #2515, opened 2026-09-05
fix(app): drain wakeup backlog completely on startup

### LINKED ISSUE #2514: [Bug]: WakeupDispatcher strands startup backlog after one batch
### Prerequisites

- [x] I have searched the existing issues and pull requests, and this is not a duplicate.

### Summary

`WakeupDispatcher` only drains one batch of at most 64 entries when it starts. If more than 64 durable wakeup entries accumulated while the service was down, the remaining entries stay in the queue. When no new wakeup signal is published afterward, those entries are never dispatched.

This is a startup backlog bug and is distinct from [#2227](https://github.com/agentscope-ai/agentscope/issues/2227), which covers duplicate/lost resume-trigger delivery caused by concurrent multi-instance consumers and queue-consumption semantics. This issue occurs with a single dispatcher and a correctly functioning queue.

### Evidence in the current code

- `WakeupDispatcher.__aenter__` performs one initial `_drain_and_dispatch()` after subscribing:
  [`src/agentscope/app/_manager/_wakeup_dispatcher.py`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/app/_manager/_wakeup_dispatcher.py#L115-L131)
- `_drain_and_dispatch()` reads at most 64 entries:
  [`src/agentscope/app/_manager/_wakeup_dispatcher.py`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/app/_manager/_wakeup_dispatcher.py#L178-L209)
- The message-bus contract says entries beyond `max_count` remain in the queue for the next call:
  [`src/agentscope/app/message_bus/_base.py`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/app/message_bus/_base.py#L130-L155)

### Reproduction

Using the real `InMemoryMessageBus` and `WakeupDispatcher` path, I pre-populated 65 durable wakeup entries and invoked the dispatcher drain once:

```text
queued entries: 65
dispatched after the startup drain: 64
remaining entries: 1
```

The remaining entry is not dispatched unless another producer publishes a new wakeup signal. With a backlog created during downtime, there may be no such signal.

The same behavior follows directly from the current code: startup invokes the drain once, and each drain is capped at 64 entries.

### Expected behavior

After subscribing at startup, the dispatcher should continue draining batches until the durable wakeup queue is empty. A backlog of 65 or more entries should not require a new producer signal before the remaining entries are processed.

The normal signal-driven loop should keep its existing bounded batch behavior where appropriate, and the change should not alter the at-most-once queue contract.

### Impact

This can leave scheduled or externally triggered agent runs permanently stuck after a normal service restart or outage followed by a burst of wakeups. The queue still contains the entries, but the dispatcher has no polling or recovery path that will read them again.

### Proposed fix and tests

Keep the existing batch size, but make the startup recovery path drain repeatedly until a call returns fewer than the batch size (or an empty batch). Add a regression test with more than 64 pre-existing entries and assert that every entry is dispatched without publishing another signal.

I would like to take this fix myself and follow up with a focused PR linked to this issue, including the regression test and the required formatting and test results. Please review the scope and proposed direction when you have a moment; I will keep the change limited to the startup backlog behavior.

CoralGarden52:
Fixes #2514

## Summary

`WakeupDispatcher` previously performed only one queue drain when starting. Because each drain is capped at 64 entries, a durable backlog larger than 64 could leave entries stranded indefinitely when no new wakeup signal was published after startup.

## Changes

- Keep the queue operation bounded at 64 entries.
- Make the startup recovery path continue draining while it receives a full batch.
- Leave the normal signal-driven path unchanged: each signal still performs one bounded drain.
- Add a regression test with 65 pre-existing wakeup entries and no new signal.

The change is limited to the startup backlog behavior and does not alter the message-bus at-most-once contract or any public API.

## Validation

Targeted regression and existing dispatcher tests:

```text
10 passed in 4.61s  # Python 3.11
```

Changed-file pre-commit checks:

```text
AST, encoding, whitespace, trailing commas, mypy, Black, flake8,
pylint, and Pyroma: all passed
```

Real in-memory message-bus smoke test using `InMemoryMessageBus` and the production `WakeupDispatcher`:

```text
queued: 65
 dispatched: 65
 remaining: 0
```

The full test suite was also run under Python 3.11:

```text
2178 passed, 125 skipped, 11 failed, 388 subtests passed
```

The 11 failures were outside the changed files: five S3 blob-store tests failed while the local Moto server returned HTTP 502 during bucket creation, and six MCP SSE/streamable-HTTP tests failed in their test-service paths. The wakeup dispatcher tests and all changed-file checks passed.

## Checklist

- [x] An issue has been created for this PR: #2514
- [x] Related documentation is not required for this internal bug fix

Thank you for reviewing this focused fix.

### DIFF
--- a/src/agentscope/app/_manager/_wakeup_dispatcher.py
+++ b/src/agentscope/app/_manager/_wakeup_dispatcher.py
@@ -61,6 +61,10 @@ _RESUME_INPUT_ADAPTER: TypeAdapter = TypeAdapter(
 # to avoid a hot re-enqueue loop while the lock is held.
 _RESUME_RETRY_BACKOFF_SECS = 0.1
 
+# Keep each queue operation bounded while allowing startup to consume a
+# backlog over multiple calls.
+_WAKEUP_DRAIN_BATCH_SIZE = 64
+
 
 class WakeupDispatcher:
     """One asyncio task per process, draining the shared trigger queue.
@@ -127,7 +131,9 @@ class WakeupDispatcher:
             name="wakeup-dispatcher",
         )
         await ready.wait()
-        await self._drain_and_dispatch()
+        while True:
+            if not await self._drain_and_dispatch():
+                break
         return self
 
     async def __aexit__(self, *exc: object) -> None:
@@ -175,17 +181,23 @@ class WakeupDispatcher:
                 "WakeupDispatcher loop crashed; subscription ended.",
             )
 
-    async def _drain_and_dispatch(self) -> None:
-        """Read up to a batch of trigger entries and dispatch each."""
+    async def _drain_and_dispatch(self) -> bool:
+        """Read a batch of trigger entries and dispatch each.
+
+        Returns:
+            `bool`:
+                Whether the queue returned a full batch and may still have
+                more entries to drain.
+        """
         try:
             raw_entries = await self._bus.queue_drain(
                 MessageBusKeys.wakeup_queue(),
-                max_count=64,
+                max_count=_WAKEUP_DRAIN_BATCH_SIZE,
             )
             entries = [payload for _, payload in raw_entries]
         except Exception:  # pylint: disable=broad-except
             logger.exception("WakeupDispatcher: dequeue_wakeups failed.")
-            return
+            return False
 
         for payload in entries:
             try:
@@ -207,6 +219,8 @@ class WakeupDispatcher:
                 kind=kind,
                 raw_input=payload.get("input"),
             )
+
+        return len(raw_entries) == _WAKEUP_DRAIN_BATCH_SIZE
 
     async def _dispatch_one(
         self,
--- a/tests/service_wakeup_dispatcher_test.py
+++ b/tests/service_wakeup_dispatcher_test.py
@@ -302,6 +302,34 @@ class TestWakeupDispatcherDispatch(IsolatedAsyncioTestCase):
             ],
         )
 
+    async def test_initial_drain_consumes_backlog_beyond_batch_size(
+        self,
+    ) -> None:
+        """Startup drains every pending entry without a new signal."""
+        bus = _FakeBus()
+        chat = _FakeChatService()
+        queue_key = MessageBusKeys.wakeup_queue()
+        for index in range(65):
+            await bus.queue_push(
+                queue_key,
+                {
+                    "user_id": "u",
+                    "session_id": f"pre-{index}",
+                    "agent_id": "a",
+                },
+            )
+
+        async with WakeupDispatcher(
+            message_bus=bus,
+            storage=_FakeStorage(),
+            chat_service=chat,
+            chat_run_registry=ChatRunRegistry(),
+        ):
+            await _yield_a_few_times()
+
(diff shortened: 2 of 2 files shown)


DavdGao:
The loop in `__aenter__` can wedge startup. `_dispatch_one` re-queues a trigger whose session lock is held (`_schedule_retry`, 0.1s), so while the loop drains, retries keep refilling the queue and the batch never comes back short. `WakeupDispatcher` is entered in the lifespan exit stack, so uvicorn never finishes startup and the port never opens. With a 1ms Redis round trip this wedges from ~300 backlog entries, and it doesn't need a second instance — the runs this loop spawns take the locks themselves, so a few hundred entries on a handful of sessions is enough.

It looks green because `_FakeBus` never awaits, so the drain loop never yields and the 0.1s retry timers cannot run at all. Any bus that does I/O interleaves them.

- Suggested to drop the loop and re-drive the existing signal path: after the initial drain, if the batch was full, publish the wakeup signal and return. `_loop` does the next bounded drain and publishes again if that was also full, so the chain self-terminates and `__aenter__` stays O(1).
- `return False` in the `except` says "drained" when the read actually failed, so one transient Redis error mid-recovery leaves the backlog stranded again — that needs to be distinguishable from a real drain.
- `len(raw_entries) == _WAKEUP_DRAIN_BATCH_SIZE` should be `>=`.

Also needs a rebase, the branch is 61 commits behind main.

CoralGarden52:
Thank you for the detailed review. I rebased the branch onto the latest `upstream/main` and changed startup recovery to perform one bounded drain only. A full batch now re-publishes the wakeup signal so the long-lived loop drains subsequent batches, keeping `__aenter__` bounded and preventing retry timers from wedging startup. Queue-read failures now return a distinct result and are retried through the signal path, while full-batch detection uses `>=`. I also added regression tests covering asynchronous locked-session retries and a transient queue-read failure. Targeted tests and all pre-commit checks pass. Please review when you have a moment. If you have any further suggestions or concerns, I’m happy to revise the patch.

CoralGarden52:
I also manually set up a local test matching the scenario you described as closely as possible.

I preloaded 300 durable wakeups across five locked sessions and added a 1 ms async delay to the queue, lock, and publish operations so that the retry timers could run while the dispatcher was working.

With the previous startup loop, `__aenter__` was still running after 0.5 seconds because the deferred retries kept adding entries back to the queue. With the updated implementation, `__aenter__` returned after about 0.074 seconds, following a single bounded drain, while the remaining entries were left for the normal signal-driven loop.

I also tested 300 unlocked wakeups separately and confirmed that the signal-driven loop processed all 300 entries, leaving the durable queue empty. Finally, I injected a one-time failure into the second queue read and confirmed that the remaining backlog was retried and fully processed.
 
All 12 `WakeupDispatcher` tests and the complete pre-commit checks pass.

Please review it again when you have time. If you notice anything else that should be adjusted, I will be happy to make the changes. Thank you for your time and guidance.

PR #2517, opened 2026-09-05
chore(model): use polymorphic block methods in count_tokens

Okabe-Rintarou-0:
## AgentScope Version

`2.0.8`

### Background

`ChatModelBase.count_tokens` used a long `isinstance` chain to walk every
content block, extract its text, and collect `DataBlock`s. The `str | list`
branching for `HintBlock.hint` and `ToolResultBlock.output` was duplicated,
and any future block type would force another edit to `count_tokens`.

### Changes

- Add two polymorphic methods to each content block (`TextBlock`,
  `ThinkingBlock`, `HintBlock`, `ToolCallBlock`, `ToolResultBlock`,
  `DataBlock`):
  - `extract_text() -> list[str]` — returns the text segments this block
    contributes.
  - `count_data_blocks() -> int` — returns the number of data blocks nested
    in this block.
- `HintBlock` and `ToolResultBlock` handle their own `str | list` field
  inline in these two methods.
- Rewrite `count_tokens` to call these methods instead of the `isinstance`
  chain. Since only the *count* of data blocks was used, it now accumulates
  an integer instead of materializing the full `data_blocks` list.

### How to test

```bash
pytest tests/model_count_tokens_test.py tests/message_test.py tests/model_base_test.py
Checklist
- An issue has been created for this PR
- I have read the CONTRIBUTING.md (https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md)
- Docstrings are in Google style
- Related documentation has been updated (e.g. links, examples, etc.) in documentation repository (https://github.com/agentscope-ai/docs)
- Code is ready for review

### DIFF
--- a/src/agentscope/message/__init__.py
+++ b/src/agentscope/message/__init__.py
@@ -2,6 +2,7 @@
 """The message module in agentscope."""
 
 from ._block import (
+    BlockBase,
     ContentBlock,
     ContentBlockTypes,
     TextBlock,
@@ -19,6 +20,7 @@ from ._base import Msg, UserMsg, AssistantMsg, SystemMsg, Usage
 
 
 __all__ = [
+    "BlockBase",
     "TextBlock",
     "ThinkingBlock",
     "HintBlock",
--- a/src/agentscope/message/_block.py
+++ b/src/agentscope/message/_block.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """The content blocks of messages."""
+from abc import abstractmethod
 from enum import StrEnum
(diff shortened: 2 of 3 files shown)


DavdGao:
The `isinstance` chain in `count_tokens` is worth cleaning up, but not this way. `extract_text()` and `count_data_blocks()` are token-counting concerns, and putting them on every block type means a `TextBlock` carries a method that returns 0 only because `count_tokens` asks. That, plus a new exported `BlockBase`, changes the framework's block interface — not a chore.

Closing this. If we want to restructure `count_tokens`, let's settle the design in its own issue first.

PR #2519, opened 2026-09-05
fix(app): clear inbox consumer on session delete

### LINKED ISSUE #2518: [Bug]: deleting a session leaves a stale inbox consumer marker
## Bug description

`SessionService._purge_session_bus()` removes a session's event log, inbox queue, and background-task registry, but it does not remove the `MessageBusKeys.inbox_consumer(session_id)` marker.

`register_inbox_consumer()` creates that marker with `SESSION_RUN_TTL_SECS` (currently 600 seconds). When a session run exits abnormally, the marker can remain behind. If the session is then deleted and later recreated with the same explicit `session_id`, `deliver_to_inbox()` sees the stale marker and does not enqueue a wakeup for the new session.

The message is written to the inbox but can remain there until the stale marker expires or another wakeup happens to be produced.

## Reproduction

Using the project's `RedisMessageBus` with a fakeredis-backed connection and calling the real `SessionService._purge_session_bus()` produced:

```text
marker before purge: '1'
marker TTL: 600
marker after purge: '1'
wakeup count after reuse: 0
inbox after reuse: [('...', {'text': 'late'})]
```

This only requires a normal session deletion followed by reuse of the same explicit session ID; it does not depend on a timing race.

## Expected behavior

Purging a session should remove its inbox-consumer marker. After the same session ID is reused, the first delivery should enqueue a wakeup so the new session can drain its inbox.

## Proposed scope

- Drop `MessageBusKeys.inbox_consumer(session_id)` in `_purge_session_bus()`.
- Add a regression test covering delete, reuse of the same ID, and wakeup delivery.

No API changes should be required.

Please review this issue when you have a moment. I would like to take this fix myself; if the scope and direction look right, please confirm that I may submit the focused PR with the regression test. Thanks!

CoralGarden52:
Fixes #2518

## Summary

- Remove the session's `inbox_consumer` marker when purging its bus state.
- Add a regression test for deleting a session, reusing the same explicit ID, and delivering a new inbox message.

Without this cleanup, a stale marker can suppress the wakeup for the reused session even though the message is successfully written to its inbox.

## Validation

- `uv run --project . --extra dev pytest tests/service_inbox_handoff_test.py tests/service_message_bus_test.py tests/in_memory_message_bus_test.py -q` — 81 passed.
- `uv run --project . --extra dev pre-commit run --all-files` — all hooks passed.
- The new regression test failed before the production change because no wakeup was emitted, and passes after the change.
- `uv run --project . --extra dev pytest tests -q` — no failures in the touched subsystem; the full run also reported 11 unrelated S3/MCP integration failures (Moto `CreateBucket` 502 responses and external MCP HTTP integration failures).

### DIFF
--- a/src/agentscope/app/_service/_session.py
+++ b/src/agentscope/app/_service/_session.py
@@ -658,6 +658,9 @@ class SessionService:
             MessageBusKeys.inbox(session_id),
         )
         await self._bus.registry_drop(
+            MessageBusKeys.inbox_consumer(session_id),
+        )
+        await self._bus.registry_drop(
             MessageBusKeys.bg_tasks(session_id),
         )
 
--- a/tests/service_inbox_handoff_test.py
+++ b/tests/service_inbox_handoff_test.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+# pylint: disable=protected-access
 """Tests for the session-inbox hand-off protocol in ``_bus_ops``.
 
 The protocol exists so a payload pushed to a session inbox is always
@@ -23,6 +24,7 @@ from agentscope.app._bus_ops import (
     has_pending_inbox_or_release,
     register_inbox_consumer,
 )
+from agentscope.app._service import SessionService
 from agentscope.app.message_bus import InMemoryMessageBus, MessageBusKeys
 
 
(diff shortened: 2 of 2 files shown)


DavdGao (approved):
lgtm

PR #2521, opened 2026-09-05
fix(channel): preserve wakeups for messages arriving during run finalization

### LINKED ISSUE #2520: fix(channel): preserve wakeups for messages arriving during run finalization
## Summary

A channel message can be left permanently in a session inbox when it arrives while the active run is finishing.

`ChatService` uses the inbox hand-off protocol introduced by #2295:

- the run registers `inbox_consumer` while it owns the session;
- the final `has_pending_inbox_or_release()` check drains the inbox and removes that marker when it is empty;
- producers use `deliver_to_inbox()` under `inbox_lock`, then enqueue a wakeup when no consumer remains.

However, `ChannelGateway._handle_message()` still checks `session_lock` and directly calls `queue_push(inbox, ...)`. It does not use `deliver_to_inbox()`.

Relevant code:

- [`ChannelGateway._handle_message`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/app/channel/_gateway.py)
- [`deliver_to_inbox`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/app/_bus_ops.py)
- [`has_pending_inbox_or_release`](https://github.com/agentscope-ai/agentscope/blob/main/src/agentscope/app/_bus_ops.py)

## Reproduction

Using the production `ChannelGateway.process()` path and production `RedisMessageBus` backed by fakeredis, I scheduled the final inbox check before the channel event was handled, while holding the session lock:

```text
PENDING_AT_FINAL_CHECK=False
CONSUMER_MARKER_AFTER_FINAL_CHECK=None
INBOX_ENTRIES=1
DURABLE_WAKEUP_ENTRIES=0
```

The same boundary using the existing `deliver_to_inbox()` helper produced:

```text
PENDING_AT_FINAL_CHECK=False
INBOX_ENTRIES=1
DURABLE_WAKEUP_ENTRIES=1
```

With 50 repetitions and no patched production methods or injected sleeps, scheduling the final check first stranded all 50 channel messages without a wakeup. Starting the channel task first let the finishing run observe the message, which confirms this is the missing hand-off protocol rather than intended inbox behavior.

## Impact

This is a normal ordering during a channel run's finalization, not a provider-specific or low-probability failure. The message is persisted in the inbox, but no dispatcher is signaled after the run releases its lock. If no later event wakes the session, the channel message is never processed.

## Proposed scope

Route the active-run ChannelGateway hint through the existing inbox hand-off helper and add a regression test covering a channel message arriving across the finalization boundary. No public API or message format change should be required.

I would like to take this fix myself and keep it focused on the channel producer path plus a regression test. Please review the scope and direction when you have a moment; I will prepare the linked focused PR with the test evidence above.

CoralGarden52:
Fixes #2520

## Summary

`ChannelGateway` used to push hints directly to a running session's inbox after checking `session_lock`. That bypassed the inbox hand-off protocol used by `ChatService` and could leave a channel message queued without a wakeup when it arrived after the run's final inbox check.

## Changes

- Route active-run Channel hints through `deliver_to_inbox()`.
- Add a deterministic regression test through the public `ChannelGateway.process()` path.
- Assert the wakeup payload and serialized inbox hint remain compatible with the existing protocol.

The change does not alter public APIs or the Channel message format.

## Validation

- Focused Channel/inbox tests: `47 passed`.
- Final regression and hand-off tests after review: `8 passed`.
- Full pre-commit suite: passed, including AST, mypy, Black, Flake8, Pylint, and Pyroma.
- Full test suite: `2178 passed, 125 skipped, 11 failed`.
  The 11 failures are existing environment-dependent failures: five S3 Moto tests received HTTP 502 during bucket creation, and six MCP SSE/streamable-HTTP tests could not complete their test-service paths.
- Production `RedisMessageBus` backed by fakeredis smoke test:

```text
PUBLIC_PROCESS_PENDING_AT_FINAL_CHECK=False
INBOX_ENTRIES=1
DURABLE_WAKEUP_ENTRIES=1
WAKEUP_SESSION_ID_MATCH=True
```

This is intentionally limited to the missing producer hand-off; no unrelated refactor is included.

### DIFF
--- a/src/agentscope/app/channel/_gateway.py
+++ b/src/agentscope/app/channel/_gateway.py
@@ -22,7 +22,7 @@ from ..._logging import logger
 from ...message import DataBlock, HintBlock, TextBlock, UserMsg
 from ...permission import PermissionContext, PermissionMode
 from ...state import AgentState
-from .._bus_ops import enqueue_run_trigger
+from .._bus_ops import deliver_to_inbox, enqueue_run_trigger
 from ..message_bus import MessageBus, MessageBusKeys
 from ..storage import (
     ChannelRecord,
@@ -205,9 +205,12 @@ class ChannelGateway:
         # A reply already in flight → inject the input as a hint so the
         # live run folds it in. Otherwise start a fresh user turn.
         if await self._bus.is_locked(MessageBusKeys.session_lock(session_id)):
-            await self._bus.queue_push(
-                MessageBusKeys.inbox(session_id),
-                HintBlock(
+            await deliver_to_inbox(
+                self._bus,
+                user_id=record.user_id,
+                session_id=session_id,
+                agent_id=agent_id,
+                payload=HintBlock(
                     hint=content,
                     source=json.dumps(
                         {
--- a/tests/channel_gateway_test.py
+++ b/tests/channel_gateway_test.py
@@ -8,10 +8,15 @@ needs a running agent and is exercised end-to-end against a real bot.
 """
 # pylint: disable=protected-access,missing-function-docstring,unused-argument
 # pylint: disable=attribute-defined-outside-init
+import asyncio
 from types import SimpleNamespace
 from typing import Any, AsyncIterator
 from unittest import IsolatedAsyncioTestCase
 
+from agentscope.app._bus_ops import (
+    has_pending_inbox_or_release,
+    register_inbox_consumer,
+)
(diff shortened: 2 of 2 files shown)


DavdGao (approved):
lgtm

PR #2528, opened 2026-09-06
fix(rag): escape Excel Markdown table cells

MaxFreedomPollard:
## AgentScope Version

2.0.8 (`main`, base commit `41ba021`)

## Description

`ExcelParser` renders a sheet through its own cell formatter, `_fmt` at `src/agentscope/rag/_parser/_excel.py:419`, which escapes the pipe character and nothing else. Two ordinary kinds of spreadsheet cell corrupt the table it produces.

A cell containing a line break (alt+enter in Excel) ends the Markdown row early. `_extract_table_data` normalises CRLF to LF at `_excel.py:74` and keeps the break, so the text after it lands on its own physical line and the row is cut short there.

A cell containing a backslash in front of a pipe loses a column. `Path \| label` is escaped to `Path \\| label`, and Markdown reads `\\` as a literal backslash and the `|` after it as an unescaped column separator, so one cell renders as two.

`WordParser` and `PPTParser` are not affected: both render through the shared `_format_markdown_table_cell` in `src/agentscope/rag/_parser/_utils.py:50`, which escapes backslashes and pipes and turns line breaks into `<br>`. The Excel path predates that helper and was never switched over to it.

Repro on `main`, with a sheet whose cells are `A|B`, `Path \| label`, `1|2` and `Line 1\nLine 2`:

```
| A\|B | Path \\| label |
| --- | --- |
| 1\|2 | Line 1
Line 2 |
```

After this change:

```
| A\|B | Path \\\| label |
| --- | --- |
| 1\|2 | Line 1<br>Line 2 |
```

## Change

`_fmt` calls `_format_markdown_table_cell` instead of its own single `replace`. The `[A1]` cell-coordinate prefix and the `table_format="json"` path are untouched, and cells with no pipe, backslash or line break render exactly as before.

## Tests

Added `ExcelParserTest.test_markdown_table_escapes_special_cells`, which mirrors the Word and PPT tests of the same name and asserts the whole parsed `Section` dump. It fails on `main` (`Path \\| label` and the split row) and passes with the change.

```bash
python -m pytest tests/rag_parser_test.py -q
# 44 passed

python -m pytest tests/ -k rag -q
# 228 passed, 1 skipped

pre-commit run --files src/agentscope/rag/_parser/_excel.py tests/rag_parser_test.py
# all hooks passed
```

Run on macOS with Python 3.11, pandas 3.0.5 and openpyxl 3.1.5. The new test uses only `_make_xlsx_simple`, the helper the existing Excel tests already build workbooks with, so it needs nothing that CI does not already install.

## Checklist

### DIFF
--- a/src/agentscope/rag/_parser/_excel.py
+++ b/src/agentscope/rag/_parser/_excel.py
@@ -192,11 +192,7 @@ class ExcelParser(ParserBase):
                 sheets are never combined into the same section.
             table_format (`Literal["markdown", "json"]``, defaults to
                 ``"markdown"``):
-                How to render tables.  ``"markdown"`` uses pipe-table
-                syntax, escaping pipes and rendering cell line breaks
-                as ``<br>``; ``"json"`` emits a JSON array prefixed with
-                a ``<system-info>`` marker and preserves extracted cell
-                strings without Markdown rendering.
+                How to render tables.
 
         Raises:
             `ValueError`: If ``table_format`` is not ``"markdown"``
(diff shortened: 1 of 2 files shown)


DavdGao (approved):
lgtm

PR #2532, opened 2026-09-07
feat(model): add Volcengine Ark chat model support

### LINKED ISSUE #2448: feat(models): add Volcengine Ark (Doubao) chat model support
**What's missing**

AgentScope supports DashScope, DeepSeek, Moonshot, Gemini, Anthropic, xAI and Ollama,
but not Volcengine Ark (火山方舟), which serves ByteDance's Doubao models. There's no
`volcengine` / `doubao` / `volces` reference anywhere in the tree today, so users on Ark
have to fall back to a generic OpenAI-compatible path and lose the provider-specific
pieces (thinking control, credential handling, model metadata).

**Why it fits**

Ark exposes an OpenAI-compatible `/chat/completions` endpoint, so this maps cleanly onto
the existing `_deepseek` shape rather than needing anything novel. Verified against the
live API today (2026-08-27):

- `POST https://ark.cn-beijing.volces.com/api/v3/chat/completions` → 200
- tool calling returns standard `tool_calls`
- reasoning is exposed as `choices[].message.reasoning_content`, and depth is graded via
  `reasoning_effort` — measured on `doubao-seed-2-1-pro-260628`: `low` → 61 reasoning
  tokens, `high` → 107, while `none`/`minimal` switch thinking off

**Proposed scope**

Following the `_deepseek` precedent, this touches three subsystems:

- `src/agentscope/model/_volcengine/` — `VolcengineChatModel` (+ `Parameters`)
- `src/agentscope/credential/_volcengine.py` — `VolcengineCredential`
- `src/agentscope/formatter/_volcengine_formatter.py` — `VolcengineChatFormatter`
- registration in the three `__init__.py` files, plus unit tests

One PR, one purpose — no refactors bundled in.

**Disclosure and what we can bring**

I work on Volcengine Ark, so this is a first-party integration rather than a
best-guess from the outside. Practically that means I can verify behaviour against the
real API instead of inferring it from docs — which matters here, because Ark's docs and
its actual behaviour diverge in places. Two examples I've already hit and would encode
in the implementation:

- model IDs must carry their full version suffix; the console's short names
  (`doubao-seed-2.1-pro`) return `404 InvalidEndpointOrModel.NotFound`
- `thinking.type = auto` is documented as valid but is rejected by 11 of 12 models

**Questions before I write anything**

1. Are you open to a Volcengine Ark provider in-tree, or would you rather this live
   outside the repo?
2. If yes — is the three-file shape above (model + credential + formatter, mirroring
   `_deepseek`) the structure you want, or is there a newer pattern I should follow?

Happy to take the `help wanted` route if you'd prefer to frame the task yourselves.
I'll hold off on code until there's a reply.

iluv7:
## AgentScope Version

`2.0.8`

## Description

Closes #2448

This PR adds first-class support for Volcengine Ark (Doubao) through Ark's OpenAI-compatible Chat Completions API. It addresses #2448 so Ark users no longer need to configure the provider through the generic OpenAI integration and can use provider-specific credentials, thinking controls, response parsing, model metadata, and formatters.

The implementation follows the existing DeepSeek provider structure and adds:

- `VolcengineChatModel` with streaming and non-streaming response handling.
- Ark thinking controls through `thinking.type` and `reasoning_effort`.
- Parsing of `reasoning_content` into `ThinkingBlock` objects.
- Tool-call parsing and tool-choice formatting.
- Usage parsing, including cached input tokens from `prompt_tokens_details.cached_tokens`.
- `VolcengineCredential`, including the default Ark endpoint and registration in `CredentialFactory`.
- `VolcengineChatFormatter` and `VolcengineMultiAgentFormatter`, including multi-turn preservation of reasoning content and tool-result formatting.
- Public exports from the model, credential, and formatter packages.
- A model card for the live-verified Doubao model.
- Unit tests for the model, formatters, credentials, model metadata, thinking, tool calls, tool results, streaming, and usage.

No new dependency is introduced. The integration reuses the existing `openai.AsyncClient` dependency and defaults to:

```text
https://ark.cn-beijing.volces.com/api/v3
```

### How to test

The following automated checks were run:

```bash
PYTHONPATH=src .venv/bin/pytest -q \
  tests/model_volcengine_test.py \
  tests/formatter_volcengine_test.py \
  tests/model_deepseek_test.py \
  tests/formatter_deepseek_test.py \
  tests/model_base_test.py
```

Result: `71 passed`.

The changed files also passed:

- Black 23.3.0 with a line length of 79.
- Flake8 6.1.0.
- mypy 1.7.0 with the repository's pre-commit arguments.

The complete test suite was also started, but collection could not complete in the local environment because several unrelated optional test dependencies were not installed, including `fakeredis`, `boto3`, `mem0`, `sqlalchemy`, and `redis`.

### Live Ark E2E verification

The integration was verified against the real Volcengine Ark API with `doubao-seed-2-1-pro-260628`. The API key was supplied only through an environment variable and was not written to the repository or test fixtures.

The following live scenarios passed:

1. **Non-streaming thinking response**
   - Sent `thinking.type=enabled` and `reasoning_effort=low`.
   - Parsed the response into one `ThinkingBlock` and one `TextBlock`.
   - Parsed usage successfully: 51 input tokens, 32 output tokens, and 0 cached input tokens.

2. **Streaming thinking response**
   - Received 15 incremental responses.
   - Aggregated them into a final response with `is_last=True`.
   - The final response contained a `ThinkingBlock` and a `TextBlock`.
   - Parsed usage successfully: 51 input tokens, 28 output tokens, and 0 cached input tokens.

3. **Forced tool call**
   - Registered a `get_weather` function schema.
   - Forced the function with `ToolChoice(mode="get_weather")`.
   - Received exactly one `ToolCallBlock` with the expected function name and non-empty arguments.
   - Parsed usage successfully: 476 input tokens and 35 output tokens.

4. **Multi-turn reasoning preservation**
   - Converted the first live response into an `AssistantMsg` containing its thinking and text blocks.
   - Sent it back as conversation history in a second request.
   - Ark accepted the preserved `reasoning_content`, and the second response again parsed into a `ThinkingBlock` and a `TextBlock`.

5. **Tool-result round trip**
   - Sent a live tool call back with its corresponding `ToolResultBlock`.
   - Ark accepted the formatted assistant tool call and tool result.
   - The follow-up request returned a final `TextBlock` successfully.

## Checklist

- [x] An issue has been created for this PR: #2448
- [ ] Related documentation has been updated in the [documentation repository](https://github.com/agentscope-ai/docs)

## Current model-card scope

This PR currently adds a model card only for `doubao-seed-2-1-pro-260628`, because it is the model explicitly identified in #2448 and verified against the live Ark API during this implementation.

`VolcengineChatModel` itself accepts any Ark model ID or endpoint ID, so the provider implementation is not restricted to Doubao Seed 2.1 Pro. Additional models should receive their own cards after their exact versioned IDs, capabilities, context/output limits, thinking behavior, and supported `reasoning_effort` values have been verified against the live API.


src/agentscope/model/_volcengine/_model.py
```diff
@@ -0,0 +1,447 @@
...
+            default=False,
+            title="Thinking",
+            description=(
+                "Whether to enable thinking mode. When enabled, the model "
+                "outputs a chain-of-thought reasoning before the final "
+                "answer via the reasoning_content field."
+            ),
+        )
+
+        reasoning_effort: (
```
DavdGao, line 49: `max` is a leftover from the DeepSeek copy — Ark documents `reasoning_effort` as `minimal` / `low` / `medium` / `high` (https://www.volcengine.com/docs/82379/1449737), and the model card in this PR narrows it to exactly those four, so the Literal contradicts your own card.

Suggest fixing it here instead: `Literal["minimal", "low", "medium", "high"]`, then the card override can be dropped entirely.

src/agentscope/model/_volcengine/_models/doubao-seed-2-1-pro-260628.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+output_types:
+  - text/plain
+  - application/x-thinking
+
+context_size: 256000
+output_size: 256000
+
+parameter_overrides:
+  reasoning_effort:
```
DavdGao: This override has no effect. The base schema of an optional Literal is `anyOf`, and `ModelCard.from_yaml` merges with a shallow `{**base, **override}`, so the result keeps the original `anyOf` (still containing `max`) and only gains a stray top-level `enum`. The frontend resolves the enum from `anyOf` (`resolveType()` in `ModelParametersPopover.tsx`), so `max` is still offered in the UI.

- If an override is really needed, use the `anyOf` form like `_anthropic/_models/claude-sonnet-4-6.yaml`; fixing the Literal makes it unnecessary.
- `test_verified_model_card` asserts on the stray `enum` key, so it passes while the override does nothing — better to assert on the schema the frontend actually reads.

src/agentscope/model/_volcengine/_models/doubao-seed-2-1-pro-260628.yaml
```diff
@@ -0,0 +1,22 @@
...
+input_types:
+  - text/plain
+  - application/x-thinking
+
+output_types:
+  - text/plain
+  - application/x-thinking
+
+context_size: 256000
+output_size: 256000
```
DavdGao, line 19: Where does `output_size: 256000` come from? It's the only card in the repo where output equals context, and public sources disagree (256k vs 128k). It's auto-injected as `max_tokens.maximum`, so a wrong value lets the UI submit a request the API will reject. Please add a source comment, like `kimi-k2.7-code.yaml` does.

src/agentscope/model/_volcengine/_model.py
```diff
@@ -0,0 +1,447 @@
...
+
+        if self.parameters.top_p is not None:
+            kwargs["top_p"] = self.parameters.top_p
+
+        if self.parameters.reasoning_effort is not None:
+            kwargs["reasoning_effort"] = self.parameters.reasoning_effort
+
+        kwargs.update(generate_kwargs)
+
+        thinking_type = (
```
DavdGao: Ark's `thinking.type` defaults to `auto` and Seed 1.6+/2.x rely on that to decide per request. Here it is forced to `disabled` on every call unless the user sets `thinking_enable`, so thinking never happens by default, and models that cannot turn thinking off will reject the request.

Suggest `thinking_enable: bool | None = None`, and only write `thinking` into `extra_body` when it is explicitly set.

src/agentscope/formatter/_volcengine_formatter.py
```diff
@@ -0,0 +1,311 @@
...
+)
+
+
+class VolcengineChatFormatter(FormatterBase):
+    """The Volcengine Ark formatter for a chatbot conversation.
+
+    The ``role`` field identifies the user and assistant participants.
+    """
+
+    input_types: list[str] = Field(
```
DavdGao, line 93: doubao-seed-2.1-pro takes image and video input, but this formatter is copied from the DeepSeek one, so `input_types` is text-only and an `ImageBlock` is silently dropped by the warning at L167 — the multi-agent formatter's description even says "Volcengine does not support multimodal input".

Suggest following `_moonshot_formatter.py` instead; Ark's `image_url` accepts both URLs and base64 data URIs, so it is simpler than the Moonshot one. The model card's `input_types` needs `image/*` as well.

src/agentscope/credential/_factory.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """The credential factory class."""
+
```
DavdGao: Unrelated whitespace change (same for the deleted blank line in `credential/__init__.py`), please revert.

DavdGao (commented):
Thanks for the work, the implementation runs and the tests/lint are green. Two things that are not tied to a specific line:

- The provider isn't registered for tracing, so `_get_provider_name()` returns `"unknown"` for `VolcengineChatModel`. Please add `ProviderNameValues.VOLCENGINE` in `middleware/_tracing/_attributes.py`, `"volcengine"` in `_CLASS_NAME_MAP`, and a `volces.com` entry in `_BASE_URL_PROVIDER_MAP` so users who still reach Ark through `OpenAIChatModel` also get the right provider name.
- Missing `scripts/model_examples/volcengine_call.py` and `volcengine_multiagent.py` plus the `Provider(...)` entry in `run_tests.py` — every other provider has them. The provider list in `README.md` / `README_zh.md` needs Volcengine too.

src/agentscope/model/_volcengine/_model.py
iluv7: Fixed in 760829c7. I narrowed `reasoning_effort` to `Literal["minimal", "low", "medium", "high"]`, matching Ark’s documented values. The parameter validation test now also verifies that max is rejected.

src/agentscope/model/_volcengine/_models/doubao-seed-2-1-pro-260628.yaml
iluv7: Fixed in 760829c7. I removed the unnecessary YAML override after correcting the `Literal` in the model parameters. I also updated the model-card test to assert against the enum inside the generated `anyOf` schema that the frontend actually reads.

src/agentscope/model/_volcengine/_models/doubao-seed-2-1-pro-260628.yaml
iluv7: Addressed in f3e87e5f. I added a source comment linking to the official Ark model documentation, which lists the maximum answer length for `doubao-seed-2-1-pro-260628` as 256k tokens.

src/agentscope/model/_volcengine/_model.py
iluv7: Fixed in 41fdeda9. `thinking_enable` now defaults to `None`, and the `thinking` field is only added to `extra_body` when the user explicitly sets it. Leaving it unset therefore preserves Ark’s default `auto` behavior. Tests cover both the default case and explicit disabling.

src/agentscope/formatter/_volcengine_formatter.py
iluv7: Fixed in e40e6ba7. I reworked the formatter following the OpenAI/Moonshot multimodal approach. Both chat and multi-agent formatters now preserve image and video inputs, including remote URLs, base64 data URIs, and local files. Multimodal data is also retained in hints and promoted tool results. The model card now declares both `image/*` and `video/*`, with tests covering the new behavior.

src/agentscope/credential/_factory.py
iluv7: Reverted in 8ce875e0.

iluv7:
Addressed both points in 4ab6a3c9. I registered Volcengine for tracing, including the provider name, class-name, and `volces.com` base-URL mappings, with tests for both `VolcengineChatModel` and Ark accessed through `OpenAIChatModel`. I also added the single-agent and multi-agent examples, registered Volcengine in `run_tests.py` using `VOLCENGINE_API_KEY`, and updated the provider lists in `README.md` and `README_zh.md`.

iluv7:
The CI failure is unrelated to the Volcengine Ark changes in this PR. `mcp 1.30.0` was released between the previous successful run and this run, and it intentionally changed the default `follow_redirects` value returned by `create_mcp_http_client()` from `True` to `False`. Since AgentScope currently installs `mcp<2.0.0` without a lock file, CI picked up the new release while `test_owning_the_client_keeps_transport_defaults` still hard-coded the old default.

I submitted #2542 to fix the test by deriving the expected transport settings from the MCP SDK factory. The fix has been verified against both MCP 1.28.1 and 1.30.0.

DavdGao (approved):
lgtm

PR #2538, opened 2026-09-07
fix(rag): preserve blank lines between paragraphs in WordParser

### LINKED ISSUE #2535: [Bug]: WordParser drops blank paragraphs when parsing DOCX files
## Prerequisites

- [x] I searched existing issues and discussions and found no duplicate.
- [x] This is a bug report, not a usage question.

## Background / Description

`WordParser` discards truly empty Word paragraphs (`<w:p>`), so blank lines between surrounding text paragraphs are lost.

## Actual Result

A DOCX containing `Paragraph one.`, an empty paragraph, and `Paragraph two.` is parsed as:

```text
Paragraph one.
Paragraph two.
```

## Error Messages

No exception is raised. The incorrect parser output above is the observable error: the empty Word paragraph is silently discarded.

## Expected Result

The empty Word paragraph should be preserved as a blank line:

```text
Paragraph one.

Paragraph two.
```

Consecutive empty paragraphs should preserve each corresponding blank line. Leading or all-empty paragraphs should not create an empty `Section`.

## Steps to Reproduce

1. Install the RAG extra: `uv pip install "agentscope[rag]"`.
2. Run:

```python
import asyncio
import io

from docx import Document
from agentscope.rag import WordParser


async def main() -> None:
    document = Document()
    document.add_paragraph("Paragraph one.")
    document.add_paragraph("")
    document.add_paragraph("Paragraph two.")

    buffer = io.BytesIO()
    document.save(buffer)
    sections = await WordParser(include_image=False).parse(
        buffer.getvalue(),
        "demo.docx",
    )
    print(sections[0].content.text)


asyncio.run(main())
```

3. Observe that the printed text contains only one newline between the paragraphs.

## Proposed Scope

- Preserve structurally empty paragraphs after existing text content.
- Do not treat image-only (`w:drawing` / `w:pict`) or other structurally non-empty paragraphs as blank paragraphs.
- Add in-memory DOCX regression coverage without binary fixtures.

## Environment

- AgentScope Version: 2.0.8
- Python Version: 3.13.9
- OS: macOS 14.8.9 (arm64)

liugy789:
## AgentScope Version

2.0.8

## Description

Closes #2535.

Word represents a blank line as a structurally empty `w:p` paragraph. `WordParser` previously discarded it, so text on both sides was joined with one newline and the blank line was lost.

This change preserves a structurally empty paragraph after existing text by appending an empty text-buffer entry. It accepts only paragraphs with no direct children other than `w:pPr`; image-only paragraphs (`w:drawing` / `w:pict`), break-only runs, and other structurally non-empty paragraphs are not treated as blank lines. Leading or all-empty paragraphs still do not create an empty `Section`.

## Testing

- `pytest tests/rag_parser_test.py` — 46 passed
- Changed-file `pre-commit` hooks — passed: mypy, black, flake8, pylint, pyroma, and repository checks
- Added in-memory DOCX regression coverage for one blank paragraph and consecutive blank paragraphs
- `pytest tests` locally: 2183 passed, 147 skipped, 3 failed. The same three failures reproduce unchanged on `main` with Python 3.13: one MCP schema docstring-format assertion and two Alembic locale-encoding failures.

## Checklist

- [x] Related documentation has been updated — n/a; this corrects existing parser behavior and adds no public API, configuration, or newly documented feature

### DIFF
--- a/src/agentscope/rag/_parser/_word.py
+++ b/src/agentscope/rag/_parser/_word.py
@@ -71,16 +71,6 @@ def _extract_text_from_paragraph(para: DocxParagraph) -> str:
     return text.strip()
 
 
-def _is_truly_blank_paragraph(para: DocxParagraph) -> bool:
-    """Return whether a paragraph has no content beyond properties."""
-    from docx.oxml.ns import qn
-
-    paragraph_properties_tag = qn("w:pPr")
-    return all(
-        child.tag == paragraph_properties_tag for child in para._element
-    )
-
-
 def _extract_table_data(table: DocxTable) -> list[list[str]]:
     """Extract table data from a python-docx Table, preserving line breaks
     within cells.
@@ -290,6 +280,8 @@ class WordParser(ParserBase):
         text_buffer: list[str] = []
 
         def flush_text() -> None:
+            while text_buffer and text_buffer[-1] == "":
+                text_buffer.pop()
             if not text_buffer:
                 return
             sections.append(
@@ -308,7 +300,9 @@ class WordParser(ParserBase):
                 text = _extract_text_from_paragraph(para)
                 if text:
                     text_buffer.append(text)
-                elif text_buffer and _is_truly_blank_paragraph(para):
+                elif text_buffer and all(
+                    child.tag == qn("w:pPr") for child in para._element
+                ):
                     text_buffer.append("")
 
                 if self.include_image:
--- a/src/agentscope/rag/_parser/_word.py
+++ b/src/agentscope/rag/_parser/_word.py
@@ -300,9 +300,10 @@ class WordParser(ParserBase):
                 text = _extract_text_from_paragraph(para)
                 if text:
                     text_buffer.append(text)
-                elif text_buffer and all(
-                    child.tag == qn("w:pPr") for child in para._element
+                elif text_buffer and not para._element.findall(
+                    ".//" + qn("w:r"),
                 ):
+                    # No runs at all, i.e. a blank line in Word
                     text_buffer.append("")
 
                 if self.include_image:
--- a/src/agentscope/rag/_parser/_word.py
+++ b/src/agentscope/rag/_parser/_word.py
@@ -71,6 +71,16 @@ def _extract_text_from_paragraph(para: DocxParagraph) -> str:
     return text.strip()
 
 
+def _is_truly_blank_paragraph(para: DocxParagraph) -> bool:
+    """Return whether a paragraph has no content beyond properties."""
+    from docx.oxml.ns import qn
+
+    paragraph_properties_tag = qn("w:pPr")
+    return all(
+        child.tag == paragraph_properties_tag for child in para._element
+    )
+
+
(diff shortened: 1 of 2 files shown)


src/agentscope/rag/_parser/_word.py
```diff
@@ -298,6 +308,8 @@ def flush_text() -> None:
                 text = _extract_text_from_paragraph(para)
                 if text:
                     text_buffer.append(text)
+                elif text_buffer and _is_truly_blank_paragraph(para):
```
DavdGao: Trailing blank paragraphs will leave trailing newlines in the flushed text, e.g. `["Paragraph one.", "", ""]` gives `"Paragraph one.\n\n"`, and a blank paragraph right before a table/image gives `"Paragraph one.\n"`. Suggested to drop the trailing empty entries in `flush_text` (e.g. `"\n".join(text_buffer).rstrip("\n")`) and add a test for it.

src/agentscope/rag/_parser/_word.py
```diff
@@ -71,6 +71,16 @@ def _extract_text_from_paragraph(para: DocxParagraph) -> str:
     return text.strip()
 
 
+def _is_truly_blank_paragraph(para: DocxParagraph) -> bool:
```
DavdGao: This helper is only used once and is a one-liner, and `qn` is already imported in `parse`. Suggested to inline it:

```python
elif text_buffer and all(c.tag == qn("w:pPr") for c in para._element):
```

src/agentscope/rag/_parser/_word.py
```diff
@@ -298,6 +308,8 @@ def flush_text() -> None:
                 text = _extract_text_from_paragraph(para)
                 if text:
                     text_buffer.append(text)
+                elif text_buffer and _is_truly_blank_paragraph(para):
+                    text_buffer.append("")
```
DavdGao, line 307: Minor: a whitespace-only paragraph (e.g. `"   "`) is also a blank line in Word but is still dropped here since it has a `w:r` child. Fine to leave as is, just noting the inconsistency.

src/agentscope/rag/_parser/_word.py
liugy789: Thanks for catching this. Fixed in f9faf4bb. `flush_text()` now removes only trailing empty buffer entries, so trailing newlines already present in table Markdown remain unchanged. Added a regression test for trailing blank paragraphs.

src/agentscope/rag/_parser/_word.py
liugy789: Thanks for the suggestion. In f9faf4bb, the one-use structural predicate is now inlined in the parsing logic.

src/agentscope/rag/_parser/_word.py
liugy789: Thanks for the clarification! We will keep the current behavior in this PR.

DavdGao (approved):

PR #2544, opened 2026-09-08
fix(pipeline): honor verifier context reset

### LINKED ISSUE #2543: [Bug]: GoalPipeline ignores verifier_reset_context after a failed verification
### Problem

`GoalPipeline.__init__` exposes `verifier_reset_context: bool = True`, documented as resetting the verifier context after each iteration. On current `main` (`ff8697e`), the value is stored but never read.

After a verifier returns `result="fail"`, the pipeline sends its feedback to the executor and starts another verifier round. The verifier keeps its prior `state.context` (and any summary), so the next verdict sees stale execution-report and verdict history even though the option defaults to resetting it.

### Expected behavior

With `verifier_reset_context=True`, a completed failed verification clears only the verifier conversation state before the next verification round. The executor state, tool/task state, and HITL resumes must remain untouched. With `False`, current accumulation behavior is preserved.

### Reproduction

A scripted executor that reports twice and a verifier that fails once then passes makes two verifier calls. Seed the verifier state with one context message before running the pipeline: it remains present before the second verifier call despite the default option being `True`.

### Proposed scope

Add a focused reset of the verifier conversation state after a completed `fail` verdict, plus offline regression coverage for the default and opt-out paths. This does not change the verifier protocol, retry accounting, or HITL behavior.

yaodong-shen:
Fixes #2543

## AgentScope Version

2.0.8

## Description

`GoalPipeline` documented and stored `verifier_reset_context`, but never read it. A verifier that rejected one execution report therefore carried its prior conversation and summary into the next verification attempt.

This PR clears only the verifier conversation state after a completed `fail` verdict when the option is enabled (the default). It leaves executor state, tool/task configuration, middleware state, structured-output retries, and HITL resumes unchanged. Setting `verifier_reset_context=False` preserves the prior accumulating behavior.

The regression coverage verifies both the default reset path and the opt-out path.

## Validation

- `PYTHONPYCACHEPREFIX=/private/tmp/agentscope_pycache .venv/bin/pytest tests/pipeline_goal_test.py -q` — 12 passed
- `PYTHONPYCACHEPREFIX=/private/tmp/agentscope_pycache .venv/bin/pre-commit run --all-files` — passed

## Checklist

- [x] No documentation update is needed: the public option already documents the intended behavior

### DIFF
--- a/src/agentscope/pipeline/_goal_pipeline.py
+++ b/src/agentscope/pipeline/_goal_pipeline.py
@@ -90,6 +90,20 @@ class GoalPipeline:
         self._iters = 0
         self._goal: None | list[TextBlock | DataBlock] = None
 
+    def _reset_verifier_context(self) -> None:
+        """Clear conversation state before the next verification round.
+
+        A failed verdict starts a new executor-verifier iteration. Keeping the
+        verifier's prior conversation in that next iteration contradicts
+        ``verifier_reset_context`` and can make a new report look like a
+        continuation of the rejected one. Deliberately leave the rest of the
+        verifier state intact: tool configuration, tasks, middleware state,
+        and an in-progress HITL reply all belong to the verifier itself rather
+        than to its conversation history.
+        """
+        self.verifier.state.context.clear()
+        self.verifier.state.summary = ""
+
     async def reply_stream(
         self,
         inputs: Msg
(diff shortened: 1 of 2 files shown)


DavdGao (approved):
lgtm

PR #2551, opened 2026-09-08
fix(agent): preserve tool data-block identity during event replay

### LINKED ISSUE #2549: [Bug]: streamed tool data blocks lose identity during event replay
### Prerequisites

- [x] I searched existing issues, pull requests, discussions, and recent history for this symptom and root cause.
- [x] This is a reproducible bug rather than a usage question.

### Background / Description

A streaming tool can return multiple `DataBlock` chunks with the same block ID. `ToolChunk` explicitly uses that stable ID to group one multimodal result, and `ToolResponse.append_chunk` merges same-ID Base64 chunks by decoded bytes.

The Agent event path loses that identity. `_convert_tool_chunk_to_event` creates each `ToolResultDataDeltaEvent` without passing `block.id`, so every event receives a new generated `block_id`. `Msg.append_event` then appends every data delta as a separate `DataBlock`.

As a result, the canonical tool response / Agent context contains one complete block, while the persisted reply and SSE-reconstructed message contain multiple partial blocks with unrelated IDs. Audio, file, or image chunks can therefore be exposed as separate incomplete resources after persistence or replay.

### Reproduction

On current `main` at `f8b40f60`, use two independently Base64-encoded chunks with the same source block ID:

```python
chunks = [
    DataBlock(
        id="audio-1",
        source=Base64Source(
            data=base64.b64encode(payload).decode("ascii"),
            media_type="audio/wav",
        ),
    )
    for payload in (b"hello", b"world")
]
```

Pass both through `Agent._convert_tool_chunk_to_event`, apply the resulting events to an `AssistantMsg` between `ToolResultStartEvent` and `ToolResultEndEvent`, and inspect the reconstructed tool result.

Observed before a fix:

```text
event block IDs: two different generated IDs
reconstructed block IDs: the same two generated IDs
reconstructed payloads: [b"hello", b"world"]
```

Expected, matching `ToolResponse.append_chunk`:

```text
event block IDs: ["audio-1", "audio-1"]
reconstructed block IDs: ["audio-1"]
reconstructed payloads: [b"helloworld"]
```

The regression test fails deterministically on current `main`; no timing or external service is involved.

### Root Cause

1. `Agent._convert_tool_chunk_to_event` omits `block_id=block.id` for both Base64 and URL `DataBlock` sources.
2. `Msg.append_event` handles every `TOOL_RESULT_DATA_DELTA` by appending a new block and has no same-ID Base64 accumulation path.

This violates the cross-layer invariant that replaying emitted events must reconstruct the same block boundaries, IDs, and bytes as the canonical `ToolResponse` stored in Agent context.

### Proposed Direction

- Preserve `DataBlock.id` when producing both Base64 and URL `ToolResultDataDeltaEvent` instances.
- When replaying a Base64 data delta into a tool result, merge it with an existing same-ID Base64 block by decoding both chunks, concatenating bytes, and re-encoding. This should mirror the behavior introduced for `ToolResponse.append_chunk` in #1901.
- Keep URL blocks one-shot and leave unrelated event behavior unchanged.
- Add regression coverage for same-ID padded Base64 chunks and URL ID preservation.

I have validated this focused direction locally with the new regression coverage plus `event_to_message_test.py` and `toolkit_test.py` (26 tests), and all file-level pre-commit hooks. I can prepare the focused PR after maintainer confirmation.

### Environment

- AgentScope Version: `2.0.8`
- Commit: `f8b40f60`
- Python Version: `3.11.16`
- OS: macOS

dvd233:
## AgentScope Version

`main` @ f8b40f6 (2.0.7)

## Description

Fixes #2549.

A streaming tool can emit multiple `DataBlock` chunks that share one block id — the documented `ToolChunk` contract ("for one multimodal data, the DataBlock instance should have the same block id, so that the agent can group them together"). Two places broke that identity:

1. `Agent._convert_tool_chunk_to_event` built each `ToolResultDataDeltaEvent` without `block.id`, so `block_id` fell back to a fresh `_generate_id()` per chunk.
2. `Msg.append_event` appended a separate `DataBlock` per delta with no grouping, while the canonical `ToolResponse.append_chunk` merges same-id Base64 chunks byte-wise.

The result: the canonical tool response held one complete block, but the persisted reply / SSE-reconstructed message contained multiple partial blocks with unrelated ids — audio/file/image chunks surfaced as separate incomplete resources after replay.

Changes:

- `_convert_tool_chunk_to_event`: pass `block_id=block.id` on both the Base64 and URL paths.
- `Msg.append_event` (TOOL_RESULT_DATA_DELTA): merge a delta into an existing same-id Base64 block (byte merge, mirroring `ToolResponse.append_chunk`); different ids still produce separate blocks. A small `_merge_base64_delta` helper lives in `message/_base.py` because the message package sits below the tool package in the dependency graph.

How to test: new `tests/tool_result_data_merge_test.py` covers the two-chunk same-id replay (merged, decodes to the concatenated bytes) and the different-id case (kept separate). Red → green: the merge test fails before the change and passes after; `event_to_message_test.py`, `event_test.py`, `agent_basic_test.py`, `compress_tool_result_test.py`, `model_response_test.py` all pass. black 23.3.0 / flake8 clean.

## Checklist

- [x] An issue has been created for this PR (#2549)
- [x] Related documentation has been updated (N/A — internal event/message behavior)

---

> This PR was prepared with AI coding assistance; the trace, patch, and red→green verification were reviewed and run locally by me.

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -3896,6 +3896,10 @@ class Agent:
                     yield ToolResultDataDeltaEvent(
                         reply_id=self.state.reply_id,
                         tool_call_id=tool_call_id,
+                        # Preserve the block id so replay can group chunks of
+                        # one multimodal payload (ToolChunk's documented
+                        # contract) instead of generating a new id per chunk.
+                        block_id=block.id,
                         media_type=block.source.media_type,
                         data=block.source.data,
                     )
@@ -3903,6 +3907,7 @@ class Agent:
                     yield ToolResultDataDeltaEvent(
                         reply_id=self.state.reply_id,
                         tool_call_id=tool_call_id,
+                        block_id=block.id,
                         media_type=block.source.media_type,
                         url=str(block.source.url),
                     )
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -3896,9 +3896,7 @@ class Agent:
                     yield ToolResultDataDeltaEvent(
                         reply_id=self.state.reply_id,
                         tool_call_id=tool_call_id,
-                        # Preserve the block id so replay can group chunks of
-                        # one multimodal payload (ToolChunk's documented
-                        # contract) instead of generating a new id per chunk.
+                        # Keep the block id so replay groups one payload.
                         block_id=block.id,
                         media_type=block.source.media_type,
                         data=block.source.data,
--- a/src/agentscope/message/_base.py
+++ b/src/agentscope/message/_base.py
@@ -465,27 +465,30 @@ class Msg(BaseModel):
                             media_type=event.media_type,
                         )
                     )
-                    # Merge same-id Base64 chunks, mirroring ToolResponse.append_chunk.
-                    existing = next(
-                        (
-                            b
-                            for b in block.output
-                            if isinstance(b, DataBlock)
-                            and b.id == event.block_id
-                            and isinstance(b.source, Base64Source)
-                        ),
-                        None,
-                    )
-                    if existing is not None and isinstance(src, Base64Source):
+                    # Merge same-id Base64 chunks, mirroring
+                    # ToolResponse.append_chunk.
+                    merge_target: DataBlock | None = None
+                    for output_block in block.output:
+                        if (
+                            isinstance(output_block, DataBlock)
+                            and output_block.id == event.block_id
+                            and isinstance(output_block.source, Base64Source)
+                        ):
+                            merge_target = output_block
+                            break
+                    if merge_target is not None and isinstance(
+                        src,
+                        Base64Source,
+                    ):
                         # Each delta is an independently encoded chunk (with
                         # its own padding); decode, concat bytes, re-encode.
                         merged = base64.b64decode(
-                            existing.source.data,
+                            merge_target.source.data,
                         ) + base64.b64decode(src.data)
-                        existing.source.data = base64.b64encode(
+                        merge_target.source.data = base64.b64encode(
                             merged,
                         ).decode("ascii")
-                        existing.source.media_type = src.media_type
+                        merge_target.source.media_type = src.media_type
                     else:
                         block.output.append(
                             DataBlock(id=event.block_id, source=src),
--- a/src/agentscope/message/_base.py
+++ b/src/agentscope/message/_base.py
@@ -55,24 +55,6 @@ def _to_blocks(content: str | list) -> list:
     return content
 
 
-def _merge_base64_delta(existing: str, incoming: str) -> str:
-    """Merge independently encoded base64 chunks without corrupting padding.
-
-    Mirrors tool._response._merge_base64_chunks (kept local here because the
-    message package sits below the tool package in the dependency graph).
-    """
-    try:
-        merged = base64.b64decode(existing, validate=True) + base64.b64decode(
-            incoming,
-            validate=True,
-        )
-    except Exception:
-        # Keep compatibility with placeholder strings that are not valid
-        # base64 payloads (e.g. some tests / hand-built fixtures).
-        return existing + incoming
-    return base64.b64encode(merged).decode("ascii")
-
-
 class Usage(BaseModel):
     """The token usage information of a message."""
 
@@ -465,8 +447,7 @@ class Msg(BaseModel):
                             media_type=event.media_type,
                         )
                     )
-                    # Merge same-id Base64 chunks, mirroring
-                    # ToolResponse.append_chunk.
+                    # Merge same-id Base64 chunks like ToolResponse.append_chunk.
                     merge_target: DataBlock | None = None
                     for output_block in block.output:
                         if (
@@ -480,13 +461,10 @@ class Msg(BaseModel):
                         src,
                         Base64Source,
                     ):
-                        # Each delta is an independently encoded chunk (with
-                        # its own padding); decode, concat bytes, re-encode.
-                        merged = base64.b64decode(
-                            merge_target.source.data,
-                        ) + base64.b64decode(src.data)
+                        existing = base64.b64decode(merge_target.source.data)
+                        incoming = base64.b64decode(src.data)
                         merge_target.source.data = base64.b64encode(
-                            merged,
+                            existing + incoming,
                         ).decode("ascii")
                         merge_target.source.media_type = src.media_type
                     else:
--- a/src/agentscope/message/_base.py
+++ b/src/agentscope/message/_base.py
@@ -465,23 +465,27 @@ class Msg(BaseModel):
                             media_type=event.media_type,
                         )
                     )
-                    # Group same-id Base64 chunks of one multimodal payload,
-                    # mirroring ToolResponse.append_chunk. Previously every
-                    # delta was appended as a separate block, splitting one
-                    # resource into multiple partial blocks on replay.
-                    existing = block.output[-1] if block.output else None
-                    if (
-                        isinstance(existing, DataBlock)
-                        and existing.id == event.block_id
-                        and isinstance(existing.source, Base64Source)
-                        and isinstance(src, Base64Source)
-                    ):
-                        existing.source.data = _merge_base64_delta(
+                    # Merge same-id Base64 chunks, mirroring ToolResponse.append_chunk.
+                    existing = next(
+                        (
+                            b
+                            for b in block.output
+                            if isinstance(b, DataBlock)
+                            and b.id == event.block_id
+                            and isinstance(b.source, Base64Source)
+                        ),
+                        None,
+                    )
+                    if existing is not None and isinstance(src, Base64Source):
+                        # Each delta is an independently encoded chunk (with
+                        # its own padding); decode, concat bytes, re-encode.
+                        merged = base64.b64decode(
                             existing.source.data,
-                            src.data,
-                        )
-                        if src.media_type:
-                            existing.source.media_type = src.media_type
+                        ) + base64.b64decode(src.data)
+                        existing.source.data = base64.b64encode(
+                            merged,
+                        ).decode("ascii")
+                        existing.source.media_type = src.media_type
                     else:
                         block.output.append(
                             DataBlock(id=event.block_id, source=src),
--- a/src/agentscope/message/_base.py
+++ b/src/agentscope/message/_base.py
@@ -55,6 +55,24 @@ def _to_blocks(content: str | list) -> list:
     return content
 
 
+def _merge_base64_delta(existing: str, incoming: str) -> str:
+    """Merge independently encoded base64 chunks without corrupting padding.
+
+    Mirrors tool._response._merge_base64_chunks (kept local here because the
+    message package sits below the tool package in the dependency graph).
+    """
+    try:
+        merged = base64.b64decode(existing, validate=True) + base64.b64decode(
+            incoming,
+            validate=True,
+        )
+    except Exception:
+        # Keep compatibility with placeholder strings that are not valid
+        # base64 payloads (e.g. some tests / hand-built fixtures).
+        return existing + incoming
+    return base64.b64encode(merged).decode("ascii")
+
+
 class Usage(BaseModel):
     """The token usage information of a message."""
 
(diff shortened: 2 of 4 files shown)


src/agentscope/message/_base.py
```diff
@@ -55,6 +55,24 @@ def _to_blocks(content: str | list) -> list:
     return content
 
 
+def _merge_base64_delta(existing: str, incoming: str) -> str:
```
DavdGao, line 58: This duplicates `_merge_base64_chunks` in `tool/_response.py`, and the `DATA_BLOCK_DELTA` branch below already does the same decode/concat/encode inline. Suggested to drop this helper and inline it the same way as `DATA_BLOCK_DELTA`.

src/agentscope/message/_base.py
```diff
@@ -55,6 +55,24 @@ def _to_blocks(content: str | list) -> list:
...
+
+    Mirrors tool._response._merge_base64_chunks (kept local here because the
+    message package sits below the tool package in the dependency graph).
+    """
+    try:
+        merged = base64.b64decode(existing, validate=True) + base64.b64decode(
+            incoming,
+            validate=True,
+        )
+    except Exception:
```
DavdGao, line 69: The string-concat fallback in the tool layer only exists to keep old tests with placeholder strings alive, no need to carry it into the message layer. Silently concatenating invalid base64 just hides bad data, and `DATA_BLOCK_DELTA` doesn't do this either. Suggested to remove the try/except.

src/agentscope/message/_base.py
```diff
@@ -447,9 +465,27 @@ def append_event(  # pylint: disable=too-many-branches,too-many-statements
                             media_type=event.media_type,
                         )
                     )
-                    block.output.append(
-                        DataBlock(id=event.block_id, source=src),
-                    )
+                    # Group same-id Base64 chunks of one multimodal payload,
```
DavdGao: Please keep the comment to one or two lines, e.g. `# Merge same-id Base64 chunks, mirroring ToolResponse.append_chunk`.

src/agentscope/message/_base.py
```diff
@@ -447,9 +465,27 @@ def append_event(  # pylint: disable=too-many-branches,too-many-statements
...
+                    # mirroring ToolResponse.append_chunk. Previously every
+                    # delta was appended as a separate block, splitting one
+                    # resource into multiple partial blocks on replay.
+                    existing = block.output[-1] if block.output else None
```
DavdGao: Only checking the last block doesn't match `ToolResponse.append_chunk`, which looks up by id across the whole content. For a stream like `data(a) -> text -> data(a)` the replay still ends up with two blocks. Suggested to look up the same-id Base64 `DataBlock` in `block.output` instead of just `output[-1]`.

src/agentscope/message/_base.py
```diff
@@ -447,9 +465,27 @@ def append_event(  # pylint: disable=too-many-branches,too-many-statements
...
+                        isinstance(existing, DataBlock)
+                        and existing.id == event.block_id
+                        and isinstance(existing.source, Base64Source)
+                        and isinstance(src, Base64Source)
+                    ):
+                        existing.source.data = _merge_base64_delta(
+                            existing.source.data,
+                            src.data,
+                        )
+                        if src.media_type:
```
DavdGao: `media_type` is a required `str` on the event, this check is always true. Just assign it directly.

src/agentscope/agent/_agent.py
```diff
@@ -3896,13 +3896,18 @@ async def _convert_tool_chunk_to_event(
                     yield ToolResultDataDeltaEvent(
                         reply_id=self.state.reply_id,
                         tool_call_id=tool_call_id,
+                        # Preserve the block id so replay can group chunks of
```
DavdGao: One line is enough here, or just drop the comment since the intent is obvious from `block_id=block.id`.

tests/tool_result_data_merge_test.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
+"""Regression tests for https://github.com/agentscope-ai/agentscope/issues/2549
```
DavdGao: The `block_id` passthrough fix in `_convert_tool_chunk_to_event` isn't covered, the tests build events with `block_id` by hand. Suggested to add a case that runs `DataBlock` chunks (Base64 and URL) through `_convert_tool_chunk_to_event` and asserts the event ids equal the block ids. The module docstring can be trimmed to one line as well.

tests/tool_result_data_merge_test.py
```diff
@@ -0,0 +1,79 @@
...
+from agentscope.message import AssistantMsg, Base64Source, DataBlock
+
+_TC = "tc-1"
+
+
+def _b64(raw: bytes) -> str:
+    return base64.b64encode(raw).decode("ascii")
+
+
+def _drive(msg: AssistantMsg, deltas: list) -> None:
```
DavdGao: Suggested to drop `_drive` and `_b64` and write the events inline in each test, it's easier to see the full input at a glance.

tests/tool_result_data_merge_test.py
```diff
@@ -0,0 +1,79 @@
...
+                media_type="audio/wav",
+                data=data,
+            ),
+        )
+    msg.append_event(
+        ToolResultEndEvent(reply_id=msg.id, tool_call_id=_TC, state="success"),
+    )
+
+
+class TestToolResultDataDeltaMerge(IsolatedAsyncioTestCase):
```
DavdGao: Both tests are sync, plain `unittest.TestCase` is enough.

tests/tool_result_data_merge_test.py
```diff
@@ -0,0 +1,79 @@
...
+        """Same-id Base64 deltas merge into a single block (issue #2549)."""
+        msg = AssistantMsg(name="assistant", content=[])
+        _drive(msg, [("audio-1", _b64(b"hello")), ("audio-1", _b64(b"world"))])
+
+        tool_blocks = [b for b in msg.content if b.type == "tool_result"]
+        self.assertEqual(len(tool_blocks), 1)
+        data_blocks = [
+            b for b in tool_blocks[0].output if isinstance(b, DataBlock)
+        ]
+        self.assertEqual(len(data_blocks), 1)
```
DavdGao: Please assert the whole `tool_blocks[0].output` with one `assertListEqual` against the expected block list, instead of checking length / type / data separately. Same for the different-id case below.

CorgiBoyG:
I ran two additional context-vs-replay consistency cases against the current head (`cb94cace`). Both reproduce deterministically with model-valid inputs:

1. `TextBlock(id="same") -> DataBlock(id="same", data=b"a") -> DataBlock(id="same", data=b"b")`. `ToolResponse.append_chunk()` assigns two fresh IDs and keeps the data blocks separate, while the event path emits the original ID and this PR merges them into one `b"ab"` block. The persisted/replayed message therefore diverges from the agent context in both identity and payload boundaries.
2. Two same-ID Base64 blocks with media types `"audio/wav"` and `""`. An empty string is accepted by the current models. `ToolResponse.append_chunk()` retains `"audio/wav"` via `incoming or existing`, while the replay path currently overwrites it with `""`.

A minimal way to address the first case without changing the event schema is to normalize only a DataBlock whose ID is already owned by a TextBlock before yielding the `ToolChunk`, then pass that same normalized chunk to `ToolResponse.append_chunk()`. The second case can mirror the tool-layer assignment: `incoming_media_type or existing_media_type`.

I intentionally excluded the invalid-Base64 compatibility case here because the review above explicitly chose not to preserve that fallback in the message layer. The two cases above do not depend on invalid payloads. I have executable regression cases for both if useful.

dvd233:
Thanks for the additional cases. I pushed d9710f38 with the follow-up fixes:

- Msg.append_event now merges same-ID Base64 deltas across the output and keeps an existing media type when an incoming delta has an empty media type.
- Toolkit.call_tool normalizes cross-type block-ID collisions before yielding and accumulating each ToolChunk, so the streamed response and event replay use the same block identities.
- Added regression coverage for empty media types, cross-type IDs, and DataBlock event IDs.

Verification:
- python -m pytest tests/tool_result_data_merge_test.py -q — 5 passed
- Black with the repository's line length 79 — passed
- Pre-commit and the Ubuntu, Windows, and macOS CI test jobs — passed

src/agentscope/tool/_toolkit.py
```diff
@@ -63,6 +68,33 @@
 """  # noqa: E501
 
 
+def _normalize_chunk_ids(
```
DavdGao, line 71: The cross-type id collision (text and data sharing one id) is invalid tool output, and `ToolResultTextDeltaEvent` doesn't carry a block id anyway, so text ids in the replay never match `ToolResponse`. Normalizing every chunk in `call_tool` for this goes beyond #2549 and rebuilds the id map on each chunk. Suggested to revert the `_toolkit.py` changes and the matching `test_cross_type_id_collisions_are_normalized_before_yield`.

tests/tool_result_data_merge_test.py
```diff
@@ -0,0 +1,273 @@
...
+    Base64Source,
+    DataBlock,
+    TextBlock,
+    ToolCallBlock,
+)
+from agentscope.state import AgentState
+from agentscope.tool import FunctionTool, ToolChunk, Toolkit, ToolResponse
+
+
+def _data_block(block_id: str, data_b64: str) -> dict[str, Any]:
```
DavdGao, line 27: Please don't build expected values with `_data_block`. Write the full dict literally in each test so the expected output is visible at a glance.

tests/tool_result_data_merge_test.py
```diff
@@ -0,0 +1,273 @@
...
+        self.assertEqual(len(tool_blocks), 1)
+        self.assertListEqual(
+            [b.model_dump() for b in tool_blocks[0].output],
+            [
+                _data_block("a", base64.b64encode(b"hello").decode("ascii")),
+                _data_block("b", base64.b64encode(b"world").decode("ascii")),
+            ],
+        )
+
+    def test_empty_media_type_does_not_clear_existing_type(self) -> None:
```
DavdGao, line 120: Suggested to assert the whole `msg.content[0].output` with one `assertListEqual`, same as the two cases above, instead of checking only `media_type`.

tests/tool_result_data_merge_test.py
```diff
@@ -0,0 +1,273 @@
...
+            [b"a", b"b"],
+        )
+
+
+class ConvertToolChunkIdentityTest(IsolatedAsyncioTestCase):
+    """_convert_tool_chunk_to_event preserves the chunk's block id."""
+
+    async def test_event_block_id_matches_data_block_id(self) -> None:
+        """Base64 and URL DataBlocks keep their id on the emitted event."""
+        from agentscope.message import URLSource
```
DavdGao, line 234: `URLSource` is imported inside the test. Move it to the top with the other `agentscope.message` imports.

DavdGao (commented):
The `block_id` passthrough and the same-id Base64 merge in `append_event` look good now. Main concern is the `_toolkit.py` change, which goes beyond #2549, plus a few test nits inline.

CorgiBoyG:
One additional issue showed up in a performance review of the valid Base64 path. The current `append_event` merge decodes and re-encodes the entire accumulated payload for every delta:

```python
existing = base64.b64decode(merge_target.source.data)
...
merge_target.source.data = base64.b64encode(existing + incoming).decode(...)
```

That makes replay quadratic in the streamed payload size and repeatedly allocates/copies the full history. I measured the current head (`d9710f38`) with valid 4 KiB independently encoded chunks:

- 100 chunks / 0.39 MiB: 0.037 s
- 250 chunks / 0.98 MiB: 0.226 s
- 500 chunks / 1.95 MiB: 0.844 s

Increasing the chunk count 5x increased runtime about 22.6x. For comparison, retaining deltas during replay and normalizing once at `ToolResultEndEvent` took 0.0019 s, 0.0041 s, and 0.0077 s for the same inputs. A separate 1-byte-chunk growth run also showed 16K -> 64K chunks (4x) growing from 0.37 s to 3.96 s (10.6x).

This is independent of the cross-type-ID discussion and uses only valid Base64. A bounded fix would be to append deltas during replay and perform same-ID byte aggregation once when the matching `ToolResultEndEvent` arrives, preserving first-occurrence order and then coalescing newly adjacent text blocks. That keeps the normal path linear and still produces the canonical final message. If intermediate `Msg` state must already expose one block, it would need a byte accumulator rather than repeatedly rebuilding Base64.

DavdGao:
Thanks for the numbers. The decode/re-encode per delta is real, but it's not introduced here: `ToolResponse.append_chunk` (`_merge_base64_chunks`) and the existing `DATA_BLOCK_DELTA` branch do exactly the same, so optimizing only the replay path doesn't change the overall cost. Deferring the merge to `ToolResultEndEvent` also leaves the intermediate `Msg` split, which is what #2549 is about. Let's keep this PR focused on the id passthrough and merge, no scope expansion here.

PR #2562, opened 2026-09-09
fix(agent): preserve reply usage in final messages

### LINKED ISSUE #2561: [Bug]: structured output reply loses model usage
## Problem

When `Agent.reply(..., structured_schema=...)` generates a structured output, the returned `Msg.usage` is `None` even when the model responses contain token usage.

```text
Model call 1: usage = 80 input / 40 output
Model call 2: usage = 100 input / 50 output
             |
             v
Current reply context: usage = 180 input / 90 output
             |
             v
Final structured Msg:   usage = None   <-- usage is lost while rebuilding it
```

Ordinary text replies expose usage correctly, but structured-output replies do not.

## Reproduction

```python
result = await agent.reply(
    UserMsg(name="user", content="Weather in Hangzhou?"),
    structured_schema=WeatherReport,
)

assert result.structured_output is not None
assert result.usage is not None  # Fails with the current implementation
```

No exception is raised. The issue is that usage already recorded for the current reply is not propagated to the final structured-output message.

## Expected behavior

The returned structured-output `Msg.usage` should contain the accumulated usage from every model call in the current reply, including input, output, and cache tokens.

## Impact

Applications that use `Msg.usage` for token, cost, or model-call metrics receive an incorrect empty value for structured-output replies.

## Scope

This is a Python Agent message-construction issue. It does not require changes to model adapters, usage accumulation, or public API design.

liugy789:
Fixes #2561

## AgentScope Version

2.0.8

## Description

Structured-output replies rebuilt the final `AssistantMsg` without carrying
over the usage accumulated during the current reply. As a result,
`Msg.usage` was `None` even though the model calls had returned usage data.

The same usage loss also affected final messages created by the
`exceed_max_iters` and interruption fallback paths.

This change:

- reuses `_get_reply_usage()` to retrieve a copy of the usage accumulated
  during the current reply;
- includes the accumulated usage in structured-output final messages;
- includes the accumulated usage in both `exceed_max_iters` exit messages;
- includes the accumulated usage in interruption fallback messages;
- verifies that the last assistant message belongs to the current `reply_id`;
- prevents a reply that adds no new assistant context from reusing the
  previous reply usage;
- preserves the current reply usage when context compression removes its
  assistant message;
- records context compression usage without losing or double-counting the
  current reply usage;
- keeps unfinished awaiting-permission messages without final usage because
  those replies have not finished yet.

No model adapter, public API, or usage accumulation interface changes are
required.

## Testing

- `pytest -q tests/compress_context_test.py tests/agent_basic_test.py tests/agent_injection_test.py tests/agent_interrupt_test.py tests/agent_structured_output_test.py` — 80 passed
- Targeted pre-commit checks — passed, including mypy, black, flake8,
  pylint, and Pyroma
- `git diff --check` — passed

## Breaking Changes

None.

## Checklist

- [x] Related documentation has been reviewed; no companion docs change is needed for this internal behavior fix

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -1823,27 +1823,12 @@ class Agent:
             )
             and not has_only_thinking_blocks
         ):
-            last_ctx = self._get_last_msg()
-            final_usage = (
-                Usage(
-                    input_tokens=last_ctx.usage.input_tokens,
-                    output_tokens=last_ctx.usage.output_tokens,
-                    cache_input_tokens=(
-                        last_ctx.usage.cache_input_tokens or 0
-                    ),
-                    cache_creation_input_tokens=(
-                        last_ctx.usage.cache_creation_input_tokens or 0
-                    ),
-                )
-                if last_ctx is not None and last_ctx.usage is not None
-                else None
-            )
             yield AssistantMsg(
                 id=self.state.reply_id,
                 name=self.name,
                 # Text only response message
                 content=list(completed_response.content),
-                usage=final_usage,
+                usage=self._get_reply_usage(),
                 # The INTERRUPTED case is excluded by the branch condition
                 finished_reason=ReplyFinishedReason.COMPLETED,
             )
@@ -3481,6 +3466,24 @@ class Agent:
             return last_msg
         return None
 
+    def _get_reply_usage(self) -> Usage | None:
+        """Get a copy of the accumulated usage for the current reply."""
+        last_msg = self._get_last_msg()
+        if (
+            last_msg is None
+            or last_msg.id != self.state.reply_id
+            or last_msg.usage is None
+        ):
+            return None
+        return Usage(
+            input_tokens=last_msg.usage.input_tokens,
+            output_tokens=last_msg.usage.output_tokens,
+            cache_input_tokens=last_msg.usage.cache_input_tokens or 0,
+            cache_creation_input_tokens=(
+                last_msg.usage.cache_creation_input_tokens or 0
+            ),
+        )
+
     def _next_action(
         self,
         final_msg: Msg | None = None,
@@ -3553,6 +3556,7 @@ class Agent:
                     id=self.state.reply_id,
                     name=self.name,
                     content="The required structured output is generated.",
+                    usage=self._get_reply_usage(),
                     finished_reason=ReplyFinishedReason.COMPLETED,
                     structured_output=deepcopy(
                         self.state.reply_context.structured_output,
--- a/tests/agent_structured_output_test.py
+++ b/tests/agent_structured_output_test.py
@@ -7,7 +7,7 @@ from pydantic import BaseModel
 from utils import AnyString, MockModel
 
 from agentscope.agent import Agent, InjectionConfig, ReActConfig
-from agentscope.model import ChatResponse
+from agentscope.model import ChatResponse, ChatUsage
 from agentscope.state import AgentState
 from agentscope.tool import ToolBase, Toolkit, ToolChunk
 from agentscope.permission import (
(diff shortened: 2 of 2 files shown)


src/agentscope/agent/_agent.py
```diff
@@ -3481,6 +3466,24 @@ def _get_last_msg(self) -> Msg | None:
...
+    def _get_reply_usage(self) -> Usage | None:
+        """Get a copy of the accumulated usage for the current reply."""
+        last_msg = self._get_last_msg()
+        if (
+            last_msg is None
+            or last_msg.id != self.state.reply_id
+            or last_msg.usage is None
+        ):
+            return None
+        return Usage(
```
DavdGao: `Usage.cache_input_tokens` and `cache_creation_input_tokens` are declared as `int = 0`, so they can never be None here and the `or 0` is dead code. The `or 0` in the other two places is needed because they convert a `ChatUsage`, which is not the case here.

- Suggested to be `return last_msg.usage.model_copy()`.

src/agentscope/agent/_agent.py
```diff
@@ -3553,6 +3556,7 @@ def _next_action(
                     id=self.state.reply_id,
                     name=self.name,
                     content="The required structured output is generated.",
+                    usage=self._get_reply_usage(),
```
DavdGao, line 3563: The root cause is that the final message is rebuilt without carrying the accumulated usage, and the structured output branch is not the only place doing that. The exit messages below lose the usage in the same way:

- `_next_action`, the two `exceed_max_iters` exit messages. I verified locally that `Msg.usage` is None there even when the model calls returned usage.
- `_reasoning_impl`, the interruption fallback message.

Suggested to pass `self._get_reply_usage()` in these as well, otherwise "structured output keeps the usage but exceeding max_iters doesn't" is hard to explain to the users. The awaiting-permission exit message should be left as is, since the reply is not finished there and the usage would be counted twice by the final message.

tests/agent_structured_output_test.py
```diff
@@ -144,6 +144,55 @@ async def test_structured_reply(self) -> None:
...
+                ),
+            ],
+        )
+
+        res = await self.agent.reply(
+            UserMsg(name="user", content="Weather in Hangzhou?"),
+            structured_schema=WeatherReport,
+        )
+
+        self.assertIsNotNone(res.usage)
```
DavdGao: Asserting the usage field by field doesn't follow the convention here. `test_structured_reply` right above compares the whole `res.model_dump()` with `assertDictEqual` in one shot, and the expected dict already contains `"usage"`.

- Suggested to assert the complete `res.model_dump()` the same way, so the expected message is visible at a glance.

tests/agent_structured_output_test.py
```diff
@@ -175,6 +224,13 @@ async def test_defaults_and_extra_fields(self) -> None:
     async def test_validation_error_retry(self) -> None:
         """An invalid structured output produces an error tool result, and
         the model retries in the next reasoning round."""
+        self.structured_tool_call.usage = ChatUsage(
```
DavdGao: Patching the shared fixture inside the test makes it hard to see what the responses actually are.

- Suggested to build the `ChatResponse` with its usage in place in this test.

DavdGao (changes requested):
The bug is real - I reverted the source change and both new tests fail, so the regression coverage is valid. Two things before this can go in:

- The same usage loss exists on the exceed_max_iters and interruption exits, please fix them together (see the inline comment).
- `_get_reply_usage` can be one line with `model_copy()`.

Also, the new `last_msg.id != self.state.reply_id` check fixes a second bug on its own: without it, a reply that appends nothing to the context would copy the previous reply's usage onto its final message. Worth mentioning it explicitly in the PR description, it's not obvious from "verifies that the usage belongs to the current reply_id".

liugy789:
Hi @DavdGao,

Thank you for the detailed review and for pointing out the additional exit
paths that were affected by the same usage loss.

I have addressed all of the requested changes:

- Added the accumulated usage to both `exceed_max_iters` exit messages.
- Added the accumulated usage to the interruption fallback message.
- Updated `_get_reply_usage()` to return `last_msg.usage.model_copy()`.
- Kept the awaiting-permission message without usage because the reply has
  not finished at that point.
- Updated the structured-output tests to assert the complete
  `res.model_dump()`.
- Reworked the validation retry test to construct the valid
  `ChatResponse` locally instead of modifying a shared fixture.
- Kept the `reply_id` check to prevent an empty reply from reusing the
  previous reply usage.

I also added regression coverage for preserving the current reply usage
when context compression removes the assistant message that stored it.

The focused test suite passes with 80 tests, and the relevant pre-commit
checks pass.

Thank you again for the careful review.

src/agentscope/agent/_agent.py
liugy789: Thanks for pointing this out. Fixed. `_get_reply_usage()` now returns `last_msg.usage.model_copy()` directly, as suggested.

src/agentscope/agent/_agent.py
liugy789: Thanks for identifying the other affected exit paths. Fixed.

Both `exceed_max_iters` exit messages and the interruption fallback message
now pass `self._get_reply_usage()`.

The awaiting-permission message remains without usage because the reply has
not finished yet.

tests/agent_structured_output_test.py
liugy789: Thanks for the suggestion. Updated as requested. The test now asserts the
complete `res.model_dump()` with `assertDictEqual`.

tests/agent_structured_output_test.py
liugy789: Thanks for pointing out the shared-fixture issue. Updated as requested. The
validation retry test now constructs the valid `ChatResponse` locally instead
of modifying the shared fixture.

DavdGao (approved):
LGTM

PR #2569, opened 2026-09-09
fix(agent): retry compression on recoverable structured-output errors

### LINKED ISSUE #2425: [Bug]: Context compression aborts the agent on recoverable structured-output schema validation errors
### Prerequisites

### Background / Description

When an agent reaches the context-compression threshold, `Agent._compress_context_impl()` calls `ChatModelBase.generate_structured_output()` with the compression summary schema.

A model can return an almost valid tool call whose field name deviates slightly from the schema (for example, `task_over` instead of the required `task_overview`). The resulting `jsonschema.ValidationError` is propagated immediately and aborts the entire agent invocation.

This happened in a long-running agent using the default `SummarySchema`. All other required fields were present; only `task_overview` was misspelled.

### Actual behavior

- `jsonschema.validate()` raises `ValidationError`.
- The error is not included in `_get_retryable_exceptions()`, so `generate_structured_output()` does not retry.
- `_compress_context_impl()` retries only when `context_overflow=True`; a schema-validation failure is re-raised.
- The whole agent invocation terminates.

### Expected behavior

Structured-output validation failures should be treated as recoverable format errors. AgentScope should retry a small number of times with concise validation feedback (missing/invalid fields and the expected schema). Only exhausted retries should raise a typed compression/structured-output error.

A one-field format deviation should not terminate a long-running agent. The retry should be generic rather than adding a special-case alias for one observed typo.

### Relevant code paths

- `agentscope/agent/_config.py`: `SummarySchema`
- `agentscope/model/_base.py`: `generate_structured_output()` and schema validation
- `agentscope/agent/_agent.py`: `_compress_context_impl()`

### Error Messages

```shell
'task_overview' is a required property

Failed validating 'required' in schema:
  required:
    - task_overview
    - current_state
    - important_discoveries
    - next_steps
    - context_to_preserve

Model tool-call input:
  {
    "task_over": "...",
    "current_state": "...",
    "important_discoveries": "...",
    "next_steps": "...",
    "context_to_preserve": "..."
  }
```

### Steps to Reproduce

1. Configure an agent with context compression enabled and enough context to exceed `trigger_ratio`.

2. Make the compression model return a `generate_structured_output` tool call with this input:

```json
{
  "task_over": "task summary",
  "current_state": "current state",
  "important_discoveries": "discoveries",
  "next_steps": "next steps",
  "context_to_preserve": "context"
}
```

3. Run:

```python
await agent.compress_context()
```

4. Observe that the first `jsonschema.ValidationError` escapes immediately and the agent is aborted, even when model retries are configured.

A regression test can use a mock model that returns an invalid summary on the first structured-output call and a valid summary on the second. The expected result is two calls and successful compression; the current behavior raises after the first call.

### Environment

- AgentScope Version: 2.0.5
- Python Version: 3.12
- OS: macOS
- Model: qwen3.7-max
- Streaming: enabled

hulincup:
## Problem fixed & how

When context compression triggers, `_compress_context_impl` calls `generate_structured_output` with the summary schema. If the model returns an almost-valid tool call with a misspelled field (e.g. `task_over` instead of the required `task_overview`), the `jsonschema.ValidationError` is wrapped into a `StructuredOutputError`. `_compress_context_impl` only retried when `context_overflow=True`; a schema-validation failure with `context_overflow=False` was re-raised (or fell back to truncation), aborting the entire long-running agent on a single one-off format deviation.

**Root cause** (`agent/_agent.py`, `_compress_context_impl`): the `generate_structured_output` call is wrapped by `except Exception as error`, whose retry branch is gated on `context_overflow`. `StructuredOutputError` is a recoverable format error, not an overflow, so it skipped the retry and propagated.

**Fix**: catch `StructuredOutputError` specifically before the generic `except Exception` and retry a bounded number of times (`_COMPRESSION_VALIDATION_RETRIES = 2`) with concise validation feedback appended — the validation error text plus the required schema fields — so the model can correct the format. The retry is generic (no per-field alias). On exhausted retries the most recent error is re-raised into the existing truncation/raise handling, preserving prior behavior for unrecoverable failures.

Closes #2425.

## Behavior modified

- **Previous**: a schema-validation failure during compression with `context_overflow=False` re-raised `StructuredOutputError` (or truncated context when `compression_fallback_to_truncation` is set), aborting the agent on a one-off format deviation.
- **Now**: the compression path retries up to 2 additional times with validation feedback before giving up; only exhausted retries fall through to the existing truncation/raise behavior.
- **Impact**: long-running agents are no longer aborted by a single misspelled summary field. Overflow / provider errors and the truncation fallback are unchanged — `RuntimeError` and `context_overflow` paths still hit the existing `except Exception` branch.

## Features added

N/A

## Code refactored

The retry is contained to `_compress_context_impl`'s compression call. No changes to `generate_structured_output` or the model abstraction's retry/strategy semantics.

## Functions optimized

N/A

## Test plan

- [x] `test_compression_retries_on_structured_output_error`: a mock model whose `generate_structured_output` raises `StructuredOutputError` on the first call and returns a valid summary on the second → asserts compression retries and succeeds with 2 calls. ✅ passes locally
- [x] `test_compression_exhausts_retries_then_aborts`: an always-failing mock → asserts the agent raises `StructuredOutputError` after the retry budget (1 initial + 2 retries = 3 calls), no infinite loop. ✅ passes locally
- [x] Full `tests/compress_context_test.py` suite (24 tests) passes locally — existing success / overflow-retry / truncation-fallback cases are unchanged (backward compatible).

### DIFF
--- a/src/agentscope/agent/_agent.py
+++ b/src/agentscope/agent/_agent.py
@@ -64,7 +64,7 @@ from ..event import (
     UserInterruptEvent,
     HintBlockEvent,
 )
-from ..exception import AgentOrientedException
+from ..exception import AgentOrientedException, StructuredOutputError
 from ..model import (
     ChatResponse,
     ChatUsage,
@@ -107,6 +107,12 @@ from ..workspace import Offloader, WorkspaceBase
 
 # The name of the tool that the agent uses to compress its own context
 _COMPRESSION_TOOL_NAME = "CompressContext"
+
+# Extra attempts to regenerate the compression summary when the model
+# returns a structurally invalid summary (a recoverable schema-validation
+# failure) rather than aborting the agent on the first try. Each retry
+# appends concise validation feedback so the model can correct the format.
+_COMPRESSION_VALIDATION_RETRIES = 2
 
 if TYPE_CHECKING:
     from ..middleware import MiddlewareBase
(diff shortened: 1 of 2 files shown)


DavdGao:
The problem this PR fixes is already handled on the latest main. Verified on `4c0547b0c` with a model whose structured output always fails schema validation:

- `generate_structured_output` already retries on `StructuredOutputError` through the strategy ladder (3 model calls in this case), and `compression_fallback_to_truncation` defaults to `True`, so the compression falls back to truncation instead of aborting the agent. `test_summary_failure_truncates_context` covers this.
- With this PR the same case ends up exactly the same (truncation, or raise when the fallback is off), only with 9 model calls instead of 3, each carrying a near-full context. The agent-level retry multiplies with the existing ladder.
- The compression path is also not the right place for this. If a validation-feedback retry is worth adding, it should live in `generate_structured_output` so every caller benefits, not as a special case for compression.

Closing this, please reproduce on the latest main first.

PR #2572, opened 2026-09-09
fix(tool): bound glob search tool output

### LINKED ISSUE #2571: [Bug]: Glob can overflow agent context with unbounded output
### Describe the bug

The built-in `Glob` tool currently joins every matching file path into one tool result. For a broad pattern such as `**/*`, the result can grow with the entire workspace and add an unnecessarily large message to the agent context.

The built-in `Grep` tool already has a default `head_limit` of 250 entries and supports `offset`, but `Glob` has no equivalent default result limit or pagination mechanism.

This is not an exception. The problem is that the tool has no default result limit and no pagination mechanism:

```text
broad Glob pattern
        |
        +-- match 1
        +-- match 2
        +-- ...
        +-- match 201
        |
        +-- current: all paths are returned in one result
        +-- expected: a bounded page with a way to fetch more
```

### To reproduce

1. Create more than 200 files matching the same pattern.
2. Call the built-in `Glob` tool without an explicit result limit:

```python
result = await Glob()(pattern="match_*.log", path="/tmp/workspace")
```

3. Observe that every matching path is returned in one result and no truncation guidance is provided.

Example observation with 201 matching files:

```text
matched_paths=201
truncation_notice=False
```

No exception is raised.

### Expected behavior

`Glob` output should remain bounded even when the caller omits optional arguments:

- enforce a tool-side default result limit;
- allow callers to request a smaller or larger page within a hard cap;
- include an explicit truncation notice when more matches exist;
- provide a way to request later pages.

### Additional context

The proposed defaults for `Glob` are a default limit of 200 paths and a hard maximum of 1000 paths. The proposed first version applies the limit at the tool-output boundary; moving the limit into helper traversal can be handled as a separate optimization.

liugy789:
## Background and motivation

The built-in `Glob` tool currently returns every matching path in one tool result. When a workspace contains many files and the pattern is broad, such as `**/*`, this can write an unnecessarily large message into the agent context.

The built-in `Grep` tool already provides a default `head_limit` of 250 entries and supports `offset`. `Glob` should provide the corresponding protection for file-path results.

The first protection should be applied before the unbounded result is exposed to the model:

```text
Unbounded Glob result
  -> every matching path enters one tool message
  -> context receives more output than the caller needs
  -> no supported way to retrieve the next page
```

This PR adds deterministic output bounds and pagination while preserving the existing matching behavior.

## Current implementation

This PR provides a focused tool-layer solution:

- add an optional `limit` parameter;
- apply a default limit of 200 paths when the caller omits `limit`;
- cap the effective limit at 1000 paths;
- support an optional non-negative `offset` for later pages;
- reject invalid `limit` and `offset` values;
- append an explicit truncation notice containing the next offset;
- report clearly when the requested offset has no remaining matches;
- preserve the existing newest-first matching order;
- add regression tests for defaults, explicit values, invalid values, the hard cap, pagination boundaries, and empty results.

The key point is that the default bound is enforced by `Glob` itself and does not depend on the model remembering to pass an optional argument.

## Scope and known limitations

- This PR changes the built-in `Glob` tool only.
- The existing `Glob(pattern, path)` call remains valid.
- The limit is applied to the final tool payload. The current helper still completes its scan, sorts the full match list, and transfers it before the tool layer slices it.
- Therefore, this version bounds model-visible output but does not yet reduce worst-case backend scan time, helper memory, or transport size. Passing `limit` and `offset` into the helper is a separate follow-up because it requires defining early-stop and pagination semantics across all backends.
- The tool schema and Google-style docstring document the new parameters; no separate companion documentation repository change is included.

## Questions for maintainers

1. Are a default of 200 paths and a hard maximum of 1000 appropriate for the AgentScope `Glob` tool?
2. Is applying the first protection at the common tool-output boundary acceptable, with helper-level early stopping handled separately?
3. Is the current plain-text truncation guidance, including `offset=<next offset>`, sufficient, or should a future version return structured pagination metadata?

## Compatibility

- Existing two-argument calls remain valid.
- Results below the default limit keep their existing path-only format.
- Local, Docker, and E2B backends continue to use the same common `Glob` output handling.

## Validation

- `pytest tests/builtin_glob_test.py tests/builtin_grep_test.py -q` — 30 passed
- `python -m compileall` for the changed Python files — passed
- `git diff --check` — passed
- Targeted pre-commit hooks — passed: AST, mypy, Black, flake8, pylint, and Pyroma

## Checklist

- [ ] Related documentation has been updated in the documentation repository — not needed for this focused tool behavior change; the schema and docstring are updated inline

Fixes #2571


src/agentscope/tool/_builtin/_glob.py
```diff
@@ -45,6 +45,12 @@ class Glob(ToolBase):
     name: str = "Glob"
     """The tool name presented to the agent."""
 
+    default_limit: int = 200
+    """The default maximum number of paths returned by the tool."""
+
+    max_limit: int = 1000
+    """The hard maximum number of paths returned by the tool."""
```
DavdGao: - `Grep` keeps this as a module-level `DEFAULT_HEAD_LIMIT` constant, here it becomes two new public class attributes. Suggested to follow `Grep` and use a module-level constant.
- `max_limit` is an arbitrary hard cap with no way to opt out, while `Grep` allows `head_limit=0` for unlimited. Suggested to drop the cap and support `0` instead.

src/agentscope/tool/_builtin/_glob.py
```diff
@@ -68,6 +74,20 @@ class Glob(ToolBase):
...
+                "description": "Maximum number of matching files to return "
+                "(default: 200, values above 1000 are capped)",
+            },
+            "offset": {
+                "type": "integer",
+                "minimum": 0,
+                "default": 0,
+                "description": "Number of matching files to skip before "
+                "returning results",
+            },
```
DavdGao: - The parameter is named `limit` here while `Grep` uses `head_limit`. Two builtin tools doing the same thing should expose the same interface, otherwise both the developer and the model have to remember two sets of rules. Suggested to rename it to `head_limit` with the same semantics.
- The tool `description` above is not updated, while `Grep` documents `head_limit` in its description. Suggested to add one line there.

src/agentscope/tool/_builtin/_glob.py
```diff
@@ -232,6 +260,36 @@ async def call(  # type: ignore[override]
...
+        ):
+            return ToolChunk(
+                content=[
+                    TextBlock(
+                        text=("Glob offset must be a non-negative integer."),
+                    ),
+                ],
+                state=ToolResultState.ERROR,
+                is_last=True,
+            )
```
DavdGao: These type checks are unnecessary — the input schema already declares `type: integer` with `minimum`, and `Grep` only checks the negative case. Suggested to keep a single `< 0` check and remove the rest.

src/agentscope/tool/_builtin/_glob.py
```diff
@@ -288,18 +346,52 @@ async def call(  # type: ignore[override]
...
-                        text=f"No files found matching pattern: {pattern}",
+                        text=(
+                            f"No more files found matching pattern: {pattern} "
+                            f"after offset {offset}."
+                        ),
                     ),
                 ],
                 state=ToolResultState.RUNNING,
                 is_last=True,
             )
```
DavdGao: The two branches return almost the same thing. Suggested to merge them into one.

src/agentscope/tool/_builtin/_glob.py
```diff
@@ -288,18 +346,52 @@ async def call(  # type: ignore[override]
...
+        if truncated:
+            output += (
+                f"\n\n[Results truncated: showing results {offset + 1}-{end} "
+                f"of {len(matches)} matches. "
+                f"Use offset={end} to retrieve more results"
+            )
+            if effective_limit < self.max_limit:
+                output += f", or increase limit up to {self.max_limit}.]"
+            else:
+                output += ".]"
```
DavdGao: - The notice uses a different format from `Grep`'s `[Showing results with pagination = limit: N, offset: M]`. Suggested to reuse the same message.
- The `or increase limit up to ...` branch is not needed, especially once the hard cap is removed.

tests/builtin_glob_test.py
```diff
@@ -102,6 +122,191 @@ async def test_recursive_pattern(self) -> None:
...
+            line
+            for line in chunk.content[0].text.splitlines()
+            if line.startswith(self.temp_dir)
+        ]
+        self.assertEqual(len(paths), 200)
+        self.assertIn(
+            "showing results 1-200 of 201 matches",
+            chunk.content[0].text,
+        )
+        self.assertIn("increase limit up to 1000", chunk.content[0].text)
```
DavdGao: The assertions match substrings and filter lines by prefix instead of comparing the full output. Suggested to build the expected text and compare it in one `assertEqual`, so the expected result is visible at a glance.

tests/builtin_glob_test.py
```diff
@@ -102,6 +122,191 @@ async def test_recursive_pattern(self) -> None:
...
+
+            self.assertEqual(chunk.state, "error")
+            self.assertEqual(
+                chunk.content[0].text,
+                "Glob limit must be a positive integer.",
+            )
+
+    async def test_limit_is_capped_at_maximum(self) -> None:
+        """Test that an excessive limit cannot bypass the hard maximum."""
+        for index in range(1001):
```
DavdGao: Creating 1001 real files makes the test slow. If the hard cap is removed this case goes away; otherwise suggested to mock the helper output rather than touching the filesystem.

DavdGao (commented):
The problem is real, but the implementation should follow the existing `Grep` tool instead of introducing a new convention. With the points below the whole change should be around 15 lines.

liugy789:
@DavdGao Thanks for the detailed review. I have updated the implementation to follow the existing `Grep` conventions:

- Renamed `limit` to `head_limit`.
- Replaced the public class-level limits with a module-level `DEFAULT_HEAD_LIMIT`.
- Removed the arbitrary maximum limit.
- Added `head_limit=0` support for unlimited output.
- Kept only the negative-value runtime checks, relying on the tool schema for type validation.
- Merged the duplicated no-results branches.
- Reused the same pagination notice format as `Grep`:
  `[Showing results with pagination = limit: N, offset: M]`
- Removed the custom “increase limit” guidance.
- Reworked the regression tests to compare the complete output.
- Removed the slow test that created 1001 files and made result ordering deterministic with explicit modification times.

Local validation:

- `tests/builtin_glob_test.py` and `tests/builtin_grep_test.py`: 30 passed
- All pre-commit hooks passed, including Black, Flake8, Pylint, and Mypy
- Independent review found no blocking issues

Thank you again for pointing me toward the existing implementation pattern.

DavdGao (approved):

PR #2575, opened 2026-09-09
fix(utils): preserve property names during schema flattening

### LINKED ISSUE #2574: [Bug]:  schema flattening drops properties named definitions or $defs
### Prerequisites

### Background / Description

_flatten_json_schema treats keys in a properties mapping as if they were JSON Schema keywords. A normal tool parameter named definitions is removed when another part of the schema introduces $defs.

### Error Messages

```shell
none
```

### Code

```python
from jsonschema import Draft202012Validator
from pydantic import BaseModel, ConfigDict

from agentscope._utils._common import _flatten_json_schema


class Entry(BaseModel):
    definition: str


class Glossary(BaseModel):
    model_config = ConfigDict(extra="forbid")
    definitions: list[Entry]


schema = Glossary.model_json_schema()
flattened = _flatten_json_schema(schema)
instance = {"definitions": [{"definition": "an explanation"}]}

print("before properties:", list(schema["properties"]))
print("after properties:", list(flattened["properties"]))
print("after required:", flattened["required"])
print("before valid:", Draft202012Validator(schema).is_valid(instance))
print("after valid:", Draft202012Validator(flattened).is_valid(instance))

```

### Actual output:

```
before properties: ['definitions']
after properties: []
after required: ['definitions']
before valid: True
after valid: False
```

The flattened object retains additionalProperties: false and required: ["definitions"], while removing the property definition.

### Expected behavior

Keys inside properties are instance property names, not schema keywords. 
Keep them unchanged and resolve each property's schema value. 
The valid instance above should remain valid after flattening.

### Root cause

In `src/agentscope/_utils/_common.py`, `_resolve_ref` recursively applies
these rules to every dictionary, including property-name mappings:

```
if "$ref" in obj:
    # Interpret this dictionary as a reference schema.
    ...

for key, value in obj.items():
    if key in ("$defs", "definitions"):
        continue
```

Consequently, a field name is interpreted as a keyword at the wrong level.

### Environment

- AgentScope Version: branch `main`, commit `3500806c29685beb4424dfb48b2ac129452a5a39`
- Python Version: 3.12.12
- OS: windows 11

Tsubaki-01:
## Summary

_flatten_json_schema currently interprets every dictionary as a schema node. That removes normal properties named definitions or $defs, and a property named $ref prevents the property mapping from being resolved.

## Changes

- Distinguish schema mappings from schema nodes in the private recursive resolver.

- Preserve keys in properties, patternProperties, dependentSchemas, and legacy dependencies; recursively resolve only their schema values.

Fixes #2574

### DIFF
--- a/src/agentscope/_utils/_common.py
+++ b/src/agentscope/_utils/_common.py
@@ -345,11 +345,26 @@ def _flatten_json_schema(schema: dict) -> dict:
     if not defs:
         return schema
 
-    def _resolve_ref(obj: Any, visited: frozenset = frozenset()) -> Any:
+    schema_map_keywords = {
+        "properties",
+        "patternProperties",
+        "dependentSchemas",
+        "dependencies",
+    }
+
+    def _resolve_ref(
+        obj: Any,
+        visited: frozenset = frozenset(),
+        is_schema_map: bool = False,
+    ) -> Any:
         if isinstance(obj, list):
             return [_resolve_ref(item, visited) for item in obj]
         if not isinstance(obj, dict):
             return obj
+        if is_schema_map:
+            return {
+                key: _resolve_ref(value, visited) for key, value in obj.items()
+            }
         if "$ref" in obj:
             ref_path = obj["$ref"]
             if isinstance(ref_path, str) and (
@@ -377,6 +392,7 @@ def _flatten_json_schema(schema: dict) -> dict:
                             resolved[key] = _resolve_ref(
                                 value,
                                 visited | {def_name},
+                                key in schema_map_keywords,
                             )
                     return resolved
             return obj
@@ -384,7 +400,11 @@ def _flatten_json_schema(schema: dict) -> dict:
         for key, value in obj.items():
             if key in ("$defs", "definitions"):
                 continue
-            result[key] = _resolve_ref(value, visited)
+            result[key] = _resolve_ref(
+                value,
+                visited,
+                key in schema_map_keywords,
+            )
         return result
 
     return _resolve_ref(schema)


qbc2016:
Please run `pre-commit run --all-files` to format the code.

Tsubaki-01:
> Please run `pre-commit run --all-files` to format the code.

Fixed the formatting in `src/agentscope/_utils/_common.py` and rebased the branch onto the latest `main`. The formatter check now passes.

qbc2016 (approved):
LGTM.

DavdGao (commented):
The `properties` fix is right, but two things:

- Same root cause is only half fixed. `default`, `const`, `enum` and `examples` hold instance values, not schemas, so the resolver still corrupts them:
  ```python
  cfg: dict = Field(default={"definitions": {"x": 1}})   # -> "default": {}
  {"enum": [{"$ref": "#/$defs/E"}]}                      # -> the enum value gets expanded as a reference
  ```
  This is exactly the bug in #2574 with a different keyword, so better to close it in the same PR.

- No regression test. `tests/model_gemini_test.py` already has a group of flatten tests, please put the repro from #2574 there.

Also `schema_map_keywords` is a constant, no need to rebuild it on every call.

I have this working locally. The `$ref` branch and the key loop were doing the same per-key work, so merging them into one loop drops the extra `is_schema_map` argument and leaves one place to handle the instance-value keys. Feel free to take it as is:

```diff
--- a/src/agentscope/_utils/_common.py
+++ b/src/agentscope/_utils/_common.py
@@
+# Keywords whose values map instance property names to schemas. Those keys are
+# user-defined names, so schema keywords must not be applied to them.
+_SCHEMA_MAP_KEYWORDS = frozenset(
+    {"properties", "patternProperties", "dependentSchemas", "dependencies"},
+)
+
+# Keywords whose values are instance data instead of schemas, so they are kept
+# verbatim rather than resolved.
+_INSTANCE_VALUE_KEYWORDS = frozenset(
+    {"default", "const", "enum", "examples"},
+)
+
 def _flatten_json_schema(schema: dict) -> dict:
@@
     def _resolve_ref(obj: Any, visited: frozenset = frozenset()) -> Any:
         if isinstance(obj, list):
             return [_resolve_ref(item, visited) for item in obj]
         if not isinstance(obj, dict):
             return obj
+
+        resolved: dict[str, Any] = {}
         if "$ref" in obj:
             ref_path = obj["$ref"]
-            if isinstance(ref_path, str) and (...):
-                ...
-            return obj
-        result: dict[str, Any] = {}
+            if not isinstance(ref_path, str) or not ref_path.startswith(
+                ("#/$defs/", "#/definitions/"),
+            ):
+                return obj
+
+            def_name = ref_path.split("/")[-1]
+            if def_name in visited:
+                logger.warning(
+                    "Circular reference detected for '%s' in tool schema",
+                    def_name,
+                )
+                return {
+                    "type": "object",
+                    "description": f"(circular: {def_name})",
+                }
+            if def_name not in defs:
+                return obj
+
+            visited = visited | {def_name}
+            resolved = _resolve_ref(defs[def_name], visited)
+
         for key, value in obj.items():
-            if key in ("$defs", "definitions"):
+            if key in ("$ref", "$defs", "definitions"):
                 continue
-            result[key] = _resolve_ref(value, visited)
-        return result
+            if key in _INSTANCE_VALUE_KEYWORDS:
+                resolved[key] = value
+            elif key in _SCHEMA_MAP_KEYWORDS and isinstance(value, dict):
+                resolved[key] = {
+                    name: _resolve_ref(sub, visited)
+                    for name, sub in value.items()
+                }
+            else:
+                resolved[key] = _resolve_ref(value, visited)
+        return resolved
```

Plus two tests in `tests/model_gemini_test.py` — `test_flatten_keeps_property_names` (a property named `definitions` and one named `$ref`) and `test_flatten_keeps_instance_values` (`default` / `const` / `enum`, including a `default` sitting next to a `$ref`). Both fail on current `main`, full suite passes with the change.

Tsubaki-01:
> The `properties` fix is right, but two things:
> 
> * Same root cause is only half fixed. `default`, `const`, `enum` and `examples` hold instance values, not schemas, so the resolver still corrupts them:
>   ```python
>   cfg: dict = Field(default={"definitions": {"x": 1}})   # -> "default": {}
>   {"enum": [{"$ref": "#/$defs/E"}]}                      # -> the enum value gets expanded as a reference
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   This is exactly the bug in #2574 with a different keyword, so better to close it in the same PR.
> * No regression test. `tests/model_gemini_test.py` already has a group of flatten tests, please put the repro from #2574 there.
> 
> Also `schema_map_keywords` is a constant, no need to rebuild it on every call.
> 
> I have this working locally. The `$ref` branch and the key loop were doing the same per-key work, so merging them into one loop drops the extra `is_schema_map` argument and leaves one place to handle the instance-value keys. Feel free to take it as is:
> 
> ```diff
> --- a/src/agentscope/_utils/_common.py
> +++ b/src/agentscope/_utils/_common.py
> @@
> +# Keywords whose values map instance property names to schemas. Those keys are
> +# user-defined names, so schema keywords must not be applied to them.
> +_SCHEMA_MAP_KEYWORDS = frozenset(
> +    {"properties", "patternProperties", "dependentSchemas", "dependencies"},
> +)
> +
> +# Keywords whose values are instance data instead of schemas, so they are kept
> +# verbatim rather than resolved.
> +_INSTANCE_VALUE_KEYWORDS = frozenset(
> +    {"default", "const", "enum", "examples"},
> +)
> +
>  def _flatten_json_schema(schema: dict) -> dict:
> @@
>      def _resolve_ref(obj: Any, visited: frozenset = frozenset()) -> Any:
>          if isinstance(obj, list):
>              return [_resolve_ref(item, visited) for item in obj]
>          if not isinstance(obj, dict):
>              return obj
> +
> +        resolved: dict[str, Any] = {}
>          if "$ref" in obj:
>              ref_path = obj["$ref"]
> -            if isinstance(ref_path, str) and (...):
> -                ...
> -            return obj
> -        result: dict[str, Any] = {}
> +            if not isinstance(ref_path, str) or not ref_path.startswith(
> +                ("#/$defs/", "#/definitions/"),
> +            ):
> +                return obj
> +
> +            def_name = ref_path.split("/")[-1]
> +            if def_name in visited:
> +                logger.warning(
> +                    "Circular reference detected for '%s' in tool schema",
> +                    def_name,
> +                )
> +                return {
> +                    "type": "object",
> +                    "description": f"(circular: {def_name})",
> +                }
> +            if def_name not in defs:
> +                return obj
> +
> +            visited = visited | {def_name}
> +            resolved = _resolve_ref(defs[def_name], visited)
> +
>          for key, value in obj.items():
> -            if key in ("$defs", "definitions"):
> +            if key in ("$ref", "$defs", "definitions"):
>                  continue
> -            result[key] = _resolve_ref(value, visited)
> -        return result
> +            if key in _INSTANCE_VALUE_KEYWORDS:
> +                resolved[key] = value
> +            elif key in _SCHEMA_MAP_KEYWORDS and isinstance(value, dict):
> +                resolved[key] = {
> +                    name: _resolve_ref(sub, visited)
> +                    for name, sub in value.items()
> +                }
> +            else:
> +                resolved[key] = _resolve_ref(value, visited)
> +        return resolved
> ```
> 
> Plus two tests in `tests/model_gemini_test.py` — `test_flatten_keeps_property_names` (a property named `definitions` and one named `$ref`) and `test_flatten_keeps_instance_values` (`default` / `const` / `enum`, including a `default` sitting next to a `$ref`). Both fail on current `main`, full suite passes with the change.

Thanks for your review! Since #2575 has already been merged, I've addressed these points in a follow-up PR based on the latest main, including regression tests. Ready for review.

PR #2588, opened 2026-09-10
fix(realtime): drop the previous session's terminal events on reconnect

### LINKED ISSUE #2587: [Bug]: Realtime event streams can terminate immediately after close and reconnect
### Prerequisites

## English

The OpenAI, Gemini, DashScope, and xAI realtime adapters allocate their event queue once in `__init__`. Calling `close()` cancels the reader task, whose `finally` block enqueues a `SessionEndedEvent` followed by the `None` end-of-stream sentinel. A later `connect()` starts a new reader without clearing or replacing that queue.

After an explicit close and reconnect, the new `events()` iterator can consume the previous session's terminal event and sentinel and stop before reading the new session's events. The provider WebSocket may be connected and receive input while its audio, transcript, and tool-call events remain unread.

Expected behavior: the same realtime model or agent instance can be explicitly closed and connected again without terminal events from the previous session ending the new event stream.

Actual behavior: the new event iterator terminates at the stale sentinel.

## 中文

OpenAI、Gemini、DashScope 和 xAI realtime adapter 都只在初始化时创建一次事件队列。`close()` 取消 reader 后，reader 的 `finally` 会写入 `SessionEndedEvent` 和 `None` 结束哨兵；再次 `connect()` 却不会清空或替换队列。

因此，同一实例显式关闭后重新连接时，新 `events()` 可能先读到旧 session 的结束哨兵并立即退出，而新 session 的事件留在队列中。预期重新连接后的事件流只受新 session 控制。

### Error Messages

```shell
No exception is required. A queue-level reproduction prints:


['SessionEndedEvent']
ResponseCreatedEvent


The ResponseCreatedEvent belongs to the new session but remains queued after the iterator has ended.

中文：该问题不需要抛出异常即可复现；上述输出表明新 session 的事件仍留在已经结束迭代的队列中。
```

### Steps to Reproduce

```python
import asyncio

from agentscope.realtime import OpenAIRealtimeModel
from agentscope.realtime import _events as events


async def main() -> None:
    model = OpenAIRealtimeModel(
        "gpt-realtime-2.1",
        OpenAICredential(api_key="unused"),
    )

    # These are enqueued by the previous reader's finally block on close.
    model._queue.put_nowait(events.SessionEndedEvent(reason="previous close"))
    model._queue.put_nowait(None)

    # Simulate the first event received by the new connection.
    model._queue.put_nowait(events.ResponseCreatedEvent(item_id="new-session"))

    print([type(event).__name__ async for event in model.events()])
    print(type(model._queue.get_nowait()).__name__)


asyncio.run(main())
```

The same queue lifecycle is present in all four provider adapters.

中文：以上代码模拟旧 reader 在关闭时留下终止事件和哨兵，再放入新连接的首个事件；迭代器会在读取新事件前结束。四个 provider adapter 均存在相同的队列生命周期。

### Environment

- AgentScope Version: 2.0.8 (main at 0fe41c07)
- Python Version: 3.12.14
- OS: macOS 26.6.2 arm64

L4XB:
## AgentScope Version

`2.0.8` (current `main`, 0fe41c0)

## Description

Fixes #2587.

**Background.** The OpenAI, Gemini, DashScope and xAI realtime adapters allocate their event queue once in `__init__`. `close()` cancels the reader task, whose `finally` block enqueues a `SessionEndedEvent` followed by the `None` end-of-stream sentinel. A later `connect()` started a new reader on the same queue without clearing it.

**Effect.** After an explicit close and reconnect on the same instance, the next `events()` iterator consumed the previous session's terminal event and sentinel and stopped, while the provider WebSocket was connected and receiving input; the new session's audio, transcript and tool-call events stayed unread.

**Change.** `connect()` drains the queue (`get_nowait` until empty) right before starting the reader, in all four adapters. The queue object itself is kept, so an `events()` iterator created before `connect()` keeps working, and a closed session that is *not* reconnected still terminates its consumer exactly as before (covered by a test).

**How to test.** `tests/realtime_reconnect_test.py` drives each adapter with a fake WebSocket (`websockets.connect` patched; for Gemini the fake replays `setupComplete` so `connect()` returns): the queue is pre-loaded with `SessionEndedEvent` + `None` as `close()` leaves it, `connect()` is called, a fresh `ResponseCreatedEvent` is enqueued, and `events()` must yield only that fresh event. Plus the no-reconnect case.

```
cd tests && python -m pytest realtime_reconnect_test.py realtime_openai_test.py realtime_dashscope_test.py realtime_gemini_test.py realtime_xai_test.py -q
# 68 passed
black --line-length=79 / flake8 on the touched files: clean
```

## Checklist

- [x] An issue has been created for this PR (#2587)
- [ ] Related documentation has been updated — no documentation describes this behaviour

### DIFF
--- a/src/agentscope/realtime/_openai/_model.py
+++ b/src/agentscope/realtime/_openai/_model.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 """The OpenAI realtime model."""
-
 import asyncio
 import base64
 import json
@@ -130,8 +129,13 @@ class OpenAIRealtimeModel(RealtimeModelBase):
             additional_headers=headers,
         )
         # A previous session leaves its terminal SessionEndedEvent and the
-        # None sentinel in the queue; a new events() iterator would stop
-        # on them before seeing anything from this session.
+        # None sentinel in the queue, and a reader that is still running
+        # would add them after the drain. Stop that reader first (its
+        # finally block enqueues the terminal events), then empty the queue
+        # so a new events() iterator only sees this session.
+        if self._reader is not None:
+            self._reader.cancel()
+            await asyncio.gather(self._reader, return_exceptions=True)
         while not self._queue.empty():
             self._queue.get_nowait()
         self._reader = asyncio.create_task(self._read(), name="openai-rt")
--- a/src/agentscope/realtime/_openai/_model.py
+++ b/src/agentscope/realtime/_openai/_model.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """The OpenAI realtime model."""
+
 import asyncio
 import base64
 import json
@@ -128,6 +129,11 @@ class OpenAIRealtimeModel(RealtimeModelBase):
             f"{url}/realtime?model={self.model}",
             additional_headers=headers,
         )
+        # A previous session leaves its terminal SessionEndedEvent and the
+        # None sentinel in the queue; a new events() iterator would stop
+        # on them before seeing anything from this session.
+        while not self._queue.empty():
+            self._queue.get_nowait()
         self._reader = asyncio.create_task(self._read(), name="openai-rt")
         await self._send(self._session_update(instructions, tools))
 
--- a/tests/realtime_reconnect_test.py
+++ b/tests/realtime_reconnect_test.py
@@ -1,0 +1,133 @@
+# -*- coding: utf-8 -*-
+"""Explicitly closing a realtime model and connecting again must give the new
+session a clean event stream (issue #2587)."""
+
+import asyncio
+import json
+from typing import Any
+from unittest.async_case import IsolatedAsyncioTestCase
+from unittest.mock import AsyncMock, patch
+
+from agentscope.credential import (
+    DashScopeCredential,
+    GeminiCredential,
+    OpenAICredential,
+    XAICredential,
+)
+from agentscope.realtime import (
+    DashScopeRealtimeModel,
+    GeminiRealtimeModel,
+    OpenAIRealtimeModel,
+    XAIRealtimeModel,
+)
+from agentscope.realtime import _events as me
+
+
+class IdleSocket:
+    """A WebSocket that replays a few frames, then stays open until closed."""
+
+    def __init__(self, frames: list[dict] | None = None) -> None:
+        self._frames = [json.dumps(_) for _ in (frames or [])]
+        self._closed = asyncio.Event()
+        self.sent: list[Any] = []
+
+    async def send(self, payload: str) -> None:
+        """Record what the model sends."""
+        self.sent.append(json.loads(payload))
+
+    async def close(self) -> None:
+        """Release the reader."""
+        self._closed.set()
+
+    def __aiter__(self) -> "IdleSocket":
+        return self
+
+    async def __anext__(self) -> str:
+        if self._frames:
+            return self._frames.pop(0)
+        await self._closed.wait()
+        raise StopAsyncIteration
+
+
+class ReconnectTest(IsolatedAsyncioTestCase):
+    """The previous session's terminal events must not end the new stream."""
+
+    async def _reconnect_and_collect(
+        self,
+        model: Any,
+        frames: list[dict] | None = None,
+    ) -> list[me.ModelEvent]:
+        # What close() leaves behind: the reader's finally block enqueues the
+        # terminal event and the end-of-stream sentinel.
+        model._queue.put_nowait(me.SessionEndedEvent(reason="closed"))
+        model._queue.put_nowait(None)
+
+        socket = IdleSocket(frames)
+        with patch("websockets.connect", new=AsyncMock(return_value=socket)):
+            await model.connect("Be brief.")
+        self.assertIsNotNone(model._reader)
+
+        fresh = me.ResponseCreatedEvent(item_id="item_new")
+        model._queue.put_nowait(fresh)
+        model._queue.put_nowait(None)
+        got = [event async for event in model.events()]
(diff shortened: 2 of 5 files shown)


wuwenbo0626:
Ops, we opened duplicate PRs, and your solution is much better and more robust than mine ;) I've closed my PR

L4XB:
Thanks for closing yours and for taking a look at this one — much appreciated. Happy to adjust anything the maintainers would like changed here.

src/agentscope/realtime/_openai/_model.py
```diff
@@ -128,6 +129,11 @@ async def connect(
             f"{url}/realtime?model={self.model}",
             additional_headers=headers,
         )
+        # A previous session leaves its terminal SessionEndedEvent and the
```
DavdGao, line 131: Draining here only covers the case where the previous reader has already exited. connect() never cancels the old reader, so when the provider closes the session and the read side hasn't noticed yet (half-open socket, _send already raised ConnectionClosed), the reconnect happens while the old reader is still running: the drain finds an empty queue, and the old reader writes SessionEndedEvent + None into the *new* session's stream when it finally dies, ending it. That's the same failure mode this PR is about. The old task is leaked too, since self._reader is overwritten and close() only cancels the newest one.

Suggested to cancel the old reader first and then drain -- the order matters, because the cancel runs the reader's finally, which enqueues the terminal events again:

```python
if self._reader is not None:
    self._reader.cancel()
    await asyncio.gather(self._reader, return_exceptions=True)
while not self._queue.empty():
    self._queue.get_nowait()
self._reader = asyncio.create_task(self._read(), name="openai-rt")
```

Same in the other three providers.

src/agentscope/realtime/_openai/_model.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """The OpenAI realtime model."""
+
```
DavdGao: Unrelated blank line after the module docstring, same in _gemini/_model.py. The other two providers don't have it -- please drop both.

tests/realtime_reconnect_test.py
```diff
@@ -0,0 +1,133 @@
...
+    """The previous session's terminal events must not end the new stream."""
+
+    async def _reconnect_and_collect(
+        self,
+        model: Any,
+        frames: list[dict] | None = None,
+    ) -> list[me.ModelEvent]:
+        # What close() leaves behind: the reader's finally block enqueues the
+        # terminal event and the end-of-stream sentinel.
+        model._queue.put_nowait(me.SessionEndedEvent(reason="closed"))
```
DavdGao: The test fakes what close() leaves behind by pushing into _queue directly, so it stays green even if the reader's finally changes. Suggested to call `await model.close()` and then connect() again, so the real lifecycle is what is under test.

tests/realtime_reconnect_test.py
```diff
@@ -0,0 +1,133 @@
+# -*- coding: utf-8 -*-
```
DavdGao, line 1: `uv run pre-commit run --all-files` fails on this file: it needs `# pylint: disable=protected-access` (every other test file that touches privates has it) and the four test methods are missing docstrings.

DavdGao (changes requested):
Verified the bug through the public API, not just the queue-level repro in the issue: agent.connect() -> agent.close() -> agent.connect() leaves the second session with _connected=False while the WebSocket is up, because close() cancels the downlink pump before model.close() enqueues the terminal events, so nobody consumes them. Good catch.

The fix works for that path but misses the other half, see the comment on _openai/_model.py. A few nits on the test as well.

src/agentscope/realtime/_openai/_model.py
L4XB: You're right, the drain alone only covered a reader that had already exited. In 5df65fe `connect()` now cancels the previous reader and awaits it (`asyncio.gather(..., return_exceptions=True)`) before emptying the queue, in exactly that order, so the reader's `finally` enqueues its terminal events before the drain and the task is no longer leaked. Applied to all four providers.

src/agentscope/realtime/_openai/_model.py
L4XB: Dropped the blank line after the module docstring in both `_openai/_model.py` and `_gemini/_model.py`.

tests/realtime_reconnect_test.py
L4XB: Rewritten to the real lifecycle: the helper now does `connect()` → `close()` → `connect()` against the idle fake socket, so whatever `close()` actually leaves in the queue is what the second session has to cope with; only the new session's own event is fed in afterwards to read the stream back. There is also a new case for the half-open path you described (`connect()` twice without `close()`), which asserts the old reader task is done and the new stream is clean; it fails on the previous commit. One detail of the fake socket: its `send()` now yields to the event loop like a real write, otherwise the reader task never starts before `close()` cancels it.

tests/realtime_reconnect_test.py
L4XB: Added `# pylint: disable=protected-access` at the top and docstrings on every test method; `pylint` with the pre-commit arguments, `black --line-length=79` (23.3.0) and `flake8` are clean on the file.

L4XB:
Thanks for the thorough check. 5df65fe addresses all four points: the old reader is cancelled and awaited before the drain in every provider, the stray blank lines are gone, the test drives the real connect/close/connect lifecycle (plus a reconnect-without-close case that fails on the previous commit), and the file passes pylint/black/flake8 with the pre-commit settings.

DavdGao (approved):

PR #2592, opened 2026-09-11
fix(utils): preserve instance values during schema flattening

Tsubaki-01:
## AgentScope Version

2.0.8

## Description

Follow-up to #2575, addressing [the review feedback](https://github.com/agentscope-ai/agentscope/pull/2575#pullrequestreview-5174928051). Refs #2574.

`_flatten_json_schema` still treated instance data under `default`, `const`, `enum`, and `examples` as schemas. This could remove ordinary `definitions` / `$defs` keys from defaults or expand literal `$ref` values inside enums.

- Preserve instance values verbatim.
- Use one field-processing loop for ordinary schemas and `$ref` siblings, and move the keyword sets to module-level constants.
- Add regression tests for schema-like property names and instance values, including values next to a schema `$ref`.

## Validation

- Before the fix, the instance-value regression failed in all 8 subcases.
- `python -m pytest tests/model_gemini_test.py -q -p no:cacheprovider`: 32 tests and 12 subtests passed.

## Checklist

- [x] An issue has been created for this PR (#2574; follow-up to #2575).
- [x] I have read the CONTRIBUTING.md.
- [ ] Related documentation has been updated (not applicable: internal schema-flattening bug fix).

### DIFF
--- a/src/agentscope/_utils/_common.py
+++ b/src/agentscope/_utils/_common.py
@@ -16,6 +16,15 @@ from .._logging import logger
 from ..exception import ToolJSONDecodeError
 
 
+# These mappings use property names as keys and schemas as values.
+_SCHEMA_MAP_KEYWORDS = frozenset(
+    {"properties", "patternProperties", "dependentSchemas", "dependencies"},
+)
+# These keywords hold instance data, which must be kept verbatim.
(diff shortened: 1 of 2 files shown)


qbc2016 (approved):
LGTM.

PR #2597, opened 2026-09-12
fix(app): pass model card's context_size to chat model

### LINKED ISSUE #2596: [Bug]: Model card's context_size is ignored by the app service chat model factory
### Prerequisites

### Background / Description

`get_model()` in `agentscope/app/_service/_model.py` builds the chat model with only `credential`, `model` and `parameters`. The matching model card's `context_size` is never forwarded to the constructor, so every chat model silently falls back to its per-provider class default.

Model cards declare much larger windows than those defaults. For example, the DeepSeek card (`deepseek-v4-flash`) declares `context_size: 1000000`, while `DeepSeekChatModel.__init__` defaults to `65536` — about 6.5% of the real window.

This is not cosmetic. `Agent` derives every context-compression threshold from `self.model.context_size` (`agent/_agent.py`, e.g. `trigger_ratio * self.model.context_size`), so an under-reported value makes compression fire far too early. The same applies to the fallback chat model, which goes through the same factory.

Evidence that this is an oversight rather than intended: the sibling embedding path in the same package does forward it. `app/_service/_embedding.py` looks up the matching card and passes `card.context_size` into the constructor. The chat path looks up the very same card but only uses it to override `formatter.input_types`. In addition, `CONTRIBUTING.md` §Chat Model lists `context_size` as a required model card field, so the declared value is meant to be authoritative.

This affects every provider whose card value differs from the class default, not just DeepSeek — e.g. DashScope (131072 vs 10000000 for `qwen-long`) and Anthropic (200000 vs 1000000 for the Claude 5 / 4.6+ cards).

The card's `output_size` is likewise never read on the chat path.

### Error Messages

```shell
No exception is raised — the defect is silent and only manifests as premature context compression.

Observed by running the reproduction below against agentscope 2.0.8:
  card declares:           1000000
  constructed model has:     65536
```

### Steps to Reproduce

import inspect
from agentscope.model._deepseek import DeepSeekChatModel

### What the model card declares
for card in DeepSeekChatModel.list_models():
    print(card.name, card.context_size)
### What the constructor falls back to
print(
    inspect.signature(DeepSeekChatModel.__init__)
    .parameters["context_size"].default
)
### Environment

- AgentScope Version: 2.0.8
- Python Version: 3.14.5
- OS: Windows 11 Pro

code-code567:
Fixes #2596

## AgentScope Version

2.0.8 (`import agentscope; print(agentscope.__version__)`)

## Description

`get_model()` built the chat model without forwarding the matching model
card's `context_size`, so every model silently kept its provider class
default — 65536 for DeepSeek while the card declares 1000000. `Agent`
derives every context-compression threshold from `model.context_size`,
so an under-reported value made compression fire far too early. The
fallback chat model goes through the same factory.

The card lookup that already patched `formatter.input_types` now also
sets `context_size` from the matching card. Both are assigned after
construction, since `ChatModelBase.__init__` only stores `context_size`
as an attribute. Models without a card keep the constructor default, and
the existing graceful degradation when the lookup fails is preserved.

I opened this PR while the issue is still open — happy to adjust the
approach or close it if you would rather fix this differently.

## How to test

```
pytest tests/service_model_test.py
```

The first case fails on main with `AssertionError: 1000000 != 65536`
and passes with this change; the second locks in the fallback for
card-less custom models.

Full suite: `2233 passed, 299 skipped` with no new failures — the 7
failures in `storage_sql_test.py` / `test_e2e_api.py` are pre-existing
on main and reproduce without this change.

## Checklist

- [x] Related documentation has been updated — not applicable, no
      user-visible API change

### DIFF
--- a/src/agentscope/app/_service/_model.py
+++ b/src/agentscope/app/_service/_model.py
@@ -51,23 +51,33 @@ async def get_model(
         if config.parameters
         else None
     )
+    # Look up the built-in model card so both the context size and the
+    # formatter's input types can be taken from it. Custom models have no
+    # card, so keep the constructor and formatter defaults.
+    card = None
+    try:
+        for candidate in model_cls.list_models():
+            if candidate.name == config.model:
+                card = candidate
+                break
+    except Exception:  # pylint: disable=broad-except
+        logger.debug(
+            "Failed to look up model card for %s, using defaults.",
+            config.model,
+        )
+
+    model_kwargs: dict = {}
+    if card is not None:
+        model_kwargs["context_size"] = card.context_size
+
     model = model_cls(
         credential=credential,
         model=config.model,
         parameters=parameters,
+        **model_kwargs,
     )
 
-    # Override the formatter's input types with the built-in model card's
-    # when one matches; custom models have no card, so keep the default.
-    try:
-        for card in model_cls.list_models():
-            if card.name == config.model:
-                model.formatter.input_types = card.input_types
-                break
-    except Exception:  # pylint: disable=broad-except
-        logger.debug(
-            "Failed to look up model card for %s, using formatter defaults.",
-            config.model,
-        )
+    if card is not None:
+        model.formatter.input_types = card.input_types
 
     return model
--- a/tests/service_model_test.py
+++ b/tests/service_model_test.py
@@ -1,0 +1,92 @@
+# -*- coding: utf-8 -*-
+"""Unit tests for :func:`get_model` — the chat model factory.
+
+The factory used to build the model with only ``credential``, ``model``
+and ``parameters``, so the matching model card's ``context_size`` never
+reached the constructor and every model silently kept its provider class
+default — 65536 for DeepSeek while the card declares 1000000. ``Agent``
+derives every context-compression threshold from
+``model.context_size``, so an under-reported value makes compression
+fire far too early.
+"""
+from types import SimpleNamespace
+from unittest import IsolatedAsyncioTestCase
+
+from agentscope.app._service._model import get_model
+from agentscope.app.storage import ChatModelConfig
+
+
+class _StubAccess:
+    """Stand-in for ``ResourceAccessService``.
+
+    ``get_model`` only reads the resolved record's ``data``.
+    """
+
+    def __init__(self, data: dict) -> None:
+        self._data = data
+
+    async def resolve_credential(
+        self,
+        user_id: str,
+        credential_id: str,
+    ) -> SimpleNamespace:
+        """Return the canned credential record.
+
+        Args:
+            user_id (`str`):
+                The viewer's user id, unused.
+            credential_id (`str`):
+                The credential id, unused.
+
+        Returns:
+            `SimpleNamespace`:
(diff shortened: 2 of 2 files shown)


code-code567:
Hi maintainers,

This is my first contribution to AgentScope, so the workflows on this PR are showing `action_required` and waiting for a maintainer to approve them.

Would someone mind approving the runs so CI can verify the change? Thanks!

For context: this fixes #2596 — the app service chat model factory never forwarded the matching model card's `context_size` to the chat model constructor, so models silently kept their provider class default (65536 for DeepSeek) while the card declares 1000000.

src/agentscope/app/_service/_model.py
```diff
@@ -51,23 +51,33 @@ async def get_model(
...
+
+    model = model_cls(
+        credential=credential,
+        model=config.model,
+        parameters=parameters,
+        **model_kwargs,
+    )
+
+    if card is not None:
+        model.formatter.input_types = card.input_types
```
DavdGao: The refactor here is unnecessary. `context_size` is only stored as an attribute in `ChatModelBase.__init__`, and we already patch `formatter.input_types` after construction, so one line in the existing loop is enough:
```python
for card in model_cls.list_models():
    if card.name == config.model:
        model.context_size = card.context_size
        model.formatter.input_types = card.input_types
        break
```

tests/service_model_test.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
+"""Unit tests for :func:`get_model` — the chat model factory.
+
+The factory used to build the model with only ``credential``, ``model``
+and ``parameters``, so the matching model card's ``context_size`` never
+reached the constructor and every model silently kept its provider class
+default — 65536 for DeepSeek while the card declares 1000000. ``Agent``
+derives every context-compression threshold from
+``model.context_size``, so an under-reported value makes compression
+fire far too early.
```
DavdGao: - Keep the module docstring to one line, the bug background belongs in the PR description.
- `_config` is only used twice, suggested to inline the `ChatModelConfig`.
- `_StubAccess` can be replaced with `AsyncMock`, e.g. `access = MagicMock(resolve_credential=AsyncMock(return_value=SimpleNamespace(data={...})))`.

DavdGao (changes requested):
Thanks for the fix, the issue is valid. Left some comments to simplify it.

code-code567:
Thanks for the review, @DavdGao — all three points are in.

`_model.py`: dropped the refactor. The lookup stays where it was and now just sets `model.context_size = card.context_size` next to `formatter.input_types`. I also updated the comment and the debug message, since they now cover both fields.

`service_model_test.py`: one-line module docstring, `ChatModelConfig` inlined, and `_StubAccess` replaced with `MagicMock(resolve_credential=AsyncMock(...))`.

Re-verified after the change: removing the `context_size` assignment makes the regression test fail with `AssertionError: 1000000 != 65536`, while the unknown-model test still passes. `pre-commit run --files` is green.

DavdGao (approved):

PR #2609, opened 2026-09-13
test(workspace): skip POSIX-shell-dependent tests on Windows hosts without coreutils

### LINKED ISSUE #2608: [Bug]: three tests fail on Windows dev machines without Git usr/bin on PATH (POSIX shell tool dependency)

### Background / Description

Three tests fail on a stock Windows dev machine where Git''s `usr/bin` (POSIX coreutils: `mv`, `mkdir`, `test`, `rm`) is not on `PATH`:

```
FAILED tests/workspace_skill_archive_test.py::AddSkillArchiveSandboxedTest::test_directory_name_is_suffixed_when_taken
FAILED tests/workspace_skill_archive_test.py::AddSkillArchiveSandboxedTest::test_flat_archive_is_accepted
FAILED tests/workspace_daytona_test.py::TestDaytonaWorkspaceBuiltinToolsMock::test_offload_context_tool_result_and_reset
```

Both clusters depend on POSIX shell tools being executable on the *host*:

1. `AddSkillArchiveSandboxedTest` deliberately exercises the base `WorkspaceBase.add_skill_archive` implementation over a `LocalBackend`. That base implementation is written for POSIX sandboxes and shells out to `mkdir -p` / `mv` (`src/agentscope/workspace/_base.py`), which do not exist on stock Windows — `LocalBackend.exec_shell` spawns programs directly without a shell. (`LocalWorkspace.add_skill_archive` itself is overridden and its own test class passes; only the base-implementation-over-local-backend tests fail.)

2. `test_offload_context_tool_result_and_reset` runs through the "host-backed Daytona fake" (`_MappedProcess`), which re-executes sandbox shell commands (`test -e`, `mkdir -p`, `rm -rf`) on the host via `asyncio.create_subprocess_shell` — i.e. `cmd.exe` on Windows, where these commands do not exist.

These pass on `windows-latest` CI because the runner image puts Git''s `usr/bin` on `PATH`, so `mv.exe`/`test.exe`/etc. resolve. On a dev machine without that on `PATH` (e.g. Git installed elsewhere, or only `Git/cmd` on `PATH`), they fail. The repo already acknowledges this class of problem: `test_add_skill_rejects_duplicate_skill` in the same Daytona test file carries `@unittest.skipIf(os.name == "nt", "host-backed Daytona fake relies on POSIX sandbox commands")`.

### Error Messages

```
RuntimeError: Failed to install skill ''pack'': [WinError 2] The system cannot find the file specified
  File "src/agentscope/workspace/_base.py", line 1399, in add_skill_archive
```

```
tests/workspace_daytona_test.py:1254: in test_offload_context_tool_result_and_reset
    self.assertTrue(await self.workspace._backend.file_exists("/home/daytona/sessions"))
E   AssertionError: False is not true
# (`test -e` runs on the host via cmd.exe and fails, so file_exists is always False)
```

### Steps to Reproduce

1. On Windows, ensure Git''s `usr/bin` is not on `PATH` (`where mv` finds nothing).
2. `uv venv --python 3.11 && uv pip install -e ".[dev]"`
3. `python -m pytest tests/workspace_skill_archive_test.py tests/workspace_daytona_test.py -q`
4. Observe the three failures above.

### Environment

- AgentScope Version: 2.0.8 (main @ b82253b)
- Python Version: 3.11.15 / 3.13.9 (both affected)
- OS: Windows 11

fszcd:
## AgentScope Version

2.0.8 (main @ b82253b)

## Description

Fixes agentscope-ai/agentscope#2608

**Background.** Three tests fail on Windows dev machines where Git''s `usr/bin` (POSIX coreutils) is not on `PATH`:

- `tests/workspace_skill_archive_test.py::AddSkillArchiveSandboxedTest::test_directory_name_is_suffixed_when_taken`
- `tests/workspace_skill_archive_test.py::AddSkillArchiveSandboxedTest::test_flat_archive_is_accepted`
- `tests/workspace_daytona_test.py::TestDaytonaWorkspaceBuiltinToolsMock::test_offload_context_tool_result_and_reset`

Root causes:

1. The base `WorkspaceBase.add_skill_archive` implementation targets POSIX sandboxes and shells out to `mkdir -p` / `mv`. The test class deliberately exercises it over a `LocalBackend`, which spawns programs directly (no shell) — so on stock Windows the executables do not exist and the install fails with `[WinError 2]`.
2. The host-backed Daytona fake (`_MappedProcess`) re-executes sandbox shell commands (`test -e`, `mkdir -p`, `rm -rf`) on the host via `asyncio.create_subprocess_shell` (`cmd.exe` on Windows), where these commands do not exist.

Both pass on `windows-latest` CI only because the runner image puts Git''s `usr/bin` on `PATH`.

**Change.** Skip the three tests when the required executables are unavailable (`shutil.which` guards), mirroring the existing `@unittest.skipIf(os.name == "nt", "host-backed Daytona fake relies on POSIX sandbox commands")` precedent on `test_add_skill_rejects_duplicate_skill` in the same Daytona file. No production code touched; coverage is unchanged on hosts that do have the tools (including CI).

**How to test.** On a Windows machine with `usr/bin` *not* on `PATH`:

```
python -m pytest tests/workspace_skill_archive_test.py tests/workspace_daytona_test.py -q
# before: 3 failed; after: 39 passed, 13 skipped
```

And with `usr/bin` on `PATH` (CI-equivalent), the guarded tests still run and pass:

```
# PATH includes D:\Git\usr\bin
python -m pytest tests/workspace_skill_archive_test.py::AddSkillArchiveSandboxedTest ^
    tests/workspace_daytona_test.py::TestDaytonaWorkspaceBuiltinToolsMock::test_offload_context_tool_result_and_reset -q
# 5 passed
```

## Checklist

- [x] Related documentation has been updated (e.g. links, examples, etc.) in [documentation repository](https://github.com/agentscope-ai/docs) (not applicable — test-only change)


tests/workspace_daytona_test.py
```diff
@@ -1514,6 +1518,10 @@ async def test_skill_paths_add_list_remove(self) -> None:
                     [skill.name for skill in await workspace.list_skills()],
                 )
 
+    @unittest.skipIf(
+        os.name == "nt" and shutil.which("test") is None,
```
DavdGao: This one is the live test, commands run in the real Daytona sandbox rather than on the host, so the skip doesn't apply here. Please remove it.

tests/workspace_daytona_test.py
```diff
@@ -1208,6 +1208,10 @@ async def test_add_skill_rejects_duplicate_skill(self) -> None:
         with self.assertRaisesRegex(ValueError, "already exists"):
             await self.workspace.add_skill(skill_dir)
 
+    @unittest.skipIf(
+        os.name == "nt" and shutil.which("test") is None,
```
DavdGao: `os.name == "nt"` is redundant here, `shutil.which("test") is None` alone is enough.

tests/workspace_skill_archive_test.py
```diff
@@ -18,6 +20,13 @@
 from agentscope.workspace import LocalWorkspace
 from agentscope.workspace._base import WorkspaceBase
 
+# The base `WorkspaceBase.add_skill_archive` implementation targets
+# POSIX sandboxes and shells out to `mv` / `mkdir`.  Exercised over
+# a LocalBackend (which spawns programs directly, without a shell),
+# that only works on Windows when Git's `usr/bin` happens to be on
+# `PATH` -- as it is on CI runners but not on every dev machine.
+_HAS_POSIX_MV = os.name != "nt" or shutil.which("mv") is not None
```
DavdGao: The `os.name` check is redundant since POSIX hosts always have `mv`. Suggested to just use `@unittest.skipIf(shutil.which("mv") is None, ...)` inline, and drop `_HAS_POSIX_MV` together with the long comment above it.

DavdGao (changes requested):
Thanks for the fix. The live Daytona test shouldn't be skipped, and the skip conditions can be simplified, see inline comments.

tests/workspace_daytona_test.py
fszcd: You're right, my mistake -- the skip landed on the live test too because I edited by method name and this class has a same-named method. Removed it; the live test runs against the real sandbox and doesn't depend on host shell tools.

tests/workspace_daytona_test.py
fszcd: Agreed, simplified to `shutil.which(\test\) is None` alone.

tests/workspace_skill_archive_test.py
fszcd: Done -- dropped `_HAS_POSIX_MV` and the comment block; both tests now use the inline `shutil.which(\mv\) is None` guard. All re-verified locally: the three tests skip cleanly on a host without coreutils, and still run and pass with Git's `usr/bin` on `PATH` (5 passed).

fszcd:
Thanks for the careful review, @DavdGao -- all three comments addressed in the latest push: the live Daytona test is untouched again, and both guard styles are simplified to inline `shutil.which(...) is None`. Re-verified locally on a Windows host without Git's `usr/bin` on `PATH` (the three tests skip cleanly; the rest of the two files still pass) and with `usr/bin` present (guarded tests run and pass, 5 passed). Ready for another look.

DavdGao (approved):

PR #2613, opened 2026-09-13
fix(formatter): keep media from tool results after all tool messages of a turn

### LINKED ISSUE #2612: [Bug]: Media from a parallel tool result splits the tool messages, so the next request is rejected
### Prerequisites

### Background / Description

When one assistant reply has two or more tool calls and a result that isn't the last one carries an image, audio or PDF, the OpenAI, DashScope, Moonshot and Volcengine chat formatters put the promoted media message between the tool messages:

```
assistant(tool_calls=[call_shot, call_title])
tool(call_shot)
user(image)        <- inserted here
tool(call_title)
```

The Chat Completions API requires every `tool_call_id` of an assistant message to be answered by tool messages before any other message, so this request is rejected with a 400. DashScope returns the same error (see #772). Since the reply is stored in memory, every later request in the session fails too.

In v2 one assistant `Msg` holds all the tool calls and their results, and each chat formatter appends the `system-reminder` user message right after the tool message that produced the media.

### Error Messages

Not reproduced against a live endpoint. The OpenAI rule is "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'". The request shape is reproduced below.

### Steps to Reproduce

Format an `AssistantMsg` with `ToolCallBlock(call_shot)`, `ToolCallBlock(call_title)`, a `ToolResultBlock` for `call_shot` whose output has a `DataBlock` image, then a text-only `ToolResultBlock` for `call_title`:

```
OpenAIChatFormatter / DashScopeChatFormatter / MoonshotChatFormatter / VolcengineChatFormatter:
  [3] assistant tool_calls=['call_shot', 'call_title']
  [4] tool tool_call_id=call_shot
  [5] user parts=['text', 'text', 'image_url', 'text']
  [6] tool tool_call_id=call_title
```

A real case is an agent reading a screenshot and a text file in parallel.

### Environment

agentscope 2.0.8 (main at b82253b), Python 3.12, macOS.

I have a fix with tests (hold the promoted media until the run of tool messages ends) and will open a PR. This issue was written with AI assistance (Claude Code).

alanhuangyoo:
## AgentScope Version

2.0.8 (main at b82253b)

## Description

Fixes #2612.

When one assistant reply has parallel tool calls and a result that isn't the last one carries media, the OpenAI, DashScope, Moonshot and Volcengine chat formatters put the promoted `system-reminder` user message between the tool messages: `assistant(tool_calls A, B)`, `tool A`, `user(image)`, `tool B`. The API requires all tool messages for an assistant's tool calls before any other message, so the request is rejected, and because the reply is in memory the session keeps failing.

Each formatter now holds the promoted media message until the run of tool results ends: it is emitted before the next block that isn't a tool result, or at the end of the message. The order becomes `assistant(tool_calls A, B)`, `tool A`, `tool B`, `user(image)`. A single tool call, or a result without media, formats exactly as before. The multi-agent formatters reuse the chat formatters for tool sequences, so they get the fix too.

Tests: one case per formatter test file with two tool calls where the first result has an image, asserting the `(role, tool_call_id)` order. All four fail on `main`. `tests/formatter_*_test.py`: 127 passed. add-trailing-comma, black 23.3.0 (line length 79) and flake8 6.1.0 are clean on the changed files, and pylint 3.0.2 with the repo's arguments rates them 10.00/10. I couldn't run the agent tests locally (`apscheduler` isn't installed).

## Checklist


src/agentscope/formatter/_openai_formatter.py
```diff
@@ -269,7 +269,16 @@ async def format(
             content_blocks = []
             tool_calls = []
 
+            # Media promoted from tool results waits until the run of tool
+            # messages ends: a user message between them would leave the
+            # remaining tool_call ids unanswered, which the API rejects.
```
DavdGao: Three-line inline comment is too long, and it's repeated in all four files. One line is enough, e.g. "Hold the promoted media until the tool messages of this turn are all emitted."

src/agentscope/formatter/_openai_formatter.py
```diff
@@ -269,7 +269,16 @@ async def format(
...
+            pending_media: list[dict] = []
```
DavdGao, line 273: The same `pending_media` handling is missing in `_ollama_formatter.py`, which builds the tool/user messages the same way.

tests/formatter_openai_chat_test.py
```diff
@@ -907,3 +907,61 @@ async def test_chat_formatter_hint_block_multimodal(self) -> None:
...
+                        state=ToolResultState.SUCCESS,
+                    ),
+                    TextBlock(text="The page is Example Domain."),
+                ],
+            ),
+        ]
+
+        res = await fmt.format(msgs)
+
+        self.assertListEqual(
```
DavdGao, line 966: The assertion only compares `(role, tool_call_id)`, so a wrong content would still pass. Assert the full structure in one `assertListEqual`, and use `AnyString()` from `tests/utils.py` for the generated identifier in the system-reminder text. Same for the other three test files.

DavdGao (changes requested):
Reproduced the wrong order on main, and this branch does fix it. Two things before merging:

- `OllamaChatFormatter` has exactly the same code (`_ollama_formatter.py:233`) and the same bug. I ran this branch and it still gives `tool -> user -> tool`. Please cover it here.
- `GeminiChatFormatter` splits the `function_response` contents the same way. That one already splits without media, so it's a different problem — just check whether Gemini accepts it, and open a separate issue if not.

src/agentscope/formatter/_openai_formatter.py
alanhuangyoo: Shortened to one line in all five formatters (87b5b7e).

src/agentscope/formatter/_openai_formatter.py
alanhuangyoo: Covered in 87b5b7e: `OllamaChatFormatter` now holds the media message the same way. Building that message moved into `_format_tool_result_media`, since `format()` was already at pylint's `max-branches=30` and the extra check would have gone over.

tests/formatter_openai_chat_test.py
alanhuangyoo: Done: all five tests now compare the full formatted list in one `assertListEqual`. I used the files' existing `_FIXED_ID` on the `DataBlock` instead of `AnyString()`, so the identifier inside the system-reminder text is checked exactly too (added the same constant to the Volcengine test file). All five fail against the formatters before this PR.

alanhuangyoo:
Thanks for the review. Pushed 87b5b7e:

- Ollama: same fix, and `tool -> tool -> user` now (new test in `formatter_ollama_test.py`).
- Comments cut to one line; tests assert the full structure.
- Gemini: it doesn't accept that shape. A model turn with N `functionCall` parts has to be answered by one turn with N `functionResponse` parts, otherwise the API returns 400 "Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn". Opened #2628 with the request body agentscope sends today.

black / flake8 / pylint (with the pre-commit args) and mypy are clean on the changed files, and the formatter tests pass (61).

DavdGao (approved):
lgtm

PR #2617, opened 2026-09-14
feat(app): standard operating procedures at the service layer

DavdGao:
Standard operating procedures at the service layer: what is stored, how a stored procedure runs, and the HTTP surface for both.

## 1. What is stored

A `SOP` holds live agents — model clients, toolkits, MCP connections — so it is code and never lands in a database. Two things do:

- **`SOPData`** — the description a service rebuilds a procedure from. New, because nothing in the SDK describes a procedure without also being able to run it.
- **`SOPRunState`** — how a run is going. Reused from the SDK verbatim; a mirror model for that shape would be a second truth.

### A step names its executor as an agent plus a session key

```python
SOPAgentRef(agent_id="agent-1", session_key="modeller")
```

Two references sharing a key share one session, and therefore one context. Sharing is something the author declares rather than something derived from which agent happened to be picked: the same agent can carry its context across several steps or stay separate in each, and picking one agent twice by accident never silently joins two conversations.

It also keeps the run record free of any assumption about a step's shape — `sessions` maps keys to session ids, so a step that grows a third role needs nothing there. `session_settings` hangs off the same key, since the model and permission mode belong to the conversation; it reuses the `SessionSettings` a channel already needs for the sessions it opens.

### A run keeps a copy of the procedure

Editing a SOP would otherwise strand every run of it still waiting on someone: the engine refuses a state whose steps no longer line up, which is right for the SDK and fatal in a UI where definitions are editable.

### Versioning

`SOPStepDataV1` carries `version: Literal["v1"]` from the start. When a step's runtime stops being able to run what an older one wrote, the new one becomes its own model beside this one, discriminated on that tag — every stored row already carries its own, so that day is additive. This pays off only if V1 is kept when V2 lands; that is the commitment the tag makes.

### Schema

`sops` and `sop_runs`, both new. `phase` is promoted to a column on `sop_runs` via `_index_paths` so "which of my runs is waiting on someone" is a query rather than a scan — read out of the run state's derived phase, so it is a copy and never a second place to write. Redis has nothing to index it by and filters after reading; that difference is in the docstring rather than papered over.

`SOPOrigin` joins `SessionOrigin` so a session a run opened says which run and which of its conversations it is.

## 2. How it runs

A step does not drive an agent. It hands the turn to `ChatService.run`, which already owns everything a turn needs — session lock, toolkit assembly, workspace, permissions, event fan-out, message persistence, failure reporting. A second assembler for SOP would have duplicated all of that and then drifted from it.

What comes back is read out of the run state, not out of the reply.

### Why a tool and not structured output

`structured_output` never travels as an event (`agent/_agent.py:3564` puts it on the final `exit_msg` only), and `_run_impl` rebuilds its reply from events — so the stored message never carries it. Reading a handover that way means reaching into `agent.state.reply_context`.

Instead the step's agent carries one of `SubmitHandover` / `SubmitVerdict` and writes the run state itself, the way the team tools reach storage themselves. Nothing changes in how `ChatService` streams or persists, the submission is visible in the transcript as a tool call, and the human case becomes the same shape as the agent case.

`SOPStepSubmitMiddleware` supplies the enforcement structured output would have given — swallowing a reply-end that submitted nothing and nudging, then failing the reply after three, exactly as `TeamMemberLoopMiddleware` holds a member to reporting. A failed attempt is then spent from the step's budget rather than hanging the run.

### Whose turn it is

A step's session is an ordinary session: a person can open it and type into it. Attaching the submit tool to every reply on such a session would hold that person's words to a submission the step never asked them for, and then send them off to be reviewed as the deliverable.

So a run claims the session in the bus registry, beside the locks, for as long as the step expects something of it — and both the tool and the middleware are attached on that claim rather than on the session's origin. In-flight state, not stored state: a claim is the same kind of thing as a lock, and it carries the step index so the toolkit does not have to work out which step is running.

A parked turn keeps its claim. Parking on a tool call is mid-attempt, not an ending: the person who approves it resumes that same reply through the ordinary chat endpoint, and still has to be able to hand the work over.

### A human verifier needs no runtime

It parks the run. Whoever is asked files the verdict through the same `SubmitVerdict` an agent reviewer calls, and the next pass picks up from it. Nothing is faked into a session and no reply or tool-call ids have to survive storage: the transcript shows interactions, the run state shows verdicts.

### The run lock

`MessageBusKeys.sop_run_lock` is held for a whole advance and always taken before any session lock. It stops two advances of one run from each reading the state, changing it in memory, and writing back over each other — a verdict would simply vanish. Filing a person's verdict takes it too, which is why that lives in `SOPService.record_verdict` rather than in the router.

Only ever one writer, therefore, and reading the run state and writing it back whole stays correct — the same thing `update_session_state` does under the session lock.

### Nothing is told to carry a run on

A person answering a tool call in a step's session answers it through the ordinary chat endpoint, so a chat run that finishes on a procedure's session advances that procedure itself — unless the run's lock is held, which means the run dispatched that reply and reads the result back anyway.

An interrupted reply is the exception: that is a person saying stop, and since the step has handed nothing over, carrying on would mean dispatching it again. A turn that ended `INTERRUPTED` leaves the run where it is.

An advance is never awaited. It can be the whole rest of the procedure, and holding a request — or a finished chat turn, and with it that session's slot in `ChatRunRegistry` — open for it would refuse the next message to that session for as long as the run had left to go. `SOPService` is entered on the lifespan's stack so those detached advances are stopped before the storage they write to is closed.

### Sessions, workspace, and cleanup

`create_run` opens one session per `session_key` up front, so dispatching never has to ask whether one exists. They share a workspace under the default `RUN` grain — deliverables move between steps as files, and handover text cannot carry a `.blend`. Assigned the way the team flow already gives a worker the leader's workspace (`_agent_create.py:447`).

Deleting a run deletes those sessions, and deleting a procedure deletes its runs and theirs. The run minted every one of them, and left behind they are sessions nobody opened — still listed under their agent, still carrying the procedure's instructions, and still wakeable by a background tool finishing long after the run is gone. `delete_team` cascades the members it created for the same reason.

## 3. The HTTP surface

```
GET    /sop/schema                 the schema the editor is drawn from
GET    /sop/                       list procedures
POST   /sop/                       create
GET    /sop/{sop_id}               read one
PATCH  /sop/{sop_id}               replace its contents
DELETE /sop/{sop_id}               delete it, its runs, and their sessions

POST   /sop/{sop_id}/runs          start a run
GET    /sop/runs                   list runs (by procedure, by phase)
GET    /sop/runs/{run_id}          read one run
DELETE /sop/runs/{run_id}          delete a run and its sessions
POST   /sop/runs/{run_id}/verdict  file a person's verdict on a step
```

Starting a run and filing a verdict both answer as soon as what they were asked about is done — the run exists, the verdict is recorded — and leave the run to proceed on its own. The client follows it through its sessions' streams and `GET /sop/runs/{run_id}`, so there is no endpoint whose job is to restart a run.

The schema is returned as pydantic emits it rather than flattened: the agent editor's is flattened because some model providers choke on `$ref`, but nothing here is shown to a model, and a verifier is a tagged union whose `discriminator.mapping` points into `$defs` — inlining the variants would leave that mapping dangling.

Two things a stored procedure can say that make a run of it impossible are refused at 422, while they are still an edit: a step naming a conversation the procedure never configured, and a procedure with no steps at all — a run of that one could reach no phase but pending, so nothing would ever end it.

## Known limitations

**Stopping a run.** `SOPService.run` accepts a `UserInterruptEvent` because the engine does, and does nothing with it; nothing passes one today. Stopping a run has to reach a turn already in flight, which means going through `ChatRunRegistry` rather than calling `ChatService.run` directly, and the stop cannot take the run lock it would be interrupting. Recorded where the signature makes the promise.

**A person typing into a worker's session.** The SOP side of this is what the dispatch claim above fixes. `TeamMemberLoopMiddleware` is still attached on "this session is a worker" rather than on "the leader dispatched this turn", so a worker's session behaves the way a step's used to. Fixing one without the other would leave two answers to one question; both are worth settling together, and neither is settled here.

## Tests

Storage, both backends: round trip with the full record asserted, owner scoping, listing by procedure and by phase, the update path that rewrites state and sessions but not the snapshot, the cascade from deleting a procedure or a run, and that a step state carrying more than the base record survives the database. `test_migrations_alone_create_the_sop_tables` runs with `create_tables=False` so a broken 0004 fails in CI rather than being masked by `metadata.create_all`, matching what 0003 did.

Service: the chat service is replaced by a script — each entry says what the session handed the turn does with it, written through the real submit tools. Covers the walk through a step, a refusal coming back with its critique, the attempt budget running out, a turn that submits nothing, a human verifier parking and then resuming from a filed verdict, one session per key, both cascades, the claim's lifetime through a parked turn and through a turn that raises, the two ways a verdict is refused, and that leaving the service ends the advances it started.

Router: a real app over fakeredis, through `TestClient`. Covers the procedure's CRUD, the 422s, that every `$ref` the schema names resolves, starting a run and finding it in the listings, that a verdict answers before the run carries on, the 409 and 404 a verdict is refused with, and that another user sees none of it.


DavdGao (commented):
Main flow works on a single node, but the failure paths and the distributed ones aren't ready yet. Inline comments cover the details; in short:

- A reviewer that doesn't submit or parks loops forever (reproduced).
- A stale dispatch claim can let the executor file its own verdict.
- Any turn in any session of the run advances it, which re-dispatches into parked sessions.
- There's no stop/resume. Runs are stranded after interrupts, crashes or redeploys, and cross-node delete blocks.
- Delete bypasses `SessionService`, so the bus state and workspaces leak.

Besides these, the comments across `_keys.py` / `_sop.py` / `_chat.py` are too long; please keep them to one or two lines. The storage tests still assert field by field instead of comparing the full structure.

src/agentscope/app/_service/_sop.py
```diff
@@ -0,0 +1,686 @@
...
+            if len(verdicts) > len(state.verifications):
+                state.verifications = verdicts
+                if verdicts[-1].passed:
+                    state.phase = SOPPhase.COMPLETED
+                else:
+                    state.submission = None
+                    state.phase = SOPPhase.PENDING
+                return
+
+            if state.submission is not None:
```
DavdGao: For a reviewer turn `state.submission` is always set, so we return here before the parked check and before `record(...)`. A reviewer that never calls `SubmitVerdict` (model error swallowed by `ChatService.run`, middleware giving up after 3 nudges) or parks on a tool approval gets re-dispatched forever without spending any attempt. Reproduced with the scripted chat: `[handover, None, None, ...]` dispatches the reviewer 50+ times with 0 verifications.

Suggest deciding by which half was dispatched (executor vs. verifier) rather than by `state.submission`, and letting the reviewer's no-submit/park go through the same failure/AWAITING branches.

src/agentscope/app/_service/_sop.py
```diff
@@ -0,0 +1,686 @@
...
+        # The engine hands over ``given`` for the former, which the
+        # briefing already carries.
+        resuming = inputs is not None and not isinstance(inputs, (Msg, list))
+        dispatch = f"{self._sop_run_id}:{self._index}"
+        try:
+            # Handed to the turn rather than left where the session can
+            # see it: a claim lying about would be picked up by whoever
+            # takes the session lock first, and that may be a person who
+            # happened to be typing into the same conversation.
+            await self._chat.run(
```
DavdGao: A turn dispatched here isn't registered in `ChatRunRegistry`, so the stop button in the session can't reach it. And even if it's interrupted, the no-submit branch below records a failure and the engine immediately starts the next attempt, which contradicts "an interrupted turn leaves the run where it is". Should check the turn's finished reason here as well.

src/agentscope/app/_service/_chat.py
```diff
@@ -1471,6 +1528,92 @@ async def _persist() -> None:
...
+        )
+        if claim is None:
+            return None
+        sop_run_id, _, step_index = claim.rpartition(":")
+        run = await self._storage.get_sop_run(user_id, sop_run_id)
+        index = int(step_index)
+        if (
+            run is None
+            or index >= len(run.state.steps)
+            or run.state.steps[index].phase
```
DavdGao: The claim is only validated by the step's phase, not by whether this session is the one the step is waiting on. And the resumed-turn path never drops the claim after it finishes. So after executor park -> approve -> handover, the executor session keeps a valid claim while the step waits for its reviewer; anyone typing into the executor session then gets `SubmitVerdict` (submission is set) plus the enforcing middleware, and the executor can pass its own work.

Suggest only reading the parked claim when the input is `UserConfirmResultEvent` / `ExternalExecutionResultEvent`, and checking the session matches the step's current role.

src/agentscope/app/_service/_chat.py
```diff
@@ -1471,6 +1528,92 @@ async def _persist() -> None:
                     trigger_text,
                 )
 
+        # An interrupted turn is a person saying stop. Carrying the
+        # procedure on from it would start the step again, which is the
+        # opposite of what they asked for — the run waits where it is
+        # until something else moves it.
+        if sop_run_id is not None and not interrupted:
```
DavdGao: This advances the run after any turn in any of the run's sessions, e.g. a person chatting in session B or a background-task wakeup, while the step is parked in session A. The advance then dispatches a fresh brief into A, which either raises `ValueError` in the agent (still awaiting tool calls) and shows up as an error in A, or queues behind the resumed turn in A and runs as a new attempt after the handover was already filed (getting `SubmitVerdict` again). Only the turn that carried the claim should trigger an advance.

src/agentscope/app/_service/_chat.py
```diff
@@ -1471,6 +1528,92 @@ async def _persist() -> None:
...
+        already driving: this reply was one it dispatched, and it reads
+        the result back itself the moment this returns.
+
+        Args:
+            user_id (`str`):
+                The owner user id.
+            sop_run_id (`str`):
+                The run this session belongs to.
+        """
+        if await self._message_bus.is_locked(
```
DavdGao: `is_locked` then `advance_later` is check-then-act: the run can be locked right after this check, and the queued advance will re-run the engine once it gets the lock. Combined with the issue above this means a duplicate dispatch. `try_lock` or doing the dedupe inside `run()` would be safer.

src/agentscope/app/_service/_sop.py
```diff
@@ -0,0 +1,686 @@
...
+            user_id (`str`):
+                The owner user id.
+            sop_run_id (`str`):
+                The run to delete.
+
+        Returns:
+            `bool`:
+                Whether there was one to delete.
+        """
+        going = list(self._advancing.get(sop_run_id, ()))
```
DavdGao: `_advancing` is per-process. In a multi-node deployment, deleting a run that's advancing on another node cancels nothing, and `acquire_lock` below blocks until that node finishes the rest of the procedure (could be hours), while `delete_sop` holds `sop_lock` the whole time and blocks `create_run` too. Needs a cross-node cancel, e.g. a bus broadcast like `session_publish_cancel`.

src/agentscope/app/_service/_sop.py
```diff
@@ -0,0 +1,686 @@
...
+        for task in going:
+            task.cancel()
+        await asyncio.gather(*going, return_exceptions=True)
+        async with self._message_bus.acquire_lock(
+            MessageBusKeys.sop_run_lock(sop_run_id),
+            ttl_secs=MessageBusKeys.SOP_RUN_TTL_SECS,
+        ):
+            return await self._storage.delete_sop_run(user_id, sop_run_id)
```
DavdGao: This goes straight to storage and bypasses `SessionService.delete_session`, so the run's sessions don't get cross-node run cancellation, bus purge (events/inbox, and the `sop_dispatch` claim, which has no TTL and stays in Redis forever), or workspace purge. The workspaces minted in `create_run` are never closed either, which leaks containers/sandboxes with docker/k8s/e2b managers. Suggest deleting the sessions through `SessionService` and closing the run's workspace(s).

src/agentscope/app/_service/_sop.py
```diff
@@ -0,0 +1,686 @@
...
+        return await self._storage.upsert_sop_run(user_id, run)
+
+    async def __aenter__(self) -> Self:
+        """Enter the service's lifetime."""
+        return self
+
+    async def __aexit__(self, *exc: object) -> None:
+        """Stop every advance still going, before its storage closes."""
+        going = [task for tasks in self._advancing.values() for task in tasks]
+        for task in going:
```
DavdGao: Cancelling here (and any node crash/redeploy) leaves the run persisted as RUNNING with nobody driving it. There's no startup recovery and no resume/retry endpoint, so it's stuck forever. Same for a run left AWAITING after an interrupt; it only moves again if someone happens to chat in one of its sessions. We need at least a way to resume a run (endpoint or recovery on startup), otherwise every rolling deploy strands in-flight runs.

src/agentscope/app/_service/_sop.py
```diff
@@ -0,0 +1,686 @@
...
+                else _generate_id()
+            )
+            session = await self._storage.upsert_session(
+                user_id=user_id,
+                agent_id=agent_id,
+                config=SessionConfig(
+                    workspace_id=workspace_id,
+                    name=f"{data.name} / {key}",
+                    chat_model_config=ChatModelConfig(
+                        **settings.chat_model_config,
```
DavdGao: `chat_model_config` is a plain dict and is only validated here, so a bad config turns `POST /runs` into a 500, and the sessions created in earlier iterations are orphaned. Agent existence isn't checked either. Suggest validating both when the SOP is created/updated.

src/agentscope/app/_service/_sop.py
```diff
@@ -0,0 +1,686 @@
...
+                Application storage.
+            workspace_manager (`WorkspaceManagerBase`):
+                Assigns the workspace a run's conversations share.
+            message_bus (`MessageBus`):
+                Serialises advances of one run against each other.
+            chat (`ChatService`):
+                Where each step's turn is taken.
+        """
+        self._storage = storage
+        self._workspace_manager = workspace_manager
```
DavdGao: `workspace_manager` is no longer used since workspace ids are minted in `create_run`. Remove the parameter.

src/agentscope/app/_service/_chat.py
```diff
@@ -249,6 +253,13 @@ def __init__(
                 pass
         self._extra_agent_tools = extra_agent_tools
         self._channel_clients = channel_clients
+        self.sop_service: SOPService | None = None
```
DavdGao: `ChatService` now knows about `SOPService` (lazy creation, back-filled by lifespan, `chat: Any` on the other side). This two-way coupling isn't great; a generic after-turn hook that the SOP service registers would keep `ChatService` SOP-agnostic.

src/agentscope/app/_router/_sop.py
```diff
@@ -0,0 +1,477 @@
...
+    Checked here rather than on the model because it is the one thing a
+    stored procedure can say that makes a run of it impossible, and the
+    editor should hear about it while it is still an edit.
+    """
+    named = {step.executor.session_key for step in data.steps}
+    for step in data.steps:
+        verifier = getattr(step.verifier, "agent", None)
+        if verifier is not None:
+            named.add(verifier.session_key)
+    missing = sorted(named - set(data.session_settings))
```
DavdGao: One `session_key` used with two different `agent_id`s isn't rejected. `create_run` keeps the last one (`agents[key] = agent_id`), so a step later dispatches its agent into a session owned by another agent. Should be a 422 here.

src/agentscope/app/middleware/_sop_step_middleware.py
```diff
@@ -0,0 +1,190 @@
...
+
+from ...agent import Agent
+from ...event import HintBlockEvent, ReplyEndEvent
+from ...message import HintBlock, ToolCallBlock, ToolResultBlock
+from ...message import ToolResultState
+from ...middleware import MiddlewareBase
+from ...types import ErrorInfo, ErrorType, ReplyFinishedReason
+
+
+class SOPStepSubmitMiddleware(MiddlewareBase):
```
DavdGao: This is basically `TeamMemberLoopMiddleware` with different tool names. Suggest generalizing into one "the reply must end with tool X" middleware instead of keeping two copies.

Also, the "a later tool call invalidates the submission" rule only exists here, since the submit tool has already written to storage, so the two can disagree.

src/agentscope/app/storage/_model/_sop.py
```diff
@@ -0,0 +1,197 @@
...
+        json_schema_extra={"format": "textarea"},
+    )
+
+
+class HumanVerifier(BaseModel):
+    """A verifier that asks a person and waits for the answer."""
+
+    type: Literal["human"] = "human"
+
+    question: str = Field(
```
DavdGao: `question` isn't used anywhere in the backend. Either use it (e.g. expose it on the AWAITING step) or drop it.

PR #2619, opened 2026-09-14
fix(app): use model card context size

### LINKED ISSUE #2596: [Bug]: Model card's context_size is ignored by the app service chat model factory

wangtaotaotao95:
Fixes #2596

## Summary
- resolve the matching built-in chat model card once in the app-service model factory
- pass its `context_size` to the chat model so context compression uses the card-declared window
- retain the existing formatter input-type override and add regression coverage

## Validation
- `pytest tests/service_model_test.py -q`
- `pre-commit run --files src/agentscope/app/_service/_model.py tests/service_model_test.py`


wangtaotaotao95:
Closing as a duplicate of #2597, which already implements the same model-card context-size fix. Apologies for the duplicate.

PR #2625, opened 2026-09-14
feat(app): run a stored procedure through ordinary chat sessions

DavdGao:
Stacked on #2617 — its three commits appear here too and should land first. This PR is the last two commits: the tools, the middleware, and `SOPService`.

## The shape

A step does not drive an agent. It hands the turn to `ChatService.run`, which already owns everything a turn needs — session lock, toolkit assembly, workspace, permissions, event fan-out, message persistence, failure reporting. Writing a second assembler for SOP would have duplicated all of that and then drifted from it.

What comes back is read out of the run state, not out of the reply.

## Why a tool and not structured output

`structured_output` never travels as an event (`agent/_agent.py:3564` puts it on the final `exit_msg` only), and `_run_impl` rebuilds its reply from events — so the stored message never carries it. Reading a handover that way means reaching into `agent.state.reply_context`.

Instead the step's agent carries one of `SubmitHandover` / `SubmitVerdict` and writes the run state itself, the way the team tools reach storage themselves. Consequences: no change to how `ChatService` streams or persists, the submission is visible in the transcript as a tool call, and the human case becomes the same shape as the agent case.

`SOPStepSubmitMiddleware` supplies the enforcement structured output would have given — swallowing a reply-end that submitted nothing and nudging, then failing the reply after three, exactly as `TeamMemberLoopMiddleware` holds a member to reporting. A failed attempt is then spent from the step's budget rather than hanging the run.

Which of the two tools an agent carries is derived once, in `get_toolkit`, from the rule the SDK's own step already uses: a step is being worked on until something is handed over, and judged afterwards. The middleware accepts whichever is there rather than deriving that a second time.

## A human verifier needs no runtime

It parks the run. Whoever is asked files the verdict through the same `SubmitVerdict` an agent reviewer calls, and the next pass picks up from it. Nothing is faked into a session, no reply or tool-call ids have to survive storage: the transcript shows interactions, the run state shows verdicts.

## Sessions

`create_run` opens one session per `session_key` — a key named by two steps is one conversation and one context, which is how a procedure says "the same one again". They share a workspace under the default `RUN` grain, assigned the way the team flow already assigns a worker the leader's (`_agent_create.py:447`), so nothing new was needed for it.

## Tests

`tests/app_sop_service_test.py` replaces the chat service with a script: each entry says what the session handed the turn does with it, written through the real submit tools. Covers the walk through a step, a refusal coming back with its critique, the attempt budget running out, a turn that submits nothing, a human verifier parking and then resuming from a filed verdict, and one session per key.

https://claude.ai/code/session_017j3JnCeguJKTXAj8FEqBdr


DavdGao:
Folded into #2617 — the whole SOP change lands as one PR so its size is visible in one place.

PR #2632, opened 2026-09-15
fix(web_ui): reconnect SSE stream and resync reply phase

lwylab:
## Summary
- The `examples/web_ui` chat hook treated a dropped `/sessions/{id}/stream` as a terminal error (`setError` only), so a missed `ReplyEndEvent` could leave `phase` stuck in `streaming`.
- Reconnect the SSE stream with exponential backoff; before each retry, re-fetch messages and correct `phase` from `is_running` / pending HITL.
- After repeated failures, toast and fall back to `idle`. This is an example resilience improvement, not a core SSE/runtime contract change (clients are expected to reconnect long-lived streams).

## Test plan
- [ ] Start `examples/web_ui`, open a session, trigger a long reply
- [ ] Kill/restart the backend (or break the Vite proxy) mid-stream and confirm a reconnect toast, then recovery
- [ ] Confirm that when the run has already finished (`is_running=false`), reconnect resets `phase` to `idle` instead of staying on Stop
- [ ] Confirm AbortController session switches still tear down cleanly without reconnect spam


lwylab:
Ready for review — CI notify/labels passed. Happy to adjust if anything looks off. cc @DavdGao

PR #2637, opened 2026-09-15
fix(workspace): resolve local search paths from the workspace directory

### LINKED ISSUE #2636: [Bug]: LocalWorkspace Glob/Grep ignore workdir for default and relative searches
## Prerequisites

- [x] I searched existing issues, pull requests, discussions, and the references to #2257 for a duplicate fix.
- [x] This report includes a reproducible bug.

## Background / Description

With `LocalWorkspace(workdir=project_dir)`, the shell runs in the project directory, but the workspace's `Glob` and `Grep` tools default to the launching Python process's directory. An agent can consequently miss files that exist in its workspace or search unrelated launcher files.

This is the pre-existing bug explicitly left for a separate change under **Known and deliberate** in #2257. It still reproduces on main at `43deddca7afdc0c65b099afb6297bb9a029abcb6`.

## Steps to Reproduce

Install `agentscope[tools]` from the commit above and run:

```python
import asyncio
import os
import tempfile
from pathlib import Path

async def main():
    initial_cwd = os.getcwd()
    with tempfile.TemporaryDirectory() as tmp:
        root = Path(tmp).resolve()
        launcher, project = root / "launcher", root / "project"
        launcher.mkdir()
        project.mkdir()
        (launcher / "launcher.py").write_text("LAUNCHER_ONLY = True\n")
        (project / "app.py").write_text("WORKSPACE_ONLY = True\n")
        os.chdir(launcher)
        try:
            async with LocalWorkspace(workdir=str(project)) as workspace:
                tools = {tool.name: tool for tool in await workspace.list_tools()}
                for name, args in [
                    ("Glob", {"pattern": "*.py"}),
                    ("Grep", {"pattern": "WORKSPACE_ONLY"}),
                    ("Glob", {"pattern": "*.py", "path": str(project)}),
                    ("Grep", {"pattern": "WORKSPACE_ONLY", "path": str(project)}),
                ]:
                    result = await tools[name](**args)
                    print(name, args, [block.text for block in result.content])
        finally:
            os.chdir(initial_cwd)


asyncio.run(main())
```

## Actual Behavior

With temporary paths abbreviated:

```text
Glob(pattern="*.py")                  -> $TMP/launcher/launcher.py
Grep(pattern="WORKSPACE_ONLY")         -> No matches found
Glob(..., path="$TMP/project")         -> $TMP/project/app.py
Grep(..., path="$TMP/project")         -> $TMP/project/app.py
```

There is no exception: both default searches complete in the wrong directory. Explicit relative paths such as `path="."` also use the launcher directory.

## Expected Behavior

Search tools returned by `LocalWorkspace.list_tools()` should resolve omitted and relative search paths against the workspace's `workdir`. Explicit absolute paths should keep their meaning.

## Environment

- AgentScope: 2.0.8, main at `43deddca7afdc0c65b099afb6297bb9a029abcb6`
- Python: 3.11.15
- OS: macOS 26.4 arm64
- ripgrep: 14.1.0

## Proposed Fix

I have prepared a focused fix and will submit the linked PR: add optional keyword-only `cwd` constructor arguments to `Glob` and `Grep`, and pass `LocalWorkspace.workdir` when constructing its tools. This preserves standalone-tool defaults and explicit absolute paths, and does not change the process-wide working directory or the UI-only `SessionConfig.cwd`.

The regression tests use real files and search tools across two workspaces. Against the main commit above, 14 of 24 subcases fail; the same tests all pass with the fix.

tomsen02:
## AgentScope Version

2.0.8; based on `43deddc`. The revision was also tested integrated with upstream main at `00d295b`.

## Description

Fixes #2636

When the launching Python process is outside `LocalWorkspace.workdir`, its shell runs in the workspace but `Glob` and `Grep` search the launcher directory. Files in the workspace can be missed, and unrelated files can be returned.

`LocalBackend` now accepts an optional `workdir`, which `getcwd()` returns. `LocalWorkspace` supplies its absolute workdir when constructing the backend. An unconfigured backend continues to read the process cwd at call time. Search tools resolve omitted and relative paths through the backend's existing path helpers; absolute paths still select the requested location. Permission suggestions and Bash path checks also receive the workspace directory through their existing `backend.getcwd()` calls.

The regression cases are in the existing backend and workspace test files. Companion documentation: https://github.com/agentscope-ai/docs/pull/69 (merged).

## Validation

Tested on macOS arm64 with Python 3.11.15.

- Both new workspace search tests fail against unmodified main and pass with the fix. They cover independent workspaces, unchanged process cwd, permission suggestions, omitted paths, `.` and `src`, and absolute path overrides. Backend tests cover an unset workdir and a configured relative workdir after the process cwd changes.
- Focused local backend/workspace, filesystem tool, shell, and permission checks on PR head `b1607f1`: **278 passed, 3 platform skips, 332 subtests passed**. The same selection integrated with current main: **282 passed, 3 platform skips, 337 subtests passed**.
- `pre-commit run --all-files` and `git diff --check`: passed.
- Broader integration check with main at `00d295b`, excluding `tests/blob_store_s3_test.py`: **2456 passed, 140 skipped, 10 failed, 442 subtests passed**. All 10 failures involve Docker persistence/offload/skills and reproduce with the same assertions on unmodified main at `00d295b`.
- The S3 test file was run separately on both versions; each completed one test before stalling and was stopped after 45 seconds. These integration checks are not claimed as passing. Windows/Linux and live model/cloud sandbox behavior were not validated locally.

## Checklist

- [x] Related documentation has been updated in a companion documentation PR


tomsen02:
Hi maintainers, a gentle follow-up on this fix for #2636. Could someone please approve the pending CI workflows and take a look when available? The PR includes regression tests for workspace-relative Glob/Grep searches, and the companion documentation is in agentscope-ai/docs#69. Happy to address feedback. Thanks!

src/agentscope/workspace/_local_workspace.py
```diff
@@ -156,7 +156,7 @@ async def list_tools(self) -> list[ToolBase]:
             shell,
             Edit(backend=backend),
             Glob(**glob_kwargs),
-            Grep(backend=backend),
+            Grep(backend=backend, cwd=self.workdir),
```
DavdGao: Among all backends only `LocalBackend.getcwd()` returns `os.getcwd()`, the sandbox ones return their workdir. Suggested to fix it there: let `LocalBackend` take an optional workdir that `getcwd()` returns, and build it with `self.workdir` in `LocalWorkspace`. Then Glob / Grep need no new `cwd` argument, and the permission suggestions and the Bash path check, which also rely on `backend.getcwd()`, become consistent as well.

src/agentscope/tool/_builtin/_grep.py
```diff
@@ -172,11 +174,29 @@ def __init__(
...
+                current-directory behavior. Relative ``cwd`` values are
+                resolved against the backend's current directory.
         """
         from ._backend import LocalBackend
 
         super().__init__(middlewares=middlewares)
         self._backend = backend or LocalBackend()
+        self._cwd = cwd
+
+    async def _resolve_path(self, path: str | None) -> str:
```
DavdGao: The same helper is duplicated in `_glob.py` and `_grep.py`, and the relative `cwd` branch is never used since `LocalWorkspace` always passes an absolute path.

src/agentscope/tool/_builtin/_grep.py
```diff
@@ -218,8 +238,9 @@ async def match_rule(
             return True
 
         path = tool_input.get("path", "")
-        if not path:
-            path = await self._backend.getcwd()
+        if path and fnmatch.fnmatch(path, rule_content):
```
DavdGao: The `match_rule` changes (here and in `_glob.py`) and the README paragraph are out of the scope of this PR.

tests/workspace_search_cwd_test.py
```diff
@@ -0,0 +1,143 @@
+# -*- coding: utf-8 -*-
```
DavdGao: Suggested to add the cases into the existing workspace test file instead of a new 143-line file with subTest matrices.

DavdGao (commented):
The problem is real, but I think it should be fixed in the backend rather than in the tools. Please see the inline comments.

tomsen02:
Thanks for the review. Updated in `b1607f1`:

- `LocalBackend(workdir=...)` owns the configured directory, and `LocalWorkspace` passes `self.workdir` when constructing it.
- Removed the added Glob/Grep constructor arguments, duplicated helper, `match_rule` changes, and console README paragraph.
- Replaced the separate test file with focused cases in the existing workspace/backend tests.

One small change remains in each search call: resolving the supplied path with `backend.abspath(..., cwd=await backend.getcwd())`. Changing `getcwd()` alone fixes omitted paths, but explicit `.` / `src` would still reach the local filesystem/subprocess relative to the launcher directory. The regression tests cover both forms.

Both workspace regressions fail on current main and pass with this revision. Focused checks on this head: 278 passed, 3 skipped, 332 subtests passed; all pre-commit hooks pass. The PR description includes the broader-suite results and baseline reproduction of the existing integration failures. I also tested the revision integrated with main at `00d295b`. Ready for another look.

PR #2641, opened 2026-09-15
fix(formatter): drop empty Gemini text parts

### LINKED ISSUE #2640: [Bug]: GeminiChatFormatter emits empty text parts rejected by Gemini
### Prerequisites

- [x] I searched existing issues and pull requests for this behavior.
- [x] This is a reproducible formatter bug rather than a usage question.

### Background / Description

`GeminiChatFormatter` unconditionally serializes every `TextBlock` as `{"text": block.text}`. An empty `TextBlock` therefore produces a Gemini content part with an empty string:

```python
import asyncio
from agentscope.formatter import GeminiChatFormatter
from agentscope.message import AssistantMsg, TextBlock

async def main():
    result = await GeminiChatFormatter().format([
        AssistantMsg(name="assistant", content=[TextBlock(text="")]),
    ])
    print(result)

asyncio.run(main())
```

Current output on `main` (`d0f4be4a`):

```python
[{"role": "model", "parts": [{"text": ""}]}]
```

Gemini rejects empty content parts with `400 INVALID_ARGUMENT` / `contents.parts must not be empty`. The formatter already handles the identical constraint for `ThinkingBlock` at `_gemini_formatter.py:160-170`, but the `TextBlock` branch at line 156 has no guard. Empty text blocks are realistic after streaming interruption, context processing, or programmatic message construction; `AnthropicChatFormatter` already has regression coverage that drops them.

### Expected behavior

Empty `TextBlock` values should be skipped. If a message contains no remaining parts, the formatter should omit that message rather than emit an empty Gemini content. Non-empty text and whitespace-bearing text should preserve existing behavior.

### Proposed fix

Guard the chat formatter `TextBlock` branch with `if block.text`, and add regression coverage for an empty-only message plus an empty block adjacent to valid text. Keep the change scoped to `GeminiChatFormatter`; system-message policy can be handled separately if maintainers want different semantics there.

### Environment

- AgentScope: current `main` at `d0f4be4a`
- Python: 3.11.16
- OS: macOS

Disclosure: I used Codex to assist the source audit and reproduction, and I will review and test any submitted patch.

wangtaotaotao95:
## Summary

- skip empty `TextBlock` values in `GeminiChatFormatter`
- omit a message when no valid Gemini parts remain
- cover empty-only and mixed empty/non-empty messages

## Validation

- `pytest -q tests/formatter_gemini_test.py` (12 passed)
- `pre-commit run --files src/agentscope/formatter/_gemini_formatter.py tests/formatter_gemini_test.py`
- `git diff --check`

Fixes #2640

Disclosure: Codex assisted with the source audit and implementation. I reviewed the diff and ran the checks above.


qbc2016 (approved):
LGTM.

PR #2643, opened 2026-09-15
fix(formatter): preserve first rendered history marker

### LINKED ISSUE #2642: [Bug]: Multi-agent formatters consume the first-history marker on an empty group
### Prerequisites

- [x] I searched existing issues and pull requests for this behavior.
- [x] The behavior is reproducible without a provider call.

### Background / Description

The multi-agent formatters track `is_first_agent_message` so the first rendered agent-history group receives `conversation_history_prompt` and `<history>` framing. Every implementation currently sets the flag to `False` after calling `_format_agent_message(...)`, even when that call returns an empty list because the group contains no blocks supported by that formatter.

A later non-empty group therefore loses some or all first-history framing. A tool sequence makes the two agent groups distinct:

```python
import asyncio
from agentscope.formatter import AnthropicMultiAgentFormatter
    AssistantMsg, UserMsg, ThinkingBlock, ToolCallBlock, ToolResultBlock,
    ToolResultState, TextBlock,
)

msgs = [
    AssistantMsg(name="A", content=[ThinkingBlock(thinking="hidden")]),
    AssistantMsg(name="A", content=[
        ToolCallBlock(id="c", name="f", input="{}"),
        ToolResultBlock(id="c", name="f", output="ok",
                        state=ToolResultState.SUCCESS),
    ]),
    UserMsg(name="B", content=[TextBlock(text="hello")]),
]

print(asyncio.run(AnthropicMultiAgentFormatter().format(msgs)))
```

Current final item on `main` (`d0f4be4a`):

```python
{"role": "user", "content": [{"type": "text", "text": "B: hello"}]}
```

The first group rendered nothing, but it consumed the marker. The actual first rendered history group lacks `conversation_history_prompt` and Anthropic’s `<history>...</history>` wrapper. OpenAI-style and Gemini multi-agent formatters retain the tags in this case but still lose `conversation_history_prompt`; the same unconditional state transition is duplicated across the formatter implementations.

This can happen when history contains provider-specific blocks that the selected formatter intentionally skips, including unsigned `ThinkingBlock` values during cross-provider replay.

### Expected behavior

`is_first_agent_message` should become `False` only after `_format_agent_message` returns at least one formatted item. Skipped/empty groups must not consume the first rendered history marker.

### Proposed direction

Store the result of `_format_agent_message`, extend `formatted_msgs`, and flip the marker only when the result is non-empty. Add a regression test with an empty/skipped group, an intervening tool sequence, and a later text group. A shared helper may be appropriate if maintainers want to remove the duplicated loop state, but the behavioral fix can remain focused.

### Environment

- AgentScope: current `main` at `d0f4be4a`
- Python: 3.11.16
- OS: macOS

Disclosure: I used Codex to assist the source audit and reproduction.

wangtaotaotao95:
## Summary

- keep `is_first_agent_message` true when an agent-message group formats to no provider messages
- apply the behavior consistently across all multi-agent formatters
- add Anthropic and Gemini regressions covering skipped leading groups

## Validation

- `uv run --python 3.11 --with '.[dev]' pytest -q tests/formatter_anthropic_test.py tests/formatter_dashscope_test.py tests/formatter_deepseek_test.py tests/formatter_gemini_test.py tests/formatter_moonshot_test.py tests/formatter_ollama_test.py tests/formatter_openai_chat_test.py tests/formatter_openai_response_test.py tests/formatter_volcengine_test.py tests/formatter_xai_test.py`\n- `uv run --python 3.11 --with '.[dev]' pre-commit run --files ...`\n\nFixes #2642\n\n## AI assistance\n\nThis change was developed with AI assistance. The implementation and tests were reviewed and executed locally before submission.


tests/formatter_anthropic_test.py
```diff
@@ -652,6 +652,39 @@ async def test_multiagent_formatter(self) -> None:
         res = await fmt.format([])
         self.assertListEqual([], res)
 
+    async def test_multiagent_empty_group_keeps_first_history_marker(
```
qbc2016, line 655: This test does not create two agent-message `groups. _group_messages` combines the `ThinkingBlock` message and this `UserMsg` into one group, so the assertion already passes on the unpatched formatter. Please insert a tool-call/result sequence between them and assert the complete output. The regression test should fail before this fix and pass afterward.

tests/formatter_gemini_test.py
```diff
@@ -529,6 +529,38 @@ async def test_multiagent_formatter(self) -> None:
         res = await fmt.format([])
         self.assertListEqual([], res)
 
+    async def test_multiagent_empty_group_keeps_first_history_marker(
```
qbc2016, line 532: This has the same grouping issue as the Anthropic test: the skipped block and later text are formatted together, so the old implementation also passes this assertion. Please separate them with a tool-call/result sequence to exercise the empty-group state transition.

qbc2016 (commented):

tests/formatter_anthropic_test.py
wangtaotaotao95: Thanks for catching this. Both regression tests now insert a tool-call/result sequence between the skipped leading message and the later user text, and assert the complete output including the tool messages.

I verified both tests fail with the pre-fix formatter implementations from d0f4be4a: Anthropic loses the history prompt and wrapper, while Gemini loses the history prompt. With the fix, all 29 tests in the Anthropic and Gemini formatter test files pass, and all pre-commit checks pass.

qbc2016 (approved):
LGTM.

PR #2656, opened 2026-09-16
fix(formatter): drop empty hint text

### LINKED ISSUE #2655: [Bug]: Empty HintBlock text bypasses Anthropic and Gemini formatter filtering
## Prerequisites

- [x] I searched existing issues and pull requests for empty formatter text and HintBlock handling.
- [x] This reproduces on current `main` after #2641.

## Background / Description

AnthropicChatFormatter and GeminiChatFormatter now drop empty `TextBlock` values, but `HintBlock` takes a separate formatting branch that still emits the same invalid empty text payload.

Both string hints and list hints are affected:

```python
HintBlock(hint="")
HintBlock(hint=[TextBlock(text="")])
```

Anthropic receives `{"type": "text", "text": ""}` and Gemini receives `{"text": ""}`. Both providers reject empty text parts. This also means wrapping an otherwise valid message with an empty hint can introduce a provider error even though an ordinary empty TextBlock is now safely ignored.

## Expected Behavior

Empty textual hint content should be skipped consistently with ordinary empty `TextBlock` content. A multimodal hint should retain its non-empty text and media parts.

## Proposed Fix

- avoid creating a standalone hint message for an empty string;
- filter empty `TextBlock` values from list hints;
- add regression tests for string and list forms in both Anthropic and Gemini formatters.

The public message and formatter APIs remain unchanged.

## AI Assistance

I am using Codex to assist with the implementation. I will review the diff and run the relevant formatter tests and repository checks before submitting the PR.

wangtaotaotao95:
## Summary

- ignore empty string hints in the Anthropic and Gemini chat formatters
- filter empty `TextBlock` entries from list-form hints
- keep adjacent assistant/model content in the same message when an empty hint is skipped

## Why

The formatters already avoid emitting empty regular text blocks, but `HintBlock` used a separate path. An empty hint could still emit an invalid empty text part or split otherwise adjacent assistant/model content into separate message groups.

## Tests

- `uv run --with pytest pytest tests/formatter_anthropic_test.py tests/formatter_gemini_test.py -q` (`31 passed`)

Fixes #2655

## AI assistance

AI assistance was used to inspect the formatter paths, implement the fix, and add regression tests. The resulting changes and test output were reviewed locally before submission.


qbc2016 (approved):
LGTM.

PR #2660, opened 2026-09-16
fix(rag): read the text inside grouped shapes in PPTParser

### LINKED ISSUE #2659: [Bug]: PPTParser drops the text inside grouped shapes
### Prerequisites

### Background / Description

`PPTParser` walks a slide's shapes and dispatches on three things
(`rag/_parser/_ppt.py`):

```python
for shape in slide.shapes:
    # 1. Pictures
    if self.include_image:
        image_bytes = _extract_image_bytes(shape)
        ...
    # 2. Tables
    if getattr(shape, "has_table", False):
        ...
    # 3. Text frames
    if getattr(shape, "has_text_frame", False):
        ...
```

A **group** is none of those. It carries no picture, no table and no text frame
of its own — its children do. So a grouped shape matches nothing, falls off the
end of the loop body, and every word inside it is dropped from the returned
Sections. No exception, nothing in the logs.

Grouping is ordinary in a deck: a label next to an arrow, a legend, a diagram
built out of boxes, a two-column callout. All of that is content a RAG index
built with `PPTParser` never sees.

### Error Messages

```shell
No exception is raised; the text is simply absent from the Sections.
```

### Steps to Reproduce

1. Code:

```python
import asyncio
import io

from pptx import Presentation
from pptx.util import Inches

from agentscope.rag._parser._ppt import PPTParser


def build() -> bytes:
    prs = Presentation()
    slide = prs.slides.add_slide(prs.slide_layouts[6])  # blank

    box = slide.shapes.add_textbox(Inches(0.5), Inches(0.2), Inches(4), Inches(0.5))
    box.text_frame.text = "PLAIN TEXTBOX"

    table = slide.shapes.add_table(2, 2, Inches(0.5), Inches(1.0), Inches(6), Inches(1.2)).table
    table.cell(0, 0).text = "HEADER A"
    table.cell(1, 1).text = "CELL B"

    a = slide.shapes.add_textbox(Inches(0.5), Inches(3.0), Inches(2), Inches(0.5))
    a.text_frame.text = "GROUPED ONE"
    b = slide.shapes.add_textbox(Inches(3.0), Inches(3.0), Inches(2), Inches(0.5))
    b.text_frame.text = "GROUPED TWO"
    slide.shapes.add_group_shape([a, b])

    buffer = io.BytesIO()
    prs.save(buffer)
    return buffer.getvalue()


async def main() -> None:
    sections = await PPTParser().parse(build(), "deck.pptx")
    print("\n".join(s.content.text for s in sections if hasattr(s.content, "text")))


asyncio.run(main())
```

2. Run: `python repro.py`

3. Observed:

```
<slide index=1>
PLAIN TEXTBOX
| HEADER A |  |
| --- | --- |
|  | CELL B |

</slide>
```

`GROUPED ONE` and `GROUPED TWO` are nowhere. The text box and the table came
through, so this is specific to the group.

Expected: the grouped text appears in reading order, after the standalone box.

### Environment

- AgentScope Version: 2.0.8 (`main` @ `0b157a31`)
- Python Version: 3.11
- OS: macOS 15 (python-pptx is platform independent, so this is not OS specific)

L4XB:
Fixes #2659.

## AgentScope Version

`2.0.8` (branched from `main` @ `0b157a31`)

## Description

`PPTParser`'s slide loop dispatches on three shape kinds — picture, table, text
frame. A **group** is none of them: it carries no picture, no table and no text
frame of its own, its children do. So a grouped shape matches nothing, falls off
the end of the loop body, and every word inside it is dropped from the Sections
with no error and nothing in the logs.

Grouping is ordinary in a deck — a label beside an arrow, a legend, a diagram
built from boxes, a two-column callout — so this is content a RAG index built
with `PPTParser` simply never sees.

Measured on one slide holding a text box, a 2x2 table and a group of two boxes:

| | on `main` | this branch |
|---|---|---|
| `PLAIN TEXTBOX` | yes | yes |
| `HEADER A` (table) | yes | yes |
| `CELL B` (table) | yes | yes |
| `GROUPED ONE` | **no** | yes |
| `GROUPED TWO` | **no** | yes |

The text box and the table come through on `main`, so this is specific to the
group.

**Change.** Walk the slide through a small `_iter_shapes`, which yields leaf
shapes in reading order and recurses into groups, since a group can hold a group:

```python
def _iter_shapes(shapes: Any) -> Iterator[Any]:
    for shape in shapes:
        # A group is the only shape with its own ``shapes`` collection.
        if hasattr(shape, "shapes"):
            yield from _iter_shapes(shape.shapes)
        else:
            yield shape
```

The loop body is untouched, so pictures, tables and text frames inside a group go
through exactly the same handling — including `include_image`, `separate_table`
and `table_format` — as they do at the top level. Duck-typing the group on its
`shapes` collection avoids importing `MSO_SHAPE_TYPE` into a module whose
python-pptx import is deliberately lazy.

A deck with no groups produces byte-identical Sections: `_iter_shapes` is the
identity for shapes that have no `shapes` collection.

## Tests

Two tests in `tests/rag_parser_test.py::PPTParserTest`, with two fixtures built
in memory beside the existing `_make_pptx_*` helpers:

- `test_group_shape_text_is_read` — grouped text is present **and in reading
  order** (the standalone box still comes first).
- `test_nested_group_shape_text_is_read` — a group inside a group, so the descent
  has to recurse rather than go one level deep.

```
$ pytest tests/rag_parser_test.py -k PPTParser -q
12 passed, 38 deselected

# same tests, src/ reverted to main
FAILED tests/rag_parser_test.py::PPTParserTest::test_group_shape_text_is_read
FAILED tests/rag_parser_test.py::PPTParserTest::test_nested_group_shape_text_is_read
2 failed, 10 passed
```

`tests/rag_parser_test.py` as a whole reports 23 failures both before and after
the change — the Word and Excel parsers' optional dependencies are not installed
in my environment; nothing in that set moves.

Style gates on the two files: `black --line-length=79` unchanged,
`flake8 --extend-ignore=E203` clean, `add-trailing-comma` no-op, `pylint` (repo
flags) `10.00/10`, `mypy` (repo flags) `Success: no issues found`.

## Checklist

- [x] An issue has been created for this PR (#2659)
- [x] Related documentation has been updated — none needed; this restores text
      the parser was already meant to return

File and line references are against `main` @ `0b157a31`.


tests/rag_parser_test.py
```diff
@@ -877,6 +956,30 @@ async def test_missing_path_raises_file_not_found(self) -> None:
...
+    async def test_group_shape_text_is_read(self) -> None:
+        """Text inside a group reaches the Sections, in reading order."""
+        parser = PPTParser(include_image=False)
+        sections = await parser.parse(_make_pptx_with_group(), "demo.pptx")
+
+        text = "\n".join(s.content.text for s in sections)
+        self.assertIn("Standalone", text)
+        self.assertIn("Grouped one", text)
+        self.assertIn("Grouped two", text)
+        self.assertLess(text.index("Standalone"), text.index("Grouped one"))
```
qbc2016: Please assert the complete `Section` output here, including the order of both grouped text boxes. An exact assertion would better protect the ordering and section behavior this change introduces.

tests/rag_parser_test.py
```diff
@@ -877,6 +956,30 @@ async def test_missing_path_raises_file_not_found(self) -> None:
...
+        parser = PPTParser(include_image=False)
+        sections = await parser.parse(
+            _make_pptx_with_nested_group(),
+            "demo.pptx",
+        )
+
+        text = "\n".join(s.content.text for s in sections)
+        self.assertIn("Innermost", text)
+        self.assertIn("Sibling", text)
+        self.assertIn("Outer", text)
```
qbc2016: Please assert the full output order (`Innermost`, `Sibling`, `Outer`) rather than only checking that each string exists. That would verify the intended depth-first traversal of nested groups.

src/agentscope/rag/_parser/_ppt.py
```diff
@@ -28,6 +29,28 @@
...
+    A group carries no picture, table or text frame of its own; its
+    children do. Iterating ``slide.shapes`` alone therefore never reaches
+    what is inside one, and the text is dropped without an error.
+
+    Args:
+        shapes (`Any`):
+            A python-pptx shape collection.
+
+    Yields:
+        `Any`: Each leaf shape, in reading order.
```
qbc2016: “Reading order” may be misleading here. `python-pptx` exposes slide shapes in z-order, and this generator traverses the shape tree depth-first; neither guarantees visual reading order. How about describing this as “depth-first shape-tree order” instead?

qbc2016 (commented):

tests/rag_parser_test.py
L4XB: Done in f76a1540. `test_group_shape_text_is_read` now pins the whole `model_dump()`:

```
"<slide index=1>\nStandalone\nGrouped one\nGrouped two\n</slide>"
```

so the order of the two grouped text boxes, the position of `Standalone` relative to them, the single section, and `metadata={"slide": 1}` are all part of the assertion. Against the pre-fix `_ppt.py` the text is `"<slide index=1>\nStandalone\n</slide>"` and the test fails on the whole list.

tests/rag_parser_test.py
L4XB: Done in f76a1540. `test_nested_group_shape_text_is_read` now pins the full text `"<slide index=1>\nInnermost\nSibling\nOuter\n</slide>"`, so the inner group being emptied before its sibling, and both before the shape that follows the outer group, are asserted rather than implied. Against the pre-fix `_ppt.py` the slide comes back empty (`"<slide index=1>\n</slide>"`).

src/agentscope/rag/_parser/_ppt.py
L4XB: Agreed, changed in f76a1540 to "in depth-first shape-tree order". python-pptx yields a shape collection in the order the shapes appear in the XML tree, which is z-order, and the recursion descends into a group before moving on to its sibling; neither says anything about where the shapes sit on the slide. The docstring now says what the generator actually guarantees.

qbc2016 (approved):
LGTM.

PR #2663, opened 2026-09-16
fix(web_ui): clear stuck running sessions in the sidebar and history

lwylab:
## Summary
- Soft-poll `useSessions` every 4s while any session has `is_running=true`, so abandoning a hung run (e.g. unreachable local model) and starting a new chat does not leave a permanent sidebar spinner after the run lock finally releases.
- Show a running indicator on session rows and add **Stop generating** to the session menu so a background run can be interrupted without opening that session.
- When loading history with `is_running=false`, close orphan assistant replies that still lack `finished_at` (crashed worker / dropped `REPLY_END`) for display only, so bubbles do not spin forever.

## Test plan
- [ ] Start a long / hung run, leave the session, create a new one — sidebar still shows running on the old session while the lock is held
- [ ] Stop the old session from its `⋯` menu (or wait until the run ends) — spinner clears within ~4s without a hard refresh
- [ ] Force an unfinished reply into history with no run lock (or kill the worker mid-reply) — reopening the session no longer leaves an endless bubble spinner
- [ ] Confirm idle sessions are unchanged (no polling traffic when nothing is running)


DavdGao:
Thanks for the PR. The web UI on main doesn't have a running indicator in the sidebar, so the stuck spinner described here doesn't exist. The indicator, the polling and the stop menu item are new features rather than a fix, and stamping a fake `finished_at` on the client only hides a backend persistence problem. Closing this one. If you think a sidebar running indicator is worth having, feel free to open an issue for it as a feature first.

PR #2669, opened 2026-09-17
fix(agent): check summary_template against summary_schema

### LINKED ISSUE #2668: [Bug]: ContextConfig does not check summary_template against summary_schema, so a trimmed schema crashes compression with KeyError
**AgentScope is an open-source project. To involve a broader community, we recommend asking your questions in English.**

**Describe the bug**

`ContextConfig.summary_template` and `ContextConfig.summary_schema` are independently configurable, but nothing checks that they agree. When a compression summary is generated, the template is rendered with the fields the schema declares:

```python
# src/agentscope/agent/_agent.py, inside _apply_change()
new_summary = cfg.summary_template.format(**res.content)
```

If the schema does not declare a field the template still references, this raises a bare `KeyError` — and it is raised while rendering the summary, i.e. in the middle of a reply, long after the configuration was built. Nothing in the traceback points at the configuration that caused it.

`_apply_change()` runs as a task and is awaited with `asyncio.shield`, with only `asyncio.CancelledError` handled, so the `KeyError` propagates and the agent's reply fails.

Related: #1388 reported this on 1.x. On 2.x the default `summary_schema` (from `SummarySchema`) declares every field the default template uses, and `jsonschema.validate` enforces `required`, so the default pair is consistent. The crash is still reachable on 2.x through the documented, independent `summary_schema` override — trimming the schema is exactly what you would do to stop paying for a field you do not want.

**To Reproduce**

```python
from agentscope.agent import ContextConfig

schema = ContextConfig().summary_schema
schema["properties"].pop("next_steps")  # drop one field the template renders

config = ContextConfig(summary_schema=schema)  # currently accepted
config.summary_template.format(**{k: "" for k in schema["properties"]})
# KeyError: 'next_steps'
```

The final line is what `_agent.py` does with the model's structured response. `jsonschema.validate` accepts the trimmed schema, so the summary is generated successfully and the failure happens only at render time.

**Expected behavior**

The mismatch should be reported when the configuration is built, naming the offending field(s), instead of surfacing as an opaque `KeyError` mid-reply. Whether that is a hard configuration error or a tolerated (logged) condition is a design call — but the current behavior, failing only once compression actually runs, is the worst of both.

Note this is deliberately **not** a proposal to silently render missing fields as empty strings; that was the approach in #2032, which was closed. Failing fast keeps the two fields honest.

**Error messages**

```text
KeyError: 'next_steps'
```

**Environment**

- AgentScope Version: 2.0.8 (`main` at `20228d5c`)
- Python Version: 3.13
- OS: macOS

wangtaotaotao95:
## AgentScope Version

2.0.8 (`main` at `20228d5c`)

Fixes #2668

## Description

`ContextConfig.summary_template` is rendered with the fields that `summary_schema` declares:

```python
# src/agentscope/agent/_agent.py, inside _apply_change()
new_summary = cfg.summary_template.format(**res.content)
```

The two fields are configured independently, but nothing checked that they agree. A schema that drops a field the template still references therefore raised a bare `KeyError` **while rendering the summary** — in the middle of a reply, long after configuration, with nothing in the traceback pointing at the config that caused it. `_apply_change()` runs as a task and is awaited with `asyncio.shield` with only `CancelledError` handled, so the `KeyError` propagates and the reply fails.

On the default configuration this is not reachable: `SummarySchema` declares every field the default template uses and `jsonschema.validate` enforces `required`. It is reachable through the documented `summary_schema` override — trimming the schema is exactly what you would do to stop paying for a field you do not want (see #1388 for the original report, and #2032, closed for silently rendering missing fields as empty strings).

This PR reports the mismatch when the configuration is built, naming the undeclared field(s), instead of tolerating it at render time:

```python
schema = ContextConfig().summary_schema
schema["properties"].pop("next_steps")
ContextConfig(summary_schema=schema)
# pydantic ValidationError: summary_template references ['next_steps'],
# which summary_schema does not declare, so formatting the summary would
# fail with a KeyError. ...
```

Design notes:

- Fail fast rather than substitute empty strings, per the reasoning that closed #2032 — the two fields should stay consistent, not silently diverge.
- Schemas without top-level `properties` (`$ref`, `allOf`, ...) cannot be introspected reliably and are left to the runtime, so no false positives.
- Extra declared fields are fine, and a template may reference a subset of them.

## Checklist

- [x] An issue has been created for this PR (#2668)

### Tests

Added `tests/context_config_test.py` (7 cases): the default pair is consistent, a trimmed schema is rejected with every missing field named, a matching custom template/schema pair is accepted, extra schema fields are accepted, a placeholder-free template is accepted, and an uninspectable schema is left alone.

- `uv run pytest tests/context_config_test.py -q` → `7 passed`
- The two rejection cases fail on the unpatched code (`2 failed, 5 passed`), confirming they exercise the fix.
- `uv run pytest tests/compress_context_test.py tests/agent_injection_test.py tests/session_config_patch_test.py -q` → `54 passed, 6 subtests passed`


src/agentscope/agent/_config.py
```diff
@@ -191,6 +198,42 @@ class ContextConfig(BaseModel):
...
+            return self
+
+        placeholders = {
+            field_name
+            for _, field_name, _, _ in string.Formatter().parse(
+                self.summary_template,
+            )
+            if field_name is not None
+        }
+        undeclared = sorted(placeholders - set(properties))
```
AmirF194: I ran this validator against `b60e931c` in a clean container.

`placeholders` collects the raw `field_name` that `string.Formatter().parse()` returns, which for a subscript or index access includes the brackets (`"stats[count]"`), not just the base name (`"stats"`). Comparing that against `set(properties)` on line 226 means a template that reaches into a declared object-typed field, like `{stats[count]}` where `summary_schema` declares `stats`, gets rejected even though `.format(**content)` resolves it fine at render time.

```
from agentscope.agent._config import ContextConfig

cfg = ContextConfig(
    summary_template="<system-info>{stats[count]}</system-info>",
    summary_schema={
        "type": "object",
        "properties": {"stats": {"type": "object", "properties": {"count": {"type": "integer"}}}},
        "required": ["stats"],
    },
)
```

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for ContextConfig
  Value error, summary_template references ['stats[count]'], which summary_schema does not declare, ...
```

but `"<system-info>{stats[count]}</system-info>".format(stats={"count": 5})` returns `"<system-info>5</system-info>"` without error, so the config this rejects would have worked.

Splitting off the base name before the diff, e.g. `field_name.split(".")[0].split("[")[0]`, before checking membership would fix this without weakening the check for the plain-field case the tests cover.

wangtaotaotao95:
Good catch — you're right, and I reproduced it before changing anything. That was a false positive in the check I added, not a limitation I had considered.

`Formatter().parse()` returns the whole accessor as `field_name`, so `{stats[count]}` came through as `"stats[count]"` and was compared against `set(properties)` verbatim. `.format(**content)` resolves the subscript at render time against whatever `stats` holds, so the template was valid and my validator rejected it anyway. Fixed in `1cb15c63` by stripping the accessor before the comparison:

```python
placeholders.add(field_name.split("[", 1)[0].split(".", 1)[0])
```

Your exact repro now constructs and renders:

```python
cfg = ContextConfig(
    summary_template="<system-info>{stats[count]}</system-info>",
    summary_schema={... "properties": {"stats": {...}} ...},
)
cfg.summary_template.format(stats={"count": 5})   # '<system-info>5</system-info>'
```

Coverage added, both directions:

- `test_reaching_into_a_declared_field_is_accepted` — covers `{stats[count]}` **and** `{stats.total}`, and asserts on the rendered result rather than only on construction, since rendering is what makes them valid.
- `test_undeclared_base_field_is_still_rejected` — `{missing[count]}` is still refused, so stripping the accessor did not turn the check into a no-op.

One boundary worth stating explicitly, since your example walks up to it: the check validates **base names only**. A template that reaches into a declared field with a key the content will not have (`{stats[missing]}` where `stats` is declared) still resolves at render time and can raise there. That is inherent to validating a template against a schema — the nested shape is the model's output, not the schema — so I have not tried to cover it; say the word if you would rather the check go deeper.

`9 passed`; `black --line-length=79`, `flake8`, `pylint` (10.00/10) and `add-trailing-comma` all clean on both changed files.

src/agentscope/agent/_config.py
AmirF194: Confirmed, thanks for the fast turnaround. The base-name strip plus the recursive check in the second commit both look right to me.

wangtaotaotao95:
@AmirF194, thanks for confirming the follow-up fix. The branch is green and still awaiting a formal review decision; an approval would be appreciated when convenient.

PR #2671, opened 2026-09-17
fix(agent): release reply_stream queue getter on cancel

### LINKED ISSUE #2665: [Bug]: Cancelling RealtimeAgent.reply_stream leaves a reader that consumes future events
### Prerequisites

- [x] I searched existing issues, pull requests, and discussions and did not find this bug reported.

### Background / Description

Cancelling a task waiting for the next `RealtimeAgent.reply_stream()` event cancels the transport uplink but leaves its `self._out.get()` task alive. The model session intentionally outlives the transport, so that orphan reader can consume an event intended for the next stream.

In the reproduction, the orphan consumes `ReplyStartEvent`; the next stream starts with `ModelCallStartEvent` instead. This can leave event consumers without the reply's start event when a client disconnects and reconnects.

Expected: cancelling/closing a stream releases its reader and uplink; a new stream receives undelivered events in order, without an old reader consuming them.

### Error Messages

No exception is surfaced after the cancellation is handled. Actual output:

```text
first event after resuming: ModelCallStartEvent
```

Expected: `first event after resuming: ReplyStartEvent`.

### Steps to Reproduce

From a repository checkout, save the following as `reproduce_stream_cancel.py` and run `PYTHONPATH=src:tests python reproduce_stream_cancel.py`. It reuses the existing offline test model/transport; no provider credentials, audio hardware, or network access are required. Private state is inspected only to synchronize the reproduction with the waiting reader and provider response.

```python
import asyncio
import contextlib

from agentscope.agent import RealtimeAgent
from agentscope.realtime import _events as me
from realtime_agent_test import FakeTransport, ScriptedModel


class IdleTransport(FakeTransport):
    def __init__(self):
        super().__init__(frames=0)
        self.ready = asyncio.Event()
        self.stop = asyncio.Event()

    async def incoming(self):
        self.ready.set()
        await self.stop.wait()
        if False:
            yield None


async def main():
    model = ScriptedModel([
        ["WAIT", me.ResponseCreatedEvent(item_id="reply-after-cancel")]
    ])
    async with RealtimeAgent("assistant", "Be brief.", model) as agent:
        transport = IdleTransport()
        stream = agent.reply_stream(transport)
        pending = asyncio.create_task(anext(stream))
        await transport.ready.wait()
        # Synchronize with the queue reader, not a wall-clock delay.
        for _ in range(20):
            if agent._out._getters:
                break
            await asyncio.sleep(0)
        assert agent._out._getters
        pending.cancel()
        with contextlib.suppress(asyncio.CancelledError):
            await pending
        await stream.aclose()

        await model.request_response()
        for _ in range(20):
            if agent._reply is not None and not agent._out._getters:
                break
            await asyncio.sleep(0)
        # Let the notified reader run before attaching the next transport.
        await asyncio.sleep(0)
        resumed = agent.reply_stream(IdleTransport())
        try:
            event = await asyncio.wait_for(anext(resumed), timeout=1)
            print("first event after resuming:", type(event).__name__)
        finally:
            await resumed.aclose()


asyncio.run(main())
```

### Environment

- AgentScope: 2.0.8, reproduced on `main` at `3ae8a1f4eca2dec6a519753e0c1f0e76ef83731f`
- Python: 3.12.13
- OS: macOS

### Root Cause / Proposed Fix

[`reply_stream`](https://github.com/agentscope-ai/agentscope/blob/3ae8a1f4eca2dec6a519753e0c1f0e76ef83731f/src/agentscope/agent/_realtime/_agent.py#L394-L423) races a newly created queue getter against the uplink using `asyncio.wait`. Cancelling the wait does not cancel its child tasks. The outer `finally` only cleans up the uplink.

I would like to contribute a focused fix that cancels and awaits the outstanding getter on exit, with deterministic regression tests for cancellation, resumption, and normal generator closure. If cancellation races with a completed dequeue, the unyielded event should remain available before later queued events; just cancelling pending getters does not cover that race. No public API or provider-session lifetime changes are intended.

kabishou11:
## AgentScope Version

2.0.8 (based on `main` @ `20228d5`)

## Description

Fixes #2665.

`RealtimeAgent.reply_stream` races an `asyncio.Queue` getter against the uplink with `asyncio.wait`. Cancelling or closing the stream cleaned up the uplink but left `self._out.get()` running. That orphan reader could consume the next stream's `ReplyStartEvent`, so a reconnect started at `ModelCallStartEvent` instead.

This PR:

- Tracks the outstanding getter across the loop.
- Cancels and awaits it on the uplink-wins path and in `finally`.
- If the getter already dequeued an event that was never yielded, puts that event back at the **front** of `_out` so later queued events stay behind it.
- Leaves uplink cleanup and public API / provider-session lifetime unchanged.

### Validation

- `PYTHONPATH=src:tests python -m pytest -q tests/realtime_agent_test.py` → **20 passed**
- New regressions cover cancel-then-resume, `aclose` releasing the getter, and completed-but-unyielded event ordering.
- Issue reproduction now prints `first event after resuming: ReplyStartEvent`.

## Checklist

- [ ] Related documentation has been updated — not applicable; behavior restore only, no API change


DavdGao:
The bug is real, and there are actually two: the `finally` only handles `uplink` and leaves the pending `getter` alive, and `getter.cancel()` in the `else` branch can drop an event the getter had already dequeued.

But draining the whole queue and pushing everything back is a lot of machinery for "put one event back at the head".

- Suggested to create the getter once before the loop and only replace it after it yields, so a getter holding a value is never cancelled mid-loop.
- For the teardown case, backing `_out` with a `deque` plus an `Event` instead of `asyncio.Queue` makes putting the event back a single `appendleft()`, and `_put_out_front` goes away.

kabishou11:
Thanks @DavdGao for the review — agreed on both points.

I redesigned along your suggestion:

- Create the `_out` getter **once** before the loop and only replace it **after** a successful yield, so a getter holding a dequeued event is never cancelled mid-loop.
- Back `_out` with a `collections.deque` plus an `asyncio.Event` instead of `asyncio.Queue`, so teardown can put a completed-but-unyielded event back with a single `appendleft()` (the full drain/re-enqueue helper is gone).

Updated in `3041ba5d0be4f12e69cae9d3a217b7980ac5983e` (rebased onto latest `main`). Related realtime cancel/requeue tests are green.

kabishou11:
Heads-up: I accidentally opened a duplicate follow-up as #2817 while iterating on review feedback. That PR is closed. Current head here (`3041ba5d`) remains the intended redesign from the earlier reply: getter-once + deque/Event + appendleft on teardown, with the realtime agent regression tests. Happy to adjust further if anything is still off.

PR #2686, opened 2026-09-19
fix(tool): read a quoted or expanded redirect target, not just a bare word

### LINKED ISSUE #2684: [Bug]: quoting or expanding a redirection target hides it from the dangerous-path check, so a write into a protected file escapes the bypass-immune ASK
## Summary

`_extract_paths_from_node` reads a redirection target only when the target's node type is `word`. A quoted or expanded target parses as `string`, `raw_string` or `concatenation`, so it is skipped and the path never reaches `_is_dangerous_path()`. The step 4 danger-path ASK is therefore skipped for a write into a protected file, and unlike its bare spelling that command is not bypass-immune, so a user allow rule can clear it silently.

## Mechanism

`src/agentscope/tool/_builtin/_bash_parser.py`, in the `file_redirect` branch:

```python
if node.type == "file_redirect":
    # Extract the target file
    for child in node.children:
        if child.type == "word":
            path = command[child.start_byte : child.end_byte]
            paths.append(("redirect", path.strip("'\"")))
```

A `file_redirect` node has exactly two children: an anonymous operator (`>`, `>>`, `<`) and the target. What the target parses as depends only on how it is written:

| written as | target node type | extracted today |
|---|---|---|
| `> ~/.zshrc` | `word` | yes |
| `> /home/u/.zshrc` | `word` | yes |
| `> "~/.zshrc"` | `string` | **no** |
| `> '$HOME/.zshrc'` | `raw_string` | **no** |
| `> $HOME/.zshrc` | `concatenation` | **no** |
| `> "$HOME/.zshrc"` | `string` | **no** |

The `.strip("'\"")` on the next line removes quotes from a node that carries none, which is what the intent looks like: the branch was written for quoted targets and only its node-type test excludes them.

## Reproduction

```python
import asyncio

from agentscope.permission import (
    PermissionBehavior, PermissionContext, PermissionEngine,
    PermissionMode, PermissionRule,
)
from agentscope.tool import Bash

CASES = [
    "echo x > ~/.zshrc",
    'echo x > "~/.zshrc"',
    "echo x > '$HOME/.zshrc'",
    "echo x > $HOME/.zshrc",
    'echo x > "~/.ssh/authorized_keys"',
]


async def main() -> None:
    engine = PermissionEngine(PermissionContext(mode=PermissionMode.DEFAULT))
    engine.add_rule(PermissionRule(
        tool_name="Bash", rule_content="echo:*",
        behavior=PermissionBehavior.ALLOW, source="user-config",
    ))
    for cmd in CASES:
        d = await engine.check_permission(Bash(), {"command": cmd})
        print(f"{cmd!r:<38} {d.behavior.value:<10} bypass_immune={d.bypass_immune}")


asyncio.run(main())
```

On current `main` (`5ff52f8`):

| command | behavior | bypass_immune |
|---|---|---|
| `echo x > ~/.zshrc` | ask | **True** |
| `echo x > "~/.zshrc"` | **allow** | False |
| `echo x > '$HOME/.zshrc'` | **allow** | False |
| `echo x > $HOME/.zshrc` | **allow** | False |
| `echo x > "~/.ssh/authorized_keys"` | **allow** | False |

The first row is the intended behaviour: the message for step 5 says such a command "cannot be auto-allowed by permission rules". Rows two to five are the same write, executed identically by bash, cleared by a rule the user wrote for printing things.

With no rule configured both spellings ask, so this only bites a user who has an allow rule covering the command. That is the normal case for `echo`, `printf` and `cat`, which are also in `SAFE_COMMANDS`.

`~/.ssh/authorized_keys` is the one worth naming in the title of the fix: a write to it is remote code execution on the host.

## Why this is not a duplicate

I searched issues and pull requests for `file_redirect`, `extract_file_paths`, quoted redirection, and the `$HOME` spelling before writing this. What exists:

- #2470 and #2471 address separator handling in `is_read_only_command()`, a different function.
- #2500 addresses the command-name comparison in `_check_dangerous_removal_path()`, also different.
- #2605 and #2629 address `tee` being classified read-only.

The parser gap itself is **known and pinned by a test**, `tests/permission_bash_parser_test.py::test_redirections_with_quoted_paths`, whose cases are `"output file.txt"`, `backup.txt` and `"file with spaces.log"`, each commented "Quoted redirections not extracted". That frames it as a filename-with-spaces limitation of `extract_redirections`. Its security consequence for `extract_file_paths` is not recorded anywhere, and no issue or PR mentions it.

## Proposed fix

Take the target whatever node type it is. The operator is an anonymous child, so skipping anonymous children is enough:

```python
if node.type == "file_redirect":
    for child in node.children:
        if not child.is_named:
            continue
        path = command[child.start_byte : child.end_byte]
        paths.append(("redirect", path.strip("'\"")))
        break
```

`_is_dangerous_path()` then does the rest: it matches on the file's basename and on path segments, so `$HOME/.zshrc` and `~/.zshrc` both match `.zshrc` and `"~/.ssh/authorized_keys"` matches the `.ssh` segment. No expansion of the variable is needed for that match to fire, which is why the fix can stay in extraction.

Two decisions worth a maintainer's word:

1. `$HOME` and `~` stay unexpanded in the extracted string. Matching is by basename and segment, so the check works, but if you would rather the reported path be resolved, that belongs in `_is_dangerous_path()` rather than here.
2. The sibling `_extract_redirections_from_node` has the same node-type test and is covered by the test above. Fixing it too would change that test's expectations from `[]` to the target. I have kept the change to the security path and left that decision to you, since `extract_redirections` is a public parser API and this issue is about the permission check.

I have the fix and regression tests ready locally and will open a PR referencing this issue. Happy to wait if you would rather the shape go another way first.

sloemo01:
Fixes #2684

## What was wrong

`_extract_paths_from_node` read a redirection target only when the target's node type was `word`:

```python
if node.type == "file_redirect":
    for child in node.children:
        if child.type == "word":
            path = command[child.start_byte : child.end_byte]
            paths.append(("redirect", path.strip("'\"")))
```

The target is a `word` only when it is written bare. What it parses as depends entirely on spelling:

| written as | target node type | extracted before |
|---|---|---|
| `> ~/.zshrc` | `word` | yes |
| `> "~/.zshrc"` | `string` | **no** |
| `> '$HOME/.zshrc'` | `raw_string` | **no** |
| `> $HOME/.zshrc` | `concatenation` | **no** |
| `> "$HOME/.zshrc"` | `string` | **no** |

A skipped path never reaches `_is_dangerous_path()`, so step 4's ASK is skipped with it, and the command stops being bypass-immune.

## Why that matters

With a user allow rule on `echo:*` in `DEFAULT` mode, before this change:

| command | behavior | bypass_immune |
|---|---|---|
| `echo x > ~/.zshrc` | ask | **True** |
| `echo x > "~/.zshrc"` | **allow** | False |
| `echo x > '$HOME/.zshrc'` | **allow** | False |
| `echo x > $HOME/.zshrc` | **allow** | False |
| `echo x > "~/.ssh/authorized_keys"` | **allow** | False |

Row one is the documented intent, from the ASK's own message: such a command "cannot be auto-allowed by permission rules". Rows two to five are the same write, executed identically by bash, cleared by a rule the user wrote for printing things. With no rule configured both spellings ask, so this only bites a user whose allow rules cover the command, which is the normal case for `echo`, `printf` and `cat`.

## The change

A `file_redirect` node has exactly two children: an anonymous operator (`>`, `>>`, `<`) and the target. Skipping anonymous children takes the target whichever node type it is, and `break` keeps it to one path per redirection:

```python
for child in node.children:
    if not child.is_named:
        continue
    path = command[child.start_byte : child.end_byte]
    paths.append(("redirect", path.strip("'\"")))
    break
```

The `.strip("'\"")` that follows now removes quotes it can actually see, which is what the original line looks like it was written for.

After the change, every row of that table returns `ask` with `bypass_immune=True`.

## Tests

`tests/permission_bash_parser_test.py`, new `test_quoted_and_expanded_redirection_targets`: seven spellings, six of which failed before the change. `"file with spaces.log"` is included, since a quoted target with spaces is the case the old behaviour was originally noticed for.

`tests/builtin_bash_test.py`, new `test_quoted_redirect_target_is_bypass_immune`: the permission outcome itself, asserting ASK and `bypass_immune` for all five spellings. This is the half that would catch a regression in the check rather than in the extraction.

All four suites touching this area pass:

```
tests/permission_bash_parser_test.py   82 tests   OK
tests/builtin_bash_test.py             30 tests   OK
tests/permission_engine_test.py        32 tests   OK
tests/permission_mode_test.py          52 tests   OK
```

## Two things left to you

The sibling `_extract_redirections_from_node` has the same node-type test, and `extract_redirections` is covered by `test_redirections_with_quoted_paths`, whose cases are commented "Quoted redirections not extracted" and whose paths are `backup.txt` and `"file with spaces.log"`. I left it alone: fixing it would flip that test's expectations, and it is a public parser API whose current framing reads as deliberate. If you want the two consistent, say so and I will do it in this PR.

`$HOME` and `~` stay unexpanded in the extracted string. The check matches on basename and path segments, so `$HOME/.zshrc` and `~/.zshrc` both match `.zshrc` and `"~/.ssh/authorized_keys"` matches the `.ssh` segment. If you would rather the reported path be resolved, that belongs in `_is_dangerous_path()` rather than in extraction.


sloemo01:
Ready for review when someone has a moment.

The `pr-guard` note above is addressed: the body now closes #2684, so the change and the problem are linked. That issue carries the mechanism, a runnable repro and the before and after.

It is a security one, which is why I would rather not sit on it. Under a user allow rule on `echo:*` in `DEFAULT` mode, `echo x > ~/.zshrc` is a bypass-immune ASK that the rule cannot touch, while `echo x > "~/.zshrc"` came back a plain ALLOW, before this change. Same write, same bash. `~/.ssh/authorized_keys` is reachable the same way.

Two things I would like a word on, both written up in the issue:

1. `_extract_redirections_from_node` carries the same node-type test, and `extract_redirections` is covered by `test_redirections_with_quoted_paths`, which pins the current behaviour with the comment "Quoted redirections not extracted". I left it alone to keep this change on the security path. Say the word and I will fold it in here rather than leaving the two functions disagreeing.
2. `$HOME` and `~` stay unexpanded in the extracted string, which is enough for the basename and segment match in `_is_dangerous_path()`. Resolving them would belong there rather than in extraction, but it is your call.

Tests: `permission_bash_parser_test` 82, `builtin_bash_test` 30, `permission_engine_test` 32, `permission_mode_test` 52, all passing. The new parser test fails six of its seven cases on `main`, which is the bug it pins.

@DavdGao mind taking a look when you get a chance?

PR #2690, opened 2026-09-19
fix(embedding): report cache source for fully cached multi-batch merges

### LINKED ISSUE #2689: [embedding] merged EmbeddingResponse reports source="api" for a fully cached multi-batch call
### Prerequisites

- [x] I have searched the existing issues and discussions, and this is not a duplicate. (`EmbeddingResponse source cache`, `embedding cache source api`, `batch merge embedding cache`, `embedding_cache` all return no match; no open PR touches `_merge_responses`.)

### Background / Description

When an embedding call is served from `embedding_cache` **and** the inputs split into more than one batch, the merged `EmbeddingResponse` reports `source="api"` even though no provider request was made. With a single batch the same call correctly reports `source="cache"`.

`EmbeddingResponse.source` is documented as "If the response comes from the cache or the API." (`src/agentscope/embedding/_embedding_response.py:31-32`), so this is a public field returning the wrong value. What makes it show up in practice is that the value flips exactly at the batch boundary: the field is preserved for 1 batch and overwritten for ≥2.

`EmbeddingModelBase._merge_responses` (`src/agentscope/embedding/_embedding_base.py:266-301`) hard-codes the field while merging:

```python
        if len(responses) == 1:
            return responses[0]          # preserves resp.source, incl. "cache"
        ...
        return EmbeddingResponse(
            embeddings=all_embeddings,
            usage=EmbeddingUsage(tokens=total_tokens, time=total_time),
            source="api",               # <- unconditional
        )
```

Every provider that can hit the cache emits `source="cache"` for a cached batch — `_openai/_model.py:148`, `_dashscope/_model.py:366` and `:451`, `_gemini/_model.py:350`, `_ollama/_model.py:116` — and each takes `embedding_cache` in its constructor, so the two sides of the contract disagree whenever more than one batch is involved. Document ingestion is that case: `rag/_knowledge.py:337` embeds `[chunk.content for chunk in chunks]` in one call, so any document larger than `batch_size` goes through the merge path.

Secondary effect of the same line: a fully cached multi-batch call reports `source="api"` together with `usage.tokens == 0`, which is a combination no provider ever returns for a live call. A caller that attributes spend by `source` will book a cache-only operation as API traffic.

### Error Messages

No exception. Observed values from the repro below (a stub `_call_api` that returns `source="cache"` for the batches named in `cached_batches`); nothing is logged either, so the mismatch is silent:

```
1 batch, cache hit                           batches=1 source='cache'  tokens=0   time=0.0
1 batch, live API                            batches=1 source='api'    tokens=14  time=0.5
2 batches, both cache hits                   batches=2 source='api'    tokens=0   time=0.0   <-- no API call, says api
3 batches, all cache hits                    batches=3 source='api'    tokens=0   time=0.0   <-- same
2 batches, one cache hit                     batches=2 source='api'    tokens=7   time=0.5
4 batches, one cache hit                     batches=4 source='api'    tokens=21  time=1.5
2 batches, both live API                     batches=2 source='api'    tokens=14  time=1.0
```

Expected: rows 3 and 4 report `source='cache'`, i.e. the same value row 1 reports for the identical cache-only call.

### Steps to Reproduce

No API key, no network, no browser. Save as `repro.py` outside the repo, `pip install agentscope` (or run inside a checkout), then:

```python
import asyncio
from agentscope.embedding import EmbeddingModelBase, EmbeddingResponse, EmbeddingUsage


class StubEmbedding(EmbeddingModelBase):
    """Returns cache hits for the batches listed in `cached_batches`."""

    def __init__(self, batch_size, cached_batches):
        super().__init__(
            credential=None, model="stub", dimensions=4, parameters=None,
            context_size=64, batch_size=batch_size, max_retries=0, retry_delay=0.0,
        )
        self.cached_batches, self.calls = cached_batches, []

    async def _call_api(self, inputs, **kwargs):
        index = len(self.calls)
        self.calls.append(list(inputs))
        cached = index in self.cached_batches
        return EmbeddingResponse(
            embeddings=[[float(index)] * 4 for _ in inputs],
            usage=EmbeddingUsage(tokens=0 if cached else 7 * len(inputs),
                                 time=0.0 if cached else 0.5),
            source="cache" if cached else "api",
        )


async def probe(label, batch_size, n_inputs, cached):
    model = StubEmbedding(batch_size, cached)
    resp = await model([f"text-{i}" for i in range(n_inputs)])
    print(f"{label:<40} source={resp.source!r:<8} tokens={resp.usage.tokens}")


async def main():
    await probe("1 batch, cache hit", 4, 2, {0})
    await probe("2 batches, both cache hits", 1, 2, {0, 1})
    await probe("3 batches, all cache hits", 1, 3, {0, 1, 2})


asyncio.run(main())
```

Printed:

```
1 batch, cache hit                       source='cache' tokens=0
2 batches, both cache hits               source='api'   tokens=0
3 batches, all cache hits                source='api'   tokens=0
```

Swap `{0, 1}` for `set()` and the same script shows the live path is unaffected. This is not covered by tests today: `grep -rn "embedding_cache" tests/` only matches the two places that set `model.embedding_cache = None` (`tests/embedding_gemini_test.py:93`, `:135`), and `tests/embedding_openai_test.py::test_multi_batch` exercises the merge path with live responses only — where `source="api"` happens to be right.

### Proposed direction (not opened as a PR yet)

The merge already walks every response for `usage`; collecting `source` in the same pass is the smallest change that restores the contract — e.g. `sources = {r.source for r in responses}`, then `"cache" if sources == {"cache"} else "api"`. A regression test belongs on the base class (`_merge_responses`), not on one provider, since all four cache-capable providers share this merge.

Alternatives considered:
1. **Add a third literal** (`"mixed"`) for a partly cached merge. More informative, but it widens `Literal["cache", "api"]` and any consumer doing `== "api"` starts missing traffic; the bug report is only about the cache-only case.
2. **Keep `"api"` and instead make the single-batch path normalise to `"api"`.** Consistent, and it would drop the only signal that a provider call was skipped — the opposite of what the field is documented to mean.
3. **Move the decision into the providers.** They already implement `_call_api` for one batch only, by design ("Subclasses only implement `_call_api` for a single batch — no batching or retry code needed", `_embedding_base.py:48-49`), so they cannot see the merge.

### Environment

- AgentScope version: `2.0.8` (source checkout at `5ff52f8`)
- Python 3.12.11, Windows 11 x64
- Reproduced with the plain `agentscope.embedding` import path; no provider credentials involved.

### Note on process

Per CONTRIBUTING (§3, step 1) I am filing this as an issue with a proposed direction and design alternatives, and waiting for core-team feedback before opening a PR. This report, the repro and the measurements were produced by an AI agent; a human has not reviewed them line by line yet, so please treat the claims above as things to verify rather than as reviewed work — the repro script is self-contained and takes seconds to run.

ZxlDragonDoctor:
### Problem
`EmbeddingModelBase._merge_responses` hard-codes `source="api"` when combining ?2 batch responses. A fully cache-hit multi-batch embedding call therefore reports `source="api"` with `usage.tokens == 0`, contradicting the documented `EmbeddingResponse.source` field ("cache or the API") and booking cache-only work as API traffic.

Single-batch cache hits already report `source="cache"`; the contract flips exactly at the batch boundary.

### Solution
While merging usage, also collect each batch `source`. Merged response uses `source="cache"` only when **every** batch was from cache; otherwise `"api"` (covers mixed cache/API and all-live). Matches the reporter's smallest proposed fix and keeps `Literal["cache","api"]` unchanged.

### Testing
- Command: `PYTHONPATH=src py -3.13 -m pytest tests/embedding_merge_source_test.py -q --noconftest`
- Result: `4 passed in 0.48s`
- Cases: single-batch cache preserved; all-cache multi-batch ? cache + tokens 0; all-api ? api; mixed ? api

### Agent dimension
evaluation / context-management (embedding cache accounting for agent RAG ingestion)

### Core value
Fully cache-hit multi-batch embeddings no longer claim API source, so usage attribution matches reality.

### Issue alignment
Addresses #2689. Reporter proposed the `sources` set + `cache iff all cache` rule; no maintainer design note on the issue yet. Single-batch path left unchanged.

Fixes #2689


ZxlDragonDoctor:
Friendly ping — this PR has been open for a few days with no review yet. Happy to rebase on main or adjust the cache-source merge rule if anything is unclear. Thanks!

ZxlDragonDoctor:
Friendly ping - this PR has been open since 2026-09-19 with no maintainer review yet. Happy to rebase on main or adjust the cache-source merge rule if anything is unclear. Thanks!

PR #2696, opened 2026-09-20
fix(model): close xAI client on setup failures

### LINKED ISSUE #2695: fix(model): close xAI AsyncClient when request setup fails
### Description

`XAIChatModel._call_api()` creates a new `xai_sdk.AsyncClient` for every model invocation, but the client is only closed after `chat.sample()` in non-streaming mode or by `_parse_stream_response()` after the streaming generator is consumed.

If an exception is raised before either path takes ownership of the client, it is never closed. Affected setup steps include:

- `formatter.format(messages)`
- tool / tool-choice formatting
- `client.chat.create(...)`
- `chat.append(...)`

Repeated invalid requests or setup failures can therefore accumulate unclosed gRPC clients/channels.

### Relevant code

`src/agentscope/model/_xai/_model.py`, in `XAIChatModel._call_api()`.

### Suggested fix

Wrap the complete setup and non-streaming execution path in `try/finally`. Track whether ownership has been handed to `_parse_stream_response()`:

```python
client = AsyncClient(...)
handed_off = False
try:
    # format messages, create chat, append messages
    if self.stream:
        result = self._parse_stream_response(start_datetime, chat, client)
        handed_off = True
        return result
    response = await chat.sample()
    return self._parse_completion_response(start_datetime, response)
finally:
    if not handed_off:
        await client.close()
```

The streaming parser already closes the client in its own `finally`, so this preserves single ownership and avoids double-closing.

### Tests

Please cover that the client is closed when:

1. message formatting raises;
2. chat creation or append raises;
3. non-streaming sampling or response parsing raises;
4. a streaming call does not close the client before generator consumption, and closes it exactly once when the generator finishes or is cancelled.

iluv7:
## AgentScope Version

2.0.8

## Description

Closes #2695

`XAIChatModel` creates a per-request asynchronous gRPC client. If request setup or non-stream response parsing raises an exception, the client is not closed, which can leak channels and related resources across failed calls.

This change makes client ownership explicit: `_call_api` closes the client for setup and non-streaming paths, while the streaming parser retains ownership until stream consumption finishes.

### Client lifecycle

- Wraps xAI request setup and non-stream response parsing in `try/finally`.
- Closes the client when message formatting, chat creation, message appending, sampling, or completion parsing fails.
- Transfers cleanup responsibility to the streaming parser only after the parser is created successfully.
- Preserves the existing streaming behavior so the client remains open while events are consumed and closes exactly once afterward.

### Tests

Added regression coverage for:

- Formatter failures.
- Chat creation failures.
- Message append failures.
- Sampling failures.
- Non-stream completion parser failures.
- Streaming ownership before, during, and after consumption.

## Testing

- `pytest tests/model_xai_test.py` — 20 passed.
- `pytest tests/formatter_xai_test.py` — 16 passed.
- `black --check --line-length=79 --target-version=py313 src/agentscope/model/_xai/_model.py tests/model_xai_test.py` — passed.
- `flake8 src/agentscope/model/_xai/_model.py tests/model_xai_test.py` — passed.
- `git diff --check` — passed.

## Checklist

- [x] Has been tested locally.
- [x] I have read the CONTRIBUTING.md.
- [x] Related documentation has been updated in the documentation repository
  (not applicable: no user-facing API or behavior change).


iluv7:
Closing this PR based on the clarification in #2695. The reported leak does not occur in practice, and the lifecycle restructuring is unnecessary. Thanks for the review.

PR #2703, opened 2026-09-20
fix(embedding): avoid blocking the event loop

### LINKED ISSUE #2702: [Bug]: Gemini and DashScope embedding calls block the event loop
### Prerequisites

### Background / Description

The text and multimodal paths in `GeminiEmbeddingModel` and `DashScopeEmbeddingModel` call synchronous SDK methods directly from `async def` methods. While an embedding request waits for its response, other coroutines on the same loop cannot run. This also serializes the batches that `EmbeddingModelBase` dispatches with `asyncio.gather`.

Gemini calls `self.client.models.embed_content`; DashScope calls `TextEmbedding.call` and `MultiModalEmbedding.call`. These waits should yield the event loop, using Gemini's async client and offloading DashScope's synchronous calls.

### Error Messages

The example below prints:

    loop progressed during SDK wait: False

Expected: `True`; the scheduled callback should run while the SDK is waiting.

### Steps to Reproduce

Save this as `repro.py` and run `python repro.py`:

```python
import asyncio
import threading
from types import SimpleNamespace

from agentscope.credential import DashScopeCredential
from agentscope.embedding import DashScopeEmbeddingModel

async def main():
    model = DashScopeEmbeddingModel(
        credential=DashScopeCredential(api_key="test"),
        model="text-embedding-v4", dimensions=1024,
    )

    def sdk_call(**kwargs):
        progressed = threading.Event()
        loop.call_soon_threadsafe(progressed.set)
        print("loop progressed during SDK wait:", progressed.wait(1))
        return SimpleNamespace(
            status_code=200,
            output={"embeddings": [{"embedding": [1.0]}]},
            usage={"total_tokens": 1},
        )

    with patch("dashscope.embeddings.TextEmbedding.call", sdk_call):
        await model(["hello"])

asyncio.run(main())
```

### Environment

- AgentScope: 2.0.8 (`5ff52f8`)
- Python: 3.11.7
- OS: Windows

Tsubaki-01:
## AgentScope Version

2.0.8

## Description

Fixes #2702.

Gemini and DashScope embedding requests currently block the event loop and serialize batches dispatched through `asyncio.gather`. Use Gemini's `client.aio.models.embed_content` and run DashScope's existing SDK calls through `asyncio.to_thread` for both text and multimodal inputs. Keeping the DashScope entrypoints also preserves compatibility with older supported SDK versions.

The provider tests check that DashScope runs its SDK call off the event loop thread and Gemini awaits the async SDK. All 19 related embedding tests and the changed-file pre-commit checks pass.

## Checklist

- [x] An issue has been created for this PR.
- [x] I have read [CONTRIBUTING.md](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md).
- [ ] Related documentation has been updated — N/A: the public API and usage are unchanged.


tests/embedding_async_test.py
```diff
@@ -0,0 +1,193 @@
...
+"""Exercise embedding SDK calls without contacting provider services."""
+import asyncio
+import threading
+from collections.abc import Iterator
+from dataclasses import asdict
+from types import SimpleNamespace
+from typing import Any
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
```
DavdGao: This test file is too heavy for an 8-line fix: pytest fixtures with a parametrized provider, conditional expected values (e.g. `14 if provider.name == "dashscope" else 0`), a threading harness, and it re-tests retry / status errors that this PR doesn't touch.

Suggested to drop this file and add one small case per provider into `embedding_dashscope_test.py` and `embedding_gemini_test.py`, in the `IsolatedAsyncioTestCase` style with literal expected values.

DavdGao (commented):
The src change looks good, only the tests need to be trimmed.

tests/embedding_async_test.py
Tsubaki-01: Thanks for your review, I'll trim the redundant tests soon.

DavdGao (approved):

PR #1998, opened 2026-07-03
feat(service): support to share resource (credential, agent, knowledge bases) across different users as a group or org

DavdGao:
## PR Title Format

Please ensure your PR title follows the Conventional Commits format:
- Format: `<type>(<scope>): <description>`
- Example: `feat(memory): add redis cache support`
- Allowed types: `feat`, `fix`, `docs`, `ci`, `refactor`, `test`, `chore`, `perf`, `style`, `build`, `revert`
- Description should start with a lowercase letter

## AgentScope Version

[The version of AgentScope you are working on, e.g. `import agentscope; print(agentscope.__version__)`]

## Description

[Please describe the background, purpose, changes made, and how to test this PR]

## Checklist

Please check the following items before code is ready to be reviewed.

- [ ]  An issue has been created for this PR
- [ ]  I have read the [CONTRIBUTING.md](https://github.com/agentscope-ai/agentscope/blob/main/CONTRIBUTING.md)
- [ ]  Docstrings are in Google style
- [ ]  Related documentation has been updated (e.g. links, examples, etc.) in [documentation repository](https://github.com/agentscope-ai/docs)
- [ ]  Code is ready for review

PR #2706, opened 2026-09-20
fix(app): resolve shared team leaders through the resource access policy

### LINKED ISSUE #2705: [Bug]: Team led by a shared (cross-owner) agent loses its leader on every leader read
### AgentScope version

main @ 5ff52f87 (same behavior in v2.0.8)

### Describe the bug

Since the resource-sharing support (#1998), session creation accepts agents shared through the resource access policy — `ResourceAccessService.resolve_agent` resolves own agents first, then user-source agents shared with the caller. `TeamCreate` then legitimately records such an agent as a team leader: it only requires a session in the team owner's namespace, with no owner equality check.

`_resolve_team_leader` (`src/agentscope/app/storage/_utils.py`) however resolves the leader's agent with an owner-scoped `storage.get_agent(user_id, agent_id)` — the docstring states the assumption outright: *"The team owner (also the leader agent's owner)"*. For a shared leader the agent record lives in another user's namespace, so creation succeeds while every later leader read fails:

| path | symptom |
|---|---|
| `TeamSay` | `TeamSay: leader records missing for team <id>` |
| `AgentCreate` | falls back to the raw leader agent id in the worker's system prompt — and since `TeamSay` routes by name, worker reports can't reach the leader |
| `AgentInvite` | same id-instead-of-name degradation in the borrowed session's first team-message hint |
| `ChatService._run_impl` | HTTP 409 `Team <id> has no resolvable leader; worker session cannot run` |
| `_build_team_detail` (session list) | `leader_agent: null` |

The roster side of the same feature is already policy-aware — #2326 resolves invited members through `access.try_resolve_agent(user_id, member.agent_id)` and re-checks revoked grants at call time. Only the leader stayed owner-scoped, so a team can be created but not driven.

### To reproduce

1. `publisher` owns an agent `Shared Leader` (`source="user"`).
2. The resource access policy grants `alice` access to that agent (e.g. a `ResourceAccessPolicyBase` whose `list_accessible` returns the `ResourceRef` for `publisher` / the shared agent).
3. `alice` creates a session on `Shared Leader` — allowed, `resolve_agent` accepts shared user-source agents.
4. In that session, `TeamCreate` succeeds and stores `leader_agent_id = <Shared Leader>.id`, `user_id = alice`.
5. `TeamSay(content="...")` from the same session returns `TeamSay: leader records missing for team <id>`.

Expected: the leader resolves for as long as alice's grant is live — team creation and leader resolution should agree. Revoking the grant (or the owner deleting the agent) should degrade to the existing missing-leader error, exactly like a revoked invited member is hidden from team detail.

### Proposed fix

Mirror #2326 for the leader: give `_resolve_team_leader` an optional `ResourceAccessService` and resolve the leader's agent through `try_resolve_agent(user_id, agent_id)` — owner-scoped read first, then the caller's live grants, `None` otherwise. Wire it the same way `AgentInvite` got its service: `TeamSay` / `AgentCreate` accept an optional `resource_access_service`, and `ChatService` / `_build_team_detail` pass the service they already hold.

Happy to open the PR — the fix with tests is already worked out locally.

iamsee:
## PR Title Format

fix(app): resolve shared team leaders through the resource access policy

## AgentScope Version

2.0.8 (branch cut from main @ 5ff52f87)

## Description

Fixes #2705

### Problem

Session creation accepts cross-owner shared agents through `ResourceAccessService.resolve_agent` (#1998), and `TeamCreate` legitimately records such an agent as a team leader. `_resolve_team_leader`, however, resolved the leader's agent with an owner-scoped `storage.get_agent(user_id, agent_id)`, so the team could be created but every leader read failed:

- `TeamSay` → `TeamSay: leader records missing`
- `AgentCreate` / `AgentInvite` → the leader's raw agent id instead of its display name (and `TeamSay` routes by name, so worker reports could not reach the leader either)
- `ChatService._run_impl` → HTTP 409, worker sessions could not run
- `_build_team_detail` (session list) → `leader_agent: null`

### Changes

- `storage/_utils.py::_resolve_team_leader` accepts an optional `access: ResourceAccessService | None`. When supplied, the leader's agent record resolves through `try_resolve_agent(user_id, agent_id)` — owner-scoped read first, then the team owner's live access grants. `None` keeps the previous owner-scoped lookup for direct integrations.
- `TeamSay` and `AgentCreate` accept the same optional `resource_access_service` constructor parameter `AgentInvite` already has; `get_toolkit` passes it for both worker and leader roles.
- `ChatService._run_impl` and `_build_team_detail` pass the access service they already hold.

No schema change and no public API break: all new parameters are optional and default to the previous behavior. Caller isolation is unchanged — session and team lookups stay in the team owner's namespace, and the leader resolves against *that* user's live grants, so a revoked or non-granted leader degrades to the existing missing-leader error instead of bypassing access. Team workers (`source="team"`) remain non-shareable: `try_resolve_agent` skips them for cross-owner refs, and a leader is always a user-source agent.

Deliberately out of scope: dissolving a team when the owner deletes the shared leader agent from the other namespace (cross-owner cascade, related to #2288), and any change to `TeamMember` semantics.

### How to test

New `TestSharedLeaderCrossOwner` in `tests/service_team_tools_test.py` builds the exact scenario on the existing fakeredis fixture, following the cross-owner invite test pattern from #2326: `publisher` owns `Shared Leader`, the policy grants `alice`, `alice` opens a session on it and creates a team.

- shared leader resolves with a live grant; without an access service it stays hidden (back-compat for owner-only integrations)
- a viewer outside the grant gets no leader (caller isolation)
- revoking the grant restores the missing-leader error
- legacy `leader_agent_id=None` records resolve via the leader session, policy-aware
- end-to-end: the shared leader spawns a worker (the worker's system prompt carries the real leader name) and `TeamSay` delivers to it; revocation degrades with the exact existing error text
- `_build_team_detail` includes the shared leader as a read-only view; `AgentInvite` names the real leader in the borrowed session's first team-message hint

`tests/service_chat_run_context_test.py` adds a worker run whose leader agent is owned by another user — it previously died with HTTP 409 during assembly and now completes with the correct leader context.

`tests/service_team_failure_report_test.py`'s `_Access` fake now mirrors `try_resolve_agent`, matching the real service contract.

Local verification on Python 3.12: `pytest tests` → **2302 passed, 306 skipped**; `pre-commit run --files <changed files>` → all hooks green. No storage-layer behavior changed (resolution is backend-agnostic and exercised on the fakeredis backend per the existing team-tools test convention), so the SQL/Redis suites are unchanged.

## Checklist

- [ ]  An issue has been created for this PR (#2705)
- [ ]  Related documentation has been updated — N/A: bug fix aligning leader resolution with the existing sharing behavior; no new public API


iamsee:
Status note for maintainers: the branch is CI-ready on my side. `pytest tests` passes locally (2302 passed, 306 skipped on Python 3.12) and `pre-commit run --files <changed files>` is green, so the pending workflow runs (Python Unittest Coverage / Pre-commit / PR Title Check) only await approval. Could a maintainer approve the runs when convenient? Happy to address review feedback or rebase onto the latest main on request.

PR #2711, opened 2026-09-21
fix(message_bus): make try_lock and acquire_lock mutually exclusive in InMemoryMessageBus

### LINKED ISSUE #2716: InMemoryMessageBus: try_lock and acquire_lock do not exclude each other on the same key
## Problem

`MessageBus` groups `acquire_lock` / `is_locked` / `try_lock` / `unlock` as Mode E — one distributed lock keyed by the same string — and `RedisMessageBus` implements it that way (both entry points `SET key ... NX` on the same key).

`InMemoryMessageBus` keeps two unrelated pieces of state: `try_lock` only writes `_lock_holders[key]`, `acquire_lock` only takes `self._locks[key]` (an `asyncio.Lock`) and never looks at `_lock_holders`. So a coroutine can enter `acquire_lock`'s critical section while another still holds a live `try_lock` claim on the same key.

## Reproduction

```python
bus = InMemoryMessageBus()
async def holder():
    await bus.try_lock("k", ttl_secs=600); log.append("A try_lock")
    await asyncio.sleep(0.15); log.append("A unlock"); await bus.unlock("k")
async def challenger():
    await asyncio.sleep(0.02)
    async with bus.acquire_lock("k"):
        log.append("B ENTERED")
await asyncio.gather(holder(), challenger())
# ['A try_lock', 'B ENTERED', 'A unlock']   <- B is inside while A holds the claim
```

Once inside, `acquire_lock` also overwrites `_lock_holders[key] = inf`, and A's later `unlock()` pops it — so `is_locked()` reports `False` while B is still in its critical section.

## Expected

Same behaviour as the Redis bus: `acquire_lock` waits behind a live `try_lock` claim, and `try_lock` is refused while `acquire_lock` holds the key.

Fix in #2711.

*Found with AI assistance (Claude); reproduction run by me.*

Linxiushen:
Fixes #2716

## What

`InMemoryMessageBus.acquire_lock()` and `InMemoryMessageBus.try_lock()` did not exclude each other on the same key. A coroutine could enter `acquire_lock`'s critical section while another still held a live `try_lock` claim on that key.

## Why it's a bug

`MessageBus` groups `acquire_lock` / `is_locked` / `try_lock` / `unlock` as **Mode E — one distributed lock keyed by the same string** (`is_locked` docs: "same key passed to acquire_lock"; `try_lock` docs: "Unlike acquire_lock (which waits until free), this returns immediately"). `RedisMessageBus` implements it that way: both entry points do `SET key ... NX` on the **same** key, so they are mutually exclusive by construction.

The in-memory bus kept two unrelated pieces of state:

- `try_lock` only wrote `self._lock_holders[key] = now + ttl` and never touched `self._locks[key]`
- `acquire_lock` only took `self._locks[key]` (an `asyncio.Lock`) and never looked at `self._lock_holders`

So `acquire_lock` sailed straight past a live `try_lock` claim. Worse, once inside it overwrote `_lock_holders[key] = inf`, and the `try_lock` holder's later `unlock()` popped that entry — so `is_locked()` reported `False` while `acquire_lock` was still in its critical section.

## Reproduction (before)

```python
bus = InMemoryMessageBus()
async def holder():
    await bus.try_lock("k", ttl_secs=600); log.append("A try_lock")
    await asyncio.sleep(0.15); log.append("A unlock"); await bus.unlock("k")
async def challenger():
    await asyncio.sleep(0.02)
    async with bus.acquire_lock("k"):
        log.append("B ENTERED")
await asyncio.gather(holder(), challenger())
```

```
['A try_lock', 'B ENTERED', 'A unlock']    # B is inside while A holds the claim
```

## Fix

- `acquire_lock` takes the `asyncio.Lock`, then checks for a live `try_lock` claim on the key; if one exists it releases and backs off (same `0.1s` poll and same `_LOCK_RETRY_DELAY_SECS` name as the Redis bus) until the claim is released or its lease expires.
- `try_lock` also refuses while `acquire_lock` holds the key (`self._locks[key].locked()`), so the exclusion holds in both directions.

After the fix the same repro gives `['A try_lock', 'A unlock', 'B ENTERED']`, the reverse direction (`acquire_lock` first → `try_lock` returns `False`) holds, and an uncontested `try_lock` still succeeds.

## Compatibility

- Behaviour when only one of the two entry points is used on a key is unchanged.
- `acquire_lock` still ignores `ttl_secs` for its own hold, as documented.
- Existing 41 tests in `tests/in_memory_message_bus_test.py` pass unchanged; two tests added (one per direction). The first fails on the old code.

`+25 −7` in the bus, `+28` in tests.

---

I used an AI assistant (Claude) while investigating and drafting this; I read the diff line by line and ran the tests and reproductions above myself, and I'm happy to explain any part of it.


DavdGao:
Thanks for the PR. `try_lock` / `unlock` have no callers in the codebase, and nothing mixes the two entry points on the same key, so this inconsistency can't be hit. We shouldn't add code, and a sleep-poll loop in particular, to defend against a bug that doesn't exist. Closing this one, and #2716 with it.

PR #2712, opened 2026-09-21
fix(_utils): keep nested arguments that schema-guided repair drops

### LINKED ISSUE #2718: _json_loads_with_repair silently drops nested arguments under additionalProperties: false
## Problem

`_json_loads_with_repair` guards against `json_repair` deleting arguments under `additionalProperties: false` — the guard's own comment says dropping an argument "is a rewrite rather than a type repair", and `test_keep_unrepairable_arguments` pins that. But the guard compares only **top-level** keys:

```python
if isinstance(parsed, dict) and parsed.keys() - res.keys():
    res = parsed
```

`json_repair` applies `additionalProperties` at every depth, so a key dropped from a **nested** object still vanishes silently, and the tool runs with an argument the model sent removed.

## Reproduction

```python
schema = {"type": "object", "properties": {
    "city": {"type": "string"},
    "opts": {"type": "object", "properties": {"unit": {"type": "string"}},
             "additionalProperties": False}}}

_json_loads_with_repair('{"city":"Beijing","extra":"keepme"}', schema)
# {'city': 'Beijing', 'extra': 'keepme'}                 <- top level: kept

_json_loads_with_repair('{"city":"Beijing","opts":{"unit":"C","nested_extra":"keepme"}}', schema)
# {'city': 'Beijing', 'opts': {'unit': 'C'}}             <- nested: dropped, no error
```

## Expected

The same rule at every depth: where a key (or list element) was dropped, keep the original container and let the caller's schema validation report it; keep the type repairs everywhere else.

Fix in #2712.

*Found with AI assistance (Claude); reproduction run by me.*

Linxiushen:
Fixes #2718

## What

`_json_loads_with_repair` guards against `json_repair` deleting arguments under `additionalProperties: false`, but the guard only compared **top-level** keys. A key dropped from a **nested** object still vanished silently, and the tool ran with an argument the model never sent removed.

## Why it's a bug (by the project's own rule)

The existing guard says it in its comment:

```python
if isinstance(parsed, dict) and parsed.keys() - res.keys():
    # Dropping arguments, e.g. under `additionalProperties:
    # false`, is a rewrite rather than a type repair.
    res = parsed
```

and `test_keep_unrepairable_arguments` pins that intent: dropping is a rewrite, and the caller's schema validation should report it with a better error. `json_repair` applies `additionalProperties` at every depth, so the same rewrite happens one level down — the check just wasn't recursing.

## Reproduction (before)

```python
schema = {"type": "object", "properties": {
    "city": {"type": "string"},
    "opts": {"type": "object", "properties": {"unit": {"type": "string"}},
             "additionalProperties": False}}}

_json_loads_with_repair('{"city":"Beijing","extra":"keepme"}', schema)
# {'city': 'Beijing', 'extra': 'keepme'}                       <- top level: kept (guard works)

_json_loads_with_repair('{"city":"Beijing","opts":{"unit":"C","nested_extra":"keepme"}}', schema)
# {'city': 'Beijing', 'opts': {'unit': 'C'}}                   <- nested: silently dropped
```

## Fix

Replace the flat comparison with `_revert_dropped_entries(original, repaired)`, which walks dicts and lists:

- a dict that lost keys → restore the original dict (same as the existing top-level rule)
- a dict that lost nothing → keep the repaired dict (including any keys repair added), recursing into shared keys
- a list that changed length → restore the original list
- a list of the same length → recurse element-wise
- anything else → keep the repaired value (that's a type repair, the thing we want)

Top-level behaviour is unchanged: when top-level keys are dropped you get `parsed` back exactly as before; otherwise you get the repaired result exactly as before.

After the fix:

```
nested extra field      -> kept:      {'city': 'Beijing', 'opts': {'unit': 'C', 'nested_extra': 'keepme'}}
nested type repair      -> still done: {'city': 'Beijing', 'opts': {'unit': '123'}}
top-level extra field   -> unchanged:  {'city': 'Beijing', 'extra': 'x'}
top-level type repair   -> unchanged:  {'city': '42'}
list element type repair -> still done: {'city': 'B', 'tags': ['1', '2']}
```

## Tests

Extends `test_keep_unrepairable_arguments` with a nested-drop case and a nested-type-repair case. The nested-drop assertion fails on the old code and passes with the fix; the existing 6 tests / 8 subtests pass unchanged.

`+35 −3` in `_common.py`, `+34` in tests. No new dependency.

---

I used an AI assistant (Claude) while investigating and drafting this; I read the diff line by line and ran the tests and reproductions above myself, and I'm happy to explain any part of it.


src/agentscope/_utils/_common.py
```diff
@@ -92,6 +92,36 @@ def _normalize_local_path(path: str) -> str:
     return os.path.abspath(os.path.expanduser(path))
 
 
+def _revert_dropped_entries(original: Any, repaired: Any) -> Any:
```
DavdGao: The partial revert isn't needed. Once a dropped key is restored, the caller's schema validation fails anyway, so the type repairs kept elsewhere are never used. Suggested to keep the existing `res = parsed`, and only make the check recursive, i.e. a small function that tells whether any nested dict lost a key.

src/agentscope/_utils/_common.py
```diff
@@ -92,6 +92,36 @@ def _normalize_local_path(path: str) -> str:
...
+        return {
+            key: (
+                _revert_dropped_entries(original[key], value)
+                if key in original
+                else value
+            )
+            for key, value in repaired.items()
+        }
+    if isinstance(original, list) and isinstance(repaired, list):
+        if len(original) != len(repaired):
```
DavdGao: Is there a case where json_repair drops list items? No test covers it. Suggested to remove this branch.

src/agentscope/_utils/_common.py
```diff
@@ -92,6 +92,36 @@ def _normalize_local_path(path: str) -> str:
     return os.path.abspath(os.path.expanduser(path))
 
 
+def _revert_dropped_entries(original: Any, repaired: Any) -> Any:
+    """Undo entries the schema-guided repair dropped, at any depth.
```
DavdGao: The docstring (and the expanded inline comment below) is too long, one or two lines are enough.

DavdGao (commented):
The problem is real, but the fix can be much simpler. Please see the inline comments.

src/agentscope/_utils/_common.py
Linxiushen: Done in 8d11eaf: the existing `res = parsed` stays, and the check is now just `_has_dropped_keys(parsed, res)`, a small predicate that recurses through nested dicts.

src/agentscope/_utils/_common.py
Linxiushen: I found no case where json_repair drops list *items*, so the length branch is gone. It does drop unknown keys from dicts *inside* arrays (`additionalProperties` is applied to array items too), so the predicate still walks into lists to look for nested dicts; added a test for that case.

src/agentscope/_utils/_common.py
Linxiushen: Trimmed to a one-line docstring, and the inline comment is back to its original two lines.

Linxiushen:
Thanks for the review, @DavdGao. Reworked in 8d11eaf as suggested: the helper is now a small recursive predicate (`_has_dropped_keys`), the revert stays `res = parsed`, the list-length branch is removed and the docstring is one line. The test still fails on main and passes with the change; black/flake8/pylint are clean on the touched lines.

PR #2713, opened 2026-09-21
perf(message): optimize streamed base64 accumulation

### LINKED ISSUE #2707: [Bug]: streamed Base64 accumulation repeatedly transcodes the full payload
### Prerequisites

- [x] I searched existing issues, discussions, PRs, and relevant history. This is not a duplicate.
- [x] This is a performance bug rather than a usage question.

### Background / Description

Streamed Base64 accumulation repeatedly decodes and re-encodes the complete payload on every chunk in two current paths:

- `ToolResponse.append_chunk()` via `_merge_base64_chunks()`
- `Msg.append_event()` for `DATA_BLOCK_DELTA`

Both paths correctly avoid concatenating independently padded Base64 strings, but each new chunk processes all bytes accumulated so far. For `n` similarly sized chunks, the total historical payload processed is proportional to `1 + 2 + ... + n`.

This is separate from #1900/#1901, which fixed padded-chunk correctness, and from #2549/#2551, which concerns tool-result identity during replay. The performance cost was acknowledged in the #2551 discussion and intentionally left out of that PR's scope.

On current `main` (`5ff52f877de12d66a30d55af279dd4f42b1590f3`), using valid independently encoded 4 KiB chunks:

| Chunks | Payload | `ToolResponse.append_chunk` | `DATA_BLOCK_DELTA` |
|---:|---:|---:|---:|
| 100 | 0.39 MiB | 0.0332 s | 0.0318 s |
| 250 | 0.98 MiB | 0.2032 s | 0.1899 s |
| 500 | 1.95 MiB | 0.8946 s | 0.7641 s |

Increasing the chunk count 5x increases runtime about 27x and 24x respectively. A one-pass decode/join/encode reference takes about 0.00325 s for the 500-chunk input, although deferring materialization is not directly compatible with the current requirement that the complete Base64 string remain visible after every delta.

The expected behavior is to preserve the current public models and immediate single-block visibility while avoiding repeated decoding and encoding of all historical bytes. Some cumulative string-copy cost may remain because `Base64Source.data` is an immutable complete string after every update.

### Error Messages

```shell
No exception is raised. The failure mode is superlinear CPU work and repeated full-payload allocation/copying during long streamed audio/file/image accumulation.
```

### Steps to Reproduce

Run the following against current `main`:

```python
import base64
import statistics
import time

from agentscope.event import (
    DataBlockDeltaEvent,
    DataBlockEndEvent,
    DataBlockStartEvent,
)
from agentscope.message import AssistantMsg, Base64Source, DataBlock

encoded = base64.b64encode(b"x" * 4096).decode()
media_type = "application/octet-stream"

for count in (100, 250, 500):
    tool_samples = []
    event_samples = []

    for _ in range(3):
        response = ToolResponse()
        start = time.perf_counter()
        for _ in range(count):
            response.append_chunk(
                ToolChunk(
                    content=[
                        DataBlock(
                            id="data",
                            source=Base64Source(
                                data=encoded,
                                media_type=media_type,
                            ),
                        ),
                    ],
                ),
            )
        tool_samples.append(time.perf_counter() - start)

        msg = AssistantMsg(id="reply", name="agent", content=[])
        msg.append_event(
            DataBlockStartEvent(
                reply_id="reply",
                block_id="data",
                media_type=media_type,
            ),
        )
        start = time.perf_counter()
        for _ in range(count):
            msg.append_event(
                DataBlockDeltaEvent(
                    reply_id="reply",
                    block_id="data",
                    data=encoded,
                    media_type=media_type,
                ),
            )
        msg.append_event(
            DataBlockEndEvent(
                reply_id="reply",
                block_id="data",
                media_type=media_type,
            ),
        )
        event_samples.append(time.perf_counter() - start)

    print(
        count,
        statistics.median(tool_samples),
        statistics.median(event_samples),
    )
```

### Proposed Scope

A focused implementation could introduce one internal incremental Base64 accumulator shared by the two existing paths:

- preserve `Base64Source.data: str` and all public schemas;
- preserve independently padded chunk semantics;
- preserve the existing `ToolResponse` invalid-placeholder fallback;
- keep the complete accumulated string visible after every append/delta;
- avoid decoding and re-encoding historical bytes on each update;
- treat any private optimization state as discardable cache, rebuilding from the public field after validation/copy/deserialization;
- cover interleaved block IDs, external field mutation, Pydantic dump/validate/copy behavior, URL transitions, cleanup, and long streams.

An incremental encoder carrying only the trailing 0–2 decoded bytes appears promising: it can append canonical Base64 without retranscoding the historical payload. It still pays the unavoidable immutable-string extension cost, but removes the larger repeated decode/byte-copy/re-encode work and avoids retaining a full duplicate byte buffer.

If this direction fits the maintainers' intended model semantics, I can prepare a focused implementation and regression/complexity tests within 3–5 days. I will wait for feedback before coding because the private-state and compatibility boundaries need agreement.

### Environment

- AgentScope: `main` at `5ff52f877de12d66a30d55af279dd4f42b1590f3`
- Python: 3.11.16
- OS: macOS arm64

AI-assisted investigation; the code paths, benchmarks, and proposed scope were reviewed locally.

iluv7:
## AgentScope Version

2.0.8

## Description

Closes #2707.

Streamed Base64 data was accumulated by decoding the complete historical payload, appending the latest independently encoded chunk, and re-encoding the entire result after every update. This affected both `ToolResponse.append_chunk()` and `Msg.append_event()` and caused superlinear CPU work and repeated full-payload allocation for long audio, image, and file streams.

This PR introduces a shared internal incremental Base64 accumulator that:

- decodes only the incoming chunk after initializing its discardable cache;
- carries only the trailing 0-2 decoded bytes between updates;
- keeps the complete canonical Base64 string visible after every update;
- preserves independently padded chunk semantics and the existing `ToolResponse` invalid-placeholder fallback;
- detects external mutation of `Base64Source.data` and rebuilds its cache;
- keeps optimization state private and out of Pydantic serialization.

The two existing accumulation paths now share this implementation. No public models, schemas, or APIs are changed.

### Tests

```text
PYTHONPATH=src .venv/bin/python -m pytest -q tests/message_test.py tests/event_to_message_test.py tests/toolkit_test.py

29 passed
```

The added regression coverage includes:

- independently padded Base64 chunks;
- interleaved data block IDs;
- long streams with 500 chunks;
- external `data` mutation;
- Pydantic dump/validate and deep-copy behavior;
- compatibility with invalid placeholder strings;
- a non-timing complexity assertion verifying that no decode operation receives the growing historical payload.

Black, Flake8, and Pylint pass for all modified files.

An additional end-to-end image check generated a 320 x 240 RGB PNG (230,921 bytes), split it into 61 independently encoded chunks, and passed it through both accumulation paths. Both reconstructed images matched the source byte-for-byte and pixel-for-pixel, had the same SHA-256 digest, and were successfully decoded by Pillow.

### Performance

The benchmark used independently encoded 4 KiB chunks and reports the median of seven runs on macOS arm64 with Python 3.13.3. It compares the previous full-payload transcode algorithm with the new incremental accumulator using identical input in the same process.

| Chunks | Previous | Incremental | Speedup | Time reduction |
|---:|---:|---:|---:|---:|
| 100 | 0.038002 s | 0.001857 s | 20.47x | 95.1% |
| 250 | 0.246269 s | 0.010190 s | 24.17x | 95.9% |
| 500 | 1.073561 s | 0.062822 s | 17.09x | 94.1% |

Some cumulative string-copy cost remains because `Base64Source.data` must continue exposing the complete immutable string after every delta, but historical bytes are no longer repeatedly decoded and re-encoded.

## Checklist

- [x] Related documentation has been updated (not applicable: this is an internal performance optimization with no public API or behavior changes)


CorgiBoyG:
Thanks again for offering to step back in the issue thread. The repository has no `/unassign me` handler, so #2707 still lists you and this draft remains open. I now have the replacement implementation completed and fully validated, but I will not publish a competing PR while #2713 is open. When convenient, could you please close this draft and remove the issue assignment through the GitHub UI? That will let me use the normal claim flow and proceed without overlap.

iluv7:
@CorgiBoyG 
Only maintainers can modify the assigned object.

CorgiBoyG:
Understood, thank you for checking. The maintainers will need to handle the issue assignment. Could you still close this draft PR from your side, as you offered in the issue thread? That will remove the active overlapping implementation while we wait for a maintainer to correct the assignee.

PR #2714, opened 2026-09-21
fix(task): reject unsatisfiable dependencies

### LINKED ISSUE #2710: [Bug]: TaskUpdate accepts cyclic and completed task dependencies
### Prerequisites

- [x] I searched the open issues and pull requests for task dependencies, cycles, and `blockedBy`.

### Summary

`TaskUpdate` accepts dependency edges that can never be satisfied. A task can block itself, two or more tasks can form a dependency cycle, and a completed task can later be added as a new prerequisite.

The task tools are presented as the agent's tracked plan: a task in `blocked_by` must complete before the dependent can start. The accepted edges below therefore leave the plan in an impossible or misleading state.

### Reproduction

Against `main` at `5ff52f877de12d66a30d55af279dd4f42b1590f3`:

1. Create task `A`, then call `TaskUpdate(task_id=A, add_blocked_by=[A])`.
   - Actual: the result reports success; `A.blocks == [A]` and `A.blocked_by == [A]`.
2. Create `A` and `B`, set `A` to block `B`, then set `B` to block `A`.
   - Actual: both tasks become mutually blocked.
3. Mark `A` completed, then set `B` to be blocked by `A`.
   - Actual: `TaskList` reports `B` as pending and blocked by an already-completed task.

### Root Cause

`TaskUpdate.call()` filters proposed `add_blocks` / `add_blocked_by` values only by membership in existing task IDs, then `_update_block_relation()` writes both reciprocal lists. It does not validate self-edges, reachability in the existing dependency graph, or task lifecycle state.

### Expected Behavior

- A task must not be its own prerequisite.
- Adding an edge must not create a direct or indirect cycle.
- Completed tasks must not be introduced into a new blocking relation.
- Invalid candidates must leave the task graph unchanged; valid edges in the same request should retain the existing behavior.

### Proposed Scope

A focused fix can validate each candidate edge before mutating task state, then add unit coverage for self-dependency, a two-task cycle, and a completed prerequisite. Expected scope: `src/agentscope/tool/_task/_update_task.py` and `tests/task_tool_test.py`.

No public API or dependency changes are needed.

AI-assisted investigation; the cited paths and three reproductions were verified locally against the stated commit.

helloJamest:
Fixes #2710

## Summary
- reject self-referential dependency edges and edges that close direct or indirect cycles
- prevent completed tasks from participating in new dependency relations
- add regression coverage for self dependencies, three-task cycles, and completed prerequisites

## Tests
- `PYTHONPATH=src .venv\\Scripts\\python.exe -m pytest tests/task_tool_test.py -q`
- `pre-commit run --files src/agentscope/tool/_task/_update_task.py tests/task_tool_test.py`


src/agentscope/tool/_task/_update_task.py
```diff
@@ -273,6 +291,35 @@ async def call(
...
+        pending = [blocked_by_id]
+        visited: set[str] = set()
+        while pending:
+            current_id = pending.pop()
+            if current_id == block_id:
+                return False
+            if current_id in visited:
+                continue
+            visited.add(current_id)
+            pending.extend(tasks_by_id[current_id].blocks)
```
DavdGao: A completed task keeps its `blocks`, since completion only clears the dependents' `blocked_by`. So the DFS walks through stale edges: with X -> A(completed) -> B, adding B -> X is rejected as a cycle. Suggested to skip completed tasks in the traversal.

src/agentscope/tool/_task/_update_task.py
```diff
@@ -179,13 +179,26 @@ async def call(
...
         if add_blocks:
-            current_blocks = _agent_state.tasks_context.tasks[index].blocks
+            current_blocks = tasks[index].blocks
             new_blocks = [
                 _
                 for _ in add_blocks
-                if _ not in current_blocks and _ in existed_ids
+                if _ not in current_blocks
+                and _ in existed_ids
+                and self._can_add_block_relation(
```
DavdGao: Invalid edges are skipped silently, so the LLM only gets "No updates were made" or a success message and never learns why. Suggested to return an explicit error, e.g. "would create a cycle" or "task X is completed". Same for `add_blocked_by` below.

src/agentscope/tool/_task/_update_task.py
```diff
@@ -179,13 +179,26 @@ async def call(
             updated_fields.append("description")
             _agent_state.tasks_context.tasks[index].description = description
 
-        existed_ids = [_.id for _ in _agent_state.tasks_context.tasks]
+        tasks = _agent_state.tasks_context.tasks
```
DavdGao: The `tasks` alias and turning `existed_ids` into a set are unrelated to this fix. Please keep the diff minimal.

tests/task_tool_test.py
```diff
@@ -1112,3 +1112,91 @@ async def test_update_nonexistent_task(self) -> None:
...
+        )
+        task = self.agent_state.tasks_context.tasks[0]
+
+        await self.task_update(
+            task_id=task.id,
+            add_blocked_by=[task.id],
+            _agent_state=self.agent_state,
+        )
+
+        self.assertEqual(task.blocks, [])
```
DavdGao: Please compare the full task dumps with `assertListEqual` rather than asserting `blocks` / `blocked_by` one by one. Same for the other new tests.

DavdGao (commented):
Thanks! Please see the inline comments.

src/agentscope/tool/_task/_update_task.py
helloJamest: Updated the traversal to stop at completed tasks and added the `X -> A(completed) -> B` regression case. Adding `B -> X` now succeeds because the completed task's stale `blocks` are not traversed.

src/agentscope/tool/_task/_update_task.py
helloJamest: TaskUpdate now reports rejected self, cycle, completed, and unknown-task dependencies with `TaskDependencyError` and `ToolResultState.ERROR`. Valid candidates in the same request remain applied; the response also lists applied fields.

src/agentscope/tool/_task/_update_task.py
helloJamest: Removed the `tasks` alias and restored `existed_ids` to a list. The remaining `_add_block_relations` helper consolidates the two duplicated validation/apply loops and keeps `TaskUpdate.call` below the project complexity limit.

tests/task_tool_test.py
helloJamest: The new tests now use `assertListEqual` against full `model_dump()` task lists. Error tests snapshot state before the invalid update, and the completed traversal test asserts every task field after the valid edge is added.

PR #2722, opened 2026-09-21
feat(rag): add HTTP-backed RAGFlow knowledge base

### LINKED ISSUE #2481: feat(rag): support RAGFlow in a new `RAGFlowKnowledgeBase` class rather than VDB

wangtaotaotao95:
## AgentScope Version

2.0.8

## Description

Fixes #2481.

This adds RAGFlow at the knowledge layer, where its managed parsing, chunking, indexing, and retrieval pipeline fits without pretending to be a vector store.

- extract the narrow KnowledgeBaseBase contract consumed by RAGMiddleware (name, description, search), and keep KnowledgeBase as its vector-store implementation
- add RAGFlowKnowledgeBase using the documented asynchronous HTTP API through the existing httpx dependency, preserving Python 3.11+ support
- support text retrieval, native metadata filters, hybrid score metadata, file upload plus parse, document status/listing, chunk pagination, deletion, explicit API errors, and client cleanup
- add an executable example and offline MockTransport coverage for requests, response mapping, failures, and both knowledge implementations through the existing middleware surface

This supersedes inactive #2485 while preserving its author credit: both commits from @ConcaSun were cherry-picked intact, then the implementation was adapted to the maintainer-requested abstraction and the HTTP approach identified in the issue discussion. No persistence schema change or optional dependency is introduced.

## Verification

- uv run --extra service --extra vdb-qdrant --with pytest pytest tests/middleware_rag_test.py tests/rag_ragflow_test.py -q (26 passed)
- uv run --with pre-commit pre-commit run --files src/agentscope/rag/_knowledge.py src/agentscope/rag/_ragflow.py src/agentscope/rag/__init__.py src/agentscope/middleware/_rag.py tests/rag_ragflow_test.py examples/rag/ragflow_knowledge.py examples/rag/README.md pyproject.toml (all passed)
- git diff --check

## Checklist

- [ ] Related documentation has been updated in the separate documentation repository (the in-repo example and README are updated here; a docs-repo follow-up can be opened if maintainers want this API documented there before merge)


src/agentscope/rag/_knowledge.py
```diff
@@ -41,7 +42,38 @@
...
+
+    description: str
+    """Description used by agents to decide when to retrieve."""
+
+    def __init__(self, name: str, description: str) -> None:
+        """Initialize the shared descriptive fields."""
+        self.name = name
+        self.description = description
+
+    @abstractmethod
```
DavdGao, line 59: `delete_document`, `list_documents` and `list_chunks` have exactly the same signatures on both implementations, and the app layer already calls them (`_knowledge_base.py`, `_index_worker.py`). Suggested to declare these three on the base as well, so the two implementations really share one contract. Only ingestion stays backend-specific.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,442 @@
...
+                )
+                if key not in best or result.score > best[key].score:
+                    best[key] = result
+        return sorted(
+            best.values(),
+            key=lambda result: result.score,
+            reverse=True,
+        )[:top_k]
+
+    async def insert_document(
```
DavdGao: Same name as `KnowledgeBase.insert_document` but a different meaning (source bytes vs pre-built chunks), and it's not on the base. That will confuse developers switching between the two. Suggested to give it a distinct name, e.g. `upload_document`, to make clear it is RAGFlow's own ingestion entry.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,442 @@
...
+        self._config = config
+        self._client = client
+        self._owns_client = client is None
+
+    @property
+    def config(self) -> RAGFlowConfig:
+        """The bound connection and retrieval configuration."""
+        return self._config
+
+    async def __aenter__(self) -> Self:
```
DavdGao: The client lifecycle machinery (`client=` injection, `_owns_client`, `__aenter__`/`__aexit__`) isn't needed. Other classes in the repo just create `httpx.AsyncClient(timeout=...)` in `__init__` (e.g. `_claw_hub.py`, the Feishu channel). Suggested to do the same, keep a single `aclose()`, and drop the rest. Tests can patch `httpx.AsyncClient` to inject a `MockTransport`.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,442 @@
...
+                client.  Primarily useful for shared connection pools and
+                offline ``MockTransport`` tests.  Caller-owned clients are not
+                closed by :meth:`aclose`.
+        """
+        super().__init__(name, description)
+        self._config = config
+        self._client = client
+        self._owns_client = client is None
+
+    @property
```
DavdGao, line 66: This exposes the api key as a plain `str`. Suggested to make `api_key` a `SecretStr`, or not expose `config` at all.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,442 @@
...
+import httpx
+from pydantic import BaseModel, Field
+
+from ._document import Chunk
+from ._knowledge import KnowledgeBaseBase
+from ._vdb import DocumentSummary, VectorSearchResult
+from ..message import DataBlock, TextBlock
+
+
+class RAGFlowError(RuntimeError):
```
DavdGao: Do callers need to catch this specifically? If not, a plain `RuntimeError` with the message is enough and we don't need a new exported exception class.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,442 @@
...
+        """Create the shared asynchronous client on first use."""
+        if self._client is None:
+            self._client = httpx.AsyncClient(timeout=self._config.timeout)
+        return self._client
+
+    def _url(self, path: str) -> str:
+        """Build an absolute RAGFlow API URL."""
+        return f"{self._config.base_url.rstrip('/')}/api/v1/{path.lstrip('/')}"
+
+    async def _request(
```
DavdGao, line 79: There are 13 `isinstance` checks on response shapes and 6 `quote()` calls on hex ids across this file. The API is documented, please trust it: `payload["code"] != 0` -> raise, then index the fields directly.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,442 @@
...
+            if not isinstance(query, DataBlock)
+        ]
+        if not query_texts:
+            return []
+
+        responses = await asyncio.gather(
+            *(self._retrieve(text, top_k) for text in query_texts),
+        )
+        best: dict[tuple[str, str], VectorSearchResult] = {}
+        chunk_indexes: dict[tuple[str, str], int] = {}
```
DavdGao: The chunk index assigned by encounter order isn't the chunk's real position, and `total_chunks` is 0 anyway. Suggested to drop `chunk_indexes` / `next_index`, dedupe by chunk id keeping the highest score, and set `chunk_index=0` with a one-line comment that RAGFlow retrieval doesn't expose the position.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,442 @@
...
+                    chunk_indexes[key] = next_index[document_id]
+                    next_index[document_id] += 1
+                metadata = {
+                    "ragflow_chunk_id": chunk_id,
+                    **{
+                        field: raw[field]
+                        for field in (
+                            "vector_similarity",
+                            "term_similarity",
+                            "highlight",
```
DavdGao, line 165: `highlight` is only returned when the request sets `highlight: true`, which we never do, so this field never appears. Either pass it or drop it.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,442 @@
...
+                        "meta_fields",
+                    )
+                    if field in document
+                },
+            )
+            for document in raw_documents
+            if document.get("id")
+        ]
+
+    async def list_chunks(
```
DavdGao, line 275: The paging loop fetches every chunk up to `offset + limit` in 100-sized pages, and the `page_size = min(100, target)` plus `total` bookkeeping is hard to follow. Since RAGFlow supports `page` / `page_size`, one request with `page=1, page_size=offset + limit` and a slice does the same in 5 lines.

src/agentscope/rag/_ragflow.py
```diff
@@ -0,0 +1,442 @@
+# -*- coding: utf-8 -*-
+"""RAGFlow-backed knowledge base using its asynchronous HTTP API.
```
DavdGao: The 8-line module docstring and the class docstrings repeat the README. One or two lines each is enough.

tests/rag_ragflow_test.py
```diff
@@ -0,0 +1,312 @@
...
+    def setUp(self) -> None:
+        """Initialize the optional client used by each test."""
+        self.client: httpx.AsyncClient | None = None
+
+    async def asyncTearDown(self) -> None:
+        """Close clients created by individual tests."""
+        if self.client is not None:
+            await self.client.aclose()
+
+    def _knowledge(self, handler: Any) -> RAGFlowKnowledgeBase:
```
DavdGao: The assertions are piecemeal (`results[0].score`, `.document_id`, `.chunk.source`, `assertIn(b'filename=...')`, a dict comprehension over selected request keys). Please compare the full structures with `assertListEqual` / `assertDictEqual` against literals: the whole request body, `[r.model_dump() for r in results]`, `[c.model_dump() for c in chunks]`. This `_knowledge` helper and the `assertIsInstance(knowledge, KnowledgeBaseBase)` can go.

examples/rag/README.md
```diff
@@ -1,11 +1,12 @@
...
 
 | Script | What it shows |
 | --- | --- |
 | [`index_and_search.py`](./index_and_search.py) | The minimal pipeline: parse → chunk → embed → insert, then `KnowledgeBase.search`. Start here. |
 | [`integrate_with_agent.py`](./integrate_with_agent.py) | Attaches the same `KnowledgeBase` to an `Agent` via `RAGMiddleware`, in both `static` (auto-inject) and `agentic` (tool-driven) modes. |
+| [`ragflow_knowledge.py`](./ragflow_knowledge.py) | Uses `RAGFlowKnowledgeBase` — a knowledge-layer integration that delegates parsing/chunking/indexing/retrieval to a RAGFlow server instead of running the pipeline locally. |
 
 Both examples use an in-memory Qdrant store (`location=":memory:"`) and the DashScope `text-embedding-v4` model, so no external services are required. The sections below show how to swap in Milvus Lite, MongoDB, or Elasticsearch instead; those backends need additional setup.
```
DavdGao: "Both examples" is now wrong with three scripts.

DavdGao (changes requested):
Thanks for picking this up! The direction is right and the API usage matches the RAGFlow HTTP docs. A few things before merging, see the inline comments. Two more:

- The description says both knowledge implementations are covered "through the existing middleware surface", but `middleware_rag_test.py` isn't touched. Please add one `RAGMiddleware` case that uses `RAGFlowKnowledgeBase` with a `MockTransport`, or drop the claim.
- The branch is based on `5ff52f8`, please merge main.

wangtaotaotao95:
@DavdGao I addressed the review feedback in the latest push: the shared base now declares the common document APIs, RAGFlow ingestion is named `upload_document`, client/error/response handling is simplified, the API key uses `SecretStr`, retrieval deduplicates by chunk id, and the tests now compare full request/result structures (including a `RAGMiddleware` + `MockTransport` case). Relevant local checks pass: 27 tests, Black, Flake8, and mypy. Could you take another look when convenient?

PR #2752, opened 2026-09-21
fix(message-bus): make try_lock unlock ownership-safe

### LINKED ISSUE #2728: [Bug]: Expired try_lock holder can unlock a successor lease
### AgentScope version

`main` at `226daf01`

### What happened?

An expired `MessageBus.try_lock()` holder can release a successor's active lease by calling `unlock(key)` late.

This became directly observable in `InMemoryMessageBus` after #2654 added TTL expiry. The Redis implementation has the same ownership gap: every non-blocking lease stores the constant value `"1"`, and `unlock()` performs an unconditional `DEL`.

A typical sequence is:

1. Task A acquires `k` with a short TTL.
2. A runs longer than the lease, so `k` expires.
3. Task B successfully acquires `k`.
4. A reaches its `finally` block and calls `unlock("k")`.
5. B's active lease is removed, allowing a third worker into the same critical section.

That breaks the documented use of `try_lock()` as a non-blocking mutex for making at-least-once queue drains effectively once.

### Minimal reproduction

For the in-memory backend, advance the mocked monotonic clock past A's deadline, acquire from B, then let A call `unlock()`. `is_locked("k")` incorrectly becomes false. For Redis, deleting A's key to simulate expiry, acquiring from B, and then letting A unlock likewise deletes B's value.

### Expected behavior

`unlock()` should release only the lease owned by the task that called `try_lock()`. If that lease has expired and the key has been reacquired, a stale cleanup must leave the successor intact.

### Suggested fix

Use an opaque ownership token for each non-blocking lease and compare it during release. The Redis comparison and delete should be one atomic Lua operation; the in-memory backend can retain the owning task alongside the expiry deadline.

ZxlDragonDoctor:
### Problem
An expired `MessageBus.try_lock()` holder can release a successor's active lease by calling `unlock(key)` late.

- `InMemoryMessageBus` after #2654 TTL expiry: late `unlock` does `_lock_holders.pop(key)` unconditionally.
- `RedisMessageBus`: non-blocking lease stores the constant `"1"`; `unlock` does unconditional `DEL`.

Sequence: A acquires k (short TTL) ? lease expires ? B acquires k ? A's finally `unlock(k)` deletes B's lease ? a third worker enters the same critical section. That breaks the documented at-least-once ? effectively-once drain contract.

### Solution
- Each non-blocking lease stores an opaque ownership token (`secrets.token_hex(16)`).
- `unlock(key, token=...)` releases only when the token still matches the live lease; omitting `token` uses the current task's successful `try_lock` context.
- Redis compares-and-deletes in one Lua script (`GET` == token then `DEL`).
- Stale unlock leaves the successor intact; `acquire_lock` context path unchanged.

Addresses the reporter analysis and suggested fix on #2728 (no maintainer design note).

### Testing
- Offline (full `pytest tests/` needs app deps not installed on this host):
  - Command: `py -3.13 offline_test_try_lock_ownership.py`
  - Result: `offline_try_lock_ownership: 7 passed`
  - Cases: exclusive until unlocked; lease expiry reacquire; **stale unlock does not release successor**; bare unlock without token keeps foreign lease; explicit token unlock; Redis Lua shape + base signature.
- In-tree tests added for CI: `tests/in_memory_message_bus_test.py` ? `test_stale_unlock_does_not_release_successor_lease`, `test_unlock_without_token_does_not_drop_foreign_lease`.
- CI command: `pytest tests/in_memory_message_bus_test.py -q`

### Agent dimension
multi-agent (distributed lease ownership for queue-drain coordination)

### Core value
Late unlock from an expired holder can no longer strip a successor's try_lock lease.

Fixes #2728


ZxlDragonDoctor:
Closing in favor of #2729 by @wangtaotaotao95, which already addresses #2728 and is the better landing vehicle.

Compared with this PR, #2729:
- was opened first and matches the reporter's ownership-token + Lua compare-and-delete approach
- covers both the in-memory and Redis backends with project tests in 	ests/
- follows the CONTRIBUTING checklist and does not carry local-only harness files

No competing implementation from me - happy to help review #2729 if useful.

PR #2755, opened 2026-09-21
fix(embedding): stop blocking the event loop on Gemini and DashScope calls

### LINKED ISSUE #2702: [Bug]: Gemini and DashScope embedding calls block the event loop

ydflow:
## Summary

Closes #2702. The text and multimodal paths in `GeminiEmbeddingModel` and `DashScopeEmbeddingModel` call synchronous SDK methods directly from `async def` methods. While an embedding request waits for its response, other coroutines on the same loop cannot run, which also serialises the batches `EmbeddingModelBase` dispatches with `asyncio.gather`.

The fix offloads the four blocking calls to a worker thread with `asyncio.to_thread`. Callers, return values, error handling and retry behaviour are unchanged.

## The four call sites

| File | Call | Before | After |
|---|---|---|---|
| `_dashscope/_model.py` | `dashscope.embeddings.TextEmbedding.call` | blocking | `await asyncio.to_thread(...)` |
| `_dashscope/_model.py` | `dashscope.MultiModalEmbedding.call` | blocking | `await asyncio.to_thread(...)` |
| `_gemini/_model.py` | `client.models.embed_content` (text) | blocking | `await asyncio.to_thread(...)` |
| `_gemini/_model.py` | `client.models.embed_content` (multimodal) | blocking | `await asyncio.to_thread(...)` |

`_call_api` is the seam the base class calls per batch under `asyncio.gather`, so wrapping the SDK call there is what actually lets batches overlap.

## Reproduction

I ran the probe from the issue against `main` before changing anything. Both providers reproduced on Windows / Python 3.13 / `dashscope` 1.27.6 / `google-genai` 2.24.0:

```
loop progressed during SDK wait: False      # DashScope, before
loop progressed during SDK wait: False      # Gemini, before
loop progressed during SDK wait: True       # both, after
```

## Tests

New: `tests/embedding_dashscope_eventloop_test.py` and `tests/embedding_gemini_eventloop_test.py`. Each has a text and a multimodal case that schedule a `call_soon_threadsafe` callback from inside the mocked SDK call and assert it fired before the call returned — i.e. the loop kept ticking.

Both new tests were checked against the unpatched code first: reverting only the `to_thread` wrapper turns all four red, so they pin the behaviour rather than the implementation.

```
$ pytest tests/embedding_dashscope_test.py tests/embedding_gemini_test.py \
         tests/embedding_openai_test.py tests/embedding_ollama_test.py \
         tests/embedding_dashscope_eventloop_test.py tests/embedding_gemini_eventloop_test.py -q
30 passed, 1 warning
```

The warning is a pre-existing `dashscope` deprecation notice emitted at import.

## Two decisions worth flagging

**I did not switch Gemini to its async client.** `google-genai` 2.24.0 exposes `client.aio`, but swapping the client object would change credential handling and lifecycle for every Gemini path, not just embedding. `asyncio.to_thread` is symmetric across both providers and keeps the diff to the four call sites. Happy to switch to `client.aio.models.embed_content` if you would rather have true async I/O here.

**I kept the exact `dashscope.embeddings.TextEmbedding.call` attribute path.** `tests/embedding_dashscope_test.py` patches that path with `@patch("dashscope.embeddings.TextEmbedding.call")`; routing through a helper would have broken those tests for no functional gain. Wrapping in a lambda preserves it.

## Notes

I did not run `black`. The repo's own source does not satisfy black's defaults — it would reformat untouched code such as the `cache_identifier` comprehension in `_dashscope/_model.py` — so I checked line length directly instead: no line in the four touched files exceeds 79 columns.


ydflow:
Thanks for the nudge — you're right, and I should have caught this before opening.

#2703 by @Tsubaki-01 is the better change and I'm closing this one as a duplicate.

Comparing the two:

- **DashScope**: equivalent. #2703 passes the function reference directly to `asyncio.to_thread` (`to_thread(TextEmbedding.call, api_key=..., **api_kwargs)`); I wrapped it in a lambda. #2703's form is cleaner.
- **Gemini**: #2703 is strictly better. It switches to `await self.client.aio.models.embed_content`, which is real non-blocking I/O. I deliberately kept the sync client behind `asyncio.to_thread`, reasoning that swapping the client object would change credential and lifecycle handling. That caution was misplaced here — the async client is the correct fix, and mine still occupies a thread-pool worker per request.

My miss during triage: I searched open PRs for `CSVLoader`-style keywords ("embedding event loop", "DashScope embedding") but never checked which PRs already referenced issue #2702. #2703's title doesn't contain either phrase, so the keyword search passed it straight through.

Leaving #2703 to land. @DavdGao's note there about trimming the tests is the right call — if any of that trimming touches the DashScope thread-offload assertion, my `test_sdk_thread` in `tests/embedding_dashscope_eventloop_test.py` covers the same invariant via `call_soon_threadsafe` and can be lifted into the existing test file if useful.

Sorry for the duplicate review cycle.

ydflow:
Closing as a duplicate of #2703, which fixes the same issue with a better Gemini approach (client.aio instead of to_thread). See my comment above for the comparison.

PR #2764, opened 2026-09-22
feat(pipeline): support leader-member delegation in a `TeamPipeline` class

### LINKED ISSUE #2763: feat(pipeline): support leader-member delegation in a `TeamPipeline` class
**Background**

Sub-agents are the most requested way to keep a main agent's context small (see #2491): the main agent delegates a task, the sub-agent works in its own context, and only the summary comes back. Supporting this at the SDK level raises one question, how a sub-agent's HITL events reach the caller of a single `reply_stream`, and how the user's response is routed back to the right agent.

Making the assignment an **external tool** of the main agent answers it without touching the `Agent` class: the main agent parks on the delegation exactly as it does on any external execution, while the pipeline runs the sub-agent, streams its events, and feeds its final reply back as the tool result.

**Changes**

- Add `TeamPipeline` and `TeamMember` to the `pipeline` module
  - members are existing `Agent` instances, the leader assigns to them through one external tool `TeamAssign(member, prompt)`, whose description lists the members and offers their names as the choices of `member`
  - members called in one round run concurrently; calls to the same member run one after another
  - a member's HITL request is streamed out unchanged, and the result sent back to the pipeline is routed to it by `reply_id`
  - members' contexts are cleared once the leader's reply ends (`reset_members`)

DavdGao:
## Description

Closes #2763

Adds `TeamPipeline` and `TeamMember` to the `pipeline` module: a leader agent that delegates tasks to existing member agents through tool calls, without touching the `Agent` class.

- The leader gets one **external tool**, `TeamAssign(member, prompt)`; its description lists the members with their descriptions and `member` is an enum of their names. The leader parks on the assignment exactly as it does on any external execution.
- The pipeline runs the assigned member in its own context, streams its events (starting from its own `ReplyStartEvent`) and feeds its final reply back as the tool result, so the leader only ever sees the summary.
- Members assigned in one round run concurrently; assignments to the same member run one after another. Results land in the leader's context in call order.
- HITL works per participant: a member's `RequireUserConfirmEvent` / `RequireExternalExecutionEvent` is streamed out unchanged, and the result sent back to `reply_stream` is routed to it by `reply_id`. An interrupt aborts every parked participant.
- `reset_members` clears the members' contexts once the leader's `ReplyEndEvent` arrives, so the leader can follow up with a member within one reply, while each member starts afresh in the next.

Nothing beyond the agents' own state needs to be persisted: routing is derived from the leader's `SUBMITTED` assignment and the member's parked `reply_id`.

```python
pipeline = TeamPipeline(
    leader=leader,
    members=[
        TeamMember(agent=researcher, description="Researches a topic and returns a sourced summary"),
        TeamMember(agent=coder, description="Edits code and runs the tests"),
    ],
)

async for event in pipeline.reply_stream(UserMsg("user", "Research A, then implement it")):
    ...
```

## Checklist

- [x] Tests added: `tests/pipeline_team_test.py` (delegation round trip, member HITL round trip, concurrent delegations, interrupt)
- [x] `pre-commit run --all-files` passes


src/agentscope/pipeline/_team_pipeline.py
```diff
@@ -0,0 +1,426 @@
...
+        self,
+        tool_calls: list[ToolCallBlock],
+        results: list[ToolResultBlock],
+    ) -> AsyncGenerator[AgentEvent, None]:
+        """Run one round of assignments, concurrently across members and
+        sequentially within one, collecting the tool results of those that
+        finish."""
+        groups: dict[str, list[ToolCallBlock]] = {}
+        for tool_call in tool_calls:
+            member = json.loads(tool_call.input)["member"]
```
DavdGao: `tool_call.input` keeps the raw string from the model, while the agent validates a *repaired* copy of it (`_json_loads_with_repair` in `_agent.py`). So an input like `{'member': 'researcher', 'prompt': 'Find A'}` is repaired, passes the schema check, becomes SUBMITTED, and then raises `JSONDecodeError` here — the exception escapes `reply_stream` and the leader stays parked on the assignment forever with no tool result. Same at :322 and :421.
- Suggested to parse with `_json_loads_with_repair(tool_call.input, tool.input_schema)` like the agent does, or parse once and reuse it.

src/agentscope/pipeline/_team_pipeline.py
```diff
@@ -0,0 +1,426 @@
...
+                yield_final_msg=yield_final_msg,
+            ):
+                # Assignments are executed by the pipeline itself, so their
+                # require events are not the caller's business
+                if isinstance(evt, RequireExternalExecutionEvent) and all(
+                    _.name == _TeamAssign.name for _ in evt.tool_calls
+                ):
+                    assignments.extend(evt.tool_calls)
+                    continue
+                yield evt
```
DavdGao: When the leader parks on an assignment, `_next_action` yields the placeholder msg "I'm waiting for your permission or the external execution to finish.", and with `yield_final_msg=True` it reaches the caller in the middle of the stream, followed by more events and the real final msg. A `Msg` is supposed to terminate the stream, so a consumer that stops at the first `Msg` loses the rest of the run, and `reply()` returns this placeholder whenever a member parks.
- `_run_member` already drops a member msg with `finished_reason is None`, the same filter is missing for the leader.
- :256 doesn't pass `yield_final_msg` either, inconsistent with here.

src/agentscope/pipeline/_team_pipeline.py
```diff
@@ -0,0 +1,426 @@
...
+            groups.setdefault(member, []).append(tool_call)
+
+        sentinel = object()
+        queue: Queue = Queue()
+
+        async def run_group(name: str, calls: list[ToolCallBlock]) -> None:
+            """Run the assignments to one member one after another."""
+            member = self.members[name]
+            for tool_call in calls:
+                if self._is_parked(member):
```
DavdGao: The class doc says assignments to the same member run one after another, but when the first one parks on HITL the second gets this ERROR and is never retried — the leader has to re-issue it. Its result also lands in the leader's context before the first one's, since the sort only orders within a round.
- Suggested to really queue it after the parked call resumes, or change the doc to match what it does.

src/agentscope/pipeline/_team_pipeline.py
```diff
@@ -0,0 +1,426 @@
...
+                reply_id=self.leader.state.reply_id,
+                execution_results=results,
+            )
+            if len(results) < len(assignments):
+                if results:
+                    async for evt in self.leader.reply_stream(inputs):
+                        yield evt
+                return
+
+    async def reply(
```
DavdGao: `reply()` and the `yield_final_msg` argument exist only here, `PipelineProtocol` and `GoalPipeline` have neither. The protocol is what the upper layers program against, so suggested to declare both there (and implement in `GoalPipeline`), or drop them here.

tests/pipeline_team_test.py
```diff
@@ -0,0 +1,496 @@
...
+        """Drain one run into (event type, owner) pairs."""
+        events = [_ async for _ in self.pipeline.reply_stream(inputs)]
+        agents = [
+            self.leader,
+            *[_.agent for _ in self.pipeline.members.values()],
+        ]
+        owners = {_.state.reply_id: _.name for _ in agents}
+        return [(_.type.value, owners[_.reply_id]) for _ in events]
+
+    def _msg_base(self, name: str) -> dict:
```
DavdGao: `_msg_base` and `_user_msg` build the expected structures with helpers. The expected result should be written literally in the test so the whole structure is visible at a glance.

tests/pipeline_team_test.py
```diff
@@ -0,0 +1,496 @@
...
+        self.assertListEqual(
+            [
+                (_.id, _.state)
+                for _ in self.leader.state.context[-1].get_content_blocks(
+                    "tool_result",
+                )
+            ],
+            [("call-1", "interrupted")],
+        )
+        self.assertListEqual(self.researcher.state.context, [])
```
DavdGao: Missing cases:
- a tool input that only `json_repair` can parse;
- a round with several members where one of them parks;
- two assignments to the same member.

DavdGao (commented):
Some problems found when running the branch locally.

Also, `examples/pipeline/goal/` ships a runnable example with a README while team has none, and `TeamPipeline`/`TeamMember` are exported public API, so the docs repo needs a section for them.

